Issue #6, opened 2020-11-07
Unite Face subtypes and Polygon type


serenity4:
Hi,
I was experimenting with the library and saw that
```
julia> Quadrangle <: Polygon
false
julia> Triangle <: Polygon
false
```
which I found a bit odd at first. Then I realized that `Polygon` is a concrete type. I might be wrong, but wouldn't any 2-face be a polygon? I also understand that the current `Polygon` struct serves to construct any kind of polygon, simple or not, including triangles and quadrangles (geometrically, not `Triangle` nor `Quadrangle` instances).

Maybe would it be better to define a `Polygon` abstract type, and have the current `Polygon` concrete type be something like a `PolygonArea` or `GenericPolygon`?

I guess it could also be possible to use a new `Polygon` abstract type in a trait-based approach instead, which would apply to `Polytope` objects.

juliohm:
I was also thinking about this a lot, and ended up naming it a `Polygon` to match what most people would expect in the GIS world. However, I agree that this is confusing, and we should probably rename it to `Ngon` or some more specialized name.

We have `Quadrangle <: Polytope` and that works as expected. If we rename `Polygon -> Ngon` I think it will be more difficult to expect people trying `Quadrangle < Ngon`.

Would you like to submit a PR with this renaming?

serenity4:
I have been thinking some more about this, and I believe the confusion _may_ come from the definition of the abstract types `Polytope` and `Face`.
In the following I assume that the current `Polytope` type refers to a geometric polytope, and not to an abstract polytope.

You can classify a k-polytope by its "dimension" or "number" k. You can have a polygon or a polyhedron for example for k=2 and k=3. But a polygon can be defined in 2D, 3D or higher, same as a polyhedron can be defined in 4D. In addition to geometry parameters (embedding space dimension `Dim` and element type `T`), the polytope should probably contain this k parameter to better describe it and e.g. differentiate between polygons and polyhedra.

Now, I believe the abstract polytope theory to be useful in bringing another level of abstraction. According to [Wikipedia](https://en.wikipedia.org/wiki/Abstract_polytope), an abstract polytope does not depend on the dimension of the Euclidian space. Only the geometric polytope does: 
> An ordinary geometric polytope is said to be a realization in some real N-dimensional space, typically Euclidean, of the corresponding abstract polytope.

[...]

PR #7, opened 2020-11-08
Rename Polygon -> PolySurface

serenity4:
Everything should be renamed properly from `Polygon` to `PolySurface`, as suggested in #6.
I also updated the docstrings and variable names to "polysurface" instead of "polygon" when it referenced the new `PolySurface` type.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 *.jl.mem
 .DS_Store
 Manifest.toml
+/.vscode
--- a/.gitignore
+++ b/.gitignore
@@ -3,4 +3,4 @@
 *.jl.mem
 .DS_Store
 Manifest.toml
-/.vscode
+.vscode
--- a/src/Meshes.jl
+++ b/src/Meshes.jl
@@ -74,7 +74,7 @@ export
   windingnumber, orientation,
 
   # polygons
-  Polygon,
+  PolySurface,
   rings, hasholes, issimple,
   windingnumber, orientation,
 
--- a/src/Meshes.jl
+++ b/src/Meshes.jl
@@ -73,7 +73,7 @@ export
   isclosed, issimple,
   windingnumber, orientation,
 
-  # polygons
+  # polysurfaces
   PolySurface,
   rings, hasholes, issimple,
   windingnumber, orientation,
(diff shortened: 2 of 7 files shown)


.gitignore
```diff
@@ -3,3 +3,4 @@
 *.jl.mem
 .DS_Store
 Manifest.toml
+/.vscode
```
juliohm: I think you can remove the `/` as it is assumed by default

src/Meshes.jl
```diff
@@ -74,7 +74,7 @@ export
   windingnumber, orientation,
 
   # polygons
```
juliohm: Missed comment update

src/discretization/fist.jl
```diff
@@ -23,7 +23,7 @@ to heuristics implemented in the algorithm.
 """
 struct FIST <: DiscretizationMethod end
 
-function discretize(polygon::Polygon, method::FIST)
+function discretize(polygon::PolySurface, method::FIST)
```
juliohm: We should also update the argument to `polysurface`

src/polygons.jl
```diff
@@ -3,9 +3,9 @@
 # ------------------------------------------------------------------
 
 """
-    Polygon(outer, [inner1, inner2, ..., innerk])
+    PolySurface(outer, [inner1, inner2, ..., innerk])
 
-A polygon with `outer` ring, and optional inner
+A polygon surface with `outer` ring, and optional inner
```
juliohm: I would write "A polygonal surface"

juliohm (changes requested):
Awesome. Thanks for the quick PR! Just a few small changes before we can merge it.

serenity4:
Ah, I missed those. Thanks!

src/plotrecipes/polygons.jl
```diff
@@ -2,7 +2,7 @@
 # Licensed under the MIT License. See LICENSE in the project root.
 # ------------------------------------------------------------------
 
-@recipe function f(polygon::Polygon)
+@recipe function f(polygon::PolySurface)
```
juliohm: Missed this one. I would do a /search/replace in the codebase to update every instance, including the tests.

serenity4:
Right, it should be fine now.

PR #8, opened 2020-11-08
Put polytope parametric dimension as a type parameter

serenity4:
Here are changes made to the `Polytope` type, to include the parametric dimension as a type parameter. I have slightly modified the `Face`s to integrate this change.

I also defined `Polygon` and `Polyhedron` aliases, which are not used at the moment (I didn't want to bother to define `PolygonFace` and `PolyhedronFace`). However, if we decide to remove this `Face` type they could be used to make the code clearer.


serenity4:
Well, there is the abstract polytope `AbstractPolytope` and the geometric polytope `Polytope`. Prefixing it with Abstract leaves room for a shorter name for the geometric one.

serenity4:
I added some tests to check that the parametric dimension is correct.

juliohm:
Makes sense. But do you feel that we will ever separate these two in the context of this package? I mean, do we need an instance of a geometric `Polytope`?

juliohm:
I am asking it because it is much nicer to dispatch on the clean name `foo(::Polytope)` than it is on `bar(::AbstractPolytope)`. Also, given that we are in a geometric context here, most people will want to dispatch on the geometric polytope. Let's rename it for consistency, at least for now.

serenity4:
The geometric polytope is the one we will use, primarily for dispatch reasons, since it has the `Dim` and `T` parameters. However, for the abstract one I don't know. You could operate on abstract polytopes in non-Euclidian spaces in you wanted, or use a totally different algebra such as Geometric Algebra. But maybe that's just not going to happen...

serenity4:
People should not have to see nor use `AbstractPolytope` at the moment. The polytope of interest is still `Polytope`, just like before.

juliohm:
Given that we are not using the `AbstractPolytope` anywhere in the code base (yet), it doesn't make much sense to introduce it in this PR. Can you just remove it for now? This will make the PR very clean for future reference, and this concept of abstract polytope can be discussed on its own in a separate issue.

serenity4:
I think it's good!

juliohm:
Amazing! Merging it :100:

PR #10, opened 2020-11-08
Make Chain/Face constructors available for all polytopes

serenity4:
I think it may be safe to assume that all the (geometric) polytopes will have vertices. In that case, the convenience constructors can be defined for all polytopes, not just `Face`s. Since a `Chain` is a 1-polytope, there is no need to specify other constructors for it, so I removed them.
I also believe that the plot recipe can apply to any polytope. They will just need to have a `facets` function defined.

### DIFF
--- a/src/geometries.jl
+++ b/src/geometries.jl
@@ -146,11 +146,11 @@ parametric dimension of the polytope. For example, a segment is a 1-face, a tria
 """
 abstract type Face{N,Dim,T} <: Polytope{N,Dim,T} end
 
-(::Type{F})(vertices::Vararg{P}) where {F<:Face,P<:Point} = F(SVector(vertices))
-(::Type{F})(vertices::AbstractVector{TP}) where {F<:Face,TP<:Tuple} = F(Point.(vertices))
-(::Type{F})(vertices::Vararg{TP}) where {F<:Face,TP<:Tuple} = F(collect(vertices))
+(::Type{F})(vertices::Vararg{TP}) where {F<:Polytope,TP<:Tuple} = F(collect(vertices))
+(::Type{F})(vertices::AbstractVector{TP}) where {F<:Polytope,TP<:Tuple} = F(Point.(vertices))
+(::Type{F})(vertices::Vararg{P}) where {F<:Polytope,P<:Point} = F(SVector(vertices))
 
-==(f1::Face, f2::Face) = f1.vertices == f2.vertices
+==(p1::Polytope, p2::Polytope) = p1.vertices == p2.vertices
 
 function Base.show(io::IO, f::Face)
   kind = nameof(typeof(f))
--- a/src/chains.jl
+++ b/src/chains.jl
@@ -12,10 +12,6 @@ struct Chain{Dim,T} <: Polytope{1,Dim,T}
   vertices::Vector{Point{Dim,T}}
 end
 
-Chain(points::Vararg{P}) where {P<:Point} = Chain(collect(points))
-Chain(points::AbstractVector{TP}) where {TP<:Tuple} = Chain(Point.(points))
-Chain(points::Vararg{TP}) where {TP<:Tuple} = Chain(collect(points))
-
 """
     isclosed(chain)
 
--- a/src/plotrecipes/faces.jl
+++ b/src/plotrecipes/faces.jl
@@ -12,8 +12,8 @@
 end
 
 # plot facets in a recursion
-@recipe function f(face::Face)
-  for fa in facets(face)
+@recipe function f(p::Polytope)
+  for fa in facets(p)
     @series begin
       primary --> false
       fa


serenity4:
In some places we use `foo.vertices` when we don't know the final structure, for polytope and face subtypes. How about using a `vertices` function and defining `vertices(p::Polytope) = p.vertices` that we would redefine for the types that don't match this, such as the `PolySurface`?

That's just an idea, I'll look into the code tomorrow to see how it would mingle with the existing `vertices` function.

juliohm:
I thought `foo.vertices` was only accessed inside implementations. I think `vertices(p::Polytope) = p.vertices` is already defined somewhere or I did it in the Face level. We can review it tomorrow.

serenity4:
> I thought foo.vertices was only accessed inside implementations.

It is indeed, it's just that any `Polytope` will need a `vertices` field. But since there is actually `vertices(p::Polytope) = p.vertices` defined it would make sense to remove `foo.vertices` with `vertices(foo)` everywhere, so that we don't hardcode the need to have a `vertices` field for a polytope (which is actually not the case for a `PolySurface`).

src/geometries.jl
```diff
@@ -146,11 +146,11 @@ parametric dimension of the polytope. For example, a segment is a 1-face, a tria
...
 
-(::Type{F})(vertices::Vararg{P}) where {F<:Face,P<:Point} = F(SVector(vertices))
-(::Type{F})(vertices::AbstractVector{TP}) where {F<:Face,TP<:Tuple} = F(Point.(vertices))
-(::Type{F})(vertices::Vararg{TP}) where {F<:Face,TP<:Tuple} = F(collect(vertices))
+(::Type{F})(vertices::Vararg{TP}) where {F<:Polytope,TP<:Tuple} = F(collect(vertices))
+(::Type{F})(vertices::AbstractVector{TP}) where {F<:Polytope,TP<:Tuple} = F(Point.(vertices))
+(::Type{F})(vertices::Vararg{P}) where {F<:Polytope,P<:Point} = F(SVector(vertices))
 
-==(f1::Face, f2::Face) = f1.vertices == f2.vertices
+==(p1::Polytope, p2::Polytope) = p1.vertices == p2.vertices
```
serenity4, line 153: Actually, I am not sure about this one. We can define this as a default, which will be fine for all common implementations _if the ordering convention is the same_, but essentially a polytope is not just 0-faces, it's all k-faces. So, to test equality of vertices, we can just stick to `vertices(p1) == vertices(p2)`, but otherwise the polytope structure may not be the same. Actually it could be interesting to define a full polytope structure, and it deserves its own issue.

juliohm:
> it would make sense to remove `foo.vertices` with `vertices(foo)` everywhere, so that we don't hardcode the need to have a `vertices` field for a polytope (which is actually not the case for a `PolySurface`)

What I try to do always is avoid calling public api inside implementations. Because the implementations are specific to a type, we will never encounter the issue you mentioned. On the other hand, sometimes we need to reuse other public function inside implementations because they are too verbose to copy/paste. I don't have a strong preference about this. If I may suggest, I would just leave the code as is, and we can come back to it later if needed.

src/geometries.jl
juliohm: It would be nice to learn more about the structure you mentioned. I will comment on the other issue you opened :+1:

serenity4:
>What I try to do always is avoid calling public api inside implementations.

Is it so that core logic can change freely without impacting the public api? In this case (taking vertices of a polytope)  I believe the public `vertices` function should always do what we want internally.

juliohm:
> Is it so that core logic can change freely without impacting the public api? In this case (taking vertices of a polytope) I believe the public `vertices` function should always do what we want internally.

I think it makes the code very clear to potential contributors. They will know that this is just an access to an existing field and that there is no additional computation involved. When someone is editing the source code of the type he/she will know exactly what is happening. When we defer this to a higher-level abstraction in a function that leaves in another source file we add an extra burden to contributors who now need to know the type hierarchy to contribute something that was initially local to that file.

juliohm:
I think calling public API inside the implementation is fine when there is computation involved. We want to reuse the computation and fix possible bugs in a single place that way.

juliohm:
I guess this PR is outdated in favor of #13 right?

serenity4:
I tried to separate this PR and #13. This one implies changes different than #13, although I just defined constructors for `Polytope` (just like in this PR) in the last commit as it was necessary to remove the need for `Face`. So basically, if this one is accepted then I just remove the last commit of #13, else I can merge #13 without this one (but losing the other changes this PR introduces).

juliohm:
I am not sure about the changes in this one. Are you suggesting a semantic change in the Chain type? Initially I was using it as a chain of geometries, but then switched to a simple polygonal chain (chain of points). I know the term chain is also used in simplicial complexes, and we need to double check that in the future. If I may suggest, we go ahead with #13 without merging this one.

serenity4:
The `Chain` type is simply a 1-polytope, with the same structure as the implementations of common polytopes. I removed the specific constructors because I found it redundant, when we make constructors available for all polytopes. We can leave it as is if you think you want to later differentiate a `Chain` from any other polytope - but I don't think this would serve any purpose.

serenity4:
To be clear, this PR would:
1. Make `Face` constructors available for all `Polytope`s
2. Apply the plot recipes to any `Polytope`, not just `Face`s
3. Remove the redundant constructor definitions from `Chain`, since 1. already takes care of that.

juliohm:
Makes sense. I will review both PRs carefully by the end of the day, and will share more comments then.

serenity4:
Yes, take your time. I just happen to have a lot of time to dedicate to this stuff, but I don't expect the same for you 😀

juliohm:
It is great that you have a lot of time to dedicate to this stuff! I am also quite free these days, so I think we will have fun :)

juliohm:
I will close this PR for now as it is doing changes to the code that can be addressed in a single PR about removal of the Face type.

As a side comment, I suggest that we activate our pedantic mode with PRs moving forward. We need to make sure that they do a single thing well, otherwise the review is time consuming and requires testing the code locally.

Issue #12, opened 2020-11-09
Define a full polytope structure


serenity4:
So far we never define a full polytope structure, but maybe it would be nice to have one. So that if you're interested in an object with all its vertices, edges, faces and so on (essentially, a mesh in 3D) without relying on conventions, there's a type available and you don't have to pass around all k-faces individually. We don't even need to encode the value of the vertices if we stick to an abstract polytope (which only describes connectivity, but not data).

juliohm:
It would be really nice to see how a prototype implementation of this. I wonder how it scales and what the public API would look like. Looking forward to learning more about it.

serenity4:
For simple shapes, when it's possible to use a mapping to find all k-faces from vertices, it's probably going to just be inefficient memory-wise. Although, if you don't really care about efficiency at that level, it would be easier for querying faces should you the indices you want. One good thing though is that the resulting polytope would be portable, particularly when the polytope is complex: in computer graphics, you have control over all faces: vertices, edges, "faces" so if you want to keep the geometry you need to save everything.

Implementing it for any number of dimensions might be tricky. In 3D, it might look something like

```julia
struct DensePolyhedron{F<:AbstractVector{<: Polytope{2, Dim, T}}, E<:AbstractVector{<: Polytope{1, Dim, T}}, V<:AbstractVector{<: Polytope{0, Dim, T}}, Dim, T} <: Polyhedron{Dim, T}
    faces::F
    edges::E
    vertices::V
end
```
which, if you only have the same polytope type inside each k-face (only `Triangle`s, `Segment`s and `Point`s), may be OK, but may quickly become a mess if you mix several k-polytopes types for k-faces. Taking simplices, though, would look reasonable for some applications. Maybe, instead of using N fields for N-polytopes, you could use a N-tuple where each item is a list:

```julia
struct DensePolyhedron{F<:AbstractVector{<: Polytope{2, Dim, T}}, E<:AbstractVector{<: Polytope{1, Dim, T}}, V<:AbstractVector{<: Polytope{0, Dim, T}}, Dim, T} <: Polyhedron{Dim, T}
    faces::Tuple{F, E, V} # in the mathematical sense (0-, 1-, 2-faces)
end
```
so that the number of fields stays fixed. But still, we would probably need N+2 type parameters at least to keep it efficient. I don't know, maybe that can be a problem. Particularly if you consider high dimensions, but then you also just have the problem that the number of k-faces (to describe useful geometry) would be insane.

[...]

Issue #11, opened 2020-11-08
Separate FEM logic from the mathematical model


serenity4:
As we discussed in #6, it would be nice to clearly separate what is related to FEM from polytope theory. Currently, a `Face` is used to denote the fact that a type is a polytope used in FEM. Then, common polytopes are defined as `Face` subtypes, with a convenience `facets` function that connects all polytope vertices into facets from the GMSH project conventions.

The specificity of `Face` subtypes (as it is in the code) that is related to FEM is just that a mapping is defined from the object type (triangle, tetrahedron...) to an array of `Connectivity` elements (holding each a set of points and the target polytope), according to the GMSH conventions.

A way to go would be to extract this FEM-specific knowledge and put it somewhere else, detached from the actual polytope. Having types for simple polytopes is useful because they have a specific structure that easily allows to build all k-faces from the vertices according to a convention. This convention should be specified by the user, with a default value that can be GMSH's.

Later on, a separate FEM part can define other polytopes that are only relevant in the field. For example, it can be some kinds of "triangles" with more than 3 vertices, possibly including one at the center. They are not triangles in a mathematical sense, because they have more than 3 edges; however, vertices are layed out so that, with a specific connectivity, edges can be defined that can be seen as a triangle. When a center point is defined, it is simply not connected to anything else.
That is, of course, if you consider all FEM points to represent actual vertices. They can just be "points of interest" that have nothing to do with the geometrical shape. In this case, an adequate sampling method, which rely on an ordering convention, should be defined for the relevant polytopes.

serenity4:
I have changes that I still need to put into separate commits before submitting PRs, but it's time to sleep now.

juliohm:
Awesome improvements here. Closing the issue as it is now fixed on master :100:

PR #13, opened 2020-11-09
Make ordering conventions explicit

serenity4:
This PR implements ordering conventions, that solves #11.
The old `Face` type is still there, but not used anymore, so it can be safely deleted (in this PR or another).
As discussed in #12, equality of polytopes imply equality of all k-faces. Therefore, I modified equality tests between polytopes so that only vertices are tested for equality. It only affects tests for `UnstructuredMesh`es. The tested behavior is exactly the same as before this PR.


juliohm:
Awesome, that looks really nice :100: I will try to test locally today, meanwhile could you please rename the folder `common_polytopes` to simply `polytopes`?

Thanks for the awesome improvements.

src/connectivities.jl
```diff
@@ -26,22 +26,22 @@ facetype(c::Connectivity) = facetype(typeof(c))
...
-into a [`Face`](@ref) of type `F`.
+into a [`Polytope`](@ref) of type `F`.
 
 ## Example
 ```
 Δ = connect((1,2,3), Triangle)
 ```
 """
-connect(list::NTuple{N,Int}, F::Type{<:Face}) where {N} = Connectivity{F,N}(list)
+connect(list::NTuple{N,Int}, F::Type{<:Polytope}) where {N} = Connectivity{F,N}(list)
```
juliohm, line 36: Although I like all your suggestions so far, it would be nice if you could draft PRs more carefully. I am spotting a lot occurrences of incomplete changes like this where you chage Face to Polytope and forget to update the type parameter from F to P.

src/Meshes.jl
```diff
@@ -84,6 +84,10 @@ export
   Connectivity,
   facetype, connect, materialize,
 
+  # ordering conventions
+  Ordering, connectivity,
```
juliohm, line 88: You are exchanging two terms in this PR: ordering and convention. Can we agree on a single term "OrderingConvention" and use it consistently in the PR?

src/conventions.jl
```diff
@@ -0,0 +1,13 @@
...
+
+"""
+Ordering convention used to construct k-faces of a polytope from vertices (0-faces).
+"""
+abstract type Ordering end
+
+"""
+Connectivity function provided by an `Ordering` subtype used to construct k-faces of a polytope from vertices (0-faces).
+"""
+function connectivity end
```
juliohm, line 13: Please configure your IDE to not remove the last EOL. VS Code does that by default unfortunately.

src/conventions/gmsh.jl
```diff
@@ -0,0 +1,103 @@
...
+#     `\.      |    ,/         
+#        `\.   '. ,/           
+#           `\. |/             
+#              `3              
+#                 `\.
+#                    ` w
+
+connectivity(::Type{<:Tetrahedron{GMSH}}, ::Type{GMSH}, ::Val{2}) = 
+    connect.([(1,2,3),(1,4,3),(1,4,2),(2,3,4)], Triangle)
+
```
juliohm, line 84: Sometimes you add 2 blank lines, sometimes you add 1 blank line between function definitions. I know this may sound picky, but it is important in the long run to be pedantic about these things.

juliohm (changes requested):
Hi @serenity4 , this PR isn't very easy to review. It is doing multiple things at the same time, and this is hard to keep track in the future. Could you maybe undo the deletion of the Face type, and just handle the ordering convention concept here? There is a lot of file movement to different folders, and a lot of s/Face/Polytope/ that is not related to ordering conventions.

We need a separate PR later to consider the removal of Face.

juliohm:
In summary, it would be nice if you could

1. Draft a new PR proposing the removal of the Face type.
2. After (1) is merged into master, update this PR to discuss the ordering conventions.

juliohm:
I've addressed item (1) in #14. Requesting your review over there :+1:

serenity4:
Appreciate the feedback. Tbh I am not used to be as clear when making commits/PRs, particularly for having a good history, but that's better practice indeed. I'll try to be as careful as I can for future PRs. Don't hesitate to continue such feedback when needed.

juliohm:
Now that #14 has been merged, we can either refactor this PR or start a clean one from a new branch out of master. It would be nice to generalize the ordering convention as you proposed here. Please let me know how you prefer to proceed.

serenity4:
I think I'll just restart anew, it will be easier.

PR #14, opened 2020-11-10
Deprecate Face type

juliohm:
This PR removes the Face type in favor of the general Polytope type.

### DIFF
--- a/src/geometries.jl
+++ b/src/geometries.jl
@@ -91,21 +91,39 @@ include("primitives/cylinder.jl")
 """
     Polytope{N,Dim,T}
 
-We say that a geometry is a N-polytope when it is a collection of "flat" sides that constitue a N-dimensional subspace.
-They are called polygon and polyhedron respectively for 2D (N=2) and 3D (N=3) subspaces, embedded in a `Dim`-dimensional space.
-The term polytope expresses a particular combinatorial structure. A polyhedron, for example, can be decomposed into faces. Each face can then be decomposed into edges, and edges into vertices.
-Some conventions act as a mapping between vertices and higher dimensional features (edges, faces, cells...), removing the need to store all features. We follow the [ordering conventions](https://gmsh.info/doc/texinfo/gmsh.html#Node-ordering) of the GMSH project.
+We say that a geometry is a N-polytope when it is a collection of "flat" sides
+that constitue a `N`-dimensional subspace. They are called polygon and polyhedron
+respectively for 2D (`N=2`) and 3D (`N=3`) subspaces, embedded in a `Dim`-dimensional
+space. The parameter `N` is also known as the rank or parametric dimension of the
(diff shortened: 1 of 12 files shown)


juliohm:
@serenity4 appreciate if you can take a look before we merge this.

src/geometries.jl
```diff
@@ -91,21 +91,39 @@ include("primitives/cylinder.jl")
...
+respectively for 2D (`N=2`) and 3D (`N=3`) subspaces, embedded in a `Dim`-dimensional
+space. The parameter `N` is also known as the rank or parametric dimension of the
+polytope: https://en.wikipedia.org/wiki/Abstract_polytope.
+
+The term polytope expresses a particular combinatorial structure. A polyhedron,
+for example, can be decomposed into faces. Each face can then be decomposed into
+edges, and edges into vertices. Some conventions act as a mapping between vertices
+and higher dimensional features (edges, faces, cells...), removing the need to
+store all features. We follow the ordering conventions of the GMSH project:
+https://gmsh.info/doc/texinfo/gmsh.html#Node-ordering
```
serenity4, line 105: Do you use a setup to cut lines automatically or do you do it by hand?

src/geometries.jl
```diff
@@ -130,56 +148,47 @@ See https://en.wikipedia.org/wiki/Facet_(geometry)
...
 
 """
-    center(polytope)
+    p1 == p2
 
-Return the center of the `polytope`.
+Tells whether or not polytopes `p1` and `p2` are equal.
 """
-center(p::Polytope) = Point(sum(coordinates.(p.vertices)) / length(p.vertices))
+==(p1::Polytope, p2::Polytope) = p1.vertices == p2.vertices
```
serenity4, line 155: I am still unsure about testing for equality only on vertices, and using `foo.vertices` to do that. It won't work for some polytopes such as `PolySurface`.
We can always add an equality method for `PolySurface`s though. But a `PolySurface` instance can represent exactly the same thing as a `Triangle` with the right parameters, making identical polytopes. But I guess it's not the place to discuss that.

serenity4 (approved):
That's neat. I have mixed feelings about moving the equality method to all `Polytope`s (see comment), but maybe we should just accept it there and deal with it in another issue, since this PR is just about removing the`Face` type.

src/geometries.jl
juliohm: I am currently doing it by hand, but VS Code definitely has an option for it. I should probably activate it to break the line after certain column.

src/geometries.jl
juliohm: Yes, I agree we can brainstorm the equality of polytopes in a separate issue. I think it will be natural to discuss it after we polish some other pressing details first like the ordering issue, and the representation of polytopes in general.

juliohm:
Awesome, I will squash and merge the PR, and continue working on the PolySurface triangulation today.

PR #15, opened 2020-11-10
Use explicit and customizable ordering conventions

serenity4:
Rewrite of PR #13. Removes the implicit ordering defined for common polytopes, in favor of an explicit ordering that can be customized by subtyping from `OrderingConvention`. The previously implicit ordering is now explicit, as defined by the `GMSH` type.

### DIFF
--- a/src/Meshes.jl
+++ b/src/Meshes.jl
@@ -25,6 +25,7 @@ include("angles.jl")
 include("geometries.jl")
 include("connectivities.jl")
 include("mesh.jl")
+include("conventions/gmsh.jl")
 
 # algorithms
 include("sampling.jl")
@@ -78,6 +79,8 @@ export
   # connectivities
   Connectivity,
   polytopetype, connect, materialize,
+  OrderingConvention, GMSH,
+  connectivity,
 
   # meshes
   Mesh,
--- a/src/Meshes.jl
+++ b/src/Meshes.jl
@@ -25,7 +25,7 @@ include("angles.jl")
 include("geometries.jl")
 include("connectivities.jl")
 include("mesh.jl")
-include("conventions/gmsh.jl")
+include("conventions.jl")
 
 # algorithms
 include("sampling.jl")
--- a/src/Meshes.jl
+++ b/src/Meshes.jl
@@ -82,7 +82,7 @@ export
 
   # conventions
   OrderingConvention, GMSH,
-  connectivity,
+  connectivities,
 
   # meshes
   Mesh,
--- a/src/Meshes.jl
+++ b/src/Meshes.jl
@@ -24,8 +24,8 @@ include("angles.jl")
 # geometries and meshes
 include("geometries.jl")
 include("connectivities.jl")
-include("mesh.jl")
 include("conventions.jl")
+include("mesh.jl")
 
 # algorithms
 include("sampling.jl")
@@ -79,6 +79,8 @@ export
   # connectivities
   Connectivity,
   polytopetype, connect, materialize,
+
+  # conventions
   OrderingConvention, GMSH,
   connectivity,
 
--- a/src/connectivities.jl
+++ b/src/connectivities.jl
@@ -48,3 +48,27 @@ end
 function Base.show(io::IO, c::Connectivity{PL}) where {PL}
   print(io, "$PL$(c.list)")
 end
+
+"""
+Convention used to reconstruct the faces of a polytope
+from a list of vertices.
+"""
+abstract type OrderingConvention end
+
+"""
+    connectivity(polytope, convention, k)
+
+Return the list of `Connectivity` elements used to construct
+all `k`-faces of a polytope, following an ordering convention.
+"""
+function connectivity(polytope::Type{<:Polytope}, convention::OrderingConvention, k::Integer)
+  connectivity(polytope, convention, Val(k))
+end
+
+function facets(p::Polytope{N}, ordering=GMSH) where {N}
+  faces(p, N-1, ordering)
+end
+
+function faces(p::Polytope{N,Dim}, rank, ordering=GMSH) where {N,Dim}
+  (materialize(ord, p.vertices) for ord in connectivity(typeof(p), ordering, rank))
+end
--- a/src/connectivities.jl
+++ b/src/connectivities.jl
@@ -48,27 +48,3 @@ end
 function Base.show(io::IO, c::Connectivity{PL}) where {PL}
   print(io, "$PL$(c.list)")
 end
-
-"""
-Convention used to reconstruct the faces of a polytope
-from a list of vertices.
-"""
-abstract type OrderingConvention end
-
-"""
-    connectivity(polytope, convention, k)
-
-Return the list of `Connectivity` elements used to construct
-all `k`-faces of a polytope, following an ordering convention.
-"""
-function connectivity(polytope::Type{<:Polytope}, convention::OrderingConvention, k::Integer)
-  connectivity(polytope, convention, Val(k))
-end
-
-function facets(p::Polytope{N}, ordering=GMSH) where {N}
-  faces(p, N-1, ordering)
-end
-
-function faces(p::Polytope{N,Dim}, rank, ordering=GMSH) where {N,Dim}
-  (materialize(ord, p.vertices) for ord in connectivity(typeof(p), ordering, rank))
-end
--- a/src/conventions.jl
+++ b/src/conventions.jl
@@ -1,0 +1,33 @@
+# ------------------------------------------------------------------
+# Licensed under the MIT License. See LICENSE in the project root.
+# ------------------------------------------------------------------
+
+"""
+Convention used to reconstruct the faces of a polytope
+from a list of vertices.
+"""
+abstract type OrderingConvention end
+
+"""
+    connectivity(polytope, convention, k)
+
+Return the list of `Connectivity` elements used to construct
+all `k`-faces of a polytope, following an ordering convention.
+"""
+function connectivity(polytope::Type{<:Polytope}, convention::OrderingConvention, k::Integer)
+  connectivity(polytope, convention, Val(k))
+end
+
+# ---------------------------
+# CONVENTION IMPLEMENTATIONS
+# ---------------------------
+
+include("conventions/gmsh.jl")
+
+function facets(p::Polytope{N}, ordering=GMSH) where {N}
+  faces(p, N-1, ordering)
+end
+
+function faces(p::Polytope{N,Dim}, rank, ordering=GMSH) where {N,Dim}
+  (materialize(ord, p.vertices) for ord in connectivity(typeof(p), ordering, rank))
+end
--- a/src/conventions.jl
+++ b/src/conventions.jl
@@ -24,10 +24,10 @@ end
 
 include("conventions/gmsh.jl")
 
-function facets(p::Polytope{N}, ordering=GMSH) where {N}
-  faces(p, N-1, ordering)
+function facets(p::Polytope{N}, convention=GMSH) where {N}
+  faces(p, N-1, convention)
 end
 
-function faces(p::Polytope, rank, ordering=GMSH)
-  (materialize(ord, p.vertices) for ord in connectivity(typeof(p), rank, ordering))
+function faces(p::Polytope, rank, convention=GMSH)
+  (materialize(ord, p.vertices) for ord in connectivity(typeof(p), rank, convention))
 end
--- a/src/conventions.jl
+++ b/src/conventions.jl
@@ -9,13 +9,13 @@ from a list of vertices.
 abstract type OrderingConvention end
 
 """
-    connectivity(polytope, convention, k)
+    connectivities(polytope, convention, k)
 
 Return the list of `Connectivity` elements used to construct all
 faces of rank `rank` of a polytope, following an ordering convention.
 """
-function connectivity(polytope::Type{<:Polytope}, rank::Integer, convention::OrderingConvention)
-  connectivity(polytope, Val(rank), convention)
+function connectivities(polytope::Type{<:Polytope}, rank::Integer, convention::OrderingConvention)
+  connectivities(polytope, Val(rank), convention)
 end
 
 # ---------------------------
@@ -29,5 +29,5 @@ function facets(p::Polytope{N}, convention=GMSH) where {N}
 end
 
 function faces(p::Polytope, rank, convention=GMSH)
-  (materialize(ord, p.vertices) for ord in connectivity(typeof(p), rank, convention))
+  (materialize(ord, p.vertices) for ord in connectivities(typeof(p), rank, convention))
 end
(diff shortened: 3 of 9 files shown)


src/conventions/gmsh.jl
```diff
@@ -0,0 +1,102 @@
+# ------------------------------------------------------------------
+# Licensed under the MIT License. See LICENSE in the project root.
+# ------------------------------------------------------------------
+
+"""
+Ordering convention of the GMSH project:
+https://gmsh.info/doc/texinfo/gmsh.html#Node-ordering
+"""
+abstract type GMSH <: OrderingConvention end
```
serenity4: I defined `GMSH` as an abstract type and dispatched on `::Type{GMSH}`, would there be any benefit of defining it as a normal struct? And, in that case, dispatch on an instance `GMSH()` rather than a `Type`?

src/connectivities.jl
```diff
@@ -48,3 +48,27 @@ end
 function Base.show(io::IO, c::Connectivity{PL}) where {PL}
   print(io, "$PL$(c.list)")
 end
+
+"""
+Convention used to reconstruct the faces of a polytope
+from a list of vertices.
+"""
+abstract type OrderingConvention end
```
juliohm: We need a new file `src/conventions.jl` to place this abstract type. This file will also include the `src/conventions/gmsh.jl`

src/conventions/gmsh.jl
juliohm: I think it is fine to dispatch on the abstract type. I am not aware of any issue.

src/connectivities.jl
```diff
@@ -48,3 +48,27 @@ end
...
+  connectivity(polytope, convention, Val(k))
+end
+
+function facets(p::Polytope{N}, ordering=GMSH) where {N}
+  faces(p, N-1, ordering)
+end
+
+function faces(p::Polytope{N,Dim}, rank, ordering=GMSH) where {N,Dim}
+  (materialize(ord, p.vertices) for ord in connectivity(typeof(p), ordering, rank))
+end
```
juliohm: I am not getting how these functions interplay with the existing `connect` function. Can you clarify? Also it seems that the `facets` and `faces` functions don't belong to the connectivities.jl source file.

juliohm (changes requested):
Attached a first round of review comments.

src/connectivities.jl
serenity4: `connectivity` is the function that needs to be overloaded with a custom `OrderingConvention` type. See [this line](https://github.com/JuliaGeometry/Meshes.jl/pull/15/files#diff-0c06e9261f736acae84a9cdcf8553f23a623c77979a6cadb535244fc3f6e2dfbR23) for an example.
Basically, you give a polytope, a convention and a rank, and it will build an array of `Connectivity` elements using `connect`, which correspond to the faces of the chosen rank. `faces`/`facets` add the materialization step to retrieve an array of faces.
I put these functions here because they are directly related to the notion of connectivity.

src/connectivities.jl
serenity4: Ok, I was wondering if it was necessary to have a file with very few lines. But that makes sense.

src/conventions.jl
```diff
@@ -0,0 +1,33 @@
...
+
+include("conventions/gmsh.jl")
+
+function facets(p::Polytope{N}, ordering=GMSH) where {N}
+  faces(p, N-1, ordering)
+end
+
+function faces(p::Polytope{N,Dim}, rank, ordering=GMSH) where {N,Dim}
+  (materialize(ord, p.vertices) for ord in connectivity(typeof(p), ordering, rank))
+end
```
serenity4: I don't really know where to put those two. If we want to have `GMSH` as a default ordering convention, it has to come after the included file. They are originally defined in `geometries.jl`, but the convention implementation needs polytope implementations..

src/conventions/gmsh.jl
juliohm: On the other hand, I think it is good to signal that this type is final. Can you make it a struct instead?

src/connectivities.jl
juliohm: I am not sure I am following the naming here. We have `connect` which effectively holds the connectivity list of the polytope. Your line example seems to implement the facets behavior returning a list of connectivity lists for segments? I am confused. Are you trying to generalize `facets` to any rank? The naming is really important here because we already have the concept of Connectivity and the function connect. A third function called connectivity doesn't help. It would be nice to find a better term for the concept you are trying to implement here. Also you use a keyword argument `ordering` to specify a OrderingConvention. Shouldn't this keyword argument be `convention` instead?

src/conventions/gmsh.jl
serenity4: I agree. I will do that tomorrow.

src/conventions.jl
juliohm: Let me give a second pass at the PR. I am still confused about where these concepts should go and the way you are organizing the changes into files.

src/Meshes.jl
```diff
@@ -25,6 +25,7 @@ include("angles.jl")
 include("geometries.jl")
 include("connectivities.jl")
 include("mesh.jl")
+include("conventions.jl")
```
juliohm, line 27: So conventions.jl could come before the mesh.jl file because it is more closely related to connectivities than it is to meshes.

src/Meshes.jl
```diff
@@ -78,6 +79,8 @@ export
   # connectivities
   Connectivity,
   polytopetype, connect, materialize,
+  OrderingConvention, GMSH,
+  connectivity,
```
juliohm: I don't think the ordering convention should be placed in the same section of the connectivities. It deserves a separate section with conventions, which is a concept in itself.

serenity4: I should probably separate that from the connectivity exports since it is related to conventions, right?

src/conventions.jl
```diff
@@ -0,0 +1,33 @@
...
+# ------------------------------------------------------------------
+
+"""
+Convention used to reconstruct the faces of a polytope
+from a list of vertices.
+"""
+abstract type OrderingConvention end
+
+"""
+    connectivity(polytope, convention, k)
```
juliohm: Perhaps the name of this function should be `connectivities` instead? You are returning a list, which is a plural, not singular concept.

src/conventions.jl
```diff
@@ -0,0 +1,33 @@
...
+# CONVENTION IMPLEMENTATIONS
+# ---------------------------
+
+include("conventions/gmsh.jl")
+
+function facets(p::Polytope{N}, ordering=GMSH) where {N}
+  faces(p, N-1, ordering)
+end
+
+function faces(p::Polytope{N,Dim}, rank, ordering=GMSH) where {N,Dim}
```
juliohm: Why request the type parameters `N,Dim` if they are not used in the implementation? Also, notice how you flip the `rank` argument from second argument in faces to 3 argument in connectivity. These details are important, and we want to strike a consistent API where users don't need to lookup signatures repeatedly.

src/conventions/gmsh.jl
```diff
@@ -0,0 +1,102 @@
...
+# |   |  +-- |-- | -> u
+# 0---+---\--1   |     
+#  \  |    \  \  |     
+#   \ |     \  \ |     
+#    \|      w  \|     
+#     4----------5     
+
+connectivity(::Type{<:Hexahedron}, ::Type{GMSH}, ::Val{2}) = 
+  connect.([(1,2,3,4),(5,6,7,8),(1,5,6,2),
+  (2,6,7,3),(3,4,8,7),(1,5,8,4)], Quadrangle)
```
juliohm: Look at the original formatting: https://github.com/JuliaGeometry/Meshes.jl/blob/f61477cfb346239642b911565b4ed930a8f6ec81/src/polytopes/hexahedron.jl#L28-L32

It is harder to read the list of vertices now because of incorrect indentation.

src/polytopes/hexahedron.jl
```diff
@@ -24,9 +24,3 @@ A hexahedron with points `p1`, `p2`, ..., `p8`.
 struct Hexahedron{Dim,T,V<:AbstractVector{Point{Dim,T}}} <: Polyhedron{Dim,T}
   vertices::V
 end
-
-function facets(hex::Hexahedron)
-  connec = connect.([(1,2,3,4),(5,6,7,8),(1,5,6,2),
-                     (2,6,7,3),(3,4,8,7),(1,5,8,4)], Quadrangle)
-  (materialize(c, hex.vertices) for c in connec)
-end
```
juliohm, line 32: Perhaps the `facets` function is a function that is more related to the polytope than it is to the convention? I think we should keep it in the original polytope file and implement it with the GMSH convention here if needed. That way, people can compare the conventions for the pyramid in a single place as opposed to opening multiple convention files and searching for pyramid.

juliohm (changes requested):

src/conventions.jl
serenity4: Sorry, I tried to be clear but obviously it's not 😅

src/conventions.jl
serenity4: I was thinking that connectivity would express the set of connections, how things connect.

src/Meshes.jl
juliohm: Yes, a section for ordering conventions, and any function related to that if necessary.

src/conventions.jl
juliohm: When you restructure the PR tomorrow by placing the functions in places near the concepts they belong, we will be able to find an appropriate name for it :ok_hand:

src/conventions.jl
serenity4: You're totally right. I really hope I'll see this kind of things more easily because even if I took care with this PR these inconsistencies passed through..

src/polytopes/hexahedron.jl
serenity4: Makes sense. Do you mean the `connectivity` functions though? `facets` is no longer defined for each polytope.

src/conventions.jl
juliohm: I am happy that you are taking the suggestions gracefully, it will really help in the long run as we evolve the package together :) We need to push each other to achieve results that we wouldn't be able to achieve alone :100:

src/polytopes/hexahedron.jl
juliohm: Perhaps I confused things again... I mean whatever function that is responsible for listing the connectivities of the polytope.

serenity4: Should we decide to do this, we'll have to include the conventions.jl file inside geometry.jl because all those polytope implementations will use whatever conventions are defined there. On the other end I feel like it puts even more stuff inside geometry.jl, which is already large. What are your thoughts on this?

src/connectivities.jl
serenity4: I agree for the name of the keyword argument.
`faces` generalizes the retrieval of faces of any rank, just like you define it for a `Mesh`. `facets` calls `faces` with rank `N-1` for a N-polytope, and is therefore also generalized. The difference with the method for `Mesh`es, is that since the connectivity information is bundled with the struct, you don't need a convention. So you only need `faces(mesh, rank)` for a mesh, and `faces(polytope, rank, convention)` for a polytope.

> Your line example seems to implement the facets behavior returning a list of connectivity lists for segments?

The GMSH convention, as you implemented it, returned only the facets of an object (it was only used to define the `facets` methods). Therefore, while a convention should allow to return any set of faces of rank k of an object based on vertices, the GMSH convention is currently only used to return facets.
As I understand it, a `Connectivity` instance associates a list of vertex indices with a polytope. `connectivity` describes how faces of an object are related to its vertices, returning a `Connectivity` for each face to materialize them later on (currently, `faces`/`facets` does that right after calling `connectivity`). I agree that both `Connectivity`/`connectivity` are a bit different and thus bring confusion, so we need to fix it.

src/conventions/gmsh.jl
```diff
@@ -0,0 +1,102 @@
...
+#       |
+# 3-----------2      
+# |     |     |      
+# |     |     |      
+# |     +---- | --> u
+# |           |      
+# |           |      
+# 0-----------1      
+
+connectivity(::Type{<:Quadrangle{GMSH}}, ::Val{1}, ::Type{GMSH}) = connect.([(1, 2), (2, 3), (3, 4), (4, 1)], Segment)
```
juliohm: Inconsistent formatting. Some functions you break the line after the equal sign, and some others like this you produce a really long line. Can you please add a breakline after the equal sign?

src/polytopes/hexahedron.jl
juliohm: I think I am getting the idea now. But I still think the name `connectivity` should be something else, maybe the plural `connectivities` would be enough.

juliohm (changes requested):

src/connectivities.jl
serenity4: The thing is, you could just define `faces` with the logic implemented in `connectivity` so that is always returns `materialize`d faces, but having to explicitly call `materialize` in each method defined by a convention (that is, for many objects/ranks) doesn't seem right to me.

juliohm:
@serenity4 can we rename the `connectivity` function to `connectivities`? I think the ready to be merged after this. :100: 

I am progressing on the polysurface api, and hopefully we will be able to see interesting cases soon.

serenity4:
I think we're good now :)

src/conventions.jl
```diff
@@ -0,0 +1,33 @@
...
+# ------------------------------------------------------------------
+
+"""
+Convention used to reconstruct the faces of a polytope
+from a list of vertices.
+"""
+abstract type OrderingConvention end
+
+"""
+    connectivities(polytope, convention, k)
```
juliohm: Need to update the order of the arguments in the docstring

src/conventions.jl
```diff
@@ -0,0 +1,33 @@
...
+
+Return the list of `Connectivity` elements used to construct all
+faces of rank `rank` of a polytope, following an ordering convention.
+"""
+function connectivities(polytope::Type{<:Polytope}, rank::Integer, convention::OrderingConvention)
```
juliohm: What about breaking a line before the last argument to emphasize the convention?

juliohm (changes requested):
The last minor details before we can go ahead and merge :)

src/conventions.jl
```diff
@@ -9,12 +9,13 @@ from a list of vertices.
...
 """
-    connectivities(polytope, convention, k)
+    connectivities(polytope, rank, convention)
 
 Return the list of `Connectivity` elements used to construct all
 faces of rank `rank` of a polytope, following an ordering convention.
 """
-function connectivities(polytope::Type{<:Polytope}, rank::Integer, convention::OrderingConvention)
+function connectivities(polytope::Type{<:Polytope}, rank::Integer,
+                        convention::OrderingConvention)
```
serenity4, line 18: Like that?

juliohm (approved):

src/conventions.jl
juliohm: Yep! Thanks! :)

PR #20, opened 2020-12-12
add boundingbox test

shreyagupta30:
Added test for bounding box function

### DIFF
--- a/test/boundingboxes.jl
+++ b/test/boundingboxes.jl
@@ -1,3 +1,4 @@
 @testset "Bounding boxes" begin
-  @test boundingbox((0,0),1) == (-1,1)
+  @test boundingbox((0,0),1) == ((-1,-1),(1,1)
+  @test boundingbox((1,1),1) == ((0,0),(2,2))
 end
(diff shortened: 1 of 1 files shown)


test/boundingboxes.jl
```diff
@@ -1,3 +1,4 @@
 @testset "Bounding boxes" begin
-  # TODO
+  @test boundingbox((0,0),1) == ((-1,-1),(1,1))
+  @test boundingbox((1,1),1) == ((0,0),(2,2))
```
juliohm: @shreyagupta30 what you are trying to do here?

shreyagupta30: I tried to test the bounding box function with two spheres. But Travis is showing me an error that function doesn't exist.

serenity4: You first have to create a Sphere for that, with `Sphere((x,y), 1)`, and the result type of the bounding box is a `Box`. So you'd need something like the following (for your first line):
```julia
boundingbox(Sphere((0,0),1)) == Box{2,Int}(Point(-1,-1),Point(1,1))
```

shreyagupta30: Thanks a lot @serenity4 . You made my life easy. :+1:

juliohm (approved):

juliohm:
Thank you!

PR #26, opened 2021-01-22
Add sideof utility

juliohm:
@serenity4 what do you think of this name and implementation? Appreciate if you can review it.


serenity4 (approved):
Looks good! `sideof` seems indeed better than `is[left/right]of`, in particular when the point is not right nor left but on.

juliohm:
Awesome! Thanks again for the prompt review. Merging it 👌🏽

Issue #29, opened 2021-01-27
Segment bug?


henry2004y:
Hi,

I am confused by the `segment` function results:
```
using Meshes

s1 = Segment((1.5,1.5), (3.0,1.5))
s2 = Segment((1.5,1.5), (3.5,1.5))
s3 = Segment((3.0,1.0), (2.0,2.0))

s1 ∩ s3  # Point(2.4, 1.5), wrong
s2 ∩ s3  # Point(2.5, 1.5), correct
```
Is it a bug? Sorry if I made a mistake here; too tired to think...

henry2004y:
Sorry I didn't check; it was already reported #27.

serenity4:
Hi, thanks for signaling the error. This is actually different. #30 fixes #27 but not this case.

juliohm:
Thank you @henry2004y for reporting this 🐞. In this case the correct output of `s1 ∩ s3` should be `Point(2.5, 1.5)`.

PR #30, opened 2021-01-28
Use atan for angle computations

### LINKED ISSUE #27: Wrong intersection of segments
I found an example where the intersection algorithm is failing:

```julia
julia> s1 = Segment((2.,1.), (1.,2.))
Segment{2,Float64}
  └─Point(2.0, 1.0)
  └─Point(1.0, 2.0)

julia> s2 = Segment((1.,0.), (1.,1.))
Segment{2,Float64}
  └─Point(1.0, 0.0)
  └─Point(1.0, 1.0)

julia> s1 ∩ s2
Point(1.0, 2.0)
```

I've read the paper multiple times, and yet I can't figure out what is wrong with the implementation. The algorithm implemented in `intersecttype` enters in case (II) as expected, but then it branches on the wrong intersection type at the last minute based on the non-zero winding number. I did the calculation by hand and all winding numbers are making sense. Also followed the steps in the VS Code debugger.

In page 332 of the paper, the author mentions the following condition inside case (II):

image

Can you spot any flaw in the implementation? Perhaps this is a typo in the paper?

Paper: https://sci-hub.se/10.1016/0167-8396(91)90019-8

cc: @serenity4

serenity4:
This is a simpler alternative to the method from _Balbes, R. and Siegel, J. 1990_ for computing angles. It relies on the two-argument version of `atan`. Also tests and fixes #27 (including the spotted `DomainError` in simple precision).

### DIFF
--- a/src/angles.jl
+++ b/src/angles.jl
@@ -8,7 +8,8 @@
 Angle ∠ABC between rays BA and BC.
 See https://en.wikipedia.org/wiki/Angle.
 
-Uses the two-argument form of `atan`. See https://en.wikipedia.org/wiki/Atan2.
+Uses the two-argument form of `atan`.
+See https://en.wikipedia.org/wiki/Atan2.
 
 ## Example
 
--- a/src/angles.jl
+++ b/src/angles.jl
@@ -14,7 +14,7 @@ See https://en.wikipedia.org/wiki/Atan2.
 ## Example
 
 ```julia
-∠(Point(0,1), Point(0,0), Point(1,0)) == π/2
+∠(Point(1,0), Point(0,0), Point(0,1)) == π/2
 ```
 """
 function ∠(A::P, B::P, C::P) where {P<:Point{2}}
--- a/src/angles.jl
+++ b/src/angles.jl
@@ -8,6 +8,8 @@
 Angle ∠ABC between rays BA and BC.
 See https://en.wikipedia.org/wiki/Angle.
 
+Uses the two-argument form of `atan`. See https://en.wikipedia.org/wiki/Atan2.
+
 ## Example
 
 ```julia
--- a/src/angles.jl
+++ b/src/angles.jl
@@ -13,12 +13,6 @@ See https://en.wikipedia.org/wiki/Angle.
 ```julia
 ∠(Point(0,1), Point(0,0), Point(1,0)) == π/2
 ```
-
-## References
-
-* Balbes, R. and Siegel, J. 1990. [A robust method for calculating
-  the simplicity and orientation of planar polygons]
-  (https://www.sciencedirect.com/science/article/abs/pii/0167839691900198)
 """
 function ∠(A::P, B::P, C::P) where {P<:Point{2}}
   BA = A - B
@@ -27,16 +21,5 @@ function ∠(A::P, B::P, C::P) where {P<:Point{2}}
   cross = BA × BC
   inner = BA ⋅ BC
 
-  R = norm(cross) / (norm(BA) * norm(BC))
-
-  # Table 1 from Balbes, R. and Siegel, J. 1990.
-  if cross ≥ 0 && inner ≥ 0
-    asin(R)
-  elseif cross ≥ 0 && inner < 0
-    π - asin(R)
-  elseif cross < 0 && inner ≥ 0
-    -asin(R)
-  elseif cross < 0 && inner < 0
-    asin(R) - π
-  end
+  atan(cross, inner)
 end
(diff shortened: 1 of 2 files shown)


juliohm:
That is great! So we don't loose generality nor the orientation of the
angles right? I am wondering why the authors decided to branch explicitly
instead of just calling atan in their paper.

serenity4:
Orientation is preserved yes, and this should generalize better since `atan` additionally handles `missing`/`Inf` cases. It's possible that at the time, this implementation of `atan` (or `atan2`, as the 2-argument form is often called) was still recent (so the authors might not know about it) or did not exist.

serenity4:
It does not fix #29 which will probably need another bug-fix PR, though the bug is likely in the intersection algorithm.

juliohm:
Thank you Cédric for this great fix. Love fixes that also reduce code and
explicit branches.

juliohm:
Is there a reference that we could cite in the docstring for the atan
method?

Also could you please refactor the test entry to read like the others where
we introduce two variables s1 and s2? At some point we may decide to
aggregate all these tests in subsets and this can facilitate refactoring it.

serenity4:
Is the formatting OK like this?

juliohm:
Perfect :) feel free to merge the PR if it is passing all tests. I'll try
to add you to the organization when I am back to my computer

src/angles.jl
```diff
@@ -8,17 +8,14 @@
 Angle ∠ABC between rays BA and BC.
 See https://en.wikipedia.org/wiki/Angle.
 
+Uses the two-argument form of `atan`.
+See https://en.wikipedia.org/wiki/Atan2.
+
 ## Example
 
 ```julia
 ∠(Point(0,1), Point(0,0), Point(1,0)) == π/2
```
juliohm: Can you also fix the typo in this example? We should swap the first and third points in order to get a positive pi/2

src/angles.jl
```diff
@@ -27,16 +24,5 @@ function ∠(A::P, B::P, C::P) where {P<:Point{2}}
...
-  if cross ≥ 0 && inner ≥ 0
-    asin(R)
-  elseif cross ≥ 0 && inner < 0
-    π - asin(R)
-  elseif cross < 0 && inner ≥ 0
-    -asin(R)
-  elseif cross < 0 && inner < 0
-    asin(R) - π
-  end
+  atan(cross, inner)
```
juliohm, line 27: Can you comment on how this atan(cross, inner) is equivalent to the previous code? I've seen [this method](https://stackoverflow.com/a/21484228) and was wondering how it relates to the above method.

serenity4: `atan` allows you to retrieve a signed angle from its sine and cosine. Here, `cross` and `inner` are the sine and cosine respectively, up to a factor of `norm(BA) * norm(BC)`. This factor is not important however, because `atan` only uses the ratio between the sine and cosine, as well as the sign of both values. The common factor is positive, so we can safely pass in the unscaled values.
Note also that `cross` is a signed value in 2D, but a vector in 3D. Therefore, for the 3D case, this would require more work: we would need to take the norm of the cross product and figure out the sign from BA and BC.

In the method you linked, the angle of both vectors from the origin is computed, and then subtracted. This will only work in 2D as well, in 3D you would have to compute a pointed plane containing both vectors and compute their angles with respect to an arbitrary origin within the plane.

In terms of speed:
```julia
julia> @benchmark ∠(Point(1,0), Point(0,0), Point(0,1)) # first method
BenchmarkTools.Trial: 
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     3.709 ns (0.00% GC)
  median time:      3.931 ns (0.00% GC)
  mean time:        3.997 ns (0.00% GC)
  maximum time:     19.856 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1000

julia> @benchmark ∠(Point(1,0), Point(0,0), Point(0,1)) # second method
BenchmarkTools.Trial: 
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     3.122 ns (0.00% GC)
  median time:      3.364 ns (0.00% GC)
  mean time:        3.570 ns (0.00% GC)
  maximum time:     24.943 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1000
```

Both methods therefore seem equivalent.

juliohm: Amazing clarification as usual @serenity4 ❤️

juliohm:
Feel free to merge @serenity4 to test your new membership in the organization :)

serenity4:
I don't have write access, although I am in the organization. Maybe a setting to tweak somewhere?

juliohm:
Oh that is interesting. Perhaps this setting is repo-based. Let me add you to the Meshes.jl project as well...

juliohm:
I've created a team called `meshes` in the organization and added you to it. You should be able to merge now. Can you please double check?

PR #31, opened 2021-01-28
Fix for intersection point calculation

### LINKED ISSUE #29: Segment bug?

henry2004y:
Hello,

I have a quick fix for the intersection point calculation, following the formula given in [Wiki](https://en.wikipedia.org/wiki/Line%E2%80%93line_intersection). All test passed locally, and it also fixed #29. Hope it works without issue.

Fixes #29

### DIFF
--- a/src/polytopes/segment.jl
+++ b/src/polytopes/segment.jl
@@ -142,9 +142,11 @@ function intersectpoint(s1::Segment{Dim,T}, s2::Segment{Dim,T}) where {Dim,T}
   v2 = y1 - y2
   v12 = x1 - y1
 
+  # Ref: https://en.wikipedia.org/wiki/Line-line_intersection#Formulas
   # find scalar along first segment
-  s = (v12[1] * v2[2] - v12[2] * v2[1]) / (v1[1] * v2[2] - v1[2] * v2[1])
-  
+  s = (v12[1] * v2[2] - v12[2] * v2[1]) /
+      (v1[1] * v2[2] - v1[2] * v2[1])
+
   x1 - s*v1
 end
 
--- a/src/polytopes/segment.jl
+++ b/src/polytopes/segment.jl
@@ -138,16 +138,12 @@ function intersectpoint(s1::Segment{Dim,T}, s2::Segment{Dim,T}) where {Dim,T}
   x1, x2 = vertices(s1)
   y1, y2 = vertices(s2)
 
-  # solve Ax = b
-  A = (y2 - y1) - (x2 - x1)
-  b = (x1 - y1)
-  s = zero(T)
-  for i in 1:length(A)
-    if !iszero(A[i])
-      s = b[i] / A[i]
-      break
-    end
-  end
+  tmp = y1 - y2
+
+  s = ((x1.coords[1] - y1.coords[1]) * tmp[2] -
+       (x1.coords[2] - y1.coords[2]) * tmp[1]) /
+      ((x1.coords[1] - x2.coords[1]) * tmp[2] -
+       (x1.coords[2] - x2.coords[2]) * tmp[1])
 
   x1 + s*(x2 - x1)
 end
--- a/src/polytopes/segment.jl
+++ b/src/polytopes/segment.jl
@@ -138,13 +138,12 @@ function intersectpoint(s1::Segment{Dim,T}, s2::Segment{Dim,T}) where {Dim,T}
   x1, x2 = vertices(s1)
   y1, y2 = vertices(s2)
 
-  tmp = y1 - y2
+  seg1 = x1 - y1
+  seg2 = x1 - x2
+  seg3 = y1 - y2
 
-  s = ((x1.coords[1] - y1.coords[1]) * tmp[2] -
-       (x1.coords[2] - y1.coords[2]) * tmp[1]) /
-      ((x1.coords[1] - x2.coords[1]) * tmp[2] -
-       (x1.coords[2] - x2.coords[2]) * tmp[1])
-
+  s = (seg1[1] * seg3[2] - seg1[2] * seg3[1]) /
+      (seg2[1] * seg3[2] - seg2[2] * seg3[1])
   x1 + s*(x2 - x1)
 end
 
--- a/src/polytopes/segment.jl
+++ b/src/polytopes/segment.jl
@@ -142,8 +142,8 @@ function intersectpoint(s1::Segment{Dim,T}, s2::Segment{Dim,T}) where {Dim,T}
   v2 = y1 - y2
   v12 = x1 - y1
 
-  # Ref: https://en.wikipedia.org/wiki/Line-line_intersection#Formulas
-  # find scalar along first segment
+  # the intersection point lies in between x1 and x2 at a fraction s
+  # (https://en.wikipedia.org/wiki/Line-line_intersection#Formulas)
   s = (v12[1] * v2[2] - v12[2] * v2[1]) /
       (v1[1] * v2[2] - v1[2] * v2[1])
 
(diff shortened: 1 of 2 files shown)


juliohm:
Thank you @henry2004y for the proposal fix! Can you please comment on the exact method you have implemented from the Wiki? it would be nice to add some references and to polish the code afterwards.

Please make sure that you can communicate the fix clearly to us first, and then we can provide a careful review of your changes.

henry2004y:
The proposed implementation of finding the intersection point in `intersectpoint(s1::Segment{Dim,T}, s2::Segment{Dim,T}) where {Dim,T}` works as follows:
1. The higher level function can already tell that the two line segments `s1` and `s2` intersect at a point or are colinear.
2. The function `vertices` is called to obtain the end points of the two segments:
```julia
x1, x2 = vertices(s1)
y1, y2 = vertices(s2)
```
3. Then we need to solve a linear equation set. Denote the intersection point as P. Let `s` be the ratio of lengths x1P to x2P, `t` be the ratio of lengths y1P to y2P. Then the equations can be written as

where I the first index of each point is the x coordinate and the second index is the y coordinate. 

`s` can then be expressed as

and the intersection point P is simply `x1 + s*(x2 - x1)`.

4. This should work not only for 2D, but also higher dimensions, since it is already guaranteed for a single point intersection.

I noticed in `intersecttype(s1, s2)` that 5 different scenarios are considered.  I think with the proposed approach type 1-3 are all covered. However, given the floating point numerical error, it is safer to keep type 2 and 3.

Besides, this is so simple math that I don't know what to add to the code comments. Actually I think it may be better to rename `x1,x2,y1,y2` to something like `p1,p2,p3,p4` so as not to confuse points with coordinates. I hesitated because the other function `intersectcolinear` follows the same names.

juliohm:
Thank you @henry2004y for the nice clarifications 👌🏽 

> this is so simple math that I don't know what to add to the code comments

There is no such thing as too simple math :) we can always benefit from a good explanation.

> I think it may be better to rename `x1,x2,y1,y2` to something like `p1,p2,p3,p4` so as not to confuse points with coordinates

I considered that when I was writing these variable names, but ended up choosing xs and ys because of the paper cited. I am trying to follow the notation in the paper so that future contributors know what the algorithm is doing.

I will now provide a careful review of your PR so that we can move forward and merge it. 💯

src/polytopes/segment.jl
```diff
@@ -138,16 +138,12 @@ function intersectpoint(s1::Segment{Dim,T}, s2::Segment{Dim,T}) where {Dim,T}
...
-      s = b[i] / A[i]
-      break
-    end
-  end
+  tmp = y1 - y2
+
+  s = ((x1.coords[1] - y1.coords[1]) * tmp[2] -
+       (x1.coords[2] - y1.coords[2]) * tmp[1]) /
+      ((x1.coords[1] - x2.coords[1]) * tmp[2] -
+       (x1.coords[2] - x2.coords[2]) * tmp[1])
```
juliohm: It seems that the proposed formula only depends on point differences of the form (x1 - y1), (x1 - x2), ... In Meshes.jl you can obtain these differences as you did with `tmp = y1 - y2`. Please try to rewrite this expression so that all differences are computed a priori and then use the function `coordinates(y1 - y2)` to extract the corresponding coordinates in the reference system. We don't use `x.coords` because that is internal undocumented behavior. It is not part of the public API of points, makes sense?

After you fix the expression, make sure the PR also includes new tests for the case you reported in #29. This way we will never reintroduce the bug.

juliohm (changes requested):
@henry2004y kindly consider these changes.

src/polytopes/segment.jl
henry2004y: I cleaned up the expression a little bit more following your advice and added the test case.

src/polytopes/segment.jl
```diff
@@ -138,13 +138,12 @@ function intersectpoint(s1::Segment{Dim,T}, s2::Segment{Dim,T}) where {Dim,T}
...
-  s = ((x1.coords[1] - y1.coords[1]) * tmp[2] -
-       (x1.coords[2] - y1.coords[2]) * tmp[1]) /
-      ((x1.coords[1] - x2.coords[1]) * tmp[2] -
-       (x1.coords[2] - x2.coords[2]) * tmp[1])
+  seg1 = x1 - y1
+  seg2 = x1 - x2
+  seg3 = y1 - y2
 
+  s = (seg1[1] * seg3[2] - seg1[2] * seg3[1]) /
+      (seg2[1] * seg3[2] - seg2[2] * seg3[1])
```
juliohm: Can you please reorder these three variables so that the first two are related to a single segment and the third is related to a "cross" term? Also, maybe we can simply call these variables `v1, v2, v12` to reflect the fact that they are vectors connecting specific vertices from segments 1 and 2.

juliohm (changes requested):

src/polytopes/segment.jl
henry2004y: Sure

src/polytopes/segment.jl
```diff
@@ -138,17 +138,11 @@ function intersectpoint(s1::Segment{Dim,T}, s2::Segment{Dim,T}) where {Dim,T}
...
-      s = b[i] / A[i]
-      break
-    end
-  end
+  v1 = x1 - x2
+  v2 = y1 - y2
+  v12 = x1 - y1
 
+  s = (v12[1] * v2[2] - v12[2] * v2[1]) / (v1[1] * v2[2] - v1[2] * v2[1])
   x1 + s*(x2 - x1)
```
```suggestion
  # find scalar along first segment
  s = (v12[1] * v2[2] - v12[2] * v2[1]) / (v1[1] * v2[2] - v1[2] * v2[1])
  
  x1 - s*v1
```

juliohm (changes requested):
Awesome 💯 Now please add some comment with links to wikipedia explaining the method used. Specifically, those equations you explained in the thread where you are solving just one of them for `s`.

src/polytopes/segment.jl
```diff
@@ -138,18 +138,14 @@ function intersectpoint(s1::Segment{Dim,T}, s2::Segment{Dim,T}) where {Dim,T}
...
-    end
-  end
-
-  x1 + s*(x2 - x1)
+  v1 = x1 - x2
+  v2 = y1 - y2
+  v12 = x1 - y1
+
+  # find scalar along first segment
+  s = (v12[1] * v2[2] - v12[2] * v2[1]) / (v1[1] * v2[2] - v1[2] * v2[1])
```
juliohm: Please break this line into two lines with numerator and denominator. After that and the comment with links to wikipedia we are ready to merge 🎉

henry2004y:
Thanks for the review! I've learned a lot through the process.

src/polytopes/segment.jl
```diff
@@ -138,18 +138,16 @@ function intersectpoint(s1::Segment{Dim,T}, s2::Segment{Dim,T}) where {Dim,T}
...
-      s = b[i] / A[i]
-      break
-    end
-  end
+  v1 = x1 - x2
+  v2 = y1 - y2
+  v12 = x1 - y1
+
+  # Ref: https://en.wikipedia.org/wiki/Line-line_intersection#Formulas
+  # find scalar along first segment
```
```suggestion
  # the intersection point lies in between x1 and x2 at a fraction s
  # (https://en.wikipedia.org/wiki/Line-line_intersection#Formulas)
```

juliohm (approved):

juliohm:
Amazing @henry2004y ! ❤️ I will submit a patch release now with the recent fixes to the intersection algorithm.

PR #35, opened 2021-02-03
Hopefully fix a recursion in Point constructors

### LINKED ISSUE #34: Point embedded in R^1
Hi,

I found another potential bug for allocating a `CartesianGrid` object. Following the [example in the document](https://juliageometry.github.io/Meshes.jl/dev/meshes.html#Meshes.CartesianGrid)
```
a = CartesianGrid((-1.,),(1.,), dims=(100,))
```
This operation can never finish, no matter what values I input for `dims`. I also tried
```
a = CartesianGrid((100,),(10.,),(1.,))
```
Same issue I guess.

FPGro:
Fix #34 

The proposed change replaces the outer parametrized Varargs constructor with the form requiring at least one argument, which should make this constructor less specific than the inner constructor for the signature `Point{Dim,T}(::SVector)`, which lead to an infinite recursion on itself before.


src/points.jl
```diff
@@ -25,7 +25,7 @@ end
 
 # convenience constructors
 Point{Dim,T}(coords::NTuple{Dim,V}) where {Dim,T,V} = Point{Dim,T}(SVector(coords))
-Point{Dim,T}(coords::Vararg{V,Dim}) where {Dim,T,V} = Point{Dim,T}(SVector(coords))
+Point{Dim,T}(coords_head::V, coords_tail::Vararg{V}) where {Dim,T,V} = Point{Dim,T}(SVector(coords_head, coords_tail...))
```
```suggestion
Point{Dim,T}(coord::V, coords::Vararg{V}) where {Dim,T,V} = Point{Dim,T}(SVector(coord, coord...))
```

juliohm (changes requested):
Thank you @FPGro for the proposed fix! The tests are all passing. Could you also add a new test constructing 1D points to see if it also passes? :)

juliohm:
@FPGro did you have a chance to take a look at this?

FPGro:
I accidentally started a review instead of answering to your comment :D I'll copy it here: @juliohm 

I'm on it. While `Point(0.1)` does work as intended, `Point{1,Float64}(0.1)` (with the proposed change) is now less specific than the default constructor and thus unreachable for 1D input.

However you view it, `Point{Dim,T}(::Vararg{V})` is directly interfering with the default inner constructor (for 1D input), which implicitly tries to convert whatever only argument it gets to a SVector. I think you have to decide how this case should be resolved:

- treating `Point{D,T}(arg1)` like `Point{D,T}(arg1, arg2, ...)` would be the most consistent option. There might be cases where `convert(SVector, arg)` is defined but not `SVector(arg)`, which would cease to function, but I doubt that. To do so, we could make the inner constructor explicit with narrower type like `::AbstractArray`, which should catch all returned values from the outer constructors but not a single number.
- keeping the default constructor behaviour of `Point{D,T}(arg1)` would be less of a change, but it would be inconsistent with `Point{D,T}(arg1, arg2, ...)` ( `Point{2,Int}(1,2)` would work but `Point{1,Int}(1)` would throw an error )

On a sidenote, should I introduce `Point1` and `Point1f` as aliases similar to the existing?

Edit: I already have some tests in place im a branch of the branch on my fork ^^ I just wanted to sort out this general issue first

juliohm:
That is a very careful analysis @FPGro , thanks for helping with this. I am trying to work my head around the design based on your comments.

juliohm:
I will open a thread on Discourse to gather more thoughts on this design issue. It is surprising to me that the Vararg method is interfering with the inner constructor.

FPGro:
I implemented the changes discussed in the discourse thread:
https://discourse.julialang.org/t/ambiguity-with-vararg-outer-constructor-and-inner-constructor/54622/7
Also added the shorthands for 1D Points and Vecs and exported them. Tests will follow

FPGro:
This is odd, construction fails for mixed inputs only

```julia
julia> Point{2,Float64}(0,0)
Point{2, Float64}([0.0, 0.0])

julia> Point{2,Float64}(0.,0)
ERROR: DimensionMismatch("Can't construct a Point{2,Float64} with an input of length 1")
Stacktrace:
 [1] Point{2, Float64}(coords::Tuple{Tuple{Float64, Int64}})
   @ Main ~\Desktop\meshes.jl:17
 [2] Point{2, Float64}(coords::Tuple{Float64, Int64})
   @ Main ~\Desktop\meshes.jl:10
 [3] top-level scope
   @ REPL[15]:1

julia> Point{2,Float64}((0.,0))
ERROR: DimensionMismatch("Can't construct a Point{2,Float64} with an input of length 1")
Stacktrace:
...
```

FPGro:
Okay, well that's reasonable too! One would have to make a better error message then and remove all implicit conversions from the test sets that fail. 
Oh and we'd have to catch this thing somehow:
```julia
julia> Point(1,1.)
Point{1, Tuple{Int64, Float64}}([(1, 1.0)])
```

You're welcome.

juliohm:
So there are old tests failing? That is actually strange! I will take a look tomorrow. I thought I was not using mixed coordinates anywhere.

FPGro:
Probably an oversight then, I think it's only here 
https://github.com/FPGro/Meshes.jl/blob/39a845d53272ac6c07a8fd7fa3fab97f827dc2ee/test/polytopes.jl#L101
and in the same file on line 104. 👍

FPGro:
Just added some tests for 1D Points to the set + tests that check, that implicit conversions are disallowed but those are commented out at the moment

test/points.jl
```diff
@@ -1,51 +1,86 @@
...
   @test coordtype(Point2(1, 1)) == Float64
   @test coordtype(Point3(1, 1, 1)) == Float64
+  @test coordtype(Point1f(1)) == Float32
   @test coordtype(Point2f(1, 1)) == Float32
   @test coordtype(Point3f(1, 1, 1)) == Float32
 
   @test coordtype(Point{2,T}((1, 1))) == T
   @test coordtype(Point{2,T}(1, 1)) == T
 
+  @test coordinates(P1(1)) == [1.0]
```
juliohm:
```suggestion
  @test coordinates(P1(1)) == T[1]
```

src/points.jl
```diff
@@ -17,28 +17,35 @@ O = Point(0.0, 0.0) # origin of 2D Euclidean space
...
 """
 struct Point{Dim,T}
   coords::SVector{Dim,T}
+  Point{Dim,T}(coords::AbstractVector) where {Dim,T} = new{Dim,T}(coords)
 end
 
 # convenience constructors
 Point{Dim,T}(coords::NTuple{Dim,V}) where {Dim,T,V} = Point{Dim,T}(SVector(coords))
-Point{Dim,T}(coords::Vararg{V,Dim}) where {Dim,T,V} = Point{Dim,T}(SVector(coords))
+Point{Dim,T}(coords...) where {Dim,T} = Point{Dim,T}(coords)
```
juliohm: So we need to remove the explicit `Vararg{V,Dim}` specification here? Could we preserve it?

test/points.jl
```diff
@@ -1,51 +1,86 @@
...
+  @test_throws DimensionMismatch Point{1,Int}((2,3))
+  @test_throws DimensionMismatch Point{-3,T}([4,5,6])
+    
+  # disabled for now, check that implicit conversion is disallowed and throws
+  # @test_throws ... Point(1, .2) # not sure what to throw yet
+  # @test_throws ... Point((3., 4))
+  # @test_throws ... Point([5., 6., 7])
+  # @test_throws ... Point{2,T}(8, 9.)
+  # @test_throws ... Point{3,Int}((-1., -2))
+  # @test_throws ... Point{4,T}([0, -1., +2, -4.])
```
juliohm, line 84: That is awesome that you are so careful with tests ❤️ ❤️ I think we can leave these without tests for now, and assume that the error will be cryptic but easily fixable.

juliohm (changes requested):
This is such a lovely PR! ❤️ Thank you very much for putting the effort and for such careful additional tests. That is simply amazing.

Can you check the small comments I left? I am happy to merge immediately after we make sure everything is in right shape. 💯

FPGro:
> So we need to remove the explicit Vararg{V,Dim} specification here? Could we preserve it?

I think that led to an infinite recursion with tuples, but I can try again if that was resolved with the other changes. :) I'll report back.

This simple constructor setup is really getting more complicated than I thought ^^

FPGro:
Okay @juliohm I re-added the `::Vararg{V}` annotations without `Dim` and that seems to work fine. I think this works now and we can worry about the error consistency later.

For reference: check if this solves #34

juliohm (approved):
That is an amazing contribution @FPGro , thanks again for your work.

@serenity4 do you have anything to add or we can merge it right away?

serenity4 (approved):
Awesome work! Nothing comes to my mind, the changes look very complete.

src/points.jl
```diff
@@ -17,28 +17,35 @@ O = Point(0.0, 0.0) # origin of 2D Euclidean space
 
 ### Notes
 
-- Type aliases are `Point2`, `Point3`, `Point2f`, `Point3f`
+- Type aliases are `Point1`, `Point2`, `Point3`, `Point1f`, `Point2f`, `Point3f`
 """
 struct Point{Dim,T}
   coords::SVector{Dim,T}
+  Point{Dim,T}(coords::AbstractVector) where {Dim,T} = new{Dim,T}(coords)
```
juliohm, line 24: Just a quick question, could we also write this as follows?

```julia
Point{Dim,T}(coords) where {Dim,T} = new(coords)
```

I tend to omit the type annotation in inner constructors, I don't know if it makes a difference.

FPGro: Me too, but it is intentional here. The inner constructor has to have more specific type requirement than the varargs constructor so we don't run into the infinite recursion again.

juliohm (approved):

juliohm:
Amazing @FPGro ! Thanks again for your work! Really welcome!

FPGro:
You're welcome! I'm gonna put up a new issue just to keep a reference for the inconsistencies with mixed types and errors. Feel free to ping me if you have an interesting issue again ^^

Issue #40, opened 2021-02-17
Benchmarking infrastructure


serenity4:
As discussed in #37, we could benefit from benchmarks for:
- informing the users of the costly and cheap functions
- tracking performance regressions during development

It can be implemented first as a simple `benchmark` folder, and possibly added as a GitHub action is desired.

juliohm:
Fixed on master ❤️

Arkoniak:
This issue looks closed, but still it may be of interest to you. You can use https://github.com/tkf/BenchmarkCI.jl to make automated benchmarking on pull requests. We used it in ParallelKMeans.jl and it was rather handy. You need benchmarks.yml in github_action folder, like here: https://github.com/PyDataBlog/ParallelKMeans.jl/blob/master/.github/workflows/benchmarks.yml

and `benchmark` directory with necessary scripts, like here https://github.com/PyDataBlog/ParallelKMeans.jl/tree/master/benchmark

Then you can either build benchmarks locally and in judge mode compare benchmarks to the master or get results as a part of PR, like here: https://github.com/PyDataBlog/ParallelKMeans.jl/pull/101#issuecomment-784559581

serenity4:
Yes, this is exactly what we did in #41 and #50. Thank you for pointing it out though, this is a very handy package.

PR #41, opened 2021-02-17
[WIP] Add benchmarks

serenity4:
Addresses #40. I added an initial benchmarking infrastructure (following the [BenchmarkTools](https://github.com/JuliaCI/BenchmarkTools.jl) recommendations) to be completed over time.

To run it, make sure you have `BenchmarkTools` in your environment and then run `julia --project=. benchmark/benchmarks.jl`

For now it produces:

image

Some ideas:
- [ ] populate the benchmarks with great inspiration from the testing suite and larger operations (suggestions are welcome).
- [ ] write these results in a file to be included elsewhere (in the README?)


juliohm:
Awesome to see a benchmark suite! I remember people integrating these benchmarks with GitHub Actions so that every PR triggers a benchmark table comparing the before and after. Are you also aware of this feature?

As a side question, why the non-macro versions of btime in the benchmarks.jl file? Is this the recommended approach to write benchmarks currently?

serenity4:
> I remember people integrating these benchmarks with GitHub Actions so that every PR triggers a benchmark table comparing the before and after. Are you also aware of this feature?

Yes, that would be ideal so that performance differences can be caught _before_ merging PRs and without effort.

> As a side question, why the non-macro versions of btime in the benchmarks.jl file? Is this the recommended approach to write benchmarks currently?

It was simply so that I could define expressions, and then benchmark them with a function. With macros it was not possible AFAIK, because it depends only on compile-time values. So, for example, I couldn't do
```julia
ex = :(Point(0., 1.))
@btime ex # no computation done
@btime $ex # no computation done either
@btime Point(0., 1.) # need to hardcode it with the macro, output not controlled
timing = btime(ex) # custom function version; benchmarks the expression exactly as if run with @btime Point(0., 1.)
```

Ultimately, the BenchmarkTools function needs an `Expr`, and the macro is used to get you that. But a macro is unnecessary in this case, you just need an `Expr`, so a function version is fine and allows you to construct your benchmarks with expressions not necessarily known at compile time. Also, `@btime` just prints to `stdout`, and I preferred to build the output myself (adding the `Expr` being benchmarked and colorizing the result).

In short: the macro version `@btime` is very handy for quick benchmarks, but is fairly limited if you want to organize them and output them differently.

juliohm:
I found the CI package: https://github.com/tkf/BenchmarkCI.jl

It looks like there is also a specialized package PkgBenchmark.jl that helps with writing benchmarks for packages. What do you think? I didn't have a chance to read carefully, but looks promising and may clean up a bit the draft.

FPGro:
Great work so far!
The CI packages look interesting, if experimental still. If you want to do this, I'd suggest making them a manual action because when you want to properly benchmark the package, this may get fairly complex and time-expensive, so you don't want to run it on every commit to a PR. 

Sidenote: you can also construct benchmarks at runtime with the macro version, but that definitely looks less clean than your version! From the BenchmarkTools docs:
```julia
for f in (sin, cos, tan)
    for x in (0.0, pi)
        suite["trig"][string(f), x] = @benchmarkable $(f)($x)
        suite["dot"][string(f), x] = @benchmarkable $(f).([$x, $x, $x])
    end
end
```

serenity4:
Things are clearer now and use BenchmarkGroup structures, printing about the same as previously (up to `"Points"` becoming `"points"` as a group name being printed).

From there minimal work is required for adding the BenchmarkCI workflow which I think could be very nice. I believe all there needs to be is to create a workflow file as described in their README.

Also note that the `Project.toml` and `Manifest.toml` are required by BenchmarkCI, which is why I added them.

juliohm:
I will be so awesome to have benchmarks running in the PRs automatically, or triggered upon request.

I will try to run the benchmarks locally over the weekend. It would be nice to create a set of files for different sections of the codebase like we do in the test suite. It helps in the long run maintaining so much code. For example, I am only able to migrate stuff from GeoStatsBase.jl to here because of this pedantic modular organization of the source 🙏🏽

benchmark/benchmarks.jl
```diff
@@ -0,0 +1,85 @@
...
+    foreach(group.data) do (str, val)
+      if val isa Trial
+        print(' '^2, rpad(str, nchars_left), " :  ")
+        display_trial(val)
+      else
+        println("\e[34;1;1m", str, "\e[m")
+        display_group(val)
+      end
+    end
+end
```
juliohm: Can we place these helper functions in a separate source file like `benchmark/utils.jl`?

benchmark/benchmarks.jl
```diff
@@ -0,0 +1,85 @@
...
+suite = BenchmarkGroup()
+
+suite["points"] = BenchmarkGroup(["microbenchmark", "points"]) # add tags for eventual filtering
+
+add_benchmark!(suite["points"], [
+  :(Point(0., 1.)),
+  :(Point([0., 1.])),
+  :(Point([0., 1])),
+  :(Point(0., convert(Float64, 1))),
+])
```
juliohm: Similarly, can we move these benchmarks about points to a separate source file like `benchmark/points.jl`?

benchmark/benchmarks.jl
```diff
@@ -0,0 +1,85 @@
...
+
+println("Julia version is $VERSION")
+t = @elapsed using Meshes
```
juliohm:
```suggestion
t = @elapsed using Meshes
println("Julia version is $VERSION")
```

benchmark/benchmarks.jl
```diff
@@ -0,0 +1,85 @@
...
+println(string("Meshes.jl loading time: \e[33;1;1m$t\e[m seconds"))
+println()
+println("Benchmarking Meshes.jl...")
+println()
+
+tune!(suite)
+result = run(suite)
+
```
juliohm:
```suggestion
```

juliohm (changes requested):
That is starting to look great 💯 Let me know when you have the CI infrastructure sorted out.

serenity4:
Getting CI to work was a bit tricky. To make it short, we won't have a passing PR here, because no baseline exists on master (`benchmark/benchmarks.jl` does not exist). See tkf/BenchmarkCI.jl#89 for a similar case. Exercising my Git skills, I could setup a preview on [a PR](https://github.com/serenity4/Meshes.jl/pull/2) on my fork that contains all but the last few commits of this PR on its master branch. That way we can see the results in a working setup.

We currently have two ways of handling the posting of benchmark results: include the report in a comment, or just print it on a CI run as in [this action](https://github.com/serenity4/Meshes.jl/pull/2/checks?check_run_id=1964635263). IMO navigating to CI to check it not that straightforward, so why not use the comment form. Please let me know what you think.

The action can be triggered on specific criteria, here it triggers on all PRs with the `run benchmark` label (that I just created for this occasion to test it). I don't know if it is possible to trigger it explicitly from a comment on a PR, I'll look into that if you think it may be preferable.

juliohm:
I am lacking words to describe my amazement here 😍 Thank you for making it happen. It is really nice to see this working!

juliohm:
The PR got mixed up with other changes in the master branch. Do you think you could start a fresh one from the latest master instead? Sorry for the noise in master, it is been a busy week to get things working quickly before the end of the month.

Issue #37, opened 2021-02-06
Low priority: consistency with implicit conversion in Point constructors


FPGro:
This is a reference to small inconsistencies that were left over from the #35 fix. Point constructors for all dimensions work now, but a small inconsistency remains with mixed-type inputs: Implicit conversion is disallowed for some constructors but not all:

```julia
julia> Point(0, 1.) # this is good
ERROR: MethodError: no method matching Point(::Int64, ::Float64)
Closest candidates are:
  Point(::V...) where V at REPL[7]:1
...

julia> Point((0, 1.)) # this is especially undesirable
Point{1, Tuple{Int64, Float64}}([(0, 1.0)])

julia> Point([0, 1.]) # implicit conversion
Point{2, Float64}([0.0, 1.0])

julia> Point{2,Float64}(0, 1.) # good
ERROR: MethodError: no method matching Point{2, Float64}(::Int64, ::Float64)
Closest candidates are:
  Point{Dim, T}(::V...) where {Dim, T, V} at REPL[5]:1
...

julia> Point{2,Float64}((0, 1.)) # needs a better error message
ERROR: DimensionMismatch("Can't construct a Point{2,Float64} with an input of length 1")
...

julia> Point{2,Float64}([0, 1.]) # implicit conversion
Point{2, Float64}([0.0, 1.0])
```

One should decide to either allow or forbid implicit conversion in all Point constructors and enforce that scheme. (But this is really not a pressing issue)

BTW: A few tests were created but disabled until consistency is enforced:
https://github.com/JuliaGeometry/Meshes.jl/blob/c010bab3913fe5a88466ad0948d0dcb1e0b4b83f/test/points.jl#L79

juliohm:
Amazing @FPGro , thanks for opening the issue. Feel free to tackle it if you have ideas. 💯

juliohm:
@FPGro do you have a suggestion to fix this one? I am working actively in other fronts of the codebase and help with these constructors would be awesome. You are the master of constructors 🥇

FPGro:
@juliohm Sure, though I leave it as a first issue for somebody else but it shouldn't be too hard, I can try it.

What's the decision on behavior? Forbid implicit conversion everywhere? And should there be restrictions on the element type of `Point`?
I'm especially thinking about weird edge cases like `Point{N, Tuple{Int, Int}}` or so, should those be allowed?

juliohm:
[...]

PR #42, opened 2021-02-18
Issue #37, allow mixed inputs again

FPGro:
And also aim for higher performance.
TODO:
- ~~check behaviour of invalid input~~ fine
- ~~check `Base.convert`s again~~ are good as they are
- ~~check tests~~ related tests pass

### DIFF
--- a/src/points.jl
+++ b/src/points.jl
@@ -21,20 +21,18 @@ O = Point(0.0, 0.0) # origin of 2D Euclidean space
 """
 struct Point{Dim,T}
   coords::SVector{Dim,T}
-  Point{Dim,T}(coords::AbstractVector) where {Dim,T} = new{Dim,T}(coords)
+  Point{Dim,T}(coords::SVector) where {Dim,T} = new{Dim,T}(coords)
 end
 
 # convenience constructors
-Point{Dim,T}(coords::NTuple{Dim,V}) where {Dim,T,V} = Point{Dim,T}(SVector(coords))
-Point{Dim,T}(coords::Vararg{V}) where {Dim,T,V} = Point{Dim,T}(coords)
-Point(coords::NTuple{Dim,T}) where {Dim,T} = Point{Dim,T}(SVector(coords))
-Point(coords::Vararg{V}) where {V} = Point(coords)
-Point(coords::AbstractVector{T}) where {T} =
-  Point{length(coords),T}(SVector{length(coords)}(coords))
+Point{Dim,T}(coords...) where {Dim,T} = Point{Dim,T}(SVector{Dim,T}(coords...))
+Point(coords::SVector{Dim,T}) where {Dim,T} = Point{Dim,T}(coords)
+Point(coords::AbstractVector{T}) where {T} = Point{length(coords),T}(coords)
+Point(coords...) = Point(SVector(coords...))
 
 # catches mismatching tuple dimensions
-Point{Dim1,T}(::NTuple{Dim2}) where {T,Dim1,Dim2} = 
-  throw(DimensionMismatch("Can't construct a Point{$(Dim1),$(T)} with an input of length $(Dim2)"))
+#Point{Dim1,T}(::NTuple{Dim2}) where {T,Dim1,Dim2} = 
+#  throw(DimensionMismatch("Can't construct a Point{$(Dim1),$(T)} with an input of length $(Dim2)"))
 
 # coordinate type conversions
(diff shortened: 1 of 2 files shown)


FPGro:
Cool, works surprisingly well:
```
Points: Test Failed at /home/runner/work/Meshes.jl/Meshes.jl/test/points.jl:75
  Expression: Point{1, Int}((2, 3))
    Expected: DimensionMismatch
      Thrown: MethodError
```
This is actually fine, just a different error type because I don't catch it manually now but let `SVector` do the work.

juliohm:
That is awesome. One of the errors in the CI is unrelated, trying to fix it in the master branch.

Printing single precision floats is incosistent:

```julia
julia> t = (1.0f0, 1.0f0)
(1.0f0, 1.0f0)

julia> println(stdout, t)
(1.0f0, 1.0f0)

julia> println(stdout, t[1])
1.0
```

Trying to figure out if I should open an issue in Julia.

FPGro:
Uhm, the issue appears to be here: 
https://github.com/JuliaGeometry/Meshes.jl/blob/0b765970714167218ecc8c247ad7a888246d33b3/test/traits.jl#L27
and probably the line 2 below, too.
```
3 DummyDomain{2,Float32}
  └─Ball{2,Float32}(Point(1.0f0, 1.0f0), 1.0f0)
  └─Ball{2,Float32}(Point(2.0f0, 2.0f0), 1.0f0)
  └─Ball{2,Float32}(Point(3.0f0, 3.0f0), 1.0f0)
# vs:
3 DummyDomain{2,Float32}
  └─Ball{2, Float32}(Point(1.0f0, 1.0f0), 1.0f0)
  └─Ball{2, Float32}(Point(2.0f0, 2.0f0), 1.0f0)
  └─Ball{2, Float32}(Point(3.0f0, 3.0f0), 1.0f0)
```

I don't have the slightest clue how the constructors changed *that* behaviour. @juliohm can you look at this and maybe fix the test? :) 

PS: and maybe delete that disabled thing here, I forgot to clean it up, sorry ^^ I'm off for a couple of hours now
https://github.com/FPGro/Meshes.jl/blob/0d40bb827c43d7a9c4754448a90b1ce8ba7b6e21/src/points.jl#L34

juliohm:
I meant that this error is not related to your work. It is a failure caused by the printing issue illustrated above. I introduced this test without knowing that single precision floats print as if they were double precision.

FPGro:
> I meant that this error is not related to your work. It is a failure caused by the printing issue illustrated above. I introduced this test without knowing that single precision floats print as if they were double precision.

Yeah sorry I wrote my comment before I read yours. If that's just a printing issue, then it should be ready for review I guess, all other tests pass. :)

src/points.jl
```diff
@@ -21,20 +21,18 @@ O = Point(0.0, 0.0) # origin of 2D Euclidean space
...
+Point{Dim,T}(coords...) where {Dim,T} = Point{Dim,T}(SVector{Dim,T}(coords...))
+Point(coords::SVector{Dim,T}) where {Dim,T} = Point{Dim,T}(coords)
+Point(coords::AbstractVector{T}) where {T} = Point{length(coords),T}(coords)
+Point(coords...) = Point(SVector(coords...))
 
 # catches mismatching tuple dimensions
-Point{Dim1,T}(::NTuple{Dim2}) where {T,Dim1,Dim2} = 
-  throw(DimensionMismatch("Can't construct a Point{$(Dim1),$(T)} with an input of length $(Dim2)"))
+#Point{Dim1,T}(::NTuple{Dim2}) where {T,Dim1,Dim2} = 
+#  throw(DimensionMismatch("Can't construct a Point{$(Dim1),$(T)} with an input of length $(Dim2)"))
```
juliohm: Should this be removed as opposed to commented out?

juliohm (changes requested):

src/points.jl
FPGro: Yes I forgot to clean up ^^ wait a second

juliohm:
Should we add new tests to catch the mixed case? I understand that this PR now allows it, so probably add a few tests to make sure they are preserved in future changes.

FPGro:
> 
> 
> Should we add new tests to catch the mixed case? I understand that this PR now allows it, so probably add a few tests to make sure they are preserved in future changes.

Sure, no problem

FPGro:
Okay tests are in place, only the unrelated printing-test fails

juliohm (approved):
Amazing work as usual! ❤️ I will merge it given that the other error is completely unrelated and is being sorted out in master.

Issue #43, opened 2021-02-18
Replace rotation conventions by actual rotation representations


juliohm:
As a quick workaround to include more features to the package, I've included a couple of rotation conventions from various software in geosciences under the `conventions` folder as we did with the `GMSH` ordering convention. These conventions should be replaced by actual types following the `ReferenceFrameRotations.jl` API so that we can easily convert back and forth between different rotation representations including quaternions, direction cosine matrices, etc.

I am super busy migrating functionality from GeoStatsBase.jl to here, so feel free to tackle this issue if you are free 👍🏽

juliohm:
Finally found the time to fix this 🙏🏽 Now we need to wait for some open issues in ReferenceFrameRotations.jl to decide on what to do next.

PR #45, opened 2021-02-18
Fix GSLIB convention

### LINKED ISSUE #44: Fix broken neighborhood tests after migration from GeoStatsBase.jl
I've migrated some rotation functionality from GeoStatsBase.jl to Meshes.jl, and some of the tests are broken.

@rmcaixeta since you are much more familiar with this part of the codebase than I am, could you please take a look?

Here are the 3 broken tests: https://github.com/JuliaGeometry/Meshes.jl/blob/61e71ec471f47c0d5a76d0ca44e16d6156ef84ec/test/neighborhoods.jl#L49-L56

I did my best to not modify the code, but had to split the rotation from the scaling functionality to gain more flexibility for the kinds of things we want to do in Meshes.jl.

rmcaixeta:
This fixes #44 
It was quick, so I made it today
The `reverse` call was ok for 2D, but need to be adjusted for 3D as just X-Y are swapped
I also migrated the rotation convention tests to check if the others were ok and it seems fine


rmcaixeta:
But it seems that this `reverse(A, start, stop)` approach is only possible if `A` is an `AbstractArray`. It doesn't work for `Tuples`. Any suggestion?

test/conventions.jl
```diff
@@ -0,0 +1,42 @@
+@testset "Conventions" begin
```
juliohm: That is awesome, thanks for the additional tests! ❤️ Could you move them to the test/neighborhood.jl file together with the other tests of similar nature? We have an open issue regarding these conventions that may justify a separate test/conventions.jl file in the future: #43

juliohm (changes requested):
Thank you for the quick fix @rmcaixeta ! That helps a lot 💯 

Besides moving the test, we need to fix the reverse for tuples. Perhaps writing it as you did in the previous version with `semiaxes[2], semiaxes[1]` to create the reverse would be fine?

It is good to know I didn't mess up the algorithm in the migration.

rmcaixeta:
Ok. I'll move the tests tomorrow

About the tuple issue, the last approach wouldn't work either because I'd need to modify the tuple. In GeoStatsBase it was constrained to AbstractVector I guess

Better create a new tuple on those cases or constrain the input to Vectors?

juliohm:
Let's create a new tuple then. In Meshes.jl we are supporting static arguments as inputs consistently (tuples of coordinates).

juliohm:
I think the cleanest solution would be to assume vectors in this outer constructor and then provide an additional outer constructor that converts the tuples to vectors 👍🏽 That way we can use the convenient `reverse(v, 1, 2)` syntax.

rmcaixeta:
Waiting for the tests to check if everything's ok. Seems alright for you?

test/neighborhoods.jl
```diff
@@ -62,4 +62,45 @@
       @test sprint(show, ellipsoid) == "Ellipsoid{Float64}((2.0, 1.0), (45.0,), GSLIB)"
     end
   end
+  
+  @testset "RotationConvention" begin
```
juliohm: I think we can remove the test set and place these tests immediately after the other tests and before the final tests with the `show` methods.

src/neighborhoods/ellipsoid.jl
```diff
@@ -28,7 +28,8 @@ function Ellipsoid(semiaxes::S, angles::A; convention::C=TaitBryanExtr) where {S
   @assert valid "invalid number of semiaxes/angles"
 
   # invert semiaxes if necessary
-  ranges = mainaxis(convention) == :Y ? reverse(semiaxes) : semiaxes
+  invert = mainaxis(convention) == :Y
+  ranges = invert ? [semiaxes[i] for i in reverse(1:Dim,1,2)] : semiaxes
```
juliohm, line 32: Nice trick!

juliohm (changes requested):
Nice trick with the list comprehension 💯

Can you just rearrange the tests in the same test set? :)

juliohm (approved):
Amazing work as usual! Merging it.

PR #46, opened 2021-02-22
Implement measure function for UnstructuredMesh

PaulDebus:
Another small utility function I was missing.

The measure (== area) of a mesh is the sum of the areas of its triangles.

Thanks for your efforts!


juliohm:
Awesome @PaulDebus ! Perhaps we need to add this function to the top `mesh.jl` file so that it works for all kinds of meshes. Also you can do `sum(measure, elements(mesh))` instead, to be more generic in 2D and 3D.

PaulDebus:
You are very right there. I always forget about elements, because I always work in 3D and the term "faces" lives deep inside my brain ;)

src/mesh.jl
```diff
@@ -50,5 +50,14 @@ of any 3-face.
 """
 elements(m::Mesh{Dim}) where {Dim} = faces(m, Dim)
 
+
+"""
+    measure(mesh)
+
+Return the surface area of a mesh by summing the area of it's triangles.
+"""
+measure(m::UnstructuredMesh) = sum(measure, elements(m))
```
juliohm:
```suggestion

"""
    measure(mesh)

Return the measure of a mesh by summing the measure of its elements.
"""
measure(m::Mesh) = sum(measure, elements(m))
```

juliohm (changes requested):
Now we need a couple of tests in the test/mesh.jl file to make sure that the function is working as expected in 2D and 3D. Can you please add such tests?

juliohm:
Also for surface area, we are still missing functionality. The plan is to implement a better data structure for the mesh types so that we can navigate the faces of any rank more easily, and compute the boundary of a mesh for example.

PaulDebus:
Yep, will look into adding tests. I will update here.

For the better data structure, I am not exactly sure what you mean by that. Is the rank of a face the number of adjacent faces, so 3 for a face somewhere in the middle and e.g. 2 if it is on the boundary?

This sounds like an interesting task, can I somehow help?

juliohm:
@PaulDebus did you have a chance to write some tests for this PR?

PaulDebus:
I started, but I did not finish yet. I am filling in the tests you have for 2D for 3D as well. Found some things to think about with 3D meshes. I pushed the current state so you can have a look if you want.

For example, this implementation is based on the `measure` of the underlying triangles, which I was working on in the other pull request #47, which is still not done and needs some attention. 

Also, I think we have to use `faces(mesh, 2)` as the `elements` function does not work with the meshes I use for testing. It returns 3D shapes, which are not what I expect.

juliohm:
You can add the measure for triangles in this PR as well since it is about measures. And to clarify a measure is the "area" in 2D and the "volume" in 3D geometries. So a 3D mesh has a 3D volume which is the sum of the element volumes. In 2D you get the sum of the 2D areas (triangles, quadrangles, etc.).

So in summary, you can add a measure function for the triangle embedded in 3D space and the tests should pass here if your mesh is a mesh of triangles.

PaulDebus:
Hey,

I was a little busy the last few days, but I kept thinking about how to implement this well. First, I will use this PR to contain everything related to measures of triangles and the other PR #47 will only be about the normal computation. 

I am still a little confused about the different measures. If I have a mesh in 3D space, I could take the 3D measure - the volume - or the 2D measure - the surface area. Am I understanding this correctly? If so, how would I distinguish what to do when calling `measure(mesh)`? 

I found a formula for the measure of simplices of different dimensions in higher dimension spaces, the [Cayley-Menger Determinant](https://en.wikipedia.org/wiki/Cayley%E2%80%93Menger_determinant). This could be used to compute many different things. And it is based on the determinant, as you thought it would be.
However, I do not think that this is appropriate for simple cases like a Triangle in 2D or 3D. For the 2D case, you already have a very good - if not optimal - implementation and for the 3D case, I checked again the simple implementation I showed recently:

```julia
function area(a::Point{3}, b::Point{3}, c::Point{3})
    A = [b-a c-a]'
    0.5sqrt(det(A*A'))
end

julia> points = Point((1., 2., 3.,)), Point((2., 2., 5.)), Point((0., 1., 4.))
julia> @time area(points...)
  0.000004 seconds (4 allocations: 112 bytes)
1.8708286933869707
```
And I inspected the execution and `A` is a `SArray` automatically, so this should not be an issue. 

My proposal would be to dispatch specialized `measure` functions based on the dimension for 2D and 3D triangles and - though I am not sure about this one - use a general dispatch for higher dimensions. This way, your implementation would stay untouched and I can add the simple 3D implementation.

In any case, I will add some tests for 3D meshes to achieve better coverage there. What are your thoughts on this? Is this suitable? If not, I think I need some more pointers to make it fit. 

Thanks for the collaboration!

juliohm:
> I am still a little confused about the different measures. If I have a mesh in 3D space, I could take the 3D measure - the volume - or the 2D measure - the surface area. Am I understanding this correctly? If so, how would I distinguish what to do when calling `measure(mesh)`?

So if you have a mesh of 3D elements, the measure of the mesh is a volume as you said. If you want the surface area of the boundary then you would need to call something like `measure(boundary(mesh))` but this boundary operation is not implemented yet.

Now if you have a 2D mesh embedded in a 3D space (e.g. mesh of triangles in 3D), then `measure(mesh)` will return an area because every element has parametric dimension 2 (triangles). Makes sense? In this case the value `measure(boundary(mesh))` would return the total length of the edges in the boundary. That is why the name `measure` is important and generic. It doesn't matter if you are in 2D or 3D space or if you have a parametric dimension 2 or 3, it will work as expected.

juliohm:
Closing this as stale. Thanks for the attempt. We will try to fix the issue in an updated PR given that you are busy with other things.

PR #47, opened 2021-02-22
Add Triangle{3} measure and normal

PaulDebus:
While I was at it, I also implemented the measure for Triangle{3}, a prerequisite for my other pull request.

Also the normal, which the measure also uses.

This time I did think of the tests ;)

### DIFF
--- a/src/Meshes.jl
+++ b/src/Meshes.jl
@@ -110,7 +110,7 @@ export
   windingnumber, chains,
   isclosed, issimple, hasholes,
   angles, innerangles, close!, open!,
-  orientation, bridge,
+  orientation, bridge, normal,
 
   # connectivities
   Connectivity,
--- a/src/polytopes/triangle.jl
+++ b/src/polytopes/triangle.jl
@@ -13,6 +13,21 @@ end
 
 measure(t::Triangle{2}) = abs(signarea(t))
 
+function measure(tri::Triangle{3})
+	v1, v2, v3 = vertices(tri)
+	e1 = v2 - v1
+	e2 = v3 - v1
+	c = cross(e1, e2)
+	norm(c) * 0.5
+end
+
+function normal(tri::Triangle{3})
+    v1, v2, v3 = vertices(tri)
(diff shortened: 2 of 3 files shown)


juliohm:
Let me share some tips with you so that you can more effectively contribute to the project. We don't merge PRs that do two things at the same time. In this case you are both introducing a method for measure and another for normal.

Also, try to implement these concepts generically unless there is an optimization reason to implement it just for triangle. Can't we just adapt the existing measure for triangles embedded in R^2 for triangles embedded in R^3? It should be a simple determinant again, right? No need to rewrite the code manually.

So, split this PR into two separate PRs, one for works on measure and another for works on normal vectors.

PaulDebus:
Alright, will split the PR. I will remove the normal from this PR and open a new one.

According to [this](https://math.stackexchange.com/questions/2152754/calculate-3d-triangle-area-by-determinant), the area in 3D can be computed via a determinant, but you also need a square root here. But I think that we can really use the multiple dispatch here. You implemented a specialized formula (expanding the determinant to a single step) for  the 2D case, which is dispatched based on the Type. I think implementing the corresponding dispatch for the 3D case is valid here. 

I have no preference between the form I first proposed using the cross product or using the determinant as following:

```julia
function area(a::Point{3}, b::Point{3}, c::Point{3})
    A = [b-a c-a]'
    0.5sqrt(det(A*A'))
end
```

Had the 3D formula been equivalent to the 2D (could be that I missed something and they are identical), then one function would have been enough, I think this case is a good example for the nice multiple dispatch in Julia

juliohm:
Yes, we need to make sure that we do not allocate memory. When you write `[b-a c-a]` you are creating a normal Julia Array instead of a static array from StaticArrays.jl.

Also, the `signarea` function can certainly be adapted for the 3D case. I think fixing this lack of area for triangles embedded in 3D can be a matter of just dropping the `{2}` restriction here and there. But please double check and add tests as many people rely on this calculation. 👍🏽

Issue #49, opened 2021-02-24
Declare collaboration practices


serenity4:
I stumbled upon SciML's [ColPrac](https://github.com/SciML/ColPrac) which is a set of guidelines aimed towards package developers to facilitate collaborative work. I found it quite interesting and it includes a lot of good practices that we already seek to apply whenever possible (such as "PRs do one thing", "code changes should be tested", etc.). Maybe we should link to those (by adding the badge) in the README? That way, contributors will know more about the kind of expectations we have.

We can also choose to create our own document, usually done in the form of a `CONTRIBUTING.md`, which can either be self-contained or build upon ColPrac (or any other guide). Personally I'm fine using ColPrac alone, having in mind that these are guidelines and not hard requirements.

juliohm:
I've seen the ColPrac repo and it is a very extensive collection of good practices for sure. I wonder if we could write down a subset that is more quick to read and relevant to this project? I find the ColPrac scary for new contributors. I think what we want is a short list of bullet points regarding PRs and issues.

serenity4:
I agree that it is maybe too extensive. If we decide to follow most of ColPrac but just decide not to show everything on the front page, we can link to it and state the points that we specifically focus on (and maybe those we explicitly ignore, or modify).

For example, it has a ton of stuff regarding package versioning, breaking changes, etc..., these are not really targeting casual contributors. So, omitting it but having a link for contributors who want a more detailed explanation, they could have a better understanding of things.

Something like:

"""
Contributors are encouraged to adhere to the [ColPrac](https://github.com/SciML/ColPrac) guidelines, and more specifically:
- practice 1
- practice 2
- practice 3
"""

juliohm:
That is a very good plan. Would you like to draft this PR? We can evolve the specific points there.

juliohm:
Fixed on master 💯

PR #51, opened 2021-02-24
Adopt ColPrac guidelines for contributing

serenity4:
Follow-up from #49.

Don't hesitate if you have any more ideas.


juliohm:
Nice, I will merge it right away. We can improve the list as we receive more contributions.

PR #52, opened 2021-02-26
Tables.jl updates

quinnj:


juliohm:
Thanks for the PR! I"ve updated the master branch to use NamedTuple for simplicity.

I will comment on Zulip with a MWE to illustrate what is still not working.

Issue #54, opened 2021-02-28
Brainstorming intersection interface


Arkoniak:
```julia
s1 = Segment(Point(4.0, 3.0), Point(0.0, -3.0))
s2 = Segment(Point(1.0, 2.0), Point(3.0, -2.0))

@code_warntype intersecttype(s1, s2)
```
returns (I leave only the beginning of the message)
```
julia> @code_warntype intersecttype(s1, s2)
MethodInstance for Meshes.intersecttype(::Segment{2, Float64, StaticArrays.SVector{2, Po
int2}}, ::Segment{2, Float64, StaticArrays.SVector{2, Point2}})
  from intersecttype(s1::Segment{2, T, V} where V<:AbstractArray{Point{2, T}, 1}, s2::Se
gment{2, T, V} where V<:AbstractArray{Point{2, T}, 1}) where T in Meshes at /home/skoffe
r/.julia/packages/Meshes/PDxIG/src/polytopes/segment.jl:69
Static Parameters
  T = Float64
Arguments
  #self#::Core.Const(Meshes.intersecttype)
  s1::Segment{2, Float64, StaticArrays.SVector{2, Point2}}
  s2::Segment{2, Float64, StaticArrays.SVector{2, Point2}}
Locals
  @_4::Union{Nothing, Tuple{Int64, Int64}}
  @_5::Union{Nothing, Tuple{Int64, Int64}}
  @_6::Tuple{StaticArrays.SOneTo{2}, Int64}
  @_7::Tuple{StaticArrays.SOneTo{2}, Int64}
  wyzero::Bool
  wxzero::Bool
  determinatey::Bool
  windingy::Float64
  determinatex::Bool
  windingx::Float64
  Q::Vector{Point2}
  y2::Point2
  y1::Point2
  x2::Point2
  x1::Point2
  ȳ::Point2
  x̄::Point2
  j@_21::Int64
  α::Float64
  β::Float64
  j@_24::Int64
  @_25::Bool
Body::Any
```

It looks like it originates from `intersectcolinear`, for two reasons:

1. this function has `@error` in one of the branches (and an output of `@error` has different type then all other branches).
2. It returns too many different types and Julia's compiler gives up on them.

juliohm:
Thanks for catching the issue. Do you have a suggestion? Perhaps eliminating the `@error` branch would be enough to solve the type instability? I am happy to review a PR.

Arkoniak:
[...]

PR #56, opened 2021-03-02
Simplify intersection between colinear segments

juliohm:
First attempt to improve #54 

The previous implementation was unnecessarily complex. The new one should be much simpler to understand and to maintain.

There is a small difference of behavior compared to the previous implementation. Previously I was returning the segment with an orientation that matched the first segment of the intersection. Now I am always returning segments that are oriented toward the positive orthant. I think it is a reasonable default, what do you think?

cc: @Arkoniak

### DIFF
--- a/test/polytopes.jl
+++ b/test/polytopes.jl
@@ -41,7 +41,7 @@
 
     s1 = Segment(P2(1,2), P2(1,0))
     s2 = Segment(P2(1,0), P2(1,1))
-    @test s1 ∩ s2 == Segment(P2(1,1), P2(1,0))
+    @test s1 ∩ s2 == Segment(P2(1,0), P2(1,1))
 
     s1 = Segment(P2(0,0), P2(2,0))
     s2 = Segment(P2(-2,0), P2(-1,0))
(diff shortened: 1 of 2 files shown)


juliohm:
@serenity4 the benchamrk CI was skipped. Do you know how we can trigger it in this PR? I can also add some entries to the benchmark suite before we merge so that we track the performance of the intersection of segments.

test/polytopes.jl
```diff
@@ -41,7 +41,7 @@
 
     s1 = Segment(P2(1,2), P2(1,0))
     s2 = Segment(P2(1,0), P2(1,1))
-    @test s1 ∩ s2 == Segment(P2(1,1), P2(1,0))
+    @test s1 ∩ s2 == Segment(P2(1,0), P2(1,1))
```
serenity4, line 44: That's because the intersection segment is built the other way around?

serenity4 (approved):
Looks a lot better indeed. For benchmarks, you have to manually add the _run benchmarks_ label.

test/polytopes.jl
juliohm: Exactly. In this case the first segment is oriented downwards and the second segment is oriented upwards. The previous implementation was preserved the orientation of the first segment. Now we always return segments oriented towards the positive orthant, which in this case is a segment that points upwards. I think it is a reasonable default, and actually it should be more easy to post-process as a user too. If any algorithm needs a specific orientation convention, it should be easy to post-process the result.

PR #58, opened 2021-03-03
Document package benchmark usage

serenity4:
This provides a description for developers who want to benchmark changes locally, and quickly presents the tools used for CI benchmarks.

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -53,7 +53,10 @@ makedocs(
       ]
   ],
     "Contributing" => [
-      "contributing/guidelines.md"
+      "contributing/guidelines.md",
+      "Developer guide" => [
+        "contributing/developer_guide/benchmarking.md"
+      ]
     ],
     "About" => [
       "License" => "about/license.md"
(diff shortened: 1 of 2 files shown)


docs/make.jl
```diff
@@ -53,7 +53,10 @@ makedocs(
       ]
   ],
     "Contributing" => [
-      "contributing/guidelines.md"
+      "contributing/guidelines.md",
+      "Developer guide" => [
+        "contributing/developer_guide/benchmarking.md"
```
juliohm, line 58: What about making this a single page without subfolders? `contributing/devguide.md`? My impression is that it won't evolve into a huge guide so it may be unnecessary to introduce a subfolder structure. We could have a single markdown file with sections. And if it grows we split it into various markdown files.

juliohm (commented):
Amazing docs, will help a lot potential new contributors.

serenity4:
Whoops, very sorry about that, wrong upstream remote. I un-merged it.

PR #60, opened 2021-03-03
Adding Toggle Option for metadata calculation to BlockPartition

atreyamaj:
Added optional parameter `calculate_metadata` which is set to `false` by default.
If it is set to `true`, the function will work like before the patch. 
While it is `false` and expensive metadata calculation is not occurring, there is a performance enhancement as shown below:

Before:
Screenshot from 2021-03-03 20-17-08

After:
Screenshot from 2021-03-03 20-17-01

The plots remain the same with both `true` and `false`:
 
Screenshot from 2021-03-03 20-18-12


src/partitioning/block.jl
```diff
@@ -16,7 +16,7 @@ BlockPartition(sides::NTuple{Dim,T}) where {Dim,T} = BlockPartition{Dim,T}(sides
 
 BlockPartition(sides::Vararg{T,Dim}) where {Dim,T} = BlockPartition(sides)
 
-function partition(object, method::BlockPartition)
+function partition(object, method::BlockPartition, calculate_metadata = false)
```
juliohm: The option should be part of the `BlockPartition` struct. And you should pass it to the constructor of the struct as a keyword argument `neighbors=false` by default.

juliohm (changes requested):
After you fix the code as suggested, please use a larger 3D partition to benchmark the difference. For example try to partition a 100x100x100 CartesianGrid into blocks of size 1x1x1 or 2x2x2 or some reasonable size that allows the calculation to finish in a few seconds to minutes.

src/partitioning/block.jl
atreyamaj: Thank you for the review. Working on it :+1:

juliohm:
@atreyamaj did you have a chance to adjust this PR?

atreyamaj:
@juliohm  Apologies for the delay, I'm still working on it. I got stuck with adding a Boolean argument to the struct after the Varargs part but upon asking in Slack Yesterday, I got a few leads. I am looking into those as we speak.

atreyamaj:
I have made the changes @juliohm 
Attaching relevant screenshots below:

Screenshot from 2021-03-08 21-34-35

Screenshot from 2021-03-08 21-35-02

Screenshot from 2021-03-08 21-35-35

Screenshot from 2021-03-08 21-35-44

src/partitioning/block.jl
```diff
@@ -10,17 +10,19 @@ A method for partitioning spatial objects into blocks of given `sides`.
...
+BlockPartition(sides::Vararg{T,Dim}; neighbors::Bool = false) where {Dim,T} = BlockPartition(SVector(sides), neighbors)
 
 function partition(object, method::BlockPartition)
   Dim = embeddim(object)
-
   psides = method.sides
   bbox = boundingbox(object)
+  calculate_metadata = method.neighbors
+  
+  println("Neighbors has been set to ", calculate_metadata)
```
juliohm: You don't want to print something to the screen unless it was asked by the user. Please remove the println statement.

Also notice the convention is to name variables without underscores when possible. I wouldn't even create a new variable for that because `method.neighbors` is already clear in a if statement: `if method.neighbors`

src/partitioning/block.jl
```diff
@@ -10,17 +10,19 @@ A method for partitioning spatial objects into blocks of given `sides`.
...
 struct BlockPartition{Dim,T} <: PartitionMethod
   sides::SVector{Dim,T}
+  neighbors::Bool
 end
 
-BlockPartition(sides::NTuple{Dim,T}) where {Dim,T} = BlockPartition{Dim,T}(sides)
-
-BlockPartition(sides::Vararg{T,Dim}) where {Dim,T} = BlockPartition(sides)
+BlockPartition(sides::NTuple{Dim,T}, neighbors = false) where {Dim,T} = BlockPartition{Dim,T}(sides, neighbors)
+BlockPartition(sides::Vararg{T,Dim}; neighbors::Bool = false) where {Dim,T} = BlockPartition(SVector(sides), neighbors)
```
juliohm:
```suggestion
BlockPartition(sides::NTuple{Dim,T}; neighbors=false) where {Dim,T} = BlockPartition{Dim,T}(sides, neighbors)
BlockPartition(sides::Vararg{T,Dim}; neighbors=false) where {Dim,T} = BlockPartition(SVector(sides), neighbors)
```
Please follow the code style convention in the code when possible.

src/partitioning/block.jl
```diff
@@ -55,16 +57,28 @@ function partition(object, method::BlockPartition)
     append!(subsets[i], j)
   end
 
+  if calculate_metadata == false
+    println("Metadata calculation has been disabled. To calculate Metadata, please enable it while calling BlockPartition with neighbors = true.") 
+  end
```
juliohm: Remove any print statement. Print statements usually appear in end user code but you are contributing to a library that people will use in high-level applications where they can print whatever they want. Makes sense?

src/partitioning/block.jl
```diff
@@ -55,16 +57,28 @@ function partition(object, method::BlockPartition)
...
+    bfinish = CartesianIndex(Dims(nblocks))
+    for (i, bcoords) in enumerate(bstart:bfinish)
+      for b in (bcoords - boffset):(bcoords + boffset)
+        if all(Tuple(bstart) .≤ Tuple(b) .≤ Tuple(bfinish)) && b ≠ bcoords
+          push!(neighbors[i], linear[b])
+        end
       end
     end
+    # Save Calculated Metadata
+    metadata = Dict(:neighbors => neighbors)
```
juliohm: Since you did initialize the `metadata` before the if statement you could simply update it instead of creating a new one from scratch: `metadata[:neighbors] = neighbors`  or something like that.

juliohm (changes requested):
Please consider the second round of reviews. Also we need tests before we can merge.

src/partitioning/block.jl
atreyamaj: Working on it!  Thank you!

src/partitioning/block.jl
atreyamaj: Yep! Fixed it! Thank you!

src/partitioning/block.jl
atreyamaj: Fixed! Thank you!

src/partitioning/block.jl
atreyamaj: Fixed! Thank you!

juliohm: I don't see the fix. Did you push to GitHub aready? Please do so that we can quickly see the changes and provide feedback.

atreyamaj: I have committed it now!

src/partitioning/block.jl
```diff
@@ -54,17 +54,25 @@ function partition(object, method::BlockPartition)
 
     append!(subsets[i], j)
   end
+  
+  #Intitialize metadata to an empty Dict.
+  #If method.neighbors is enabled, we will populate it.
```
juliohm:
```suggestion
  # initialize metadata
```
Less is more

src/partitioning/block.jl
```diff
@@ -54,17 +54,25 @@ function partition(object, method::BlockPartition)
...
+    bstart  = CartesianIndex(ntuple(i -> 1, Dim))
+    boffset = CartesianIndex(ntuple(i -> 1, Dim))
+    bfinish = CartesianIndex(Dims(nblocks))
+    for (i, bcoords) in enumerate(bstart:bfinish)
+      for b in (bcoords - boffset):(bcoords + boffset)
+        if all(Tuple(bstart) .≤ Tuple(b) .≤ Tuple(bfinish)) && b ≠ bcoords
+          push!(neighbors[i], linear[b])
+        end
       end
     end
```
juliohm, line 73: Can you please double check the indentation? I find it strange that GitHub only shows changes for a part of the loop. Double check that the indentation is correct locally and is not using TAB

juliohm (changes requested):
Nice. Now we need tests to make sure that the option is working as expected. For example, I would create a partition with and without the option enabled and check that the results make sense and that the option enabled takes more time.

src/partitioning/block.jl
atreyamaj: Checking.

atreyamaj:
> Nice. Now we need tests to make sure that the option is working as expected. For example, I would create a partition with and without the option enabled and check that the results make sense and that the option enabled takes more time.

How would you like me to do this? Should I write a script or should I do it in a notebook and post screenshots?

src/partitioning/block.jl
atreyamaj: The indentation seems to be correct locally. I have re-written it as well just to make sure.
Screenshot from 2021-03-09 02-33-48

src/partitioning/block.jl
atreyamaj: Fixed! Thank you!

atreyamaj:
I ran the function and found the following results:
The time taken seems to be a third of what was required before, with less memory as well.
Screenshot from 2021-03-09 13-07-09

juliohm:
Thank you for the updates @atreyamaj. The way you write tests in a Julia package is in the `test` folder. There should be a file already called `test/partitions.jl` or a similar name with tests for the `BlockPartition` method. Add more lines of tests there with and without the option and preferably with 3D grids if you want to compare the time difference. You can use the macro `@elapsed f()` to get the time as a number to compare and the macro `@allocated f()` to get the memory allocated.

I would focus on first adding a few tests that are just to make sure that the results make sense, that is with the option `neighbors=true` you get the correct metadata in the results and with options `neighbors=false` you get an empty dictionary.

Then, after the tests above are ready I would consider tests with 3D grids to make sure that the performance difference is visible.

To run the tests you activate the project environment with `] activate /path/to/Meshes.jl` and then `] test`

atreyamaj:
The previously written tests are passing:
Screenshot from 2021-03-10 14-38-48

I will now be adding 3D tests.

juliohm:
@atreyamaj let me know when the tests are ready, looking forward to merging this fix.

atreyamaj:
> @atreyamaj let me know when the tests are ready, looking forward to merging this fix.

Will do :+1:

atreyamaj:
@juliohm  I have written one test for 3D computation, and appended a few tests for the previous ones and added variations for the neighbors option. 
Computation speed seems to be affected quite a bit when the partition sizes are small, for 3D blocks:

Screenshot from 2021-03-10 21-29-57


Screenshot from 2021-03-10 20-29-12

Screenshot from 2021-03-10 20-31-22

Test cases:
Screenshot from 2021-03-10 21-28-22

test/partitioning.jl
```diff
@@ -82,9 +82,39 @@
     @test length(p) == 4
     @test all(nelements.(p) .== 25)
 
+    p = partition(grid, BlockPartition(T(5),T(5), neighbors = false))
+    @test length(p) == 4
+    @test all(nelements.(p) .== 25)
```
juliohm: This test is redundant right? It already happens right above. Please remove it.

juliohm (changes requested):
Hi @atreyamaj please erase the tests you added, and add a new set of tests at the end specifically for the option added. That is, you don't to replicate all previous tests with the option enabled. You just need a couple of tests that specifically check the result of the partition and try to access `metadata(p)`. Makes sense?

juliohm:
As a general comment, you should use GitHub markdown syntax to share your updates instead of screenshots which are not searchable. So instead of copying/pasting a screenshot, try to start a markdown section with Julia hightlight here.

atreyamaj:
> As a general comment, you should use GitHub markdown syntax to share your updates instead of screenshots which are not searchable. So instead of copying/pasting a screenshot, try to start a markdown section with Julia hightlight here.

Thank you! I will keep this in mind for future PRs!

atreyamaj:
@juliohm I have made the required changes. I removed the additional duplicate tests that I added, and only put in 2 new tests, both based on a 3D grid. All tests are passing.

`
Test Summary:       | Pass  Total
Meshes.jl (Float32) |  927    927
Test Summary:       | Pass  Total
Meshes.jl (Float64) |  927    927
    Testing Meshes tests passed 
`

test/partitioning.jl
atreyamaj: Removed!

test/partitioning.jl
```diff
@@ -85,6 +85,16 @@
...
+
+    grid = CartesianGrid{T}(100,100,100)
+
+    p = partition(grid, BlockPartition(T(1.), T(1.), T(1.), neighbors = true))
+    @test length(p) == 1000000
+    @test Set(nelements.(p)) == Set(1)
+
+    p = partition(grid, BlockPartition(T(5.), T(5.), T(5.), neighbors = false))
+    @test length(p) == 8000
+    @test Set(nelements.(p)) == Set(125)
```
juliohm: Can you share the times for these two tests specifically? We also don't want to increase the time that it takes to run the test suite considerably so I would just increase the size of the blocks to something else if it is taking more than 2s for example. Also you need to check what this PR is actually doing, which is providing a new option that computes or not the metadata. So you need to add tests to check `metadata(p)` in both cases.

juliohm (changes requested):

test/partitioning.jl
atreyamaj: Working on it now!

atreyamaj: When neighbors is true, blocksize of 1 is making the function take 37.043399 seconds. However, if I set neighbors to false for blocksize (1,1,1) and true for size (5,5,5), it finishes in 7.201889 seconds. Would the latter be alright?

atreyamaj: Adding metadata test now

atreyamaj: Looking into it now!

atreyamaj:
```julia
function testing123()
    grid = CartesianGrid(50,50,50)
    p = partition(grid, BlockPartition(1., 1., 1., neighbors = false))
    println(length(p) == 1000000)
    println(Set(nelements.(p)) == Set(1))

    p = partition(grid, BlockPartition(5., 5., 5., neighbors = true))
    println(length(p) == 8000)
    println(Set(nelements.(p)) == Set(125))
end
```
Takes a total of 0.836927 seconds to run. Should I update this?

atreyamaj: When I check for metadata, should I go for a complete value to value check, or should I only check whether the Dict is empty or not? I am having a bit of trouble reading the values since the matrix is quite large, which is leading to a lot of metadata calculation which I am not able to read.
If we need a value by value comparison, I am thinking of adding a different simple 2D test for it. Would that be okay?

atreyamaj: Okay :+1:

juliohm: Regarding the metadata test, you should see that the neighbors key will contain a vector with the same length of the number of blocks returned. Each entry will contain the neighboring blocks. For example, the block at index (1,1) has neighbors [(2,1),(1,2)]. You can just check the following:

```julia
p = partition(...)
m = metadata(p)
n = m[:neighbors]
@test length(n) == length(p)
@test all(0 .< length.(n) .<= 27)
```

juliohm: So I am checking that the vector `neighbors` has the same length of the number of blocks returned and also checking that each block returned has at most 3x3x3 = 27 blocks nearby.

atreyamaj: @juliohm I have added the metadata test for the 3D (50,50,50) grid itself with a blocksize of (5,5,5). Do let me know if the code looks too messy because of the size of the dict, I will try using something else instead in that case.

atreyamaj: Ohhhh Okay, this makes a lot more sense than what I was doing earlier. Thank you!

atreyamaj: I have added the changes.

atreyamaj: I have added the changes! Please do let me know if I should add some more tests.

test/partitioning.jl
```diff
@@ -85,6 +85,20 @@
...
+    @test length(p) == 125000
+    @test Set(nelements.(p)) == Set(1)
+    @test metadata(p) == Dict{Any,Any}()
+
+    p = partition(grid, BlockPartition(T(5.), T(5.), T(5.), neighbors = true))
+    @test length(p) == 1000
+    @test Set(nelements.(p)) == Set(125)
+    n = metadata(p)[:neighbors]
+    @test length(n) == length(p)
+    @test all(0 .< length.(n) .<= 125)
```
juliohm: The maximum number of adjacent blocks is 3x3x3=27 right? Can you please adjust the test? Also can you use the LaTeX symbol `\le` instead of `<=`?

atreyamaj: Oh I am so sorry, I don't know how I missed that. Changing it now. Also changed `<=` to ` ≤`

test/partitioning.jl
```diff
@@ -85,12 +85,27 @@
...
+    @test length(p) == 125000
+    @test Set(nelements.(p)) == Set(1)
+    @test metadata(p) == Dict{Any,Any}()
+
+    p = partition(grid, BlockPartition(T(5.), T(5.), T(5.), neighbors = true))
+    @test length(p) == 1000
+    @test Set(nelements.(p)) == Set(125)
+    n = metadata(p)[:neighbors]
+    @test length(n) == length(p)
+    @test all(0 .< length.(n) .≤ 27)
```
juliohm:
```suggestion
    @test all(0 .< length.(n) .< 27) 
```
I think we can guarantee strict inequality since the center block is not included.

test/partitioning.jl
```diff
@@ -85,12 +85,27 @@
...
+    @test length(n) == length(p)
+    @test all(0 .< length.(n) .< 27) 
+    
   end
 
   @testset "BisectPointPartition" begin
     grid = CartesianGrid((10,10), T.((-0.5,-0.5)), T.((1.0, 1.0)))
 
-    p = partition(grid, BisectPointPartition(T.((0.0,1.0)), T.((5.0,5.1))))
+    p = partition(grid, BisectPointPartition(T.((0.0,1.0)), T.((5.0,5.1))))
```
juliohm:
```suggestion
    p = partition(grid, BisectPointPartition(T.((0.0,1.0)), T.((5.0,5.1))))
```
Please be careful with empty spaces

atreyamaj: Sorry, will be more mindful :+1:

juliohm:
Ok, now we need just a final detail, which is to move the variable `neighbors` defined in line 39 to inside the if branch you introduced in this PR. It is being defined too early in the code. Could you please move it down?

atreyamaj:
> Ok, now we need just a final detail, which is to move the variable `neighbors` defined in line 39 to inside the if branch you introduced in this PR. It is being defined too early in the code. Could you please move it down?

Looking into it!

atreyamaj:
> > Ok, now we need just a final detail, which is to move the variable `neighbors` defined in line 39 to inside the if branch you introduced in this PR. It is being defined too early in the code. Could you please move it down?
> 
> Looking into it!

I have moved it, running tests now.

juliohm:
And we also need to update the docstring. Can you please add a sentence that says that one can optionally specify the option `neighbors`?

atreyamaj:
> > > Ok, now we need just a final detail, which is to move the variable `neighbors` defined in line 39 to inside the if branch you introduced in this PR. It is being defined too early in the code. Could you please move it down?
> > 
> > 
> > Looking into it!
> 
> I have moved it, running tests now.

@juliohm Since we are using `neighbors` outside of the `if` condition, the tests are failing.
We are using neighbors like this:

```julia
if method.neighbors == true
    neighbors = [Int[] for i in 1:prod(nblocks)]
    bstart  = CartesianIndex(ntuple(i -> 1, Dim)).......
```

and later on:
```julia 
empty = isempty.(subsets)
  subsets = subsets[.!empty]
  neighbors = neighbors[.!empty]
  for i in findall(empty)
    for n in neighbors
      setdiff!(n, i)
    end
  end
```

Which is why when I run the tests, I get this:
```julia
ERROR: LoadError: Some tests did not pass: 914 passed, 0 failed, 4 errored, 0 broken.
in expression starting at /home/atreyamaj/.julia/dev/Meshes/test/runtests.jl:75
ERROR: Package Meshes errored during testing
```

Example error:
```julia
BlockPartition: Error During Test at /home/atreyamaj/.julia/dev/Meshes/test/partitioning.jl:78
  Got exception outside of a @test
  UndefVarError: neighbors not defined
```

atreyamaj:
> And we also need to update the docstring. Can you please add a sentence that says that one can optionally specify the option `neighbors`?

Doing this now :+1:

juliohm:
You can move it to right before the `if method.neighbors` statement then. Makes sense?

atreyamaj:
> You can move it to right before the `if method.neighbors` statement then. Makes sense?

Oh okay, doing that.

atreyamaj:
> > And we also need to update the docstring. Can you please add a sentence that says that one can optionally specify the option `neighbors`?
> 
> Doing this now



> And we also need to update the docstring. Can you please add a sentence that says that one can optionally specify the option `neighbors`?

By docstring, do you mean docs/src/algorithms/partitioning.md ?

juliohm:
I mean the documentation right before the `BlockPartition` struct. This is what you get when you type `?BlockPartition` in the terminal.

atreyamaj:
> I mean the documentation right before the `BlockPartition` struct. This is what you get when you type `?BlockPartition` in the terminal.

Done!

src/partitioning/block.jl
```diff
@@ -7,14 +7,15 @@
     BlockPartition(side₁, side₂, ...)
 
 A method for partitioning spatial objects into blocks of given `sides`.
+Metadata calculation is disabled by default. Enable it by calling BlockPartition with neighbors = true.
```
juliohm:
```suggestion
A method for partitioning spatial objects into blocks of given `sides`.
Optionally, compute the `neighbors` of a block as the metadata.
```

juliohm:
Please add the option `neighbors` to the docstring header:

```julia
BlockPartition(sides; neighbors=false)
BlockPartition(side1, side2, ...; neighbors=false)
```

atreyamaj:
> ```julia
> BlockPartition(sides; neighbors=false)
> ```

Added!

src/partitioning/block.jl
```diff
@@ -3,18 +3,19 @@
 # ------------------------------------------------------------------
 
 """
-    BlockPartition(sides)
-    BlockPartition(side₁, side₂, ...)
+    BlockPartition(sides; neighbors=false)
+    BlockPartition(side1, side2, ...; neighbors=false)
```
juliohm:
```suggestion
    BlockPartition(side₁, side₂, ...; neighbors=false)
```

test/partitioning.jl
```diff
@@ -85,6 +85,21 @@
...
+    @test Set(nelements.(p)) == Set(1)
+    @test metadata(p) == Dict{Any,Any}()
+
+    p = partition(grid, BlockPartition(T(5.), T(5.), T(5.), neighbors = true))
+    @test length(p) == 1000
+    @test Set(nelements.(p)) == Set(125)
+    n = metadata(p)[:neighbors]
+    @test length(n) == length(p)
+    @test all(0 .< length.(n) .< 27) 
+
```
juliohm:
```suggestion
```

juliohm (approved):

juliohm:
Thanks!

atreyamaj:
Thank you for helping me through the process!

PR #62, opened 2021-03-06
Replace coordinates! by centroid

### LINKED ISSUE #61: Replace coordinates! by centroid
I've migrated the trait `coordinates!` from GeoStatsBase.jl to Mehes.jl for computing the coordinates of the centroid of the i-th element of the mesh. Most algorithms follow this pattern:

```julia
function algorithm(mesh::{Dim,T})
  x = MVector{Dim,T} # pre-allocate memory
  for i in 1:nelements(mesh)
    coordinates!(x, mesh, i)
    # do something with x
  end
end
```

Now I realized that this pre-allocation of the coordinates with `MVector` doesn't make any difference in performance. I could have simply used a non-mutating version `x = coordinates(mesh, i)` in the loop for clarity. The benchmark bellow shows that there is no difference in CPU time, nor in memory allocations. Because the names `coordinates!` and `coordinates` are already taken, I am using the names `centroid!` and `centroid` in the benchmark:

```julia
using Meshes
using StaticArrays
using BenchmarkTools

function centroid!(x, Ω, i)
  for j in 1:embeddim(Ω)
    x[j] = 2.0*i + j
  end
  x
end

function centroid(Ω, i)
  SVector(ntuple(j->2.0*i+j, embeddim(Ω)))
end


function f1(Ω)
  Dim = embeddim(Ω)
  T = coordtype(Ω)
  x = MVector{Dim,T}(undef)
  s = zero(T)
  for i in 1:nelements(Ω)
    centroid!(x, Ω, i)
    s += sum(x) / Dim
  end
  s
end

function f2(Ω)
  Dim = embeddim(Ω)
  T = coordtype(Ω)
  s = zero(T)
  for i in 1:nelements(Ω)
    x = centroid(Ω, i)
    s += sum(x) / Dim
  end
  s
end

Ω = CartesianGrid(100,100)

@btime f1(Ω)
@btime f2(Ω)
```
```julia
8.738 μs (1 allocation: 16 bytes)
1.00025e8
8.932 μs (1 allocation: 16 bytes)
1.00025e8
```

In this proposal I would like to eliminate the mutating version `coordinates!` and replace the name coordinates by the name `centroid(mesh, i)` that allocates an `SVector` in the stack. This name is more appropriate and we can make this function return an actual `Point` instead of a raw vector.

I will prepare a PR with these modifications, but please let me know if you spot any failure in this analysis. After this PR is reviewed and merged, I will be able to release another series of versions for the GeoStats.jl stack to use the more explicit `centroid` name.

cc: @serenity4

juliohm:
Fix #61.

### DIFF
--- a/src/partitioning/block.jl
+++ b/src/partitioning/block.jl
@@ -18,7 +18,6 @@ BlockPartition(sides::Vararg{T,Dim}) where {Dim,T} = BlockPartition(sides)
 
 function partition(object, method::BlockPartition)
   Dim = embeddim(object)
-  T = coordtype(object)
 
   psides = method.sides
   bbox = boundingbox(object)
@@ -36,21 +35,18 @@ function partition(object, method::BlockPartition)
   start   = @. ce - nleft * psides
   nblocks = @. nleft + nright
 
-  subsets   = [Vector{Int}() for i in 1:prod(nblocks)]
-  neighbors = [Vector{Int}() for i in 1:prod(nblocks)]
+  subsets   = [Int[] for i in 1:prod(nblocks)]
+  neighbors = [Int[] for i in 1:prod(nblocks)]
 
   # Cartesian to linear indices
   linear = LinearIndices(Dims(nblocks))
 
-  coords = MVector{Dim,T}(undef)
   for j in 1:nelements(object)
-    coordinates!(coords, object, j)
+    coords = coordinates(centroid(object, j))
 
     # find block coordinates
     c = @. floor(Int, (coords - start) / psides) + 1
-    @inbounds for i in 1:Dim
-      c[i] = clamp(c[i], 1, nblocks[i])
-    end
+    c = @. clamp(c, 1, nblocks)
     bcoords = CartesianIndex(Tuple(c))
 
     # block index
--- a/src/Meshes.jl
+++ b/src/Meshes.jl
@@ -155,14 +155,12 @@ export
 
   # point sets
   PointSet,
-  coordinates, coordinates!,
   nelements,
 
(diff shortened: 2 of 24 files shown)


src/partitioning/block.jl
```diff
@@ -36,21 +35,18 @@ function partition(object, method::BlockPartition)
...
   for j in 1:nelements(object)
-    coordinates!(coords, object, j)
+    coords = coordinates(centroid(object, j))
 
     # find block coordinates
     c = @. floor(Int, (coords - start) / psides) + 1
-    @inbounds for i in 1:Dim
-      c[i] = clamp(c[i], 1, nblocks[i])
-    end
+    c = @. clamp(c, 1, nblocks)
```
serenity4, line 49: Why use the `@.` macro here instead of broadcasting `clamp` manually?

serenity4 (approved):
That's a lot of changes in places I am absolutely not familiar with. Overall it looks fine, and I find the result cleaner. I can try a thorough review if you want, but I would need some time. Right now I'd say the replacement logic is good, so if tests and benchmarks are OK, it's fine for me.

src/partitioning/block.jl
juliohm: I think just for consistency with the above line, I could have used it directly in the clamp call indeed. Does it make a difference in terms of performance?

juliohm:
Yes, I realized that after I requested the review. The code was added so quickly and now we are finally polishing the final details.

I will merge it here and continue the fixes in downstream packages. Thanks for the quick reply 💯

PR #66, opened 2021-03-09
Invalidate adjacent ears after pop in FIST (fix #65)

### LINKED ISSUE #65: Corner cases in FIST algorithm
The FIST implementation we have is reasonably robust, but it is still failing with some difficult polygonal areas. In our test suite we have a bunch of polygonal area examples in the `data` directory that are loaded for testing. Tests are passing for most polygonal areas in the family `poly1...5.line`, for example:

# poly1.line
image
image

# poly2.line
image
image

# poly3.line
But some cases like `poly3.line` are not passing the visual test:

image
image

Notice how some edges cross. That shouldn't be the case.

With the latest changes, we can also handle polygonal areas with holes in the `holes1...5.line` family, for example:

# hole2.line
image
image

# hole1.line
But some cases like `hole1.line` have crossing edges too:

image
image

Notice how the left inner hole is crossed by edges.

Finally we have polygonal areas in the `smooth1...5.line` family that are working fine:

# smooth2.line
image
image

# smooth1.line
But similarly, some cases fail with crossing edges:

image
image

It would be nice to dive in debug mode to figure out what is happening. Both `PolyArea` and `UnstructuredMesh` can be plotted directly with our current plot recipes so that helps too in the debugging process.

If someone can read the FIST paper by Held and then read our implementation, which basically follows the exact same notation, that would be great. I think I've implemented all the ingredients except for one or two corner cases they discuss, which may be related to the issues we are seeing.

If someone wants to dive in, I can help explaining the current implementation in a video chat for example.

serenity4:
The bug was a missing update step for the ears list. Adjacent ears were not discarded after one was popped.
I will update the tests tomorrow.

### DIFF
--- a/src/discretization/fist.jl
+++ b/src/discretization/fist.jl
@@ -50,8 +50,13 @@ function discretize(polyarea::PolyArea, ::FIST)
       𝒫 = Chain(points[inds])
       n = nvertices(𝒫)
       # 3. update 𝒬 near clipped ear
-      update_adjacent_ear!(𝒬, 𝒫, i-1, n)
-      update_adjacent_ear!(𝒬, 𝒫, i, n)
+      for j in (i-1, i)
+        if isear(𝒫, j)
+          𝒬 = 𝒬 ∪ [mod1(j,n)]
+        else
+          𝒬 = setdiff(𝒬, [mod1(j,n)])
+        end
+      end
       clipped = true
     elseif clipped # recompute all ears
       𝒬 = ears(𝒫)
@@ -141,16 +146,4 @@ function isearccw(𝒫::Chain{Dim,T}, i) where {Dim,T}
   incones = incone(i-1, i+1) && incone(i+1, i-1)
 
   isconvex && !intersects && incones
-end
-
-function update_adjacent_ear!(𝒬, 𝒫, i, n)
-  if !isear(𝒫, i)
-    ind = findfirst(==(i), 𝒬)
-    !isnothing(ind) && deleteat!(𝒬, ind)
-  else
-    if mod1(i,n) ∉ 𝒬
-      ind = something(findfirst(>(mod1(i,n)), 𝒬), 1)
-      insert!(𝒬, ind, mod1(i,n))
-    end
-  end
 end
--- a/src/discretization/fist.jl
+++ b/src/discretization/fist.jl
@@ -50,8 +50,8 @@ function discretize(polyarea::PolyArea, ::FIST)
       𝒫 = Chain(points[inds])
       n = nvertices(𝒫)
       # 3. update 𝒬 near clipped ear
-      isear(𝒫, i)   && (𝒬 = 𝒬 ∪ [mod1(i,n)])
-      isear(𝒫, i+1) && (𝒬 = 𝒬 ∪ [mod1(i+1,n)])
+      update_adjacent_ear!(𝒬, 𝒫, i-1, n)
+      update_adjacent_ear!(𝒬, 𝒫, i, n)
       clipped = true
     elseif clipped # recompute all ears
       𝒬 = ears(𝒫)
@@ -141,4 +141,16 @@ function isearccw(𝒫::Chain{Dim,T}, i) where {Dim,T}
   incones = incone(i-1, i+1) && incone(i+1, i-1)
 
   isconvex && !intersects && incones
+end
+
+function update_adjacent_ear!(𝒬, 𝒫, i, n)
+  if !isear(𝒫, i)
+    ind = findfirst(==(i), 𝒬)
+    !isnothing(ind) && deleteat!(𝒬, ind)
+  else
+    if mod1(i,n) ∉ 𝒬
+      ind = something(findfirst(>(mod1(i,n)), 𝒬), 1)
+      insert!(𝒬, ind, mod1(i,n))
+    end
+  end
 end
--- a/test/discretization.jl
+++ b/test/discretization.jl
@@ -1,4 +1,13 @@
 @testset "Discretization" begin
+  function is_fully_connected(mesh::UnstructuredMesh)
+    inds = 1:length(vertices(mesh))
+    connected = Iterators.flatten(getproperty.(mesh.connec, :list))
+    all(inds .== sort(unique(connected)))
+  end
+  function has_same_vertices(𝒫::PolyArea, mesh::UnstructuredMesh)
+    Set(vcat(vertices.(chains(𝒫))...)) == Set(vertices(mesh))
+  end
+
   @testset "FIST" begin
     𝒫 = Chain(P2[(0,0),(1,0),(1,1),(2,1),(2,2),(1,2),(0,0)])
     @test Meshes.ears(𝒫) == [2,4,5]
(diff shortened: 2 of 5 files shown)


src/discretization/fist.jl
```diff
@@ -142,3 +144,15 @@ function isearccw(𝒫::Chain{Dim,T}, i) where {Dim,T}
 
   isconvex && !intersects && incones
 end
+
+function remove_adjacent_ears!(𝒬, i)
+  ind = 1
+  while ind <= length(𝒬)
```
serenity4: Instead of traversing all the ears, assuming they are ordered, we can just check for the i-2 and i+2 ones.

serenity4:
Hmm, tests pass for Float32 but not for Float64. The FIST algorithm does not create the same connectivity for both types (only in the smooth1.line case), probably because of floating point errors.
Maybe we should remove these tests with hardcoded connectivities? We now already check that we get a triangulation that covers all the mesh and with the same vertices as the initial `PolyArea`. Changes to the algorithm may trigger valid triangulations, but will almost always require re-tweaking these connectivity values, to the point that I don't think it makes sense to test that.

src/discretization/fist.jl
juliohm: I am confused because the original algorithm doesn't treat the neighbor ears immediately after the pop, but we can discuss over a call today the details.

serenity4:
I refactored a bit the algorithm, it does different but valid triangulations. Which one is better may be subjective, although I prefer the old version. See the new version (not pushed to this PR yet):
Scene

And see the old version (with this PR's commits):
Scene_old

serenity4:
I ended up removing the tests with hardcoded triangulations, since the triangulation may change and still have a correct result. This is an implementation detail, and re-generating the triangulations by hand every time a change occurs would not be very useful. Instead we can definitely add more testing (check that triangles don't overlap for example).

serenity4:
Refactored (new) version: 
ezgif com-gif-maker

Old version:
ezgif com-gif-maker(2)

serenity4:
I may have spotted the reason for this change of behavior. The initial fix deleted ears ranging from `i-2` to `i+2`, while the most recent commit deletes only ears at `i-1` and `i+1` (and only if they are not ears of the new polygon).

serenity4:
Visually, we can also notice that the most recent version is consistent with the expected behavior, where triangle fans appear more naturally in the clipping process. I think we can review what should be the final version of this PR.

juliohm:
These are all super cool pictures! I will try to look at them carefully tomorrow morning 👍🏽

src/discretization/fist.jl
```diff
@@ -142,3 +142,18 @@ function isearccw(𝒫::Chain{Dim,T}, i) where {Dim,T}
...
   isconvex && !intersects && incones
 end
+
+function update_adjacent_ear!(𝒬, 𝒫, i, n)
+  if !isear(𝒫, i)
+    ind = findfirst(==(i), 𝒬)
+    !isnothing(ind) && deleteat!(𝒬, ind)
+    if !isnothing(ind)
+      println("$(i) was deleted from 𝒬 (index $ind)")
+    end
```
juliohm: I think this is left from your debugging process?
```suggestion
```

test/discretization.jl
```diff
@@ -1,4 +1,13 @@
 @testset "Discretization" begin
+  function is_fully_connected(mesh::UnstructuredMesh)
+    inds = collect(1:length(vertices(mesh)))
```
juliohm:
```suggestion
    inds = collect(1:nelements(mesh))
```

test/discretization.jl
```diff
@@ -1,4 +1,13 @@
...
+    connected = collect(Iterators.flatten(getproperty.(mesh.connec, :list)))
```
juliohm: We need to refactor this line because it is using internal behavior. Why not just compare the `Connectivity` objects directly? We shouldn't need to access the internal list of integers.

juliohm: Also it seems that the variable `connected` is not used later in the function. A typo?

test/discretization.jl
```diff
@@ -1,4 +1,13 @@
...
+    all(inds .== sort(unique(inds)))
+  end
+  function has_same_vertices(𝒫::PolyArea, mesh::UnstructuredMesh)
+    Set(vcat(vertices.(chains(𝒫))...)) == Set(vertices(mesh))
```
juliohm, line 8: If I remember correctly the most intuitive thing should work: `Set(vertices(P)) == Set(vertices(mesh))`

juliohm:
@serenity4 besides the suggestions above, I will try to get back to the PR tomorrow. I thought I would find the time to review it carefully today, but a couple of meetings today will inhibit me from coding.

Thanks for the incredible help as usual. My impression is that you nailed the bug 💯

test/discretization.jl
serenity4: The number of elements is the number of triangles. Here, we really want the number of vertices. However that makes me think that the number of triangles will exactly be `n-2`, `n` being the number of points, so I'll add this test.

test/discretization.jl
serenity4: `vertices` does not work on `PolyArea` objects:
```julia
Test threw exception
  Expression: Set(vertices(𝒫)) == Set(vertices(mesh))
  type PolyArea has no field vertices
```

test/discretization.jl
serenity4: Yes, that was a big typo, we have to compare to `connected`. We need the indices of each `Connectivity` object, to make sure they include all indices of the mesh. AFAIK there is no public way of doing that, which is why I rely on the `list` field. Since this is in tests, I don't think this poses problems, right? I'd rather avoid adding a public method in this PR that already eventually changes a lot of things.

serenity4:
There are a couple of things I spotted. `nvertices` does not work on `UnstructuredMesh` objects, and `vertices` does not work on `PolyArea` objects. Maybe something to fix in another PR? Also, AFAIK we don't have a nice way for accessing the indices of a `Connectivity` element.

src/discretization/fist.jl
```diff
@@ -50,8 +50,8 @@ function discretize(polyarea::PolyArea, ::FIST)
       𝒫 = Chain(points[inds])
       n = nvertices(𝒫)
       # 3. update 𝒬 near clipped ear
-      isear(𝒫, i)   && (𝒬 = 𝒬 ∪ [mod1(i,n)])
-      isear(𝒫, i+1) && (𝒬 = 𝒬 ∪ [mod1(i+1,n)])
+      update_adjacent_ear!(𝒬, 𝒫, i-1, n)
+      update_adjacent_ear!(𝒬, 𝒫, i, n)
```
juliohm: Can we replace these two calls by a single function call `update_adjacent_ears!(Q, P, i)` that takes care of both adjacent years? That way the logic is fully encapsulated in the function without having to keep track of a sequence of calls with changing indices.

src/discretization/fist.jl
```diff
@@ -142,3 +142,15 @@ function isearccw(𝒫::Chain{Dim,T}, i) where {Dim,T}
...
+function update_adjacent_ear!(𝒬, 𝒫, i, n)
+  if !isear(𝒫, i)
+    ind = findfirst(==(i), 𝒬)
+    !isnothing(ind) && deleteat!(𝒬, ind)
+  else
+    if mod1(i,n) ∉ 𝒬
+      ind = something(findfirst(>(mod1(i,n)), 𝒬), 1)
+      insert!(𝒬, ind, mod1(i,n))
+    end
+  end
```
juliohm: Perhaps we can simplify this code considerably. The logic seems very simple assuming we have a single function call to update both years:

```julia
function update_adjacent_ears!(Q, P, i)
  # check if adjacent ears are valid ears
  for j in [i-1, i+1]
    if valid(j)
      # add to Q
    else
      # remove from Q
    end
  end
end
```
I will try to refactor it locally.

juliohm:
Regarding the suggestions for another PR, I fully agree.

juliohm:
Sorry I messed up the PR with an incorrect push. Git is so unnecessarily complicated :/

juliohm:
Can I simply start a fresh PR with the minimum changes we discussed initially? It would consist of only updating the Q correctly, nothing more. Sorry again for the Git mess. I am surprised how I can still get it wrong.

serenity4:
Fine, let's get #68 merged and add the additional FIST tests in another PR then.

src/discretization/fist.jl
serenity4: You could also use broadcasting on `[i-1, i]` or map, if you want a single function call, but that's mostly a matter of style.

PR #68, opened 2021-03-12
Fix FIST updates

### LINKED ISSUE #65: Corner cases in FIST algorithm

juliohm:
Fix #65 Supersede #66

### DIFF
--- a/src/discretization/fist.jl
+++ b/src/discretization/fist.jl
@@ -50,8 +50,13 @@ function discretize(polyarea::PolyArea, ::FIST)
       𝒫 = Chain(points[inds])
       n = nvertices(𝒫)
       # 3. update 𝒬 near clipped ear
-      isear(𝒫, i)   && (𝒬 = 𝒬 ∪ [mod1(i,n)])
-      isear(𝒫, i+1) && (𝒬 = 𝒬 ∪ [mod1(i+1,n)])
+      for j in (i-1, i)
+        if isear(𝒫, j)
+          𝒬 = 𝒬 ∪ [mod1(j,n)]
+        else
+          𝒬 = setdiff(𝒬, [mod1(j,n)])
+        end
+      end
       clipped = true
     elseif clipped # recompute all ears
       𝒬 = ears(𝒫)
--- a/src/discretization/fist.jl
+++ b/src/discretization/fist.jl
@@ -54,7 +54,7 @@ function discretize(polyarea::PolyArea, ::FIST)
         if isear(𝒫, j)
           𝒬 = 𝒬 ∪ [mod1(j,n)]
         else
-          𝒬 = setdiff(𝒬, [mod1(j,n)])
+          setdiff!(𝒬, [mod1(j,n)])
         end
       end
       clipped = true
--- a/test/discretization.jl
+++ b/test/discretization.jl
@@ -32,93 +32,54 @@
     points = P2[(0,0),(1,0),(1,1),(2,1),(2,2),(1,2),(0,0)]
     connec = connect.([(4,5,6),(3,4,6),(3,6,1),(1,2,3)], Triangle)
     target = UnstructuredMesh(points[1:end-1], connec)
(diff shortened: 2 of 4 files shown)


juliohm:
@serenity4 I've started this clean PR with a simplified version of the fix you proposed in #66 . I messed up the other PR trying to improve it. Also, we can work on the missing features before we add tests for this one. For example, the missing `vertices(polyarea)` method.

Right now the plan is to make sure that the results look correct visually. There is a case `smooth2.line` that is hanging in single precision, and I realized that some tests in the test/polytopes.jl are still failing in single precision.

Can we merge this simple fix here, and continue working on separate PRs before we come back to handle `smooth2.line`?

serenity4 (approved):
Looks good!

src/discretization/fist.jl
```diff
@@ -50,8 +50,13 @@ function discretize(polyarea::PolyArea, ::FIST)
...
       # 3. update 𝒬 near clipped ear
-      isear(𝒫, i)   && (𝒬 = 𝒬 ∪ [mod1(i,n)])
-      isear(𝒫, i+1) && (𝒬 = 𝒬 ∪ [mod1(i+1,n)])
+      for j in (i-1, i)
+        if isear(𝒫, j)
+          𝒬 = 𝒬 ∪ [mod1(j,n)]
+        else
+          𝒬 = setdiff(𝒬, [mod1(j,n)])
+        end
+      end
```
serenity4: Quick note: this is not in-place, so likely slower that the original PR with unnecessary allocations. But we can do an optimization pass later and not worry about that now.

src/discretization/fist.jl
```diff
@@ -50,8 +50,13 @@ function discretize(polyarea::PolyArea, ::FIST)
       𝒫 = Chain(points[inds])
       n = nvertices(𝒫)
       # 3. update 𝒬 near clipped ear
-      isear(𝒫, i)   && (𝒬 = 𝒬 ∪ [mod1(i,n)])
-      isear(𝒫, i+1) && (𝒬 = 𝒬 ∪ [mod1(i+1,n)])
+      for j in (i-1, i)
+        if isear(𝒫, j)
+          𝒬 = 𝒬 ∪ [mod1(j,n)]
+        else
+          𝒬 = setdiff(𝒬, [mod1(j,n)])
```
serenity4:
```suggestion
          setdiff!(𝒬, [mod1(j,n)])
```

src/discretization/fist.jl
juliohm: Yep. I think we can work on performance optimizations in a separate pass. We are still in the process of making sure the algorithm is working smoothly with various example polygons.

PR #72, opened 2021-03-12
Fix co-linear check in intersection algorithm

### LINKED ISSUE #71: Intersection of segments in single precision
Found an issue with the last branch of the intersection algorithm:

```julia
s1 = Segment(Point(0.94495744f0, 0.53224397f0), Point(0.94798386f0, 0.5344541f0))
s2 = Segment(Point(0.94798386f0, 0.5344541f0), Point(0.9472896f0, 0.5340202f0))

s1 ∩ s2
Segment{2,Float32}
  └─Point(0.9472896f0, 0.5340202f0)
  └─Point(0.94798386f0, 0.5344541f0)
```
image

In this plot `s1` is the black segment and `s2` is the red segment. The intersection algorithm reaches the `CASE (III)` correctly, and then tries to decide whether or not these two segments are co-linear:

```julia
elseif !determinatex && !determinatey # CASE (III)
    if !isapprox((x2 - x1) × (y2 - y1), zero(T), atol=atol(T))
      # configuration (3)
      CornerTouchingSegments(intersectpoint(s1, s2))
    else
      # configuration (3), (4) or (5)
      intersectcolinear(s1, s2)
    end
  end
```

I am using the cross product to check for co-linearity, but the tolerance seems to be too tight. The `atol(T)` should return a reasonable tolerance for distances, not areas or volumes, I think that is causing the issue. Based on the plot, we would like to return `configuration (3)` because the segments aren't overlapping and aren't co-linear.

How should we set the tolerance in this case? Or should we use a different condition to check for co-linearity that doesn't use the cross product?

juliohm:
Fix #71


juliohm:
Interesting, the intersections.jl benchmarks are not showing. I did something wrong?

And the build is failing on Windows for some unrelated reason with the CI bot.

serenity4:
That's odd, the benchmark from `intersections.jl` is shown but not on the first table. Configuration wise it looks OK, so I don't know why it's doing that.
I'll take a look at the PR tomorrow, I am away from the computer tonight.

src/intersections/segments.jl
```diff
@@ -79,7 +79,9 @@ function intersecttype(s1::Segment{2,T}, s2::Segment{2,T}) where {T}
       NonIntersectingSegments()
     end
   elseif !determinatex && !determinatey # CASE (III)
-    if !isapprox((x2 - x1) × (y2 - y1), zero(T), atol=atol(T))
+    u, v = (x2 - x1), (y2 - y1)
+    α = (u ⋅ v) / (v ⋅ v)
+    if !isapprox(norm(u - α*v), zero(T), atol=atol(T))
```
serenity4:
```suggestion
    if !isapprox(u × v, zero(T), atol=atol(T)^2)
```
It passes tests on my machine (including the additional one from the PR), and is way simpler. Would there be any reason for not using `atol(T)^2`?

serenity4 (changes requested):

src/intersections/segments.jl
serenity4: Also, now that `u` and `v` are not used multiple times, it may be better to just leave this as it was, with the only change being that we test for `atol(T)^2`, not introducing `u` and `v`.

juliohm: Nice. I was not sure if the `atol(T)^2` was safe. Perhaps there is a material somewhere explaining how these floating point errors propagate in area calculations? I will commit your suggestion and we see if we bump into this again in the future. 👍🏽

serenity4: The beginning of [this video](https://www.youtube.com/watch?v=zHPXGBiTM5A) explains very well how floats behave in numerical operations. If I understood correctly, when comparing to `zero(T)`, it's just going to check whether the result is between `[-atol(T), atol(T)]` since `zero(T)` is represented perfectly on floats, so there should be no problem here. Problems would arise if, for example, you do an area calculation, then take the square root of that.

serenity4 (approved):

src/intersections/segments.jl
```diff
@@ -79,7 +79,8 @@ function intersecttype(s1::Segment{2,T}, s2::Segment{2,T}) where {T}
       NonIntersectingSegments()
     end
   elseif !determinatex && !determinatey # CASE (III)
-    if !isapprox((x2 - x1) × (y2 - y1), zero(T), atol=atol(T))
+    u, v = (x2 - x1), (y2 - y1)
+    if !isapprox(u × v, zero(T), atol=atol(T)^2)
```
```suggestion
    if !isapprox((x2 - x1) × (y2 - y1), zero(T), atol=atol(T)^2)
```

PR #79, opened 2021-03-15
Add segments function for chains

serenity4:
This PR adds a utility function that returns all the consecutive segments of a `Chain` object.

### DIFF
--- a/src/polytopes/chain.jl
+++ b/src/polytopes/chain.jl
@@ -39,9 +39,9 @@ end
 Return the segments linking consecutive points of the `chain`.
 """
 function segments(c::Chain)
-  vs = vertices(c)
+  vs = c.vertices
   n = length(vs)
-  [Segment(view(vs, [i,i+1])) for i in 1:n-1]
+  (Segment(view(vs, [i,i+1])) for i in 1:n-1)
 end
 
 """
@@ -63,7 +63,7 @@ intersect at end points.
 """
 function issimple(c::Chain)
   vs = c.vertices
-  ss = segments(c)
+  ss = collect(segments(c))
   for i in 1:length(ss)
     for j in i+1:length(ss)
       I = intersecttype(ss[i], ss[j])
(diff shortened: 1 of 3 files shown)


serenity4:
Hmm, I have no clue on why the PR failed. Is this a CI bug? I can't even see the log.

src/polytopes/chain.jl
```diff
@@ -33,6 +33,17 @@ function vertices(c::Chain)
   end
 end
 
+"""
+    segments(chain)
+
+Return the segments linking consecutive points of the `chain`.
+"""
+function segments(c::Chain)
+  vs = vertices(c)
```
juliohm: I think in this case we need to use `c.vertices` to always get all the points including the last one which may be equal to the first one in case of closed chains. The function `vertices(c)` does a more complicated logic to remove the repeated one.

src/polytopes/chain.jl
```diff
@@ -33,6 +33,17 @@ function vertices(c::Chain)
...
+  n = length(vs)
+  [Segment(view(vs, [i,i+1])) for i in 1:n-1]
```
juliohm: We could return a generator instead of materializing it. Just replace the brackets by parenthesis.

test/polytopes.jl
```diff
@@ -133,6 +133,10 @@
   end
 
   @testset "Chains" begin
+    # segments
+    c = Chain(P2[(1,1),(2,2),(3,3)])
+    @test segments(c) == [Segment(P2(1,1),P2(2,2)),Segment(P2(2,2),P2(3,3))]
+
```
juliohm, line 142: We should also add a test with closed chains to make sure things are working as expected.

juliohm (changes requested):

juliohm:
Yes, the CI build on Windows is problematic. I've seen this before where the error message is not even shown. I guess it is an issue on the GitHub side.

serenity4:
I think it's good now. Tests pass locally.

juliohm (approved):

PR #83, opened 2021-03-16
Parametrize Chain vector type

### LINKED ISSUE #81: Additional chain constructors
The `Chain` type is missing a few constructors. For example, we can't initialize one with a static vector:
```julia
julia> using Meshes, StaticArrays

julia> Chain(@SVector [Point2(0.,0.),Point2(1.,1.)])
ERROR: MethodError: no method matching Chain(::SVector{2, Point2})
Closest candidates are:
  (::Type{PL})(::TP...) where {PL<:Polytope, TP<:Tuple} at /home/belmant/.julia/dev/Meshes/src/polytopes.jl:29
  (::Type{PL})(::P...) where {PL<:Polytope, P<:Point} at /home/belmant/.julia/dev/Meshes/src/polytopes.jl:27
  Chain(::CircularArrays.CircularVector{T, A} where {T, A}) at /home/belmant/.julia/dev/Meshes/src/polytopes/chain.jl:15
  ...
Stacktrace:
 [1] top-level scope
   @ REPL[7]:1
```

nor can we with just points (as varargs), although the docstring mentions `Chain(p1, p2, ..., pn)` at the struct definition:
```julia
julia> Chain(Point2(0.,0.),Point2(1.,1.))
ERROR: MethodError: no method matching Chain(::SVector{2, Point2})
Closest candidates are:
  (::Type{PL})(::TP...) where {PL<:Polytope, TP<:Tuple} at /home/belmant/.julia/dev/Meshes/src/polytopes.jl:29
  (::Type{PL})(::P...) where {PL<:Polytope, P<:Point} at /home/belmant/.julia/dev/Meshes/src/polytopes.jl:27
  Chain(::CircularArrays.CircularVector{T, A} where {T, A}) at /home/belmant/.julia/dev/Meshes/src/polytopes/chain.jl:15
  ...
Stacktrace:
 [1] Chain(::Point2, ::Vararg{Point2, N} where N)
   @ Meshes ~/.julia/dev/Meshes/src/polytopes.jl:27
 [2] top-level scope
   @ REPL[8]:1
```

The second case should be fixed if the first one is too.

serenity4:
Fixes #81. Adds two constructors to `Chain`, namely one that takes arbitrary vectors (`SVector`s being a special case), and one that takes vararg points.


src/polytopes/chain.jl
```diff
@@ -15,6 +15,9 @@ end
 Chain(vertices::CircularVector) =
   Chain([collect(vertices); vertices[begin]])
 
+Chain(vertices::AbstractVector) = Chain(collect(vertices))
```
serenity4: I wasn't sure whether to juxtapose it to the previous definition, but I think it's more readable this way.

juliohm: I wonder if we need this one explicitly given that `Polytope` already defines a constructor for AbstractVector. Is it am ambiguity issue?

src/polytopes/chain.jl
```diff
@@ -15,6 +15,9 @@ end
...
+Chain(vertices::AbstractVector{<:Tuple}) = Chain(Point.(vertices))
+
```
juliohm: Similar for this one. Why the `Polytope` constructors are not being propagated?

src/polytopes/chain.jl
serenity4: `Polytope` defines this constructor for vectors:
```julia
(::Type{PL})(vertices::AbstractVector{TP}) where {PL<:Polytope,TP<:Tuple} = PL(Point.(vertices))
```
But it won't work for `AbstractVector{<:Point}`, it works only for tuples. Would it be better maybe to extend to above `Chain` constructor for all polytopes?

src/polytopes/chain.jl
serenity4: This one came from the ambiguity error above. This can be removed if we restrict the above to only handle points though, which may be better.

src/polytopes/chain.jl
serenity4: The problem is that certain polytopes may have a `vertices` field that is a `Vector` (like `Chain`s), or `AbstractVector` which may be `SVector`s or not. Unless we change `Chain` to parametrize its vector type, we will have to redefine a special constructor for `Chain`.

juliohm: Yes, I think I tried adding this constructor to all polytopes at some point but it caused ambiguities down the line. Can you give it a try? I think that was the reason why I didn't add these constructors to Chain so far. Ideally we would be able to have them in all Polytope.

serenity4: We could either
- assume that all polytopes will prefer `SVector`s
- have polytopes that don't want a `SVector` redefine their own constructor
or
- assume that all polytopes will prefer `Vector`
- have polytopes that don't want a `Vector` redefine their own constructor
or
- make no assumptions on the preferred vector type
- have all polytopes rely on their internal constructor
- have all polytopes redefine their own version of the definition below (which would be removed)

```julia
(::Type{PL})(vertices::Vararg{P}) where {PL<:Polytope,P<:Point} = PL(SVector(vertices))
```

juliohm: Taking the time to process this important design decision. I understand from your bullet points that we have 3 main alternative designs, correct? Could you please group them differently so that we can reason more clearly about it. :)

juliohm: Oh I see the "or" you typed there now. Nevermind.

serenity4: It's actually a bit more nasty. I'll make a few experiments and report the possibilities. Feel free to do the same if you are at ease with constructor shenanigans.

juliohm: Thanks for experimenting with these ideas. I am not very good at constructor shenanigans so I am glad you can help on this front :)

serenity4:
There we go. Ready to merge upon approval.
Note that the `Segment` constructor tests are not related to the change, since `Segment` is unaffected. I left them there because I believe it is good to have for the future.

serenity4:
Feel free to remove them if you prefer @juliohm.

test/polytopes.jl
```diff
@@ -1,5 +1,12 @@
 @testset "Polytopes" begin
   @testset "Segment" begin
+    # constructors
+    s1 = Segment(P2[(1,1),(2,2),(1,1)])
+    s2 = Segment(P2(1,1),P2(2,2),P2(1,1))
+    s3 = Segment(@SVector P2[(1,1),(2,2),(1,1)])
+    s4 = Segment(T.((1,1.)),T.((2.,2.)),T.((1.,1.)))
+    @test s1 == s2 == s3 == s4
```
juliohm: But Segment is only a straight segment between two points. Why we would like to add tests with three points?

juliohm (changes requested):
The Segment is between two points only. Can you elaborate on why you added these tests? Maybe there is something we are interpreting differently here.

juliohm:
Cool. Time to mess up another GitHub PR with my poor Git skills. :D I think I know what happened the other time. Let me try this time...

test/polytopes.jl
serenity4: Ooh. That's a remnant of a copy-paste that I didn't correct. It was originally to test that the constructors we have for polytopes work on something else than a `Chain`.

serenity4:
Rebase, don't pull. I think that's what happened.

test/polytopes.jl
```diff
@@ -1,5 +1,12 @@
...
+
```
```suggestion
```

juliohm:
I love the GitHub commit on PR feature. Saves a lot of Git headache. Just erased the tests using the browser. 👌🏽

src/polytopes/chain.jl
```diff
@@ -8,10 +8,11 @@
 A polygonal chain from a sequence of points `p1`, `p2`, ..., `pn`.
 See https://en.wikipedia.org/wiki/Polygonal_chain.
 """
-struct Chain{Dim,T} <: Polytope{1,Dim,T}
-  vertices::Vector{Point{Dim,T}}
+struct Chain{Dim,T,V<:AbstractVector{Point{Dim,T}}} <: Polytope{1,Dim,T}
+  vertices::V
 end
 
+Chain(vertices::Vararg{<:Point}) = Chain(collect(vertices))
```
juliohm, line 16:
```suggestion
# specialize Polytope outer constructor to use standard vector
Chain(vertices::Vararg{<:Point}) = Chain(collect(vertices))

```

juliohm (approved):

juliohm:
Just waiting for the tests now. 👍🏽

Issue #84, opened 2021-03-19
Make parametric objects callable


serenity4:
Some geometric objects possess a natural parametric space. For example, segments linearly interpolate between two points, rays can be evaluated from the origin to the infinity, bezier curves can be evaluated between its endpoints... I had the idea of making these objects callable on their parametric inputs, so a segment could be called with `t` going from 0 to 1 to make a linear interpolation of its two points. Considering objects with a single parametric dimension it should be fine, unambiguous and easy to do. I can draft a PR to see how it goes for one-dimensional parametric objects.

When there is more than one parametric dimension, I am not sure how this should be implemented. For triangles, for example, it is possible to use barycentric coordinates (from 0 to 1), but also two-dimensional coordinates within a certain range (since this is a polygon after all, with parametric dimension 2). Feel free to comment if you have any ideas.

- [x] Segment
- [x] Ray
- [x] BezierCurve
- [x] Line
- [x] Triangle
- [x] Quadrangle

juliohm:
Yes, it would be great to parameterize these geometries when this is possible. We can start with the ones of parametric dimension 1 as you suggested 👍🏽

juliohm:
As an update, we currently have `Segment`, `Ray` and `BezierCurve` as callables.

GuoHaoxuan:
For `Triangle` and `Quadrangle` is should be solved via barycentric coordinate system. (Same as `Segment`)

But in `Triangle`, it will has 2 parameters due to u+v+w=1. Which point is special that not included in the parameters in a triangle? Is it reasonable to use the 2 points in the front?

It is the same question in `Quadrangle`.

juliohm:
@HaoxuanGuo the first two points of the triangle are a good simple choice :) Would you mind submitting a PR for that? We have barycentric coordinates in some methods like [this one](https://github.com/JuliaGeometry/Meshes.jl/blob/982b902e635c1789850400cdcd3b2781592c8c67/src/sampling/homogeneous.jl#L38-L44). If you could add a functor method to the `Triangle` with two arguments and could replace the function linked by a simple call with random arguments, that would be great.

I forgot to check the quadrangle case, it is already solved [here](https://github.com/JuliaGeometry/Meshes.jl/blob/982b902e635c1789850400cdcd3b2781592c8c67/src/polytopes/ngon.jl#L168-L171) 👌🏽

PR #85, opened 2021-03-19
Make 1D parametric types callable

serenity4:
Done for `Segment`.

### DIFF
--- a/src/polytopes/segment.jl
+++ b/src/polytopes/segment.jl
@@ -6,11 +6,21 @@
     Segment(p1, p2)
 
 An oriented line segment with end points `p1`, `p2`.
+The segment can be called as `s(t)` with `t` between
+`0` and `1` to interpolate linearly between its endpoints.
 
 See also [`Line`](@ref).
 """
 struct Segment{Dim,T,V<:AbstractVector{Point{Dim,T}}} <: Polytope{1,Dim,T}
   vertices::V
+end
+
+function (s::Segment)(t)
+  if t < 0 || t > 1
+    throw(DomainError(t, "s(t) is not defined for t outside [0, 1]."))
+  end
+  p1, p2 = s.vertices
+  p1 + t * (p2 - p1)
 end
 
 measure(s::Segment) = norm(s.vertices[2] - s.vertices[1])
--- a/src/primitives/ray.jl
+++ b/src/primitives/ray.jl
@@ -6,10 +6,19 @@
     Ray(p, v)
 
 A ray originating at point `p`, pointed in direction `v`.
+It can be called as `r(t)` with `t > 0` to cast it at
+`p + t * v`.
 """
 struct Ray{Dim,T} <: Primitive{Dim,T}
   p::Point{Dim,T}
   v::Vec{Dim,T}
+end
+
+function (r::Ray)(t)
+  if t < 0
+    throw(DomainError(t, "r(t) is not defined for t < 0."))
+  end
+  r.p + t * r.v
 end
 
 paramdim(::Type{<:Ray}) = 1
--- a/test/polytopes.jl
+++ b/test/polytopes.jl
@@ -97,6 +97,7 @@
     @test s1(T(0.)) == P2(0., 0.)
     @test s1(T(1.)) == P2(1., 1.)
     @test_throws DomainError(T(1.2), "s(t) is not defined for t outside [0, 1].") s1(T(1.2))
+    @test_throws DomainError(T(-0.5), "s(t) is not defined for t outside [0, 1].") s1(T(-0.5))
   end
 
   @testset "Triangles" begin
--- a/test/polytopes.jl
+++ b/test/polytopes.jl
@@ -92,6 +92,11 @@
     s1 = Segment(P2(0.94495744, 0.53224397), P2(0.94798386, 0.5344541))
     s2 = Segment(P2(0.94798386, 0.5344541), P2(0.9472896, 0.5340202))
     @test s1 ∩ s2 == s2 ∩ s1 == P2(0.94798386, 0.5344541) 
+
+    s1 = Segment(P2(0.,0.), P2(1., 1.))
+    @test s1(T(0.)) == P2(0., 0.)
+    @test s1(T(1.)) == P2(1., 1.)
+    @test_throws DomainError(T(1.2), "s(t) is not defined for t outside [0, 1].") s1(T(1.2))
   end
 
   @testset "Triangles" begin
--- a/test/primitives.jl
+++ b/test/primitives.jl
@@ -8,6 +8,10 @@
   @testset "Rays" begin
     r = Ray(P2(0,0), V2(1,1))
     @test paramdim(r) == 1
+    @test r(T(0.)) == P2(0,0)
+    @test r(T(1.)) == P2(1,1)
+    @test r(T(Inf)) == P2(Inf,Inf)
+    @test_throws DomainError(T(-1), "r(t) is not defined for t < 0.") r(T(-1))
   end
 
   @testset "Boxes" begin


src/polytopes/segment.jl
```diff
@@ -6,13 +6,23 @@
...
 See also [`Line`](@ref).
 """
 struct Segment{Dim,T,V<:AbstractVector{Point{Dim,T}}} <: Polytope{1,Dim,T}
   vertices::V
 end
 
+function (s::Segment)(t)
+  if t < 0 || t > 1
+    throw(DomainError(t, "s(t) is not defined for t outside [0, 1]."))
+  end
```
juliohm, line 21: Would this if-else branch create a Union result as we seen in the intersection algorithm? Maybe we can just assume the user will call with any t of interest? Extrapolation should be fine if the performance is compromised with the if-else branch.

juliohm (changes requested):
Perhaps we should avoid the if-else check to not compromise the performance of the evaluation. What do you think?

src/polytopes/segment.jl
serenity4: No, this does not affect performance. The compiler assumes that the branch will never be executed (as far as type inferrence is concerned) because it throws an error. I benchmarked it locally and there is no performance loss. If you call `asin(-2)`, you can see that it is done for trigonometric functions in `Base`, so I think there is no downside to making this check.

serenity4: In the intersection algorithm, it had an impact only when `@error` was present in the branch, because that is not an exception and it actually returned a value. When transformed to an `error` (to throw an exception), the inference worked fine. Though in the end, you just removed the `@error` (which I agree was a good thing), so we did not have the trouble anymore.

serenity4: From `Base` at `special/trig.jl:425`:
```julia
@noinline asin_domain_error(x) = throw(DomainError(x, "asin(x) is not defined for |x|>1."))
function asin(x::T) where T<:Union{Float32, Float64}
    # Since  asin(x) = x + x^3/6 + x^5*3/40 + x^7*15/336 + ...
    # we approximate asin(x) on [0,0.5] by
    #     asin(x) = x + x*x^2*R(x^2)
    # where
    #     R(x^2) is a rational approximation of (asin(x)-x)/x^3
    # and its remez error is bounded by
    #     |(asin(x)-x)/x^3 - R(x^2)| < 2^(-58.75)
    absx = abs(x)
    if absx >= T(1.0) # |x|>= 1
        if absx == T(1.0)
            return flipsign(T(pi)/2, x)
        end
        asin_domain_error(x)
    elseif absx < T(1.0)/2
        # if |x| sufficiently small, |x| is a good approximation
        if absx < ASIN_X_MIN_THRESHOLD(T)
            return x
        end
        return muladd(x, arc_tRt(x*x), x)
    end
    # else 1/2 <= |x| < 1
    t = (T(1.0) - absx)/2
    return asin_kernel(t, x)
end
```

serenity4:
Good for `Segment` and `Ray`. For other geometric types, we'll have to think about it on #84 since there may be multiple possibilities, or sometimes it's just not possible without additional information. I think we should merge this if it is OK can think about the rest later.

src/polytopes/segment.jl
juliohm: Very nice. It is good to know that the compiler can figure out the throw expressions and avoid these Union types as result.

juliohm:
Just one more question regarding the if-else branch. Isn't it true that the performance will be affected by the branch? Even if the result type is not a Union, we are still forcing the check 0 <= t <= 1 on every function call right?

serenity4:
Ready to merge.

serenity4:
Yes, but I think this is very cheap. On local benchmarks, I didn't see anything standing out from the noise around ~15ns.

juliohm:
Can you comment on the last question above? Isn't the check a performance bottleneck in the extreme case where we have many evaluations with thousands of segments? I like these checks just wondering how much we loose in terms of performance as this is also part of our goals :)

juliohm:
Got it, and assuming that Base functions like `asin` and other do the same, it should indeed very negligible.

juliohm (approved):

serenity4:
Comparing the output from `@code_llvm`, I get for the unchecked case:
```julia
julia> @code_llvm s(0.5)
;  @ /home/belmant/.julia/dev/Meshes/src/polytopes/segment.jl:18 within `Segment'
define void @julia_Segment_2699([1 x [1 x [2 x double]]]* noalias nocapture sret %0, [1 x [1 x [2 x [1 x [1 x [2 x double]]]]]]* nocapture nonnull readonly align 8 dereferenceable(32) %1, double %2) {
top:
;  @ /home/belmant/.julia/dev/Meshes/src/polytopes/segment.jl:23 within `Segment'
; ┌ @ /home/belmant/.julia/dev/Meshes/src/points.jl:76 within `-' @ /home/belmant/.julia/packages/StaticArrays/LJQEe/src/linalg.jl:15
; │┌ @ /home/belmant/.julia/packages/StaticArrays/LJQEe/src/mapreduce.jl:37 within `map'
; ││┌ @ /home/belmant/.julia/packages/StaticArrays/LJQEe/src/mapreduce.jl:40 within `_map'
; │││┌ @ /home/belmant/.julia/packages/StaticArrays/LJQEe/src/mapreduce.jl:75 within `macro expansion'
; ││││┌ @ /home/belmant/.julia/packages/StaticArrays/LJQEe/src/SVector.jl:39 within `getindex' @ tuple.jl:29
       %3 = getelementptr inbounds [1 x [1 x [2 x [1 x [1 x [2 x double]]]]]], [1 x [1 x [2 x [1 x [1 x [2 x double]]]]]]* %1, i64 0, i64 0, i64 0, i64 1, i64 0, i64 0, i64 0
; ││││└
; ││││┌ @ float.jl:329 within `-'
       %4 = bitcast double* %3 to <2 x double>*
       %5 = load <2 x double>, <2 x double>* %4, align 8
       %6 = bitcast [1 x [1 x [2 x [1 x [1 x [2 x double]]]]]]* %1 to <2 x double>*
       %7 = load <2 x double>, <2 x double>* %6, align 8
       %8 = fsub <2 x double> %5, %7
; └└└└└
; ┌ @ /home/belmant/.julia/packages/StaticArrays/LJQEe/src/linalg.jl:20 within `*'
; │┌ @ broadcast.jl:821 within `broadcast'
; ││┌ @ broadcast.jl:883 within `materialize'
; │││┌ @ /home/belmant/.julia/packages/StaticArrays/LJQEe/src/broadcast.jl:26 within `copy'
; ││││┌ @ /home/belmant/.julia/packages/StaticArrays/LJQEe/src/broadcast.jl:99 within `_broadcast'
; │││││┌ @ /home/belmant/.julia/packages/StaticArrays/LJQEe/src/broadcast.jl:125 within `macro expansion'
; ││││││┌ @ float.jl:332 within `*'
         %9 = insertelement <2 x double> undef, double %2, i32 0
         %10 = shufflevector <2 x double> %9, <2 x double> undef, <2 x i32> zeroinitializer
         %11 = fmul <2 x double> %8, %10
; └└└└└└└
; ┌ @ /home/belmant/.julia/dev/Meshes/src/points.jl:85 within `+' @ /home/belmant/.julia/packages/StaticArrays/LJQEe/src/linalg.jl:11
; │┌ @ /home/belmant/.julia/packages/StaticArrays/LJQEe/src/mapreduce.jl:37 within `map'
; ││┌ @ /home/belmant/.julia/packages/StaticArrays/LJQEe/src/mapreduce.jl:40 within `_map'
; │││┌ @ /home/belmant/.julia/packages/StaticArrays/LJQEe/src/mapreduce.jl:75 within `macro expansion'
; ││││┌ @ float.jl:326 within `+'
       %12 = fadd <2 x double> %11, %7
; └└└└└
  %13 = bitcast [1 x [1 x [2 x double]]]* %0 to <2 x double>*
  store <2 x double> %12, <2 x double>* %13, align 8
  ret void
}
```

and for the checked version

```julia
julia> @code_llvm s(0.5)
;  @ /home/belmant/.julia/dev/Meshes/src/polytopes/segment.jl:18 within `Segment'
define void @julia_Segment_2701([1 x [1 x [2 x double]]]* noalias nocapture sret %0, [1 x [1 x [2 x [1 x [1 x [2 x double]]]]]]* nocapture nonnull readonly align 8 dereferenceable(32) %1, double %2) {
top:
  %3 = alloca [2 x {}*], align 8
  %gcframe10 = alloca [3 x {}*], align 16
  %gcframe10.sub = getelementptr inbounds [3 x {}*], [3 x {}*]* %gcframe10, i64 0, i64 0
  %.sub = getelementptr inbounds [2 x {}*], [2 x {}*]* %3, i64 0, i64 0
  %4 = bitcast [3 x {}*]* %gcframe10 to i8*
  call void @llvm.memset.p0i8.i32(i8* nonnull align 16 dereferenceable(24) %4, i8 0, i32 24, i1 false)
  %thread_ptr = call i8* asm "movq %fs:0, $0", "=r"() #7
  %ptls_i8 = getelementptr i8, i8* %thread_ptr, i64 -32768
;  @ /home/belmant/.julia/dev/Meshes/src/polytopes/segment.jl:19 within `Segment'
; ┌ @ float.jl:416 within `<' @ float.jl:371
   %5 = bitcast [3 x {}*]* %gcframe10 to i64*
   store i64 4, i64* %5, align 16
   %6 = getelementptr inbounds [3 x {}*], [3 x {}*]* %gcframe10, i64 0, i64 1
   %7 = bitcast i8* %ptls_i8 to i64*
   %8 = load i64, i64* %7, align 8
   %9 = bitcast {}** %6 to i64*
   store i64 %8, i64* %9, align 8
   %10 = bitcast i8* %ptls_i8 to {}***
   store {}** %gcframe10.sub, {}*** %10, align 8
   %11 = fcmp olt double %2, 0.000000e+00
; └
  %12 = fcmp ogt double %2, 1.000000e+00
  %value_phi.in = or i1 %11, %12
  br i1 %value_phi.in, label %L23, label %L26

L23:                                              ; preds = %top
;  @ /home/belmant/.julia/dev/Meshes/src/polytopes/segment.jl:20 within `Segment'
  %13 = call noalias nonnull {}* @jl_gc_pool_alloc(i8* %ptls_i8, i32 1400, i32 16) #1
  %14 = bitcast {}* %13 to i64*
  %15 = getelementptr inbounds i64, i64* %14, i64 -1
  store atomic i64 139965098456848, i64* %15 unordered, align 8
  %16 = bitcast {}* %13 to double*
  store double %2, double* %16, align 8
  %17 = getelementptr inbounds [3 x {}*], [3 x {}*]* %gcframe10, i64 0, i64 2
  store {}* %13, {}** %17, align 16
  store {}* %13, {}** %.sub, align 8
  %18 = getelementptr inbounds [2 x {}*], [2 x {}*]* %3, i64 0, i64 1
  store {}* inttoptr (i64 139962983444624 to {}*), {}** %18, align 8
  %19 = call nonnull {}* @jl_apply_generic({}* inttoptr (i64 139965143912816 to {}*), {}** nonnull %.sub, i32 2)
  call void @jl_throw({}* %19)
  unreachable

L26:                                              ; preds = %top
;  @ /home/belmant/.julia/dev/Meshes/src/polytopes/segment.jl:23 within `Segment'
; ┌ @ /home/belmant/.julia/dev/Meshes/src/points.jl:76 within `-' @ /home/belmant/.julia/packages/StaticArrays/LJQEe/src/linalg.jl:15
; │┌ @ /home/belmant/.julia/packages/StaticArrays/LJQEe/src/mapreduce.jl:37 within `map'
; ││┌ @ /home/belmant/.julia/packages/StaticArrays/LJQEe/src/mapreduce.jl:40 within `_map'
; │││┌ @ /home/belmant/.julia/packages/StaticArrays/LJQEe/src/mapreduce.jl:75 within `macro expansion'
; ││││┌ @ /home/belmant/.julia/packages/StaticArrays/LJQEe/src/SVector.jl:39 within `getindex' @ tuple.jl:29
       %20 = getelementptr inbounds [1 x [1 x [2 x [1 x [1 x [2 x double]]]]]], [1 x [1 x [2 x [1 x [1 x [2 x double]]]]]]* %1, i64 0, i64 0, i64 0, i64 1, i64 0, i64 0, i64 0
; ││││└
; ││││┌ @ float.jl:329 within `-'
       %21 = bitcast double* %20 to <2 x double>*
       %22 = load <2 x double>, <2 x double>* %21, align 8
       %23 = bitcast [1 x [1 x [2 x [1 x [1 x [2 x double]]]]]]* %1 to <2 x double>*
       %24 = load <2 x double>, <2 x double>* %23, align 8
       %25 = fsub <2 x double> %22, %24
; └└└└└
; ┌ @ /home/belmant/.julia/packages/StaticArrays/LJQEe/src/linalg.jl:20 within `*'
; │┌ @ broadcast.jl:821 within `broadcast'
; ││┌ @ broadcast.jl:883 within `materialize'
; │││┌ @ /home/belmant/.julia/packages/StaticArrays/LJQEe/src/broadcast.jl:26 within `copy'
; ││││┌ @ /home/belmant/.julia/packages/StaticArrays/LJQEe/src/broadcast.jl:99 within `_broadcast'
; │││││┌ @ /home/belmant/.julia/packages/StaticArrays/LJQEe/src/broadcast.jl:125 within `macro expansion'
; ││││││┌ @ float.jl:332 within `*'
         %26 = insertelement <2 x double> undef, double %2, i32 0
         %27 = shufflevector <2 x double> %26, <2 x double> undef, <2 x i32> zeroinitializer
         %28 = fmul <2 x double> %25, %27
; └└└└└└└
; ┌ @ /home/belmant/.julia/dev/Meshes/src/points.jl:85 within `+' @ /home/belmant/.julia/packages/StaticArrays/LJQEe/src/linalg.jl:11
; │┌ @ /home/belmant/.julia/packages/StaticArrays/LJQEe/src/mapreduce.jl:37 within `map'
; ││┌ @ /home/belmant/.julia/packages/StaticArrays/LJQEe/src/mapreduce.jl:40 within `_map'
; │││┌ @ /home/belmant/.julia/packages/StaticArrays/LJQEe/src/mapreduce.jl:75 within `macro expansion'
; ││││┌ @ float.jl:326 within `+'
       %29 = fadd <2 x double> %28, %24
; └└└└└
  %30 = bitcast [1 x [1 x [2 x double]]]* %0 to <2 x double>*
  store <2 x double> %29, <2 x double>* %30, align 8
  %31 = load i64, i64* %9, align 8
  store i64 %31, i64* %7, align 8
  ret void
}
```

There does seem to be a lot more going on, which is normal since it includes a different control flow with two exit points, but yeah, since performance wise it is fine and it is done in `Base`, I'd say it's fine also here. Plus, I don't think there would actually be situations that would cause a bottleneck around these function calls.

juliohm:
Tests are failing on Ubuntu for an unrelated reason. We need to update some of the figures in the test/data folder apparently.

Merging this one 💯

serenity4:
Another possibility would be to define a special function `evaluate(s::Segment)` without checks, and only include the check inside the call, but that is probably over-engineering (that's my opinion at least).

juliohm:
Yes I think it is fine to have the check. I just asked to follow the pedantic tradition :)

PR #86, opened 2021-03-20
Add Bezier curve primitive

serenity4:
Adds a new `BezierCurve` type, which can be evaluated by calling on its parametric variable. I implemented two methods, namely Horner's method and De Casteljau's algorithm, that can be used to evaluate it. The first is a fast approximation, useful with many control points, while the second is very accurate but also more expensive. I made the second one be the default, because with few control points (I tested with 3) the performance is very similar.


src/primitives/bezier.jl
```diff
@@ -0,0 +1,89 @@
...
+"""
+struct Horner <: BezierEvalMethod end
+
+(curve::BezierCurve)(t) = curve(t, DeCasteljau())
+
+function check_domain(::BezierCurve, t)
+  if t < 0 || t > 1
+    throw(DomainError(t, "b(t) is not defined for t outside [0, 1]."))
+  end
+end
```
serenity4: I wasn't sure whether I should define a function for this or not. The idea was not to copy paste this for the two algorithms. The first argument type is just used for dispatch. If that makes sense, I can extend this function with other methods for other types which have a restricted domain (all callable parametric types for now), in a separate PR. To have atomic changes I could just remove this here, copy-paste the error handling code, and recreate this function in a separate PR, if you deem it preferable.

src/primitives/bezier.jl
```diff
@@ -0,0 +1,89 @@
...
+BezierCurve(points::Vararg) = BezierCurve(collect(points))
+BezierCurve(points::AbstractVector{<:Tuple}) = BezierCurve(Point.(points))
+
+ncontrols(b::BezierCurve) = length(b.controls)
+
+"""
+Evaluation method used to obtain a point along
+a Bézier curve from a parametric expression.
+"""
+abstract type BezierEvalMethod end
```
serenity4, line 38: This abstract type is not concretely used, but I found it was clearer to appropriately subtype both method types.

serenity4:
The code for Horner's method was adapted from [Bezier.jl](https://github.com/arlk/Bezier.jl). Since this is essentially a copy-paste with minor modifications, should I specify it somewhere? Maybe in the dosctring? The package is MIT-licensed.

src/primitives/bezier.jl
```diff
@@ -0,0 +1,89 @@
...
+  bᵢ = coordinates(first(cs)) .* tᶜ
+  n = ncontrols(curve) - 1
+  for i in 1:n-1
+    p = coordinates(cs[i+1])
+    tᵢ = tᵢ * t
+    cᵢ = (n - i + 1) * cᵢ / i
+    bᵢ = (bᵢ + cᵢ * p * tᵢ) * tᶜ
+  end
+  Point(bᵢ + coordinates(last(cs)) * tᵢ * t)
+end
```
serenity4: I took the same notation as the author's, but I don't know where it comes from. Should I change it? This one has the advantage of being concise, but without context I admit it is difficult to decipher it.

src/primitives/bezier.jl
juliohm: Yes, maybe we should introduce this helper function in a separate PR and then add methods to it for the different geometries like Segment, Ray and now Bezier :)

src/primitives/bezier.jl
```diff
@@ -0,0 +1,89 @@
...
+  ss = collect(segments(Chain(curve.controls)))
+  points = [s(t) for s in ss]
+  if length(points) == 1
+    points[1]
+  else
+    BezierCurve(points)(t)
+  end
+end
+
+function (curve::BezierCurve{<:Any,T})(t, ::Horner) where {T}
```
juliohm: Why can't we use `BezierCurve{Dim,T}` here?

src/primitives/bezier.jl
```diff
@@ -0,0 +1,89 @@
...
+  cs = curve.controls
+  bᵢ = coordinates(first(cs)) .* tᶜ
+  n = ncontrols(curve) - 1
+  for i in 1:n-1
+    p = coordinates(cs[i+1])
+    tᵢ = tᵢ * t
+    cᵢ = (n - i + 1) * cᵢ / i
+    bᵢ = (bᵢ + cᵢ * p * tᵢ) * tᶜ
+  end
+  Point(bᵢ + coordinates(last(cs)) * tᵢ * t)
```
juliohm: Perhaps this `coordinates(last(cs))` could be moved up in the function body near the `coordinates(first(cs))` ? Perhaps these two could be computed together?

src/primitives/bezier.jl
juliohm: It would be nice if we could find a good reference to cite here, and then follow the same notation there. It is difficult to know what the code is doing otherwise, but it is indeed very elegant.

src/primitives/bezier.jl
```diff
@@ -0,0 +1,89 @@
+# ------------------------------------------------------------------
+# Licensed under the MIT License. See LICENSE in the project root.
+# ------------------------------------------------------------------
+
+"""
+    BezierCurve(points)
+
+A Bézier curve with control points `points`.
+See https://en.wikipedia.org/wiki/Bézier_curve.
```
juliohm, line 9: From the Wikipedia article, what is the type of Bezier curve you implemented? Is it of a fixed order (linear, quadratic, cubic) or a recursive definition? I need to learn about these curves more seriously at some point. They are really nice.

src/primitives/bezier.jl
serenity4: We can, it's just that `Dim` won't be used, so there is no need to request the parameter.

src/primitives/bezier.jl
serenity4: It's a recursive definition, for both evaluation algorithms. If you have a curve with 3 points it's quadratic, 4 it's cubic, and so on.

src/primitives/bezier.jl
juliohm: Let's use Dim explicitly then, it makes the code more readable.

src/primitives/bezier.jl
juliohm: I would state this explicitly in the doc string then: `A recursive Bezier curve with control points`.

src/primitives/bezier.jl
serenity4: The Wikipedia page + the new docstring should be enough. Actually, it's not complicated at all when you introduce context.

serenity4:
I ended up refactoring the evaluation with Horner's method, at such point that it does not look like the original version at all. It is more natural, follows Wikipedia's notations and is 3x faster as a side-effect.

src/primitives/bezier.jl
```diff
@@ -0,0 +1,103 @@
...
+  t̄ⁿ⁻ⁱ = one(T)
+  for i in n:-1:1
+    cᵢ₋₁ *= i / (n - i + one(T))
+    pᵢ₋₁ = @views coordinates(cs[i])
+    t̄ⁿ⁻ⁱ *= t̄
+    aᵢ₋₁ = cᵢ₋₁ * pᵢ₋₁ * t̄ⁿ⁻ⁱ
+    bᵢ₋₁ = aᵢ₋₁ + bᵢ₋₁ * t
+  end
+
+  b₀ = bᵢ₋₁
```
serenity4, line 101: I would agree that this is useless, but it makes it clearer when comparing to the original algorithm.

src/primitives/bezier.jl
```diff
@@ -0,0 +1,103 @@
...
+    throw(DomainError(t, "b(t) is not defined for t outside [0, 1]."))
+  end
+  cs = curve.controls
+  t̄ = one(T) - t
+  n = degree(curve)
+  pₙ = @views coordinates(cs[end])
+  aₙ = pₙ
+
+  # initialization with i = n + 1
+  bᵢ₋₁ = bₙ = aₙ
```
serenity4: `bn` is not used at all, but it adds clarity when looking at the notations (cf docstring). And, thanks to dead code elimination, there is no performance loss at all when doing that (I checked).

serenity4:
Because Horner's method is now faster, there is a 3x performance difference when evaluating a Bezier curve even for a quadratic one (3 points). So, either we provide accuracy with OK performance as a default, or a fast approximation. I still think we should prioritize accuracy, so that people don't encounter dark bugs when we can avoid it, and if they want speed then they are warned about stability issues and make a conscious choice.

src/primitives/bezier.jl
```diff
@@ -0,0 +1,103 @@
...
+```julia
+BezierCurve(Point2[(0.,0.),(1.,-1.)])
+```
+"""
+struct BezierCurve{Dim,T,V<:AbstractVector{Point{Dim,T}}} <: Primitive{Dim,T}
+  controls::V
+end
+
+BezierCurve(points::Vararg) = BezierCurve(collect(points))
+BezierCurve(points::AbstractVector{<:Tuple}) = BezierCurve(Point.(points))
```
juliohm:
```suggestion
BezierCurve(points::AbstractVector{<:Tuple}) = BezierCurve(Point.(points))
BezierCurve(points::Vararg) = BezierCurve(collect(points))
```

src/primitives/bezier.jl
```diff
@@ -0,0 +1,103 @@
...
+"""
+struct Horner <: BezierEvalMethod end
+
+(curve::BezierCurve)(t) = curve(t, DeCasteljau())
+
+function (curve::BezierCurve)(t, ::DeCasteljau)
+  if t < 0 || t > 1
+    throw(DomainError(t, "b(t) is not defined for t outside [0, 1]."))
+  end
+  ss = collect(segments(Chain(curve.controls)))
```
juliohm: I think we don't need to collect the segments here, right?

serenity4: Ooh, right.

src/primitives/bezier.jl
```diff
@@ -0,0 +1,103 @@
...
```
serenity4:
```suggestion
  ss = segments(Chain(curve.controls))
```

src/primitives/bezier.jl
```diff
@@ -0,0 +1,103 @@
...
+    BezierCurve(points)(t)
+  end
+end
+
+"""
+Apply Horner's method on the monomial representation of the
+Bézier curve B = ∑ᵢ aᵢtⁱ with i ∈ [0, n], n the degree of the
+curve, aᵢ = binomial(n, i) * pᵢ * t̄ⁿ⁻ⁱ and t̄ = (1 - t).
+Horner's rule recursively reconstructs B from a sequence bᵢ
+with bₙ = aₙ and bᵢ₋₁ = aᵢ₋₁ + bᵢ * t until b₀ = B.
```
juliohm, line 77: Love these docstrings ❤️

src/primitives/bezier.jl
```diff
@@ -0,0 +1,103 @@
...
+struct BezierCurve{Dim,T,V<:AbstractVector{Point{Dim,T}}} <: Primitive{Dim,T}
+  controls::V
+end
+
+BezierCurve(points::Vararg) = BezierCurve(collect(points))
+BezierCurve(points::AbstractVector{<:Tuple}) = BezierCurve(Point.(points))
+
+ncontrols(b::BezierCurve) = length(b.controls)
+
+degree(b::BezierCurve) = ncontrols(b) - 1
```
juliohm, line 32: Just double checking, degree is the most accurate term right? It is not order or something else...

src/primitives/bezier.jl
```diff
@@ -0,0 +1,103 @@
...
+with bₙ = aₙ and bᵢ₋₁ = aᵢ₋₁ + bᵢ * t until b₀ = B.
+"""
+function (curve::BezierCurve{Dim,T})(t, ::Horner) where {Dim,T}
+  if t < 0 || t > 1
+    throw(DomainError(t, "b(t) is not defined for t outside [0, 1]."))
+  end
+  cs = curve.controls
+  t̄ = one(T) - t
+  n = degree(curve)
+  pₙ = @views coordinates(cs[end])
```
juliohm: Can you explain this `@views` along with the coordinates of a single point? Also, maybe we should use `last(cs)` to be consistent with other access patterns in the codebase?

src/primitives/bezier.jl
juliohm: Is this code elimination guaranteed or something that the compiler may fail in a future release? I like code clarity but unused seem more annoying. Perhaps just a comment instead of introducing a variable name would be better. Can we elaborate on the comment instead?

src/primitives/bezier.jl
juliohm: At least this one is being used later in the `Point` return. 👍🏽

test/runtests.jl
```diff
@@ -9,6 +9,9 @@ using StaticArrays
 using Test, Random, Plots
 using ReferenceTests, ImageIO
 
+# fix import conflict with Plots
+const BezierCurve = Meshes.BezierCurve
+
```
juliohm: Should we add this constant definition inside the testset? That way we don't need to keep track of the main runtests.jl file when things change in the future.

juliohm (changes requested):
This PR will be super useful for a tracing algorithm I am currently implementing ❤️ Let's just fix the minor details before we merge it.

juliohm:
> Because Horner's method is now faster, there is a 3x performance difference when evaluating a Bezier curve even for a quadratic one (3 points). So, either we provide accuracy with OK performance as a default, or a fast approximation. I still think we should prioritize accuracy, so that people don't encounter dark bugs when we can avoid it, and if they want speed then they are warned about stability issues and make a conscious choice.

Yes, I think it is a reasonable choice. Make the accurate method the default and then use the approximation only as an option. 👍🏽

src/primitives/bezier.jl
serenity4: Hmm I actually find both terms to be used. Although I quickly googled "order vs degree", and found that the order of a polynomial is the number of terms in a series expansion, and so it equals the degree + 1. So, here it is the degree, and `ncontrols` actually equals the order.

src/primitives/bezier.jl
serenity4: I agree for `last(cs)`, will a view work with this though? And, I can replace `@views` with `@view`, yes.

src/primitives/bezier.jl
serenity4: The compiler should never fail such a basic optimization. I mean, this would be a serious problem that I think the developers won't let happen.

src/primitives/bezier.jl
juliohm: Yes, just wanted to double check. I think degree is perfectly fine 💯

src/primitives/bezier.jl
serenity4: Well, I removed the view there, I don't think it actually makes a difference for just 1 point.

serenity4:
I think it is ready to merge :100:

juliohm (approved):

juliohm:
That is an amazing PR. Thanks! Let's just wait for the tests to pass before we merge.

PR #88, opened 2021-03-22
Add `Plane` primitive

stillyslalom:


src/primitives/plane.jl
```diff
@@ -0,0 +1,15 @@
+# ------------------------------------------------------------------
+# Licensed under the MIT License. See LICENSE in the project root.
+# ------------------------------------------------------------------
+
+"""
+    Plane(o, n)
```
juliohm:
```suggestion
    Plane(p, n)
```

src/primitives/plane.jl
```diff
@@ -0,0 +1,15 @@
...
+    Plane(p, n)
+
+A plane coincident with point `p`, with normal vector `n`.
+"""
+struct Plane{Dim,T} <: Primitive{Dim,T}
+  p::Point{Dim,T}
+  n::Vec{Dim,T}
+end
+
+paramdim(::Type{<:Plane}) = 2
```
juliohm: The parametric dimension of a plane is Dim-1 more generally.

juliohm (changes requested):
Awesome. We just need to fix the parametric dimension to make it more general. Also fix the tests accordingly.

src/primitives/plane.jl
stillyslalom: I'd taken the plane as a general 2-dimensional subspace in the same way that `paramdim(::Line{Dim}) == 1` for arbitrary `Dim`. Would this generalized treatment fit better under a new `Hyperplane` abstract type containing both `Line` and `Plane`?

juliohm: A `Line` is an actual 1D object defined by a direction. I don't see it as a plane. On the other hand, a Plane is defined here in terms of all points perpendicular to a direction, including a given point of reference, and this leads to a parametric dimension Dim-1.

serenity4: I would disagree. The orthogonal vector is like a dual of the plane in 3D. But inherently a plane is a combination of two vectors, and you need two parameters to navigate on it.

serenity4: My comment above is not consistent with the code I realized. I would suggest to define a plane with two vectors.

juliohm: So we have a lot of flexibility with these definitions, but we need to think in terms of the application. We can represent a plane in multiple ways and give it different semantics depending on the usage. I think we don't want to overcomplicate things. Should we open an issue to discus this Line/Plane situation in more detail?

serenity4: To add to the mix, a plane can also be seen as a sphere with infinite radius. In that case, it requires 4 points including one at infinity (so, 3 because the last one is trivial).

I agree that opening an issue on the matter would be helpful. Especially for such a basic primitive, we want to get it right and not suffer from making the wrong choices. I think a lot of stuff can be built on top of it.

stillyslalom: Following [this](https://github.com/JuliaGeometry/Meshes.jl/issues/90#issuecomment-804445499) discussion, I kept the original `paramdim(::Type{<:Plane}) = 2`. Is that okay, or do you still want to go with `paramdim(::Type{<:Plane{Dim,T}}) where {Dim} = Dim - 1`?

juliohm (approved):

juliohm:
I think we can merge this right away so that you keep working on your downstream efforts. We can then continue the discussion in #90 to polish these concepts over time. Nice additions BTW 💯

PR #103, opened 2021-03-30
Line intersection

### LINKED ISSUE #102: Add intersection between two Line objects
We have a method for intersecting two Segment objects. We could add a similar method for Line objects.

juliohm:
Fix #102 

I've added two broken tests because we need to implement equality for Line still. We could either merge this one as is or add a equality with a suggested method.


serenity4 (approved):
Nice! Thanks @juliohm.

Issue #98, opened 2021-03-30
Refactor distances between geometries


juliohm:
Currently we have a single `evaluate(distance, g1, g2)` but as pointed out by @lassepe over Zulip, we could refactor this into at least two types of distances between geometries: `mindistance` and `maxdistance`. We could also consider Hausdorff distances in the future.

I think we can preserve `evaluate(distance, p1, p2)` where `p1` and `p2` are points, and also provide the `colwise` and `pairwise` variants like in Distances.jl.

serenity4:
What should `maxdistance` do? Return the distance between a point and e.g. the farthest part of another geometry? Or is it intended for a particular set of metrics (non-Euclidean for example)?

Also, from a naming perspective, why was `evaluate` chosen instead of simply `distance`?

serenity4:
For example, I could have seen `distance(metric, p1, p2)`, or am I wrong about the use of `metric` and `distance`?

juliohm:
> What should `maxdistance` do? Return the distance between a point and e.g. the farthest part of another geometry? Or is it intended for a particular set of metrics (non-Euclidean for example)?

Exactly. The distance between the point and the farthest point in the geometry.

> Also, from a naming perspective, why was `evaluate` chosen instead of simply `distance`?

I tried to follow the Distances.jl API so that whenever we pass points as inputs we can call `evaluate`, `colwise` and `pairwise` with any of the distances in that package.

lassepe:
Here are two cents from my perspective:

1. The correct way to implement the `Distances` seems to be to add a dispatch to the functor for the metric, rather than `Distances.evaluate`. By this means, the generic fallbacks for `evaluate` and `pairwise` et al should work out of the box.

2. I would vote for only using the metric types from `Distances` but not implement their distance API for anything beyond `Meshes.Point` because I would consider it function piracy to add dispatches beyond vector-like types. Instead, it seems cleaner to have package internal `mindistance` and `maxdistance` that share the metric types with `Distances`.

3. It seems natural to have something like `closest` that returns a `NamedTuple` of the farthest point and distance since both pieces of information are naturally computed in one pass. Then `mindistance` can just fallback to a call of that function and discard the point.

juliohm:
All very good points @lassepe. Can you submit a PR to each item separately so that we can more easily review and brainstorm? That would be great. There are lot of breaking changes, and we want to make sure we are not missing anything in the transition.

lassepe:
[...]

PR #108, opened 2021-04-04
Refactor distances between geometries | Part 1

lassepe:
First PR addressing #98 

This allows `Distances.evaluate` et al. to work without a custom dispatch for the `Point` type.

Let me know if the dependency on `Lazy.@forward` is okay or whether we want to just forward manually for the three functions.


lassepe:
The CI error on macos seems not related. It is some error during instantiation of the plotting packages.

juliohm:
@lassepe unfortunately we can't merge this PR as it attempts to solve multiple things at the same time. See our contributing guidelines and start a PR with just the distance refactoring. Here you add a new dependency on Lazy.jl, you forward operations from points to coordinates, and other things that I didn't spend too much time reviewing.

We appreciate if you can follow the plan we brainstormed over Zulip.

lassepe:
This is only attempting to solve one point of my comment in #98:

> 1. The correct way to implement the Distances seems to be to add a dispatch to the functor for the metric, rather than Distances.evaluate. By this means, the generic fallbacks for evaluate and pairwise et al should work out of the box.

If you look at the way Distances is implemented, you can't really add a dispatch to the metric functors because it generates the code:

https://github.com/JuliaStats/Distances.jl/blob/fa867d59098dd848fd71bc48005a1bf858928a47/src/metrics.jl#L328

Thus, I decided to implement a minimal iterator interface to make the generic methods from Distances.jl work.
That said, I'm happy to remove the dependency on Lazy.jl as stated in the PR description. Alternatively, one could also not change this part at all since `Distances.evaluate` would be deprecated by one of my next PR's anyway.

juliohm:
I think Distances.jl evolved a lot since the first time I used it. It is worth checking what is the current approach to implementing new methods and if we need to follow it or just provide a high-level `evaluate(D, point, point)`.

Regarding the iterator interface for points, we can't add it. A point is not the same thing as its coordinates in a coordinate reference system. A tuple of coordinates can be seen as iterable, but a point is an abstract concept with multiple types of coordinates.

lassepe:
> Regarding the iterator interface for points, we can't add it. A point is not the same thing as its coordinates in a coordinate reference system. A tuple of coordinates can be seen as iterable, but a point is an abstract concept with multiple types of coordinates.

Okay, that makes a lot of sense. I'll go another route then. I'm sorry for the noise.

PR #110, opened 2021-04-05
Fix duplicate method warning

### LINKED ISSUE #53: Warning related to SimpleTraits.jl
Do you know what is happening here? How to fix this warning during precompilation?

```julia
julia> using Meshes
[ Info: Precompiling Meshes [eacbb407-ea5a-433e-ab97-5258b1ca43fa]
WARNING: Method definition _linear(D, Any) where {D} in module Meshes at /home/juliohm/.julia/packages/SimpleTraits/1wYi7/src/SimpleTraits.jl:331 overwritten on the same line (check for duplicate calls to `include`).
  ** incremental compilation may be fatally broken for this module **
```

lassepe:
Fixes #53 

As per the comment in the SimpleTraits README, negated trait functions must use the same argument
names in order to avoid duplicate definitions:

  > Long story short: define the two methods of a trait and its negation using the same argument
  > names and no warning should be issued.

### DIFF
--- a/src/viewing.jl
+++ b/src/viewing.jl
@@ -61,7 +61,7 @@ end
 # convert from Cartesian to linear indices if needed
 @traitfn _linear(domain::D, inds) where {D; IsGrid{D}} =
   vec(LinearIndices(size(domain))[inds])
-@traitfn _linear(::D, inds) where {D; !IsGrid{D}} = inds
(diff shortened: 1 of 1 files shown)


juliohm:
That is an awesome fix! Thanks for reading their readme carefully 💯

PR #112, opened 2021-04-06
Refactor distances between geometries | Part 1: mindistance

lassepe:
This is the first out of a sequence of PR's to address #98 

This PR realizes the steps outlined in https://github.com/JuliaGeometry/Meshes.jl/issues/98#issuecomment-813926820

As such, it

- implements `mindistance`
- deprecates `Distances.evaluate(` for our types.
- adds tests for deprecation warnings and invariance of the computed `mindistance` to argument order.

### DIFF
--- a/src/distances.jl
+++ b/src/distances.jl
@@ -1,27 +1,27 @@
 # ------------------------------------------------------------------
 # Licensed under the MIT License. See LICENSE in the project root.
 # ------------------------------------------------------------------
-
-# flip arguments so that points always come first
-evaluate(d::PreMetric, g::Geometry, p::Point) = evaluate(d, p, g)
-
 """
-    evaluate(Euclidean(), point, line)
+    mindistance(metric::PreMetric, g::Union{Geometry,Point}, p::Point)
 
-Evaluate the Euclidean distance between `point` and `line`.
+Returns the minimum distance between the the point `p` and the closest point in geometry `g` as
+measured by the `metric`.
 """
-function evaluate(::Euclidean, p::Point, l::Line)
-  a, b = l(0), l(1)
-  u = p - a
-  v = b - a
-  α = (u ⋅ v) / (v ⋅ v)
-  norm(u - α*v)
+function mindistance end
+
+# flip arguments to always have geometry be the first argument.
+mindistance(metric::PreMetric, p::Point, g::Geometry) = mindistance(metric, g, p)
+
+function mindistance(metric::Union{Euclidean,SqEuclidean}, l::Line, p::Point)
+    a, b = l(0), l(1)
+    u = p - a
+    v = b - a
+    α = (u ⋅ v) / (v ⋅ v)
+    metric(u, α * v)
 end
 
-"""
-    evaluate(::PreMetric, point1, point2)
+mindistance(metric::PreMetric, p1::Point, p2::Point) =
+    evaluate(metric, coordinates(p1), coordinates(p2))
 
-Evaluate pre-metric between coordinates of `point2` and `point2`.
-"""
-evaluate(d::PreMetric, p1::Point, p2::Point) =
-  evaluate(d, coordinates(p1), coordinates(p2))
+@deprecate evaluate(d::PreMetric, g::Union{Geometry,Point}, p::Point) mindistance(d, g, p)
+@deprecate evaluate(d::PreMetric, p::Point, g::Geometry) mindistance(d, g, p)
--- a/src/Meshes.jl
+++ b/src/Meshes.jl
@@ -15,7 +15,7 @@ using Random
 using IterTools: ivec
 using StatsBase: Weights
 using SpecialFunctions: gamma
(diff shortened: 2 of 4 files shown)


src/distances.jl
```diff
@@ -1,27 +1,27 @@
 # ------------------------------------------------------------------
 # Licensed under the MIT License. See LICENSE in the project root.
 # ------------------------------------------------------------------
-
-# flip arguments so that points always come first
-evaluate(d::PreMetric, g::Geometry, p::Point) = evaluate(d, p, g)
-
 """
-    evaluate(Euclidean(), point, line)
+    mindistance(metric::PreMetric, g::Union{Geometry,Point}, p::Point)
```
juliohm, line 5: The function signature should be more intuitive, meaning we shouldn't pass a metric as the first argument. We should instead have something like

```julia
mindistance(p::Point, g::Geometry; metric=Euclidean())
```
that focuses on the objects and leaves the metric as an option.

Also notice that we should implement a version with `p` as the first argument always and flip the argument to handle the symmetry. This is for organizational purposes and consistency across the codebase.

src/distances.jl
```diff
@@ -1,27 +1,27 @@
...
+    a, b = l(0), l(1)
+    u = p - a
+    v = b - a
+    α = (u ⋅ v) / (v ⋅ v)
+    metric(u, α * v)
 end
 
-"""
-    evaluate(::PreMetric, point1, point2)
+mindistance(metric::PreMetric, p1::Point, p2::Point) =
```
juliohm, line 23: Same holds for the case of two points. We need to pass a metric as an option because most users will be in Euclidean geometry with distances as straight lines.

src/distances.jl
```diff
@@ -1,27 +1,27 @@
...
+    evaluate(metric, coordinates(p1), coordinates(p2))
 
-Evaluate pre-metric between coordinates of `point2` and `point2`.
-"""
-evaluate(d::PreMetric, p1::Point, p2::Point) =
-  evaluate(d, coordinates(p1), coordinates(p2))
+@deprecate evaluate(d::PreMetric, g::Union{Geometry,Point}, p::Point) mindistance(d, g, p)
+@deprecate evaluate(d::PreMetric, p::Point, g::Geometry) mindistance(d, g, p)
```
juliohm, line 27: I think we don't need to add deprecation warnings at this point of the project. We are still breaking a lot of things to get the design right, and these deprecation warnings will pile up.

juliohm (changes requested):
Thank you @lassepe for working on this improvement. Please take a look at the suggested changes before we continue with the review.

Also, we may need to think more carefully about the function name when one of the arguments is a point because there is no much use in other type of distance other than the `mindistance`. So we could probably call it `distance` directly. This `mindistance` versus `maxdistance` issue only becomes relevant when we have two geometries as the arguments I guess.

juliohm:
As a side comment, I noticed that you started other PRs assuming this PR would be quickly merged. I think the best approach is to work on a single PR at a time in this case. We will likely need to rebase the other PRs in order to get them merged and may also need to change them drastically based on the reviews on this present PR.

src/distances.jl
lassepe: > Also notice that we should implement a version with p as the first argument always and flip the argument to handle the symmetry. This is for organizational purposes and consistency across the codebase.

This is implemented in line 13:
`mindistance(metric::PreMetric, p::Point, g::Geometry) = mindistance(metric, g, p)`

lassepe: > Also, we may need to think more carefully about the function name when one of the arguments is a point because there is no much use in other type of distance other than the `mindistance`. So we could probably call it `distance` directly. This `mindistance` versus `maxdistance` issue only becomes relevant when we have two geometries as the arguments I guess.

I would argue that `mindistance` vs. `maxdistance` is also relevant as soon as one of the objects is not a singleton set (i.e. a Point). However, I am somewhat convinced that `maxdistance` almost never makes sense. For example, for `Line`, `Ray`, and `Plane` `maxdistance` would have to return `Inf`.

I would therefore vote for only implementing `mindistance` (and maybe call it just `distance`) and worry about other types of distances if someone actually needs them.

lassepe:
> As a side comment, I noticed that you started other PRs assuming this PR would be quickly merged. I think the best approach is to work on a single PR at a time in this case. We will likely need to rebase the other PRs in order to get them merged and may also need to change them drastically based on the reviews on this present PR.

The only reason why I created all of them in one go is that I had a 5 hour train ride with a lot of time. Thus, I created them in a batch to outline what the rest of the features could look like. I'm happy to rebase/rewrite as we go. No need to merge things quickly from my POV.

juliohm:
@lassepe I will close this series of PRs because we need to refactor the code and discuss better what this interface will be. Right now the code can't be merged as is and there is a lot to be discussed before we accept a change, specially if it is gonna be erased again in the future with another refactoring.

PR #117, opened 2021-04-10
FilteredSearch refactored to Meshes

rmcaixeta:
This relates to https://github.com/JuliaEarth/GeoStats.jl/issues/158 and to the conversation with @hameye in GeoStats.jl gitter
This extends the previous named `BoundedSearch` to `FilteredSearch`, adding extra search constraints
I refactored what I had started in https://github.com/JuliaEarth/GeoStatsBase.jl/pull/70 and adjusted the sectors constraints errors when data was exactly in the limits between two sectors. It's compiling and passing in some tests I made in 2D. Some additional steps for this PR:

- [ ] Review style and organization
- [x] Check for errors and create CI tests for the new constraints
- [x] Add ASCII illustration for the `FilteredSearch` docstring
- [ ] Check for errors downstream and rename to `FilteredSearch` other occurrences of `BoundedSearch` (e.g. GeoEstimation.jl/src/kriging.jl)


src/neighborsearch.jl
```diff
@@ -27,12 +27,12 @@ specify a `mask` for all indices of the domain.
...
 
 """
-    BoundedNeighborSearchMethod
+    FilteredNeighborSearchMethod
 
 A method for searching neighbors with the property that the number of neighbors
 is bounded above by a known constant (e.g. k-nearest neighbors).
 """
-abstract type BoundedNeighborSearchMethod <: NeighborSearchMethod end
+abstract type FilteredNeighborSearchMethod <: NeighborSearchMethod end
```
juliohm: I think we want to stick to BoundedNeighborSearchMethod here, it is just an abstract type that informs when a search method is guaranteed to return a bounded number of indices.

src/Meshes.jl
```diff
@@ -26,7 +26,7 @@ import Tables
 import Random
 import Base: values, ==, +, -, *
 import StatsBase: sample
-import Distances: evaluate
+import Distances: colwise, evaluate
```
juliohm: I think we can move this to `using Distances: colwise` instead. The import is to extend the method table and we are not doing it in this PR.

src/neighborsearch/filtered.jl
```diff
@@ -0,0 +1,145 @@
...
+"""
+struct FilteredSearch{M<:NeighborSearchMethod} <: FilteredNeighborSearchMethod
+  method::M
+  nmax::Int
+  maxpercategory
+  maxpersector
+  order
+end
+
+FilteredSearch(method::M, nmax=0; maxpercategory=nothing, maxpersector=nothing, order=Euclidean()) where {M} =
```
juliohm: Let's break this line into two after the `mexpercategory=nothing,` argument.

juliohm (changes requested):
Thanks @rmcaixeta ! This is a must needed feature ❤️  I have a few quick fixes before we can dive into the code in more detail.

rmcaixeta:
Thanks @juliohm ! Changes made

src/neighborsearch/filtered.jl
```diff
@@ -0,0 +1,146 @@
...
+function search!(neighbors, pₒ::Point, method::FilteredSearch; mask=nothing)
+  meth = method.method
+  obj  = meth.domain
+  inds = search(pₒ, meth, mask=mask)
+  nmax = method.nmax == 0 ? Inf : method.nmax
+
+  # get distances and give priority to closest neighbors if necessary
+  if !(meth isa KNearestSearch)
+    indcoords = reduce(hcat, coordinates.([centroid(obj, s) for s in inds]))
+    dists     = colwise(method.order, coordinates(pₒ), indcoords)
```
juliohm: I think we can avoid this major allocation of points and can instead calculate the distances with a single list comprehension with the simple `evaluate` function:

```julia
dists = [evaluate(method.order, coordinates(p0), coordinates(centroid(obj, ind))) for ind in inds]
```
```

File: src/neighborsearch/filtered.jl [hunk 5]
```diff
@@ -0,0 +1,146 @@
...
+maxneighbors(method::FilteredSearch) = method.nmax
+
+function search!(neighbors, pₒ::Point, method::FilteredSearch; mask=nothing)
+  meth = method.method
+  obj  = meth.domain
+  inds = search(pₒ, meth, mask=mask)
+  nmax = method.nmax == 0 ? Inf : method.nmax
+
+  # get distances and give priority to closest neighbors if necessary
+  if !(meth isa KNearestSearch)
```
juliohm: Maybe we need to introduce a new subtype of SearchMethod? We are just trying to catch all search methods that are based on balls?

src/neighborsearch/filtered.jl
```diff
@@ -0,0 +1,146 @@
...
+  end
+
+  # initialize category and sectors constraints if necessary
+  categs  = initcategories(obj, inds, method.maxpercategory)
+  sectors = initsectors(meth, method.maxpersector)
+
+  # check about initialization size
+  nneigh = 0
+  for i in inds
+    nneigh == nmax && break
```
juliohm: Should this break be moved to the end of the loop?

juliohm (changes requested):
Thanks @rmcaixeta for the quick fix. A few more comments before we jump into the new functions added in the PR.

src/neighborsearch/filtered.jl
rmcaixeta: Ok, I'll change it. The colwise seems to be faster only if order = Mahalanobis (e.g. Ellipsoid.metric)

src/neighborsearch/filtered.jl
rmcaixeta: This is also new. The idea is to sort data according to distance, giving priority to closest points when filtering data. I'm checking here, currently `KNearestSearch` (and also `KBallSearch` that I forgot here) return the `inds` sorted by distance. In a `Ellipsoid` for example, the output of `FilteredSearch` will not guarantee that the priority will be given to the closest points

rmcaixeta: Solutions to it: (1) set `sortres=true` in `NearestNeighbors.inrange` function too. (2) sort during `FilteredSearch`.  I'd say that's better to sort during `FilteredSearch` because the filter priority could be given to a different metric than the first seacrh. e.g. an Ellipsoid would give priority to the closest "anisotropic distance", while in a filtered search with order=Euclidean() would give priority to the closest euclidean dist.

rmcaixeta: `if !(meth isa KNearestSearch)` was an attempt to avoid sorting if it was already sorted. But this might not be enough, since `KNearestSearch` and `FilteredSearch` metrics might not be the same. Moreover, `KBallSearch` would need to be considered too as it's already sorted with `sortres=true` too

src/neighborsearch/filtered.jl
rmcaixeta: The final results will be the same, I can move there if you prefer

src/neighborsearch/filtered.jl
juliohm: I am trying to grasp the bits, but I get that this if statement was introduce to sort the indices returned by the inner search method differently. I think a better design would provide a sorting method in the `FilteredSearch` if that is important. I would start very simple with the options you originally considered, possibly sharing examples in the PR so that we understand what each section of the code is doing.

My feeling is that we can improve the design to accommodate the filters proposed and to make the code more readable for future contributors.

src/neighborsearch/filtered.jl
juliohm: I know, I think it is just a matter style :) The break of the while loop is usually placed near the end as we read the algorithm, specially because in this case we have `nneigh = 0` before the loop and the condition is clearly still false in the first iteration.
A break is usually placed at the beginning when you expect to save some computation.

rmcaixeta: No problem, I'll change that. About the `nneigh = 0` and `nmax=0`, the "correct" default would be `nmax=Inf` in case only other constraints are desired. I set `nmax=0` and change it to `nmax = method.nmax == 0 ? Inf : method.nmax` in the beginning because the `search` method for all `BoundedNeighborSearchMethod` preallocate the `neighbors` size to `nmax` (https://github.com/JuliaGeometry/Meshes.jl/blob/master/src/neighborsearch.jl). Not a much elegant solution, but suggestions are welcome

src/neighborsearch/filtered.jl
rmcaixeta: I'll change it to make it simpler. Might sort twice the data, but I think it's more important to sort it initially to avoid inconsistencies. If the user wants to preserve original order and prevent sorting, just need to set `order=nothing`. Do you think it's better to change the name of `order` to `metric` to match other functions style?

juliohm: Yes, if `order` is a distance from Distances.jl than it should be named a `metric`.

src/neighborsearch/filtered.jl
juliohm: I need to find the time to review the code more carefully. More verbose comments could help here.

rmcaixeta:
@hameye , I made my tests here, could not find problems so far. If you could check there in your workflow too, it'd be great

juliohm:
@rmcaixeta love the new ASCII art ❤️

src/neighborsearch/filtered.jl
```diff
@@ -0,0 +1,189 @@
...
+struct FilteredSearch{M<:NeighborSearchMethod} <: BoundedNeighborSearchMethod
+  method::M
+  nmax
+  maxpercategory
+  maxpersector
+  metric
+end
+
+FilteredSearch(method::M, nmax=Inf; maxpercategory=nothing,
+               maxpersector=nothing, metric=Euclidean()) where {M} =
```
hameye: It would make more sense if the metric was the one of the method if there is one

src/neighborsearch/filtered.jl
```diff
@@ -0,0 +1,189 @@
...
+  obj  = meth.domain
+  nmax = method.nmax
+
+  # initial search, using mask if necessary
+  inds = search(pₒ, meth, mask=mask)
+
+  # `inds` is not guranteed to be sorted by distance. also, a different
+  # metric might be preferred to give priority to the FilteredSearch.
+  # unless `metric = nothing`, the indices are sorted here using metric
+  if !isnothing(method.metric)
```
hameye: Maybe we should automatically sort in the search ? 
If the method is kball or knn, the search already outpouts sorted data. 
Do you have an advice on that point @juliohm ?

hameye (commented):
Just some suggestions after trying it in my workflow
I'm trying to use maxpercategory it in my workflow in order to validate, but the algorithm seems ok to me ! 
I don't have any opinion on the maxpersector algorithm yet.

src/neighborsearch/filtered.jl
rmcaixeta: That'd probably be more intuitive indeed.
Maybe we could let as default `metric=:same` or something similar.

src/neighborsearch/filtered.jl
rmcaixeta: In the case of preserving original metric, we could set metric to `metric = method isa NeighborhoodSearch ? method.tree.metric : nothing`. That would also avoid sorting twice if search method is a `KBallSearch` or `KNearestSearch`

juliohm:
Thank you @hameye and @rmcaixeta for the comments. I will try to find some time next week to review the PR more carefully. Super busy week around here.

src/neighborsearch/filtered.jl
```diff
@@ -0,0 +1,193 @@
...
+# initialize categories constraints if necessary. e.g. if `maxpercategory =
+# (holeid = 2,)`, dict[:count][:holeid] will have all the possible values of
+# holeid, initiliazed to zero for further counting.
+function initcategories(obj, inds, catgs)
+  catgs == nothing && return Dict(:use => false)
+  tab   = Tables.columns(values(obj))
+  catvals = Dict(k => unique(Tables.getcolumn(tab, k)[inds]) for k in keys(catgs))
+  counter = Dict(k => Dict(zip(v, zeros(Int,length(v)))) for (k, v) in catvals)
+  Dict(:use => true, :max => catgs, :count => counter)
+end
```
juliohm: We need to refactor this. Dictionaries don't seem to be the correct data structure to count categories. Also, we have too much nesting in this dictionary, which is an indication that the design should be reconsidered.

Let's think about the original problem. We start with a set of candidate `inds` returned by the inner method. Now we want to filter these indices such that a maximum number of indices is retained for each specified categorical property. So we need to define a filtering operation. That is clear.

The way we write this filter is the key here. Let's introduce a counter for each specified categorical property:

```julia
countpercat = Dict(keys(maxpercategory) .=> 0)
```

For each candidate index, we check if it will be retained by checking if all the counts did not surpass the maximum counts allowed:

```julia
filtered = Int[]
for ind in inds
  retain = true
  for (k, v) in maxpercategory
    if countpercat[k] >= v
      retain = false
    end
  end
  if retain
    push!(filtered, ind)
    # update countpercat by inspecting the line `ind` only
  end
end
```

I think this pseudo-code should be enough. Can you double check?

src/neighborsearch/filtered.jl
```diff
@@ -0,0 +1,193 @@
...
+    extrinsic = isextrinsic(method.neigh.convention)
+    invertfac = extrinsic ? -1 : 1
+    rotangs   = invertfac * (method.neigh.angles .+ rotangs)
+  end
+
+  # rotation matrix
+  rotmatx = rotmat(rotangs, convention)'
+
+  Dict(:use => true, :count => zeros(Int, 2^N), :rotmat => rotmatx, :max => maxpersector)
+end
```
juliohm: I think the same logic I wrote down for `initcategories` could be applied here. We usually don't need to introduce dictionaries for filtering. We can save memory allocations by writing for loops manually that are well-documented so that people can follow the logic easily. The dictionaries are hiding the simple logic explained in the docstring.

juliohm (changes requested):
Hi @rmcaixeta thanks for this awesome PR as usual. I finally got a chance to sit down and read the code more carefully. You will see that I added a set of comments and suggestions for improving the code. We need to rewrite some internal logic to make it simpler and more efficient memory-wise.

Let's try to incorporate this initial round of reviews before continuing with the second part regarding the maxpersector option, which is more complicated. I am predicting that we will need to refactor the rotation API before using it here. Currently our rotation API is too cumbersome to use, and we really need to tackle #43 before it beats us.

rmcaixeta:
Hey Julio, thanks for the checks
I adjusted the docstrings as recommended. I won't have time to adjust the remaining parts. If you or @hameye  want to follow with that, feel free to continue it, use it as it is and improve it later, or abandon it

Just about the metric discussion in your comments, `nothing` and `:same` have different uses (see discussion with @hameye). As it is right now, there are three options. metric = nothing does not sort anything before filtering. metric = :same sort inds using same metric of primary search method. And metric = some other Distance.metric will sort it with custom distance (e.g. Mahalanobis Ellipsoid in the primary search but Euclidean rank for filtering)

juliohm:
I tried to refactor the PR to include a few changes, but ended up messing up the commit history with a lot of changes from the master branch.

In any case, I realized that options such as `maxpercategory` only make sense for geospatial data, not for geospatial domains where variables are not present. We really need to redesign the API of search methods to accommodate these nuances. We already have a bunch of variants of search methods that differ by a small detail. I think we could instead introduce a filtering language that combines different types of filters more flexibly. I will close the PR and we can try to tackle this issue in the future with more time.

PR #118, opened 2021-04-12
Fix radius call for ellipsoid search

hameye:
Linked to GeoStats issue https://github.com/JuliaEarth/GeoStats.jl/issues/176


juliohm:
That is an awesome fix @hameye, thanks for working on it. We just need a few tests to make sure it is not reintroduced in the future.

juliohm:
Amazing code style by the way 👌🏽

PR #119, opened 2021-04-13
Feature/line equality

### LINKED ISSUE #105: Implement equality for Line
We need a method `l1 == l2` that tells when two lines `l1` and `l2` are equal.

hyrodium:
This PR fixes #105.

### DIFF
--- a/src/primitives/line.jl
+++ b/src/primitives/line.jl
@@ -29,3 +29,7 @@ function Base.in(p::Point, l::Line)
   # d ≈ 0.0 will be too precise, and d < atol{T} can't scale.
   d+w ≈ w
 end
+
+function Base.:(==)(l1::Line, l2::Line)
+  l1.a in l2 && l1.b in l2 && l2.a in l1 && l2.b in l1
+end
--- a/src/primitives/line.jl
+++ b/src/primitives/line.jl
@@ -19,3 +19,13 @@ Line(a::Tuple, b::Tuple) = Line(Point(a), Point(b))
 paramdim(::Type{<:Line}) = 1
 
 (l::Line)(t) = l.a + t * (l.b - l.a)
+
+function Base.in(p::Point, l::Line)
+  q = p-l.a
+  w = norm(l.b-l.a)
+  v = (l.b-l.a)/w
+  # d is a distance between p and l
+  d = norm(q - v*dot(q, v))
+  # d ≈ 0.0 will be too precise, and d < atol{T} can't scale.
+  d+w ≈ w
+end
--- a/src/primitives/line.jl
+++ b/src/primitives/line.jl
@@ -27,6 +27,4 @@ function Base.in(p::Point, l::Line)
   d+w ≈ w
 end
 
-function Base.:(==)(l1::Line, l2::Line)
-  l1.a in l2 && l1.b in l2 && l2.a in l1 && l2.b in l1
-end
+==(l1::Line, l2::Line) = l1.a ∈ l2 && l1.b ∈ l2 && l2.a ∈ l1 && l2.b ∈ l1
--- a/src/primitives/line.jl
+++ b/src/primitives/line.jl
@@ -21,11 +21,8 @@ paramdim(::Type{<:Line}) = 1
 (l::Line)(t) = l.a + t * (l.b - l.a)
 
 function Base.in(p::Point, l::Line)
-  q = p-l.a
   w = norm(l.b-l.a)
-  v = (l.b-l.a)/w
-  # d is a distance between p and l
-  d = norm(q - v*dot(q, v))
+  d = evaluate(Euclidean(), p, l)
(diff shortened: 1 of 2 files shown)


hyrodium:
Here's an example of intransitivity of `==(::Line, ::Line)` which is discussed in [this comment](https://github.com/JuliaGeometry/Meshes.jl/issues/105#issuecomment-818398453).
```julia
l1 = Line{2, Float64}(Point(0.0, 0.0), Point(1.0, 2.01027582027942e-9))
l2 = Line{2, Float64}(Point(0.0, 0.0), Point(1.0, 0.0))
l3 = Line{2, Float64}(Point(0.0, 0.0), Point(1.0, -1.4337470895339735e-8))

l1 == l2  # true
l2 == l3  # true
l1 == l3  # false
```

hyrodium:
The ci `Julia 1 - macos-latest - x64 - pull_request` is failing, but it seems like this PR is not relevant.

src/primitives/line.jl
```diff
@@ -19,3 +19,14 @@ Line(a::Tuple, b::Tuple) = Line(Point(a), Point(b))
 paramdim(::Type{<:Line}) = 1
 
 (l::Line)(t) = l.a + t * (l.b - l.a)
+
+function Base.in(p::Point, l::Line)
+  w = norm(l.b-l.a)
+  d = evaluate(Euclidean(), p, l)
+  # d ≈ 0.0 will be too precise, and d < atol{T} can't scale.
+  d+w ≈ w
```
juliohm, line 27: Can you comment a bit more about the scale issue you mention? I like that you are writing very clean and concise code 💯 👍🏽 Just double checking that you are aware of `isapprox(d, zero(T), atol = atol(T))`.

src/primitives/line.jl
```diff
@@ -19,3 +19,14 @@ Line(a::Tuple, b::Tuple) = Line(Point(a), Point(b))
...
+end
+
+function Base.:(==)(l1::Line, l2::Line)
+  l1.a in l2 && l1.b in l2 && l2.a in l1 && l2.b in l1
+end
```
juliohm:
```suggestion
==(l1::Line, l2::Line) = l1.a ∈ l2 && l1.b ∈ l2 && l2.a ∈ l1 && l2.b ∈ l1
```

juliohm: I copied pasted the `\in` operator in Julia on GitHub and it may have changed the character to some unicode that is invalid Julia. Appreciate if you can double check after accepting  the suggestion.

test/primitives.jl
```diff
@@ -14,8 +14,8 @@
 
     l1 = Line(P2(0,0), P2(1,0))
     l2 = Line(P2(1,0), P2(2,0))
-    @test_broken l1 == l2
-    @test_broken l1 ∩ l2 == l2 ∩ l1 == l1
+    @test l1 == l2
+    @test l1 ∩ l2 == l2 ∩ l1 == l1
```
juliohm, line 18: 🎉

juliohm (changes requested):
Amazing PR @hyrodium , thanks for the very nice and clean code. Appreciate if you can take a look at the comments before we merge it. The build failing on MacOS is indeed unrelated.

src/primitives/line.jl
hyrodium: For `Sphere` example, if we change whole size, then the thresold will change together.
```julia
julia> Point(0.0,1+1e-8) in Sphere(Point(0.0,0.0),1.0)
false

julia> Point(0.0,1+1e-9) in Sphere(Point(0.0,0.0),1.0)
true

julia> Point(0.0,(1+1e-8)/100) in Sphere(Point(0.0,0.0),1/100)
false

julia> Point(0.0,(1+1e-9)/100) in Sphere(Point(0.0,0.0),1/100)
true
```

I thought this prpperty is important, because we will deal with both large and small scale.
Here's the example for `Line`:

```julia
julia> Point(0.0,1e-7) in Line(Point(0.0,0.0),Point(1.0,0.0))
false

julia> Point(0.0,1e-8) in Line(Point(0.0,0.0),Point(1.0,0.0))
true

julia> Point(0.0,1e-7/100) in Line(Point(0.0,0.0),Point(1/100,0.0))
false

julia> Point(0.0,1e-8/100) in Line(Point(0.0,0.0),Point(1/100,0.0))
true
```

If we use `isapprox(d, zero(T), atol = atol(T))`, then the thresold parameter will be fixed for specific lengh.

hyrodium: The following result seems weird, but this is inevitable, I guess:wink:
```julia
julia> Point(0.0,1e-8) in Line(Point(0.0,0.0),Point(1.0,0.0))
true

julia> Point(0.0,1e-8) in Line(Point(0.0,0.0),Point(1/100,0.0))
false
```

juliohm:
Thanks! Nice to have you around. Looking forward to more thoughtful discussions and improvements.

PR #121, opened 2021-04-13
Add topological structures

juliohm:
This PR introduces new possibilities for representations of general meshes. In particular, it introduces topological data structures that are independent of the coordinates of the vertices. In practice, this means that we can convert between different topological representations to navigate the mesh efficiently if necessary for a particular algorithm.

At this point we have two structures available:

- `ExplicitStructure`: what we have been using so far in the package. This is a structure that is common in computer graphics because we can easily loop over triangles explicitly, visualize, save to disk, etc.
- `HalfEdgeStructure`: an edge-based data structure for orientable 2-manifolds that is useful for navigating the boundary of cells, the star of vertices, etc.


src/topostructures/conversions.jl
```diff
@@ -0,0 +1,93 @@
...
+
+    # connect vertices into a polytope type
+    nv = length(v)
+    if nv == 3
+      celltype = Triangle
+    elseif nv == 4
+      celltype = Quadrangle
+    else
+      throw(Error("Polytope type not implemented."))
+    end
```
serenity4, line 87: I feel this is a bit hacky. This is required because a `Connectivity` is attached to a polytope. I think we may want to revisit that in the future.

src/conventions.jl
```diff
@@ -2,39 +2,6 @@
...
-# ----------------
-# IMPLEMENTATIONS
-# ----------------
-
-include("conventions/GMSH.jl")
-
-# ---------------------
-# ROTATION CONVENTIONS
-# ---------------------
-
```
serenity4, line 37: Why was it deleted? Also, there are a few places where `GMSH` is still referred to, if you really want to remove that. However I believe this is unrelated to topological structures; these conventions are just a way to build predefined sets of faces from predefined polytopes.

serenity4: A solution to keep the old functionality would be to change these `connectivities` into constructors for an `ExplicitStructure` instead.

src/topostructures/conversions.jl
```diff
@@ -0,0 +1,93 @@
...
+
+  # map vertices to inner half-edges
+  edgeonvertex = Vector{Int}(undef, nvertices)
+  for i in eachindex(edgeonvertex)
+    e = findfirst(he -> he.head == i, inneredges)
+    edgeonvertex[i] = e
+  end
+
+  HalfEdgeStructure(halfedges, edgeoncell, edgeonvertex)
+end
```
serenity4, line 64: Even if those large blocks of code (that could be functions) are only needed here and never elsewhere, factoring them out into functions could make the core more readable. However these "flattened" chunks of code are quite recurrent in other parts of the codebase, which hints me that there is a conscious choice for this.

serenity4 (changes requested):
Here's a first round. I haven't looked into the new conversions in detail yet, but I feel there may be some potential for a refactor.
The important thing to deal with are ordering conventions which are definitively lost in this PR.

src/topostructures/halfedge.jl
```diff
@@ -0,0 +1,68 @@
+# ------------------------------------------------------------------
+# Licensed under the MIT License. See LICENSE in the project root.
+# ------------------------------------------------------------------
+
+"""
+    HalfEdge(head, cell, prev, next, half)
+
+Stores the indices of the `head` vertex, the `prev`
+and `next` edges in the left `cell`, and the opposite
+`half`-edge. For some half-edges the `cell` may be
```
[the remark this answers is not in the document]
juliohm: Yes, I thought I had added the case of border edges. Will review the doc string to add this example 👍🏽

src/topostructures/conversions.jl
```diff
@@ -0,0 +1,93 @@
...
+    # select a half-edge on the cell
+    e = edgeoncell(s, c)
+
+    # retrieve vertices of the cell
+    n = e.next
+    v = [e.head]
+    while n != e
+      push!(v, n.head)
+      n = n.next
+    end
```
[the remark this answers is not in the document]
juliohm: Yes, the plan is to implement the `boundary` `coboundary` and `adjacency` functions in a separate PR and use them here later. I just wanted to get this working first without implementing all the topological relations.

src/topostructures/conversions.jl
juliohm: I felt the same. Maybe we could introduce a 2D cell type with an arbitrary number of vertices and no particular shape. We have to think carefully before doing that.

src/conventions.jl
juliohm: The GMSH convention was really buggy in the sense that it was using local indices as opposed to global indices in a vector of points. So getting the connectivities with local indices wasn't very useful. We can build that more generally with these data structures. So I think I will erase all occurrences of GMSH and we can see if it makes sense to bring it back in the future. Right now it was just hanging there with no particular good use case.

src/topostructures/conversions.jl
juliohm: I think these chunks will naturally emerge as we evolve the API. Right now I am trying to not commit to particular functions because we are still exploring the design space. If we add functions and people start using them we will have users complaining about things breaking too often.

juliohm:
I will merge this one quickly to keep working on related features and avoid Git complications.

src/topostructures/conversions.jl
serenity4: Fair enough. For the last part, you can always chose not to export them until the API is figured out.

PR #122, opened 2021-04-15
More useful constructor interface

hyrodium:
This PR provides more useful constructor interface for `Point` and `Line`.

# `Point` constructor
## Before this PR
```julia
julia> Point{3,Float64}(Point(1,2,3))
ERROR: DimensionMismatch("No precise constructor for StaticArrays.SVector{3, Float64} found. Length of input was 1.")
Stacktrace:
 [1] StaticArrays.SVector{3, Float64}(x::Tuple{Tuple{Tuple{Tuple{Point{3, Int64}}}}})
   @ StaticArrays ~/.julia/packages/StaticArrays/rdb0l/src/convert.jl:1
 [2] StaticArray (repeats 4 times)
   @ ~/.julia/packages/StaticArrays/rdb0l/src/convert.jl:4 [inlined]
 [3] Point3(coords::Point{3, Int64})
   @ Meshes ~/.julia/dev/Meshes/src/points.jl:28
 [4] top-level scope
   @ REPL[2]:1

julia> Point(Point(1,2))
Point(Point(1, 2),)
```

## After this PR
```julia
julia> Point{3,Float64}(Point(1,2,3))
Point(1.0, 2.0, 3.0)

julia> Point(Point(1,2,3))
Point(1, 2, 3)
```

## Discussion
This behavior is same as `StaticArrays.SVector`.
```julia
julia> SVector{3,Float64}(SVector(1,2,3))
3-element SVector{3, Float64} with indices SOneTo(3):
 1.0
 2.0
 3.0

julia> SVector(SVector(1,2,3))
3-element SVector{3, Int64} with indices SOneTo(3):
 1
 2
 3
```

# `Line` constructor
## Before this PR
```julia
julia> Line(Point(1,2),Point(2.0,3.0))
ERROR: MethodError: no method matching Line(::Point{2, Int64}, ::Point2)
Closest candidates are:
  Line(::Point{Dim, T}, ::Point{Dim, T}) where {Dim, T} at /home/hyrodium/.julia/dev/Meshes/src/primitives/line.jl:13
Stacktrace:
 [1] top-level scope
   @ REPL[3]:1
```

## After this PR
```julia
julia> Line(Point(1,2),Point(2.0,3.0))
Line{2, Float64}(Point(1.0, 2.0), Point(2.0, 3.0))
```

### DIFF
--- a/src/points.jl
+++ b/src/points.jl
@@ -26,6 +26,7 @@ end
 
 # convenience constructors
 Point{Dim,T}(coords...) where {Dim,T} = Point{Dim,T}(SVector{Dim,T}(coords...))
+Point{Dim,T}(p::Point) where {Dim,T} = Point{Dim,T}(p.coords)
 Point(coords::SVector{Dim,T}) where {Dim,T} = Point{Dim,T}(coords)
 Point(coords::AbstractVector{T}) where {T} = Point{length(coords),T}(coords)
 Point(coords...) = Point(SVector(coords...))
--- a/src/points.jl
+++ b/src/points.jl
@@ -27,6 +27,7 @@ end
 # convenience constructors
 Point{Dim,T}(coords...) where {Dim,T} = Point{Dim,T}(SVector{Dim,T}(coords...))
 Point{Dim,T}(p::Point) where {Dim,T} = Point{Dim,T}(p.coords)
+Point(p::Point) = Point(p.coords)
 Point(coords::SVector{Dim,T}) where {Dim,T} = Point{Dim,T}(coords)
 Point(coords::AbstractVector{T}) where {T} = Point{length(coords),T}(coords)
 Point(coords...) = Point(SVector(coords...))
--- a/src/points.jl
+++ b/src/points.jl
@@ -31,6 +31,9 @@ Point(coords::SVector{Dim,T}) where {Dim,T} = Point{Dim,T}(coords)
 Point(coords::AbstractVector{T}) where {T} = Point{length(coords),T}(coords)
 Point(coords...) = Point(SVector(coords...))
 
+# promote rule for Point
+Base.promote_rule(::Type{Point{Dim,T}}, ::Type{Point{Dim,S}}) where {Dim, T, S} = Point{Dim, promote_type(T,S)}
+
 # coordinate type conversions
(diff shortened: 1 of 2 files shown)


juliohm:
Can you comment on why we would like to add these constructors? What is the use case you have in mind where you want to pass a point to another point constructor? This is a pattern of use that is a bit strange.

In general, this question is valid for all types in this package. We don't provide constructors for Circle(Circle()), Triangle(Triangle()), ... These are unnatural, maybe you are coming from a OOP view of the world?

hyrodium:
Julia's documentation has the following [description](https://docs.julialang.org/en/v1/manual/conversion-and-promotion/).

>Note that the behavior of `convert(T, x)` appears to be nearly identical to `T(x)`.

We can convert types with `convert(Point{3,Float64}, Point(1,2))` explicitly, but I think most other julia users may write `Point{3,Float64}(Point(1,2))`.

I thought this kind of writing is common in Julia.

* `Int(1.0) === 1`
* `Float64(1) === 1.0`
* `Rational(Rational(1,2)) === 1//2`
* `SVector(SVector(2,3)) ===  SVector(2,3)`

Doesn't that seem reasonable?

juliohm:
This is the case usually for primitive types or array types. Custom types in packages generally do not provide self-constructors just for conversion of type parameters. This introduces an additional layer of variability to maintain in the long-term in user code and so we should focus our attention to other more pressing issues.

As a side comment, are you aware of the helpers `Point3` and `Point3f` etc? They are useful to construct points with particular literals for people who just want to type Int entries and get back floats:

```julia
Point3(1,2,3)
Point2(1,2)
```

juliohm:
I like that you try to follow Julia's docs to write PRs, this is a good thing. Just ask first in our forum to see if it makes sense before committing time to it. We are on Zulip.

hyrodium:
Thanks for the review!
From the next time, I'll ask on Zulip.

juliohm:
And the `convert` methods we provide for `Point` only exist so that people can write `Point2[(1,2), (3,4)]` in arrays according to the Julia docs. They are not used explicitly in user code generally.

hyrodium:
Ah, I see! I was wondering why we defined `convert(::Type{Point{Dim,T}}, coords)`, but now it's clear!

PR #130, opened 2021-04-29
Introduce Ngon

juliohm:
This PR introduces a new type called `Ngon` that is a `Polygon` with known number of vertices. It generalizes the `Triangle`, `Quadrangle`, ..., we had before. The motivation behind this generalization is to avoid [code like this](https://github.com/JuliaGeometry/Meshes.jl/blob/e81efb1025f2a43dd295cea8bc9281af8cb453bd/src/topostructures/conversions.jl#L81-L87) where we are inserting branches just to figure out the correct name of the geometry. With this PR names are just aliases for Ngon with specific number of vertices. This information is known at compile time so we can safely construct `Ngon{N}` in the snippet above  and it will behave as if it was a specific subtype.

Appreciate if you can review this one @serenity4. I think we are ok assuming that all polygons are uniquely determined by their number of vertices and that we always make them follow counter-clockwise convention.


serenity4 (approved):
Looks great, thanks! I don't see any modifications to add.

PR #142, opened 2021-06-02
Add vertices method for Box

serenity4:
A convenience method for retrieving all corners of a box. It is as fast as doing it by hand (~20ns for retrieving all 4 corners of a 2D box) thanks to the use of a `@generated` function.


src/primitives/box.jl
```diff
@@ -33,6 +33,23 @@ measure(b::Box) = prod(b.max - b.min)
 diagonal(b::Box) = norm(b.max - b.min)
 sides(b::Box) = b.max - b.min
 
+"""
+Return all corners of the box.
+"""
+@generated function vertices(b::Box)
```
juliohm: That is awesome. As usual you playing with very fancy Julia constructs 🕶️ Can we also pass type parameters `b::Box{Dim,T}` instead of querying with `embeddim` and `coordtype` inside the function?

src/primitives/box.jl
```diff
@@ -33,6 +33,23 @@ measure(b::Box) = prod(b.max - b.min)
...
+  list = sort(collect(distinct(IterTools.subsets(repeat(1:2, paramdim(b)), Val{paramdim(b)}()))))
```
juliohm: Can you elaborate or maybe split this line into multiple lines? It is a bit complicated to grasp, at least reading it for the first time. Perhaps split into multiple lines explaining what is going on.

src/primitives/box.jl
```diff
@@ -33,6 +33,23 @@ measure(b::Box) = prod(b.max - b.min)
...
+  P = Point{embeddim(b),coordtype(b)}
+  ex = Expr(:ref, P)
+  map(list) do idxs
+    next_tuple = Expr(:tuple, (:(coords[$(idxs[i])][$i]) for i in eachindex(idxs))...)
+    push!(ex.args, next_tuple)
+  end
```
juliohm: So you are looping over all possible indices right? Can´t we use CartesianIndices for that in a cleaner way?

```julia
julia> for I in CartesianIndices((2,2))
         @show I
       end
I = CartesianIndex(1, 1)
I = CartesianIndex(2, 1)
I = CartesianIndex(1, 2)
I = CartesianIndex(2, 2)
```

I think it already uses @generated behind the scenes.

juliohm (changes requested):
Awesome to have ❤️

src/primitives/box.jl
serenity4: Sure.

src/primitives/box.jl
serenity4: Oooh I didn't know that. Indeed that simplifies everything :sweat_smile:

src/primitives/box.jl
serenity4: Isn't it clear? We take the sorted version of distinct subsets of indices 1:2 repeated over the parametric dimension of b :laughing:

src/primitives/box.jl
serenity4: Ah maybe not exactly, in 3D I need to generate
```julia
[[1, 1, 1], [1, 1, 2], [1, 2, 1], [1, 2, 2], [2, 1, 1], [2, 1, 2], [2, 2, 1], [2, 2, 2]]
```

juliohm: Also we need to make sure that the CCW orientation is respected in 2D I guess? So that we can feed a Box into a mesh algorithm for example. Perhaps this level of automation needs to be more manual than what we would like it to be. Maybe we should cover the 2D and 3D cases manually for safety?

src/primitives/box.jl
juliohm: 🤣 now that you read it out loud I can start processing it hahaha. These codes with fancy iterations are always tricky to grasp.

src/primitives/box.jl
serenity4: Got it working. That's awesome, thanks for suggesting that.

serenity4: I'll look into that.

src/primitives/box.jl
```diff
@@ -33,6 +33,23 @@ measure(b::Box) = prod(b.max - b.min)
 diagonal(b::Box) = norm(b.max - b.min)
 sides(b::Box) = b.max - b.min
 
+"""
+Return all corners of the box.
+"""
+@generated function vertices(b::Box{Dim,T}) where {Dim,T}
+  list = CartesianIndices(Tuple(repeat([1:2], Dim)))
```
juliohm: Maybe `CartesianIndices(ntuple(i->2, Dim))` to avoid heap allocations.

serenity4: That's cleaner too, nice.

src/primitives/box.jl
serenity4: In 2D it should work out well with the ordering given by `CartesianIndices`. In 3D not so much if you want triangles to be CCW and recover all the cube. We need a non-trivial ordering scheme (unlike in 2D where it's rather easy) to link the vertices in that case.

serenity4: To be more accurate, in 2D it works well only if you consider triangle strips (like [shown here](https://www.khronos.org/registry/vulkan/specs/1.2-khr-extensions/html/chap20.html#drawing-triangle-strips)); that is, a particular connectivity scheme defined from a list of vertices.

juliohm: For the sake of clarity, should we use simple code for 2D and 3D for now? We are still working out these CCW details and I have no idea yet how they will play in 3D. So if we could have simple code that is easy to read we could think about generalizations in the future. It is a cool code but won´t necessarily attract many contributors :)

serenity4: I am not sure what you want, do you mean writing what the generated function does by hand for 2D/3D cases or you're talking about triangulating the box?

serenity4: If you're talking about the generated function, then we can put the expression processing code inside another function and expand it like [there](https://stackoverflow.com/questions/66402105/any-way-to-expand-a-generated-function-in-julia) to get the generated expression.

juliohm: I mean we could keep it simple, I don´t see a real need for generated functions and more advanced-looking code when we can't deal with higher dimensions yet. Keeping it simple will help new contributors to suggest improvements. Let's keep it simple for now. That way we reduce the need for external libraries such as IterTools which is currently only used for `ivec`.

serenity4: I don't think it would be useful. Generated functions are very good when you know the logic up-front and don't really want to type in all the code yourself. That way you don't need to look at two different functions for 2D and 3D, particularly when `Box` is already defined very generically. And we support 4D if someone decides he needs to go in 4D (unlikely, but possible); writing 3 functions (2D, 3D by hand and keeping this generated function for higher-D) would be a mess to maintain, and not keeping the generated function would lead to a loss of (admittedly minor) features.

Also, it does quite a simple thing and gives natural results, and I don't really see why we'd want to change it in the future. And if someone decides to change it, a manual version may not be easier since you would need to reconstruct this logic in your head anyway. This `CartesianIndices` logic is very central and if one knows it upfront it will be easier to see how we loop over indices.

Lastly we don't use IterTools at all there (`CartesianIndices` removed the dependency).

serenity4: To mitigate the difficulty that one may have to understand this function, we can add an example in the docstring which shows the logic that it does in 2D and 3D.

src/primitives/box.jl
```diff
@@ -33,6 +33,23 @@ measure(b::Box) = prod(b.max - b.min)
...
+  ex = Expr(:ref, P)
+  map(list) do idxs
+    next_tuple = Expr(:tuple, (:(coords[$(idxs[i])][$i]) for i in 1:Dim)...)
+    push!(ex.args, next_tuple)
+  end
+  quote
+    coords = coordinates(b.min), coordinates(b.max)
+    $ex
+  end
+end
```
juliohm: Again, I don't feel that this is the kind of code that is easy to read and maintain. Metaprogramming has its strengths but this is too big of a hammer to solve a very simple and specific feature with 2D and 3D boxes. You are trying to write code for a future we don't even know we will be able to support (higher dimensions). Let's keep it simple, please.

Suppose we share this code in any forum. Do you think people will know what it is doing in less than 1min? I am positive that people will struggle. Please rewrite it like we do in the CartesianGrid for example. Premature generalizations can be harmful too.

juliohm (changes requested):

src/primitives/box.jl
juliohm: Remember the [Zen of Python](https://www.python.org/dev/peps/pep-0020/#id2):

- Simple is better than complex.
- Explicit is better than implicit.
- Readability counts.
- Special cases aren't special enough to break the rules.
- Although practicality beats purity.
- If the implementation is hard to explain, it's a bad idea.

These are the main ones related to this issue.

serenity4: To be honest, the following snippet (from CartesianGrid) is not easy to understand for me. Locally it is, but I quickly lose sight of the global picture.

Snippet:
```julia
  if Dim == 1 # segment
    p1 = (o[1] + (i[1] - 1) * s[1],)
    p2 = (o[1] + (    i[1]) * s[1],)
    Segment(p1, p2)
  elseif Dim == 2 # quadrangle
    p1 = (o[1] + (i[1] - 1) * s[1],
          o[2] + (i[2] - 1) * s[2])
    p2 = (o[1] + (i[1]    ) * s[1],
          o[2] + (i[2] - 1) * s[2])
    p3 = (o[1] + (i[1]    ) * s[1],
          o[2] + (i[2]    ) * s[2])
    p4 = (o[1] + (i[1] - 1) * s[1],
          o[2] + (i[2]    ) * s[2])
    Quadrangle(p1, p2, p3, p4)
  elseif Dim == 3 # hexahedron
    p1 = (o[1] + (i[1] - 1) * s[1],
          o[2] + (i[2] - 1) * s[2],
          o[3] + (i[3] - 1) * s[3])
    p2 = (o[1] + (i[1]    ) * s[1],
          o[2] + (i[2] - 1) * s[2],
          o[3] + (i[3] - 1) * s[3])
    p3 = (o[1] + (i[1]    ) * s[1],
          o[2] + (i[2]    ) * s[2],
          o[3] + (i[3] - 1) * s[3])
    p4 = (o[1] + (i[1] - 1) * s[1],
          o[2] + (i[2]    ) * s[2],
          o[3] + (i[3] - 1) * s[3])
    p5 = (o[1] + (i[1] - 1) * s[1],
          o[2] + (i[2] - 1) * s[2],
          o[3] + (i[3]    ) * s[3])
    p6 = (o[1] + (i[1]    ) * s[1],
          o[2] + (i[2] - 1) * s[2],
          o[3] + (i[3]    ) * s[3])
    p7 = (o[1] + (i[1]    ) * s[1],
          o[2] + (i[2]    ) * s[2],
          o[3] + (i[3]    ) * s[3])
    p8 = (o[1] + (i[1] - 1) * s[1],
          o[2] + (i[2]    ) * s[2],
          o[3] + (i[3]    ) * s[3])
    Hexahedron(p1, p2, p3, p4, p5, p6, p7, p8)
  else
    throw(ErrorException("not implemented"))
  end
```

I brought a few changes to the generated function. It should be more readable now. I also added the manual versions. They are faster to process (like, you directly see the result), but my opinion is that it's more error-prone because you put yourself at the place of the machine and have to run the program in your head to confirm its validity instead of relying on structured logic. I still think that the generated function is better from a maintainability perspective, but if you still insist, I can delete it as we just probably just have two conflicting opinions on the matter.

juliohm:
Hi @serenity4 , as I commented quickly in Zulip, the tests aren't following CCW order. That is another reason we should keep it simple and eliminate the general metaprogramming piece there for which orientation is extremely difficult to grasp.

serenity4:
How should we do it in 3D then? CCW does not make sense in this case. But we can have a CCW convention for 2D only.

juliohm:
Maybe we should just follow the same ordering in the current CartesianGrid implementation in the 3D case. Later we can figure out how to deal with orientation in 3D across the codebase. The development is still mostly concentrated in 2-manifold meshes.

src/primitives/box.jl
```diff
@@ -33,6 +33,37 @@ measure(b::Box) = prod(b.max - b.min)
 diagonal(b::Box) = norm(b.max - b.min)
 sides(b::Box) = b.max - b.min
 
+"""
+Return all corners of a 2D box in CCW convention.
+"""
```
juliohm:
```suggestion
```

src/primitives/box.jl
```diff
@@ -33,6 +33,34 @@ measure(b::Box) = prod(b.max - b.min)
...
+    (B[1], A[2]),
+    (B[1], B[2]),
+    (A[1], B[2]),
+  ]
+end
+
+"""
+Return all corners of a 3D box with a
+cartesian ordering.
+"""
```
juliohm:
```suggestion
```

src/primitives/box.jl
```diff
@@ -33,6 +33,30 @@ measure(b::Box) = prod(b.max - b.min)
...
 sides(b::Box) = b.max - b.min
 
+function vertices(b::Box{2,T}) where {T}
+  A, B = coordinates(b.min), coordinates(b.max)
+  Point{2,T}[
+    (A[1], A[2]),
+    (B[1], A[2]),
+    (B[1], B[2]),
+    (A[1], B[2]),
+  ]
```
juliohm:
```suggestion
  A = coordinates(b.min)
  B = coordinates(b.max)
  Point.([
    (A[1], A[2]),
    (B[1], A[2]),
    (B[1], B[2]),
    (A[1], B[2]),
  ])
```

src/primitives/box.jl
```diff
@@ -33,6 +33,31 @@ measure(b::Box) = prod(b.max - b.min)
 diagonal(b::Box) = norm(b.max - b.min)
 sides(b::Box) = b.max - b.min
 
+function vertices(b::Box{2,T}) where {T}
```
juliohm:
```suggestion
function vertices(b::Box{2})
```

src/primitives/box.jl
```diff
@@ -33,6 +33,31 @@ measure(b::Box) = prod(b.max - b.min)
...
+    (A[1], A[2], A[3]),
+    (B[1], A[2], A[3]),
+    (A[1], B[2], A[3]),
+    (B[1], B[2], A[3]),
+    (A[1], A[2], B[3]),
+    (B[1], A[2], B[3]),
+    (A[1], B[2], B[3]),
+    (B[1], B[2], B[3]),
+  ]
+end
```
juliohm:
```suggestion
function vertices(b::Box{3})
  A = coordinates(b.min)
  B = coordinates(b.max)
  Point.([
    (A[1], A[2], A[3]),
    (B[1], A[2], A[3]),
    (A[1], B[2], A[3]),
    (B[1], B[2], A[3]),
    (A[1], A[2], B[3]),
    (B[1], A[2], B[3]),
    (A[1], B[2], B[3]),
    (B[1], B[2], B[3]),
  ])
end
```

test/primitives.jl
```diff
@@ -68,6 +68,12 @@
     @test Meshes.center(b) == P2(1.5,1.5)
     @test diagonal(b) == √T(2)
 
+    b = Box(P2(1,2), P2(3,4))
+    @test vertices(b) == P2[(1,2),(3,2),(3,4),(1,4)]
+
+    b = Box(P3(1,2,3), P3(4,5,6))
+    @test vertices(b) == P3[(1,2,3),(4,2,3),(1,5,3),(4,5,3),(1,2,6),(4,2,6),(1,5,6),(4,5,6)]
```
juliohm: This second test is following a different convention than the Hexahedron case in CartesianGrid. The current convention is CCW in the front face and then another CCW in the back face.

juliohm (changes requested):

juliohm:
@serenity4 did you have chance to take a second look into this?

serenity4:
Not yet, I'll try to get some time in the following days.

serenity4:
I think we're good now.

juliohm (approved):

juliohm:
That is awesome. I had missed the notification. Merged ❤️

PR #146, opened 2021-07-10
Replace GeoStats.jl with Meshes.jl in docs

serenity4:
A small migration artifact it seems :smile:


juliohm (approved):
That is true, thanks for fixing :)

PR #150, opened 2021-07-25
Implement edges generator for elements

mohamed82008:
In this PR, I implement the function `edges` which returns a generator. The generator returns Segments when collected or iterated over.

### DIFF
--- a/src/polytopes/hexahedron.jl
+++ b/src/polytopes/hexahedron.jl
@@ -13,3 +13,12 @@ end
 
 nvertices(::Type{<:Hexahedron}) = 8
 nvertices(h::Hexahedron) = nvertices(typeof(h))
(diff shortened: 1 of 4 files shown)


mohdibntarek:
Needs tests.

juliohm:
This functionality should go in the more general topological relations in `toporelations`. It is a `boundary` relation where you give a hexahedron element (rank 3) and ask for the edges (rank 1). Try to take a look at the test suite for the topological relations and the main reference in the docstring to understand the generality. We can iterate over vertices, edges, faces, etc using these topological relations, and the code is much cleaner that way.

juliohm:
The issue is being discussed in the Zulip channel. Closing the PR and looking forward to new additions to the topological relations 💯

PR #162, opened 2021-08-08
Sampling from a rectangle with Int vertices throws DomainError

### LINKED ISSUE #161: Constructors with Integer literals should convert to floating point coordinates
If I try to sample from a triangle whose vertices are float points, nothing is wrong. 
```julia 
julia> Ω = Ngon(Point(0., 1.), Point(1., 0.), Point(1., 1.))
Triangle{2,Float64}
  └─Point(0.0, 1.0)
  └─Point(1.0, 0.0)
  └─Point(1.0, 1.0)

julia> pnt = sample(Ω, HomogeneousSampling(5)) |> collect
5-element Vector{Point2}:
 Point(0.445212832411942, 0.9489938469778333)
 Point(0.43267110681743415, 0.8483702741628089)
 Point(0.9689446652731993, 0.49281098786470057)
 Point(0.940827342417384, 0.2071168869475296)
 Point(0.5334816086641955, 0.9112811451205238)
```
However if types of the vertices are integer, then I get a DomainError. 
```julia 
ulia> Ω = Ngon(Point(0, 1), Point(1, 0), Point(1, 1))
Triangle{2,Int64}
  └─Point(0, 1)
  └─Point(1, 0)
  └─Point(1, 1)

julia> pnt = sample(Ω, HomogeneousSampling(5)) |> collect
ERROR: DomainError with -3.3173935348431923e18:
sqrt will only return a complex result if called with a complex argument. Try sqrt(Complex(x)).
Stacktrace:
  [1] throw_complex_domainerror(f::Symbol, x::Float64)
    @ Base.Math ./math.jl:33
  [2] sqrt
    @ ./math.jl:582 [inlined]
  [3] sqrt
    @ ./math.jl:608 [inlined]
  [4] (::Meshes.var"#randpoint#172"{Int64, StaticArrays.SVector{2, Int64}, StaticArrays.SVector{2, Int64}, StaticArrays.SVector{2, Int64}})()
    @ Meshes ~/.julia/packages/Meshes/0vOB2/src/sampling/homogeneous.jl:38
  [5] #171
    @ ./generator.jl:0 [inlined]
  [6] iterate
    @ ./generator.jl:47 [inlined]
  [7] iterate
    @ ~/.julia/packages/IterTools/0dYLc/src/IterTools.jl:869 [inlined]
  [8] collect_to!
    @ ./array.jl:724 [inlined]
  [9] collect_to_with_first!(dest::Vector{Point2}, v1::Point2, itr::IterTools.IVec{Base.Generator{UnitRange{Int64}, Meshes.var"#171#173"{Meshes.var"#randpoint#172"{Int64, StaticArrays.SVector{2, Int64}, StaticArrays.SVector{2, Int64}, StaticArrays.SVector{2, Int64}}}}}, st::Int64)
    @ Base ./array.jl:702
 [10] _collect(c::UnitRange{Int64}, itr::IterTools.IVec{Base.Generator{UnitRange{Int64}, Meshes.var"#171#173"{Meshes.var"#randpoint#172"{Int64, StaticArrays.SVector{2, Int64}, StaticArrays.SVector{2, Int64}, StaticArrays.SVector{2, Int64}}}}}, #unused#::Base.EltypeUnknown, isz::Base.HasLength)
    @ Base ./array.jl:696
 [11] collect
    @ ./array.jl:602 [inlined]
 [12] |>(x::IterTools.IVec{Base.Generator{UnitRange{Int64}, Meshes.var"#171#173"{Meshes.var"#randpoint#172"{Int64, StaticArrays.SVector{2, Int64}, StaticArrays.SVector{2, Int64}, StaticArrays.SVector{2, Int64}}}}}, f::typeof(collect))
    @ Base ./operators.jl:858
 [13] top-level scope
    @ REPL[59]:1
julia> 
```
I think this looks awkward and worths opening an issue here,

zekeriyasari:
Fixes #161

### DIFF
--- a/src/points.jl
+++ b/src/points.jl
@@ -28,7 +28,7 @@ end
 Point{Dim,T}(coords...) where {Dim,T} = Point{Dim,T}(SVector{Dim,T}(coords...))
 Point(coords::SVector{Dim,T}) where {Dim,T} = Point{Dim,T}(coords)
 Point(coords::AbstractVector{T}) where {T} = Point{length(coords),T}(coords)
-Point(coords...) = Point(SVector(coords...))
+Point(coords...) = Point(float(SVector(coords...)))
 
 # coordinate type conversions
 Base.convert(::Type{Point{Dim,T}}, coords) where {Dim,T} = Point{Dim,T}(coords)
--- a/test/points.jl
+++ b/test/points.jl
@@ -2,7 +2,8 @@
   @test embeddim(Point(1)) == 1
   @test embeddim(Point(1, 2)) == 2
   @test embeddim(Point(1, 2, 3)) == 3
-  @test coordtype(Point(1, 1)) == Int
+  # @test coordtype(Point(1, 1)) == Int
+  @test coordtype(Point(1, 1)) == Float64  # Internal conversion to Float64
   @test coordtype(Point(1.,1.)) == Float64
   @test coordtype(Point(1f0, 1f0)) == Float32
   @test coordtype(Point1(1)) == Float64
@@ -67,7 +68,8 @@
   @test coordtype(Point([1.,2.,3.])) == Float64
   
   # check all 1D Point constructors, because those tend to make trouble
-  @test Point(1) == Point((1,)) == Point([1])
+  # @test Point(1) == Point((1,)) == Point([1])
+  @test Point(1) == Point((1,)) == Point([1.0]) # Internal conversion to Float64
   @test Point{1,Int}(-2) == Point{1,Int}((-2,)) == Point{1,Int}([-2])
   @test Point{1,T}(0) == Point{1,T}((0,)) == Point{1,T}([0])
  


src/points.jl
```diff
@@ -28,7 +28,7 @@ end
 Point{Dim,T}(coords...) where {Dim,T} = Point{Dim,T}(SVector{Dim,T}(coords...))
 Point(coords::SVector{Dim,T}) where {Dim,T} = Point{Dim,T}(coords)
 Point(coords::AbstractVector{T}) where {T} = Point{length(coords),T}(coords)
-Point(coords...) = Point(SVector(coords...))
+Point(coords...) = Point(float(SVector(coords...)))
```
juliohm, line 31: What we really want is a more specific method that dispatches on `Integer` and converts to `Float64`, otherwise will be ignoring user input. Only the specific case where the user types integer literals needs to be amended.

juliohm (changes requested):
We need additional methods to dispatch on Integer.

src/points.jl
zekeriyasari: > What we really want is a more specific method that dispatches on Integer and converts to Float64

Sorry, but I could not get what you really want. 

> Only the specific case where the user types integer literals needs to be amended.

This is just the case. Here are sample constructors. 
```julia 
julia> Point(1, 2) 
Point(1.0, 2.0)

julia> Point(1., 2.)
Point(1.0, 2.0)

julia> Point(1f0, 2f0) 
Point(1.0f0, 2.0f0)

ulia> Point([1, 2])
Point(1, 2)

julia> Point([1., 2.])
Point(1.0, 2.0)

julia> Point(@SVector [1, 2])
Point(1, 2)

julia> Point(@SVector [1., 2.])
Point(1.0, 2.0)

julia> Point(@SVector [1f0, 2f0])
Point(1.0f0, 2.0f0)
```
Note that if the argument to the constructor is a vector, the int values are not converted to Float64. I think those integers must be converted to Float64, too. but, after your review, 
>otherwise will be ignoring user input.

I am confused a little bit.

> What we really want is a more specific method that dispatches on Integer and converts to Float64 

May be want you is something like. 
```suggestion
Point(coords::Integer...) = Point(float(SVector(coords...)))
```

juliohm: Consider now that someone types `Point(1°, 2°)` or any instance of a more complicated type that is not `Number`, we don't want to convert that to float. Makes sense? That is why we need to leave the existing constructors untouched and make sure that we have an additional method to catch the Integer literal case. Right now you converting anything that is passed in to float.

juliohm:
Also notice that we have `Point2` and `Point3` with forced float conversion already. So I am not 100% sure we should always convert Integer to float. Still thinking about it... What do you think @serenity4 ?

zekeriyasari:
Not to derail the thread, but vectorized construction of points also seemed odd to me as well.
```julia 
julia> [Point2(p) for p in [[0, 0], [1, 0], [1, 1], [1, 1]]]
4-element Vector{Point2}:
 Point(0.0, 0.0)
 Point(1.0, 0.0)
 Point(1.0, 1.0)
 Point(1.0, 1.0)

julia> pts = Point2.([0, 0], [1, 0], [1, 1], [1, 1])
ERROR: DimensionMismatch("No precise constructor for StaticArrays.SVector{2, Float64} found. Length of input was 4.")
Stacktrace:
 [1] StaticArrays.SVector{2, Float64}(x::Tuple{Tuple{Tuple{NTuple{4, Int64}}}})
   @ StaticArrays ~/.julia/packages/StaticArrays/aMe0r/src/convert.jl:1
 [2] StaticArray (repeats 4 times)
   @ ~/.julia/packages/StaticArrays/aMe0r/src/convert.jl:4 [inlined]
 [3] Point2(::Int64, ::Vararg{Int64, N} where N)
   @ Meshes ~/.julia/packages/Meshes/wFYhe/src/points.jl:28
 [4] _broadcast_getindex_evalf
   @ ./broadcast.jl:648 [inlined]
 [5] _broadcast_getindex
   @ ./broadcast.jl:621 [inlined]
 [6] getindex
   @ ./broadcast.jl:575 [inlined]
 [7] copy
   @ ./broadcast.jl:922 [inlined]
 [8] materialize(bc::Base.Broadcast.Broadcasted{Base.Broadcast.DefaultArrayStyle{1}, Nothing, Type{Point2}, NTuple{4, Vector{Int64}}})
   @ Base.Broadcast ./broadcast.jl:883
 [9] top-level scope
   @ REPL[74]:1

julia> [Point(p) for p in [[0, 0], [1, 0], [1, 1], [1, 1]]]
4-element Vector{Point{2, Int64}}:
 Point(0, 0)
 Point(1, 0)
 Point(1, 1)
 Point(1, 1)

julia> pts = Point.([0, 0], [1, 0], [1, 1], [1, 1])
2-element Vector{Point{4, Int64}}:
 Point(0, 1, 1, 1)
 Point(0, 0, 1, 1)
```

juliohm:
Check the tests for the expected usage.

juliohm:
I will close this PR so that you can start it fresh 👍🏽

Issue #125, opened 2021-04-21
Calculate ray mesh intersection


jackbeagley:
We can already find the intersection between lines, segments and boxes. It would be good to find the intersection between a mesh and a line too (naturally this would involve first finding the intersection, if it exists, between a line and a triangle).
Could use https://github.com/JuliaGeometry/TriangleIntersect.jl

juliohm:
Thank you @jackbeagley for the feature request. Would you like to submit a PR with the proposed method? Notice that we have three different concepts in the codebase already: Line, Segment, Ray. These are not the same concept and it seems that the TriangleIntersect.jl package provides an intersection between Ray and Triangle. Make sure you implement the correct method in the PR if you are willing to contribute.

jackbeagley:
Point taken regarding the different concepts! I'll have a crack at it and hopefully have a PR at some point to it.

jackbeagley:
So I have written most of the code for this, but while testing `intersect(segment, mesh)` or equivalently `intersect(ray, mesh)`. I found that `SimpleMesh` is not a subtype of `Geometry`, which seems strange to me, surely an object made up of `Geometry` is itself `Geometry`? At any rate is it best to define another method for `intersect`, something like `Base.intersect(g::Geometry, m::SimpleMesh)` or does it make sense to make `SimpleMesh` a subtype of `Geometry`?

juliohm:
Currently we have methods for `intersect(geometry, domain)` [here](https://github.com/JuliaGeometry/Meshes.jl/blob/6d5d758adf98bf75545e4ab0b89561b7e4d11a31/src/intersections.jl#L164). The current design makes a distinction between collections of geometries (a domain) such as meshes and the geometries themselves. We can always revise the design if there is a good reason. Experience has shown that we don't gain much by trying to subtype everything as a Geometry.

jackbeagley:
That method only checks if there is an intersection though (returning `true` or `false`), rather than returning the actual intersection. Seems like the way forward might be to implement `Base.intersect(g::Geometry, d::Domain)`?

juliohm:
Yes, I think it is a good way to get it started.

juliohm:
@jackbeagley did you solve this issue already or still missing some methods? :)

jackbeagley:
@juliohm I think everything else is implemented in my fork (for segment and mesh at least), but I'm leaving civilisation tomorrow for four weeks, so even though it's close I just really don't have time to sort it out till then!

[...]

PR #163, opened 2021-08-09
Segment, Triangle intersections

jackbeagley:
Implemented intersection of Segments and Triangles, on the way to fulling implementing #125


jackbeagley:
Remaining codecov is Base.intersect(d::Domain, g::Geometry) = intersect(g, d) which has not yet been implemented. To come shortly!

juliohm:
Thank you @jackbeagley for the really nice work. In order to review things more carefully, we will need to split this PR into multiple PRs tackling different objects. Can you please start with a PR that introduces intersection between segment and triangle? I don't think we covered all possible outputs for example.

Remember that as reviewers, we are trying to spot all possible corner cases, and so currently it is hard to reason about the changes because they are related to various different features (i.e. intersections between various objects, segments, triangles, meshes, ...).

jackbeagley:
> Thank you @jackbeagley for the really nice work. In order to review things more carefully, we will need to split this PR into multiple PRs tackling different objects. Can you please start with a PR that introduces intersection between segment and triangle? I don't think we covered all possible outputs for example.
> 
> Remember that as reviewers, we are trying to spot all possible corner cases, and so currently it is hard to reason about the changes because they are related to various different features (i.e. intersections between various objects, segments, triangles, meshes, ...).

@juliohm I have removed the unrelated changes, so it should just be limited to the triangle segment intersect now. That was laziness on my behalf, I will be more particular in the future!

src/Meshes.jl
```diff
@@ -294,6 +294,7 @@ export
   OverlappingBoxes,
   FaceTouchingBoxes,
   CornerTouchingBoxes,
+  IntersectingSegmentTri,
```
juliohm: I can't figure out what kind of intersection you have in mind from the name. I understand that the result can be a point, a segment or no intersection, correct?

src/intersections.jl
```diff
@@ -73,6 +73,15 @@ struct CornerTouchingBoxes{P<:Point} <: Intersection
   value::P
 end
 
+# -------------------------------
+# SEGMENT-TRIANGLE INTERSECTIONS
+# -------------------------------
+
+struct IntersectingSegmentTri{P<:Point, T} <: Intersection
+  value::P
+  t::T
```
juliohm: What is this parameter t? Reading this struct I am assuming that IntersectingSegmentTri refers to the point case?

src/intersections/segments.jl
```diff
@@ -167,3 +167,86 @@ function intersectcolinear(s1::Segment{Dim,T}, s2::Segment{Dim,T}) where {Dim,T}
     NoIntersection()
   end
 end
+
+# intersection of Segment and Triangle
+function intersecttype(s::Segment{3,T}, t::Triangle{3,T}) where {T}
+  # get the triangle's and segment's vertices
+  s_v = s.vertices
+  t_v = t.vertices
```
juliohm: Notice that whenever we are using these objects in external functions that are not defined in the same source file of the object we can't use internal fields. We need to call the public API `vertices(s)` instead of `s.vertices`.

src/intersections/segments.jl
```diff
@@ -167,3 +167,86 @@ function intersectcolinear(s1::Segment{Dim,T}, s2::Segment{Dim,T}) where {Dim,T}
...
+
+  s₁ = s_v[1]
+  s₂ = s_v[2]
+
+  # simple checks to see if the triangle overlaps with the bounding box of the segment
+  # this check is much faster than calculating an intersection
+  if any(i -> all(j -> (s₁.coords[i] > t_v[j].coords[i]) && (s₂.coords[i] > t_v[j].coords[i]), 1:3), 1:3)
```
juliohm: Any chance this algorithm can be agnostic to embedding dimension? It only works for triangles embedded in 3D space or also works for 2D space?

Again, the same comments apply about using public API instead of `s.coords`

juliohm (changes requested):
Thank you @jackbeagley for the quick update. I left a few important comments about usage of public API in functions that are not "owned" by the objects. When you wrote `normal(triangle)` it was fine to use internal constructs because that function was defined next to the struct itself and also it is "owned" in a sense by the object. In the case of `intersecttype` we have two inputs so none of them owns the function.

Also, regarding the algorithm, we need some good references in the docstring. It is hard to maintain and update the code in the future otherwise. Do you have a paper or well-known algorithm in mind?

src/intersections/segments.jl
jackbeagley: This part here is certainly agnostic to embedding dimension, I could split out this into a separate function perhaps?

juliohm: No worries, I was just wondering if the entire algorithm was agnostic to dimension.

juliohm:
@jackbeagley did you have a chance to take a look at this?

jackbeagley:
Hi @juliohm , some other stuff has come up over the last couple of weeks but I'm hoping to get back to it this coming week!

jackbeagley:
So... I would really love some help here. The algorithm in `intersecttype(s::Segment{3,T}, t::Triangle{3,T})` is working really nicely and is fast. _However_ the paper I referenced considers the case of a coplanar segment and triangle to not be an intersection, which I don't really want to do. I did some research on algorithms for the case of coplanar segments and triangles in 3D but couldn't find anything, so I tried to consider all the cases and wrote `intersectcoplanar(s::Segment{Dim,T}, t::Triangle{Dim,T})` anyway. I think it covers all the cases but there are issues:
- The code is type unstable and I can't think of a good, elegant way to fix this (might just need to come back to it for a fresh perspective?)
- I'm not sure it covers all the cases so would really love another set of eyes on it

Also, you might notice that the tests don't all pass... One of them might benefit from `isapprox` being implemented for `Segment`. The test also begs the question as to whether `Segment(p1, p2) == Segment(p2, p1)`. Is this something that has been covered?

Really keen to hear your thoughts @juliohm

src/Meshes.jl
```diff
@@ -294,6 +294,8 @@ export
...
```
```suggestion
  IntersectingSegmentTri,
```

Please replace `Tri` by the explicit `Triangle` everywhere.

src/intersections/segments.jl
```diff
@@ -167,3 +167,268 @@ function intersectcolinear(s1::Segment{Dim,T}, s2::Segment{Dim,T}) where {Dim,T}
...
   end
 end
+
+"""
+    intersecttype(s, t)
+
+Calculate the intersection of a segment and triangle in 3D.
+Algorithm from Jiménez et al 2009 (A robust segment/triangle intersection algorithm for interference tests. Efficiency study). 
+"""
+function intersecttype(s::Segment{3,T}, t::Triangle{3,T}) where {T}
```
juliohm: Let's add this implementation to a separate source file about triangles, `triangles.jl`

src/intersections/segments.jl
```diff
@@ -167,3 +167,268 @@ function intersectcolinear(s1::Segment{Dim,T}, s2::Segment{Dim,T}) where {Dim,T}
     NoIntersection()
   end
 end
+
+"""
+    intersecttype(s, t)
+
+Calculate the intersection of a segment and triangle in 3D.
+Algorithm from Jiménez et al 2009 (A robust segment/triangle intersection algorithm for interference tests. Efficiency study).
```
juliohm: Love to see contributions with references! ❤️

juliohm (changes requested):
That is an amazing PR as usual @jackbeagley 💯 I will try to review it more carefully over the next weeks, but feel free to go ahead and fix some small naming issues I pointed out.

The code is really easy to read, so I will try to read the paper and understand what the implementation is doing. Thank you again for the very organized PR.

juliohm:
@jackbeagley did you have a chance to take a look into this? Sorry for pinging, I am just trying to keep things moving. I know we are all busy and alternating different tasks. This PR is a very nice addition after we sort out the minor details.

jackbeagley (commented):
> @jackbeagley did you have a chance to take a look into this? Sorry for pinging, I am just trying to keep things moving. I know we are all busy and alternating different tasks. This PR is a very nice addition after we sort out the minor details.

Just submitted those small changes now! Haven't had time to improve the `intersectcoplanar` function though unfortunately.

Just wondering though, what are your thoughts of projecting the coplanar case into 2D and determining intersections from there. Would that make any sense?

juliohm:
> Just wondering though, what are your thoughts of projecting the coplanar case into 2D and determining intersections from there. Would that make any sense?

At first it seems like a reasonable approach. I will try to read the paper in between the multiple projects...

juliohm:
@jackbeagley do you think you will have time to refine this PR in the following months?

jackbeagley:
Hey @juliohm, as of yesterday I am in MIQ, so finally have some time away from the lab to work on this again. After a big break I want to propose that for now we do not return an intersection for the degenerate case of a segment co-planar with the triangle. I haven't found anything in literature for **calculating and returning** the intersection between a segment and triangle in 2D (equivalent to the degenerate case), likely due to the issues we encountered here. This behavior would match that of the original algorithm in _Jimenez2010_, in which the author notes that this is unlikely to be an issue as if the segment is co-planar, it will intersect conventionally with a neighboring element. I will have a crack at implementing the algorithm as described in this reply so you can take a look.

If segment-triangle intersection is implemented in 2D in the future then it should be simple enough to implement for the 3D degenerate case.

juliohm:
Awesome @jackbeagley , that sounds like a good plan. I think we will only be able to address these intersection algorithms properly with geometric algebra in the future. Let me know when the PR is ready for review again and I will try to finally find time to read the paper. Hectic days around here.

GuoHaoxuan:
I want to use this feature in my work recently and have focused on this for a long time. It looks a bit slow to go. Can I join this to accelerate the progress? If so, how?

juliohm:
@HaoxuanGuo contributions are very welcome. What exactly would you like to contribute? Improvements to this PR or geometric algebra in general? The Meshes.jl project has many missing features and I could open issues to list some easy features to add as a starter.

GuoHaoxuan:
@juliohm Although I will use ray/triangle intersection in my work (this PR). But both this PR or other geometric algebra in general is either OK.

GuoHaoxuan:
@jackbeagley There are some thinking about type-stable and performance.

I think (1. checking if a ray and a triangle intersects) and (2. find the intersection point or line) is different condition.

## 1. Just Checking the Intersection

In (1. just checking), we may want to find the the hit with a ray and a geometry roughly. May want to render a image, trace a particle, or click objects on mouse click.

In mouse click condition, Microsoft wrote this code (Sorry, in C++, not Julia):

<details>
<summary>Code in DirectX SDK Demo</summary>

```C++
// Determine whether a ray intersect with a triangle
// Parameters
// orig: origin of the ray
// dir: direction of the ray
// v0, v1, v2: vertices of triangle
// t(out): weight of the intersection for the ray
// u(out), v(out): barycentric coordinate of intersection

bool IntersectTriangle(const Vector3& orig, const Vector3& dir,
    Vector3& v0, Vector3& v1, Vector3& v2,
    float* t, float* u, float* v)
{
    // E1
    Vector3 E1 = v1 - v0;

    // E2
    Vector3 E2 = v2 - v0;

    // P
    Vector3 P = dir.Cross(E2);

    // determinant
    float det = E1.Dot(P);

    // keep det > 0, modify T accordingly
    Vector3 T;
    if( det >0 )
    {
        T = orig - v0;
    }
    else
    {
        T = v0 - orig;
        det = -det;
    }

    // If determinant is near zero, ray lies in plane of triangle
    if( det < 0.0001f )
        return false;

    // Calculate u and make sure u <= 1
    *u = T.Dot(P);
    if( *u < 0.0f || *u > det )
        return false;

    // Q
    Vector3 Q = T.Cross(E1);

    // Calculate v and make sure u + v <= 1
    *v = dir.Dot(Q);
    if( *v < 0.0f || *u + *v > det )
        return false;

    // Calculate t, scale parameters, ray intersects triangle
    *t = E2.Dot(Q);

    float fInvDet = 1.0f / det;
    *t *= fInvDet;
    *u *= fInvDet;
    *v *= fInvDet;

    return true;
}
```

</details>

The code above describe möller's algotithm, and it **returns a bool**. In julia, `Union{T, Nothing}` is considered specially with fast performance (type-stable). Thus, in this condition we roughly omit co-planar and get a huge performance imporvement.

## 2. Need Exactly Intersect Result

Of course, we may want to get some exact result. Such as Boolean operation in geometry and etc.. (Or want to get the result when have known they intersects.) In this condition, a bit slower is acceptable.

---

In conclusion, consider all possibility in one function may be not elegant. Separate the two condition in different option functions may be better. Because the scenes to be used is different.

jackbeagley:
Hi @HaoxuanGuo,

I would encourage you to check for yourself that `intersecttype(s::Segment{3,T}, t::Triangle{3,T})` is type-stable (using `@code_warntype(s, t)` where `s` and `t` are a segment and triangle that you have defined). While `Body::Union{NoIntersection, IntersectingSegmentTriangle{Point3}}` shows up in red, Julia should optimise for this with _Union-splitting_, so it won't be an issue (see [https://julialang.org/blog/2018/08/union-splitting/](url)).

The code I have written is specifically for segments so I am not sure how helpful it is for you. Reading between the lines it sounds like you aren't trying to find an exact intersection, but rather just whether one exists? In that case, I suggest you look into [https://juliageometry.github.io/Meshes.jl/stable/algorithms/intersection.html#Meshes.hasintersect](url) which I think is what you need? You may need to implement this for `ray` and `triangle`.

jackbeagley:
@juliohm, I am pretty happy now and would love for you to take a look at it! I have hit 100% codecov of the diff, but I added a lot of tests to get there. Are you happy with the way I have implemented all the tests?

juliohm:
Wow, that is an incredible work @jackbeagley! Looking forward to merging it right away. Can you just split the tests with multiple @test calls instead? That way we can always nail down the specific test that is failing and copy/paste it to the REPL for further investigation. When we place the tests in a loop like this it is hard to find out which one exactly failed.

jackbeagley:
> Wow, that is an incredible work @jackbeagley! Looking forward to merging it right away. Can you just split the tests with multiple @test calls instead? That way we can always nail down the specific test that is failing and copy/paste it to the REPL for further investigation. When we place the tests in a loop like this it is hard to find out which one exactly failed.

Ah I hadn't at all considered that!
Given how I have structured the tests do you think it would be acceptable to make a function that tests each element of the test vector I constructed? Something like:

```
  function test_sᵥᵢtⱼ(i, sᵥtⱼ, tⱼ)
    @test isa(intersecttype(sᵥtⱼ[1], tⱼ), sᵥtⱼ[3])
    @test isa(intersecttype(reverse_segment(sᵥtⱼ[1]), tⱼ), sᵥtⱼ[3])

    if isnothing(sᵥtⱼ[2])
      @test isnothing(sᵥtⱼ[1] ∩ tⱼ)
    else
      @test sᵥtⱼ[1] ∩ tⱼ == sᵥtⱼ[2]
    end
  end

 test_sᵥᵢtⱼ(1, sᵥt₁, t₁)
 test_sᵥᵢtⱼ(2, sᵥt₁, t₁)
 test_sᵥᵢtⱼ(3, sᵥt₁, t₁)
```

Obviously it isn't perfect, but would help with maintainability of these tests

juliohm:
I think the functional approach is still very implicit. It makes it hard for anyone that is not the author of the tests to review or extend the existing tests. It is also an approach that is not present in the other test sets we have written.

I appreciate if you can do this additional refactoring of the tests even if it feels a bit repetitive. That way we can actually improve the comments for specific sections like:

```julia
# the following triangles have this property, therefore XYZ should happen
@test ...
@test ...
@test ...

# the following triangles on the other hand do not respect ABC
@test ...
@test ...
@test ...
```

We try to follow the rule of thumb "explicit is better than implicit", particularly in the tests.

jackbeagley:
> I think the functional approach is still very implicit. It makes it hard for anyone that is not the author of the tests to review or extend the existing tests. It is also an approach that is not present in the other test sets we have written.
> 
> I appreciate if you can do this additional refactoring of the tests even if it feels a bit repetitive. That way we can actually improve the comments for specific sections like:
> 
> ```julia
> # the following triangles have this property, therefore XYZ should happen
> @test ...
> @test ...
> @test ...
> 
> # the following triangles on the other hand do not respect ABC
> @test ...
> @test ...
> @test ...
> ```
> 
> We try to follow the rule of thumb "explicit is better than implicit", particularly in the tests.

No worries, I'm on it!

jackbeagley:
@juliohm All tests are explicit now

GuoHaoxuan:
@jackbeagley Thank you for your inform on *Union-splitting*. I post the thinking above is because I want to try to accelerate it via separating the two conditions. Now, due to Union-splitting, there is no huge cost on finding methods. Thus, the separating is not needed. Your code is elegant. 😄

src/intersections/triangles.jl
```diff
@@ -0,0 +1,127 @@
+# ------------------------------------------------------------------
+# Licensed under the MIT License. See LICENSE in the project root.
+# ------------------------------------------------------------------
+
+"""
+    intersecttype(s, t)
+
+Calculate the intersection of a segment and triangle in 3D.
+Algorithm from Jiménez et al 2009 (A robust segment/triangle intersection algorithm for interference tests. Efficiency study).
```
juliohm: Please follow our citation style, here is an example: https://github.com/JuliaGeometry/Meshes.jl/blob/d2e312300eb8ab75c8d8baebe600aaf56004dd6b/src/intersections/segments.jl#L19-L23

test/intersections.jl
```diff
@@ -130,6 +130,123 @@ using Base: hash_integer
...
+    reverse_segment(s) = Segment(vertices(s)[2], vertices(s)[1])
+
+    # segments and triangles in 3D
+    t₁ = Triangle(P3(0, 0, 0), P3(1, 0, 0), P3(0, 1, 0))  # Triangle lying in XY plane
+    t₂ = Triangle(P3(0, 0, 0), P3(2, 0, 0), P3(0, 2, 2))  # Inclined triangle
+    
+    ## Intersections with t₁
+    # Intersects through t₁
+    s₁ = Segment(P3(0.2, 0.2, 1.0), P3(0.2, 0.2, -1.0))
+    @test isa(intersecttype(s₁, t₁), IntersectingSegmentTriangle)
```
juliohm: Please use the infix version `intersecttype(s1, t1) isa IntersectingSegmentTriangle`

test/intersections.jl
```diff
@@ -130,6 +130,123 @@ using Base: hash_integer
...
+    @test s₁ ∩ t₁ == P3(0.2, 0.2, 0.0)
```
juliohm: I noticed that in most tests t1 is fixed. Should we maybe define a single t for this section of tests? Also, I think we don't need a different variable name s1, s2, ... for each test. We can just overwrite a single variable name s.

My understanding is that you picked a given `t` and wrote a collection of tests varying `s`. I would just avoid numbering to avoid the impression that there are tests involving multiple variables at the same type.

juliohm (changes requested):
Awesome update @jackbeagley ! Thanks for refactoring the tests. I left a few more suggestions.

Should we also add a method for intersecttype(triangle, segment) for symmetry?

GuoHaoxuan:
> Should we also add a method for intersecttype(triangle, segment) for symmetry?

@juliohm Maybe it should be added on general intersecttype instead of only `intersecttype(triangle, segment)`. All the intersects is symmetry.

```julia
intersecttype(a::T1, b::T2) where {T1, T2} = intersecttype(b, a)
```

for general fallback.

juliohm:
Thank you @HaoxuanGuo , you are right, we should add this symmetry for any input type.

jackbeagley:
Just implemented those suggestions @juliohm, I presume the symmetry suggestion should be done in its own PWR?

src/intersections/triangles.jl
```diff
@@ -0,0 +1,131 @@
...
+
+"""
+    intersecttype(s, t)
+
+Calculate the intersection of a segment and triangle in 3D.
+
+## References
+
+* Jiménez, J., Segura, R. and Feito, F. 2009. [A robust segment/triangle intersection algorithm for interference tests. Efficiency study]
+(https://www.sciencedirect.com/science/article/pii/S0925772109001448?via%3Dihub)
```
```suggestion
* Jiménez, J., Segura, R. and Feito, F. 2009. [A robust segment/triangle
  intersection algorithm for interference tests. Efficiency study]
  (https://www.sciencedirect.com/science/article/pii/S0925772109001448?via%3Dihub)
```

juliohm:
This is an amazing PR @jackbeagley ! 🎉 Merging it ❤️

juliohm:
We can work on the symmetry in a separate PR 👍🏽

jackbeagley:
> This is an amazing PR @jackbeagley ! 🎉 Merging it ❤️

That's great, we finally got there! Thanks for all the help.

PR #164, opened 2021-08-12
Add convexhull.jl

Andre2957:
Added the file  src/convexhull.jl

### DIFF
--- a/src/convexhull.jl
+++ b/src/convexhull.jl
@@ -10,7 +10,7 @@ This is a modification of the Graham's Scan algorithm, devised by A. M. Andrew, 
 See https://en.wikipedia.org/wiki/Graham_scan
 """
 function upperhull(pset::PointSet)
-  xs = sort(coordinates.(pset), by = first)
+  xs = sort(coordinates.(pset), by=first)
   upperHull = [xs[1],xs[2]]
   for i in 3:length(xs)
     push!(upperHull,xs[i])
--- a/src/convexhull.jl
+++ b/src/convexhull.jl
@@ -2,35 +2,30 @@
     upperhull(pset)
 
 Upper part of the convex hull of a given set of 2D points `pset`. 
-The upperhull is computed by the "Graham Scan" algorithm.
+The upperhull is computed according to the Andrew's Monotone Chain Algorithm.
+See https://en.wikipedia.org/wiki/Convex_hull_algorithms.
+
+The input is sorted on x-coordinate, and the hull is computed in two steps producing the upper and the lower parts of the hull respectively. 
+This is a modification of the Graham's Scan algorithm, devised by A. M. Andrew, known as Andrew's Monotone Chain Algorithm. It has the same basic properties as Graham's scan. 
+See https://en.wikipedia.org/wiki/Graham_scan
 """
 function upperhull(pset::PointSet)
-
-  xs=sort(coordinates.(pset), by=first)
-  
-  upperHull=[xs[1],xs[2]] 
-  
+  xs = sort(coordinates.(pset), by = first)
+  upperHull = [xs[1],xs[2]]
   for i in 3:length(xs)
     push!(upperHull,xs[i])
-    
-    A=Point(upperHull[end-2])
-    B=Point(upperHull[end-1])
-    C=Point(upperHull[end])
-
+    A = Point(upperHull[end-2])
+    B = Point(upperHull[end-1])
+    C = Point(upperHull[end])
     # remove the second to last point in upperHull while the last 3 points make a "left turn"
-    while length(upperHull)>2 && ∠(A,B,C)<0 
-      
+    while length(upperHull)>2 && ∠(A,B,C)<0
       splice!(upperHull,lastindex(upperHull)-1)
-      
       if length(upperHull)>2 
-
-        A=Point(upperHull[end-2])
-        B=Point(upperHull[end-1])
-        C=Point(upperHull[end])
+        A = Point(upperHull[end-2])
+        B = Point(upperHull[end-1])
+        C = Point(upperHull[end])
       end
-    
     end
-
   end 
   upperHull
 end 
--- a/src/convexhull.jl
+++ b/src/convexhull.jl
@@ -1,20 +1,30 @@
+"""
+Compute the upper part of the convex hull of a given set of 2D points 'pset' using the "Graham Scan" algorithm.
+
+*) Sort the points lexicographic
+
+*) Assess the last 3 points in 'upperHull'. If the the clockwise angle is <180° remove the second to last point.
+
+"""
+
 function upperhull(pset::PointSet) 
 
-  xs=Vector{Float64}[] 
   xs=sort(coordinates.(pset), by=first)
 
-  convHull=[xs[1],xs[2]] 
+  upperHull=[xs[1],xs[2]] 
   
-  for i in 3:size(xs,1)
-    push!(convHull,xs[i])
-
-    # remove the second to last point in convHull while the last 3 points make a "left turn"
-    while size(convHull)[1]>2 && ∠(Point(xs[i-2]),Point(xs[i-1]),Point(xs[i]))<0 
-      splice!(convHull,i-1)
+  for i in 3:length(xs)
+    push!(upperHull,xs[i])
+    
+    # remove the second to last point in upperHull while the last 3 points make a "left turn"
+    while length(upperHull)>2 && ∠(Point(upperHull[lastindex(upperHull)-2]),Point(upperHull[lastindex(upperHull)-1]),Point(upperHull[lastindex(upperHull)]))<0 
+      
+      splice!(upperHull,lastindex(upperHull)-1)
+    
     end 
 
   end
 
-  convHull
+  upperHull
 
 end
--- a/src/convexhull.jl
+++ b/src/convexhull.jl
@@ -19,7 +19,7 @@ function upperhull(pset::PointSet)
     C = Point(upperHull[end])
     # remove the second to last point in upperHull while the last 3 points make a "left turn"
     while length(upperHull)>2 && ∠(A,B,C)<0
-      splice!(upperHull,lastindex(upperHull)-1)
+      splice!(upperHull, length(upperHull)-1)
       if length(upperHull) > 2 
         A = Point(upperHull[end-2])
         B = Point(upperHull[end-1])
--- a/src/convexhull.jl
+++ b/src/convexhull.jl
@@ -1,30 +1,36 @@
 """
-Compute the upper part of the convex hull of a given set of 2D points 'pset' using the "Graham Scan" algorithm.
+    upperhull(pset)
 
-*) Sort the points lexicographic
-
-*) Assess the last 3 points in 'upperHull'. If the the clockwise angle is <180° remove the second to last point.
-
+Upper part of the convex hull of a given set of 2D points `pset`. 
+The upperhull is computed by the "Graham Scan" algorithm.
 """
-
-function upperhull(pset::PointSet) 
+function upperhull(pset::PointSet)
 
   xs=sort(coordinates.(pset), by=first)
-
+  
   upperHull=[xs[1],xs[2]] 
   
   for i in 3:length(xs)
     push!(upperHull,xs[i])
     
+    A=Point(upperHull[end-2])
+    B=Point(upperHull[end-1])
+    C=Point(upperHull[end])
+
     # remove the second to last point in upperHull while the last 3 points make a "left turn"
-    while length(upperHull)>2 && ∠(Point(upperHull[lastindex(upperHull)-2]),Point(upperHull[lastindex(upperHull)-1]),Point(upperHull[lastindex(upperHull)]))<0 
+    while length(upperHull)>2 && ∠(A,B,C)<0 
       
       splice!(upperHull,lastindex(upperHull)-1)
+      
+      if length(upperHull)>2 
+
+        A=Point(upperHull[end-2])
+        B=Point(upperHull[end-1])
+        C=Point(upperHull[end])
+      end
     
-    end 
+    end
 
-  end
-
+  end 
   upperHull
-
-end
+end 
--- a/src/convexhull.jl
+++ b/src/convexhull.jl
@@ -1,0 +1,32 @@
+function UpperConvexHull(mySet::PointSet) 
+  
+    x=Array{Float64, 1}[] # Vector of Array{Float64,2}
+   
+    for p in mySet 
+     push!(x,coordinates(p))   
+    end
+   
+    x=sort(x, by=first)
+    
+    convHull=[x[1],x[2]] 
+    for i in 3:size(x,1)
+     push!(convHull,x[i])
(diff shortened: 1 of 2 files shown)


src/convexhull.jl
```diff
@@ -0,0 +1,32 @@
+function UpperConvexHull(mySet::PointSet)
```
juliohm: Function names in Julia are lowercase, can you please rename to `upperhull` and rename the argument to `pset`?

src/convexhull.jl
```diff
@@ -0,0 +1,32 @@
+function UpperConvexHull(mySet::PointSet) 
+  
+    x=Array{Float64, 1}[] # Vector of Array{Float64,2}
```
juliohm: `Array{Float64,1}` is equivalent to `Vector{Float64}`

src/convexhull.jl
```diff
@@ -0,0 +1,32 @@
...
+   
+    for p in mySet 
+     push!(x,coordinates(p))   
+    end
```
juliohm: You can do `xs = coordinates.(pset)` to get all coordinates at once.

src/convexhull.jl
```diff
@@ -0,0 +1,32 @@
...
+   
+    x=sort(x, by=first)
+    
+    convHull=[x[1],x[2]] 
+    for i in 3:size(x,1)
+     push!(convHull,x[i])
+         while size(convHull)[1]>2 && ∠(Point(x[i-2]),Point(x[i-1]),Point(x[i]))<0
```
juliohm: Incorrect indentation, please use 2 whitespaces in all lines.

src/convexhull.jl
```diff
@@ -0,0 +1,32 @@
...
+ 
+              # os ultimos 3 pontos fazem volta pra esquerda i,i-1,i-2
```
juliohm: Please write comments in English as this is an international effort.

src/convexhull.jl
```diff
@@ -0,0 +1,32 @@
...
+         end 
+    
+     end
+     convHull
+     #=
+     conn=Connectivity{Ngon,size(convHull,1)}
+     print(size(convHull,1))
+     print(conn)
+     SimpleMesh(PointSet(convHull),conn)
+ =#
```
juliohm: Please eliminate dead code from the PR

juliohm (changes requested):
Hi @Andre2957 thank you for the PR. Added a first round of reviews. You can continue commiting changes to your local branch and pushing to GitHub and it will update the PR.

src/convexhull.jl
```diff
@@ -0,0 +1,20 @@
+function upperhull(pset::PointSet) 
+
+  xs=Vector{Float64}[]
```
juliohm: This line is unnecessary, you don't need to preallocate

src/convexhull.jl
```diff
@@ -0,0 +1,20 @@
+function upperhull(pset::PointSet) 
+
+  xs=Vector{Float64}[] 
+  xs=sort(coordinates.(pset), by=first)
+
+  convHull=[xs[1],xs[2]] 
+  
+  for i in 3:size(xs,1)
```
juliohm: For a 1-dimensional vector you can use `length(xs)` instead of `size(xs, 1)`

src/convexhull.jl
```diff
@@ -0,0 +1,20 @@
...
+    push!(convHull,xs[i])
+
+    # remove the second to last point in convHull while the last 3 points make a "left turn"
+    while size(convHull)[1]>2 && ∠(Point(xs[i-2]),Point(xs[i-1]),Point(xs[i]))<0
```
juliohm: Can you please rewrite the comment with more details about the algorithm? Also, do you think it is possible to simplify the stop criterion? The first `size(convHul)[1] > 2` part seems redundant. Maybe it is always true?

src/convexhull.jl
```diff
@@ -0,0 +1,20 @@
...
+    push!(convHull,xs[i])
+
+    # remove the second to last point in convHull while the last 3 points make a "left turn"
+    while size(convHull)[1]>2 && ∠(Point(xs[i-2]),Point(xs[i-1]),Point(xs[i]))<0 
+      splice!(convHull,i-1)
+    end 
+
+  end
+
+  convHull
```
juliohm: I need to review the algorithm, but isn't this code only returning the upper part of the hull? Can you choose a better variable name?

test/convexhull.jl
```diff
@@ -0,0 +1,7 @@
+@testset "upperhull" begin
```
juliohm: The name of this test set is `Convex Hull`, upper hull is just a piece of the solution, right?

test/convexhull.jl
```diff
@@ -0,0 +1,7 @@
...
+  
+  pset1=PointSet((0.,0.),(1.,1.),(1.5,1.0))
```
juliohm: Please use the same format of the other tests with `P2` which will be changed to single and double floating point precision.

juliohm (changes requested):

src/convexhull.jl
Andre2957: Actually, I don't think it is redundant. The collection might end up with just 2 points after  removal of the second to last point in the array. Aside from this I don't really know how to simplify the stopping criteria. The angle criterion seems necessary.

src/convexhull.jl
```diff
@@ -0,0 +1,30 @@
+"""
+Compute the upper part of the convex hull of a given set of 2D points 'pset' using the "Graham Scan" algorithm.
+
+*) Sort the points lexicographic
+
+*) Assess the last 3 points in 'upperHull'. If the the clockwise angle is <180° remove the second to last point.
+
+"""
+
```
juliohm: Check the other docstrings in the project. They follow a specific format with the name of the function followed by a description. Indentation matters.

Also there shouldn't be white space between the docstring and the function otherwise Julia won't be able to associate the docstring to the function in help mode.

src/convexhull.jl
```diff
@@ -0,0 +1,30 @@
...
+function upperhull(pset::PointSet) 
+
+  xs=sort(coordinates.(pset), by=first)
+
```
juliohm: Please read the collaboration guidelines, specifically the code style of the project: https://github.com/JuliaGeometry/Meshes.jl#contributing

src/convexhull.jl
```diff
@@ -0,0 +1,30 @@
...
+  upperHull=[xs[1],xs[2]] 
+  
+  for i in 3:length(xs)
+    push!(upperHull,xs[i])
+    
+    # remove the second to last point in upperHull while the last 3 points make a "left turn"
+    while length(upperHull)>2 && ∠(Point(upperHull[lastindex(upperHull)-2]),Point(upperHull[lastindex(upperHull)-1]),Point(upperHull[lastindex(upperHull)]))<0
```
juliohm: This line is very difficult to read. Notice how long it is. Please reorganize the code so that others can quickly understand what is going on. Create additional veriables if necessary outside the while loop. Also the syntax `upperHull[lastindex(upperHull)-2]` can be replaced by `upperHull[end-2]`

juliohm (changes requested):

juliohm:
@Andre2957 did you have a chance to look into this?

src/convexhull.jl
Andre2957: I've shortened the stopping criterion, but had to define new variables inside the while loop, since they get updated inside the loop as well.

src/convexhull.jl
Andre2957: Sorry, I don't know what I should change here exactly

juliohm: I will give it a another round of reviews so that we can align the code style further.

src/convexhull.jl
```diff
@@ -0,0 +1,36 @@
+"""
+    upperhull(pset)
+
+Upper part of the convex hull of a given set of 2D points `pset`. 
+The upperhull is computed by the "Graham Scan" algorithm.
```
juliohm: Can you add a reference to the algorithm like in other doc strings?

src/convexhull.jl
```diff
@@ -0,0 +1,36 @@
...
+"""
+function upperhull(pset::PointSet)
+
```
juliohm:
```suggestion
```
Fix code style.

src/convexhull.jl
```diff
@@ -0,0 +1,36 @@
...
+  xs=sort(coordinates.(pset), by=first)
+  
+  upperHull=[xs[1],xs[2]]
```
juliohm: Please rewrite all assignments with whitespaces `a = 1` between left and right hand side.

src/convexhull.jl
```diff
@@ -0,0 +1,36 @@
...
+      
+      splice!(upperHull,lastindex(upperHull)-1)
+      
+      if length(upperHull)>2 
+
+        A=Point(upperHull[end-2])
+        B=Point(upperHull[end-1])
+        C=Point(upperHull[end])
+      end
+
```
juliohm: We do not use white space in the codebase like you are using here. Can you please eliminate the black lines?

juliohm (changes requested):
Thank you for incorporating the suggestions. This review process is crucial to align code style and to improve your experience with the language.

After you incorporate this set of reviews we can come back to the actual content of the PR and discuss the algorithm itself. Notice that the function that you implemented does not solve the original convex hull problem, but only part of it. If I remember correctly, the algorithm solves the upper and lower hull and then combines the results into the final result.

src/convexhull.jl
Andre2957: Is this [wikipedia page](https://en.wikipedia.org/wiki/Graham_scan) good for this?

Andre2957: In wikipedia the algorithm is from bottom to top, whereas I'm implementing it  from left to right.

Andre2957: This comment seems fitting to our discussion (also form wikipedia) : 

"The same basic idea works also if the input is sorted on x-coordinate instead of angle, and the hull is computed in two steps producing the upper and the lower parts of the hull respectively. This modification was devised by A. M. Andrew[2] and is known as Andrew's Monotone Chain Algorithm. It has the same basic properties as Graham's scan."

juliohm: Nice. Please cite the exact algorithm that is being implemented and add this excerpt from Wikipedia to the docstring so that future readers understand what is going on.

juliohm:
@Andre2957 looking forward to another round of review. Please consider the code style changes and then we can start working on  other aspects to make the code more general and easy to maintain.

juliohm:
@Andre2957 do you think you will still have time to keep working on this PR? I will submit a new round of reviews and you let me know if you can work on it.

src/convexhull.jl
```diff
@@ -0,0 +1,31 @@
+"""
+    upperhull(pset)
+
+Upper part of the convex hull of a given set of 2D points `pset`.
```
juliohm, line 4: I didn't have time to look into the modification of the original algorithm but  I had understood that it returned the complete hull, not just the upper part. Can you please clarify?

src/convexhull.jl
```diff
@@ -0,0 +1,31 @@
+"""
+    upperhull(pset)
+
+Upper part of the convex hull of a given set of 2D points `pset`. 
+The upperhull is computed according to the Andrew's Monotone Chain Algorithm.
+See https://en.wikipedia.org/wiki/Convex_hull_algorithms.
+
+The input is sorted on x-coordinate, and the hull is computed in two steps producing the upper and the lower parts of the hull respectively. 
+This is a modification of the Graham's Scan algorithm, devised by A. M. Andrew, known as Andrew's Monotone Chain Algorithm. It has the same basic properties as Graham's scan.
```
juliohm, line 9: Long lines such as this should be split into multiple lines. Try to follow the number of columns of the previous paragraph.

src/convexhull.jl
```diff
@@ -0,0 +1,31 @@
...
+Upper part of the convex hull of a given set of 2D points `pset`. 
+The upperhull is computed according to the Andrew's Monotone Chain Algorithm.
+See https://en.wikipedia.org/wiki/Convex_hull_algorithms.
+
+The input is sorted on x-coordinate, and the hull is computed in two steps producing the upper and the lower parts of the hull respectively. 
+This is a modification of the Graham's Scan algorithm, devised by A. M. Andrew, known as Andrew's Monotone Chain Algorithm. It has the same basic properties as Graham's scan. 
+See https://en.wikipedia.org/wiki/Graham_scan
+"""
+function upperhull(pset::PointSet)
+  xs = sort(coordinates.(pset), by = first)
```
juliohm:
```suggestion
  xs = sort(coordinates.(pset), by=first)
```

src/convexhull.jl
```diff
@@ -0,0 +1,31 @@
...
+  upperHull = [xs[1],xs[2]]
+  for i in 3:length(xs)
+    push!(upperHull,xs[i])
+    A = Point(upperHull[end-2])
+    B = Point(upperHull[end-1])
+    C = Point(upperHull[end])
+    # remove the second to last point in upperHull while the last 3 points make a "left turn"
+    while length(upperHull)>2 && ∠(A,B,C)<0
+      splice!(upperHull,lastindex(upperHull)-1)
```
juliohm:
```suggestion
      splice!(upperHull, length(upperHull)-1)
```

src/convexhull.jl
```diff
@@ -0,0 +1,31 @@
...
+      if length(upperHull)>2
```
juliohm:
```suggestion
      if length(upperHull) > 2 
```

test/convexhull.jl
```diff
@@ -0,0 +1,14 @@
+@testset "Convex Hull" begin
+  
+  # Upper part of the Convex Hull
+  pset1=PointSet(P2(0,0),P2(1,1),P2(1.5,1.0))
+  pset2=PointSet(P2(0,0),P2(1,1),P2(1.5,1.0),P2(1.5,3.))
+  pset3=PointSet(P2(0,0),P2(1,1),P2(1.5,1.0),P2(1.5,3.),P2(2,3))
+  pset4=PointSet(P2(0,0),P2(1,1),P2(1.5,1.0),P2(1.5,3.),P2(2,3),P2(2,4))
```
juliohm:
```suggestion
  pset1 = PointSet(P2(0,0),P2(1,1),P2(1.5,1.0))
  pset2 = PointSet(P2(0,0),P2(1,1),P2(1.5,1.0),P2(1.5,3.))
  pset3 = PointSet(P2(0,0),P2(1,1),P2(1.5,1.0),P2(1.5,3.),P2(2,3))
  pset4 = PointSet(P2(0,0),P2(1,1),P2(1.5,1.0),P2(1.5,3.),P2(2,3),P2(2,4))
```

juliohm (changes requested):

src/convexhull.jl
Andre2957: I won't be able to keep working on this. It was a nice try though!

juliohm: Thanks for starting it out @Andre2957 , good luck with your qualification exam!

PR #165, opened 2021-08-16
Update mesh.md with topology and geometry example

moyner:
Extending the existing example with boundary+coboundary and basic geometry.

### DIFF
--- a/docs/src/meshes.md
+++ b/docs/src/meshes.md
@@ -60,3 +60,47 @@ mesh = SimpleMesh(points, connec)
 ```@example meshes
 viz(mesh, showfacets = true)
 ```
+
+```@example meshes
+# Convert topology to half edge topology for fast lookup
+ht = convert(HalfEdgeTopology, topology(mesh))
+# We can create the boundary facets of the first Ngon.
+# The boundary is parametrized on the dimension we are mapping from,
+# and the lower dimension we are mapping to.
+# The Ngons are of order 2 and are made up of segments with order 1:
+b = Boundary{2,1}(ht)
+# Show boundary of Ngon number one:
+b(1)
+```
+
+```@example meshes
+## The co-boundary provides the connectivity from a lower to higher
+# dimension.
+cb = Coboundary{1, 2}(ht)
+# From this, we can look up the pair of Ngons that share a given 
+# segment in their boundary.
+cb(3)
+```
+
+```@example meshes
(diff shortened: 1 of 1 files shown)


docs/src/meshes.md
```diff
@@ -60,3 +60,47 @@ mesh = SimpleMesh(points, connec)
 ```@example meshes
 viz(mesh, showfacets = true)
 ```
+
+```@example meshes
+# Convert topology to half edge topology for fast lookup
```
juliohm:
```suggestion
# convert topology to half-edge topology for fast lookup
```

docs/src/meshes.md
```diff
@@ -60,3 +60,47 @@ mesh = SimpleMesh(points, connec)
...
+ht = convert(HalfEdgeTopology, topology(mesh))
+# We can create the boundary facets of the first Ngon.
+# The boundary is parametrized on the dimension we are mapping from,
+# and the lower dimension we are mapping to.
```
juliohm, line 69: Convert this to normal text in between example code sections.

docs/src/meshes.md
```diff
@@ -60,3 +60,47 @@ mesh = SimpleMesh(points, connec)
...
+# The Ngons are of order 2 and are made up of segments with order 1:
+b = Boundary{2,1}(ht)
```
juliohm, line 71: I would use the symbol `\partial21` for the boundary relation so that it is explicit in the variable name.

docs/src/meshes.md
```diff
@@ -60,3 +60,47 @@ mesh = SimpleMesh(points, connec)
...
+# Show boundary of Ngon number one:
+b(1)
+```
+
+```@example meshes
+## The co-boundary provides the connectivity from a lower to higher
+# dimension.
+cb = Coboundary{1, 2}(ht)
```
juliohm, line 79: Similarly, I would use `\scrC11` for the variable name to make the relation explicit in the variable name.

juliohm (changes requested):
Thank you @moyner for this great addition. Please consider this first round of reviews. Maybe it would make sense to start the example with a preamble text explaining what the example is about.

juliohm:
@moyner did you have a chance to take a look at this?

juliohm:
Fixed on master.

moyner:
Thank you for fixing this. Sorry for not following up this PR.

PR #172, opened 2021-09-19
Remove type instability for PolyArea

serenity4:
A missing type constructor in PolyArea made it unable to infer the `Chain` type parameter `V`.

### DIFF
--- a/src/polytopes/polyarea.jl
+++ b/src/polytopes/polyarea.jl
@@ -54,8 +54,8 @@ struct PolyArea{Dim,T,C<:Chain{Dim,T}} <: Polygon{Dim,T}
   end
 end
 
-PolyArea(outer::C, inners=[]; fix=true) where {Dim,T,C<:Chain{Dim,T}} =
-  PolyArea{Dim,T,Chain{Dim,T}}(outer, inners, fix)
+PolyArea(outer::C, inners=C[]; fix=true) where {Dim,T,V,C<:Chain{Dim,T,V}} =
+  PolyArea{Dim,T,Chain{Dim,T,V}}(outer, inners, fix)
 
 PolyArea(outer::AbstractVector{P}, inners=[]; fix=true) where {P<:Point} =
   PolyArea(Chain(outer), [Chain(inner) for inner in inners]; fix=fix)
--- a/test/polytopes.jl
+++ b/test/polytopes.jl
@@ -320,6 +320,10 @@
     @test P2(0.75,0.25) ∉ poly
     @test P2(0.75,0.75) ∈ poly
 
+    point = P2(0.5,0.5)
+    bytes = @allocated point ∈ poly
+    @test bytes == 0
+
     # area
     outer = P2[(0,0),(1,0),(1,1),(0,1),(0,0)]
     hole1 = P2[(0.2,0.2),(0.4,0.2),(0.4,0.4),(0.2,0.4),(0.2,0.2)]
--- a/test/polytopes.jl
+++ b/test/polytopes.jl
@@ -319,7 +319,6 @@
     @test P2(0.25,0.25) ∉ poly
     @test P2(0.75,0.25) ∉ poly
     @test P2(0.75,0.75) ∈ poly
-
     point = P2(0.5,0.5)
     bytes = @allocated point ∈ poly
     @test bytes == 0


serenity4:
I'm not sure why the tests are failing. They also fail on master (with Julia 1.6.2). I don't think it is related to these changes, maybe CircularArrays.jl or StaticArrays.jl broke something?

Edit: it's probably not the case since we have compat bounds set correctly and there does not appear to be releases related to those bounds from the latest master commit.

juliohm:
I am surprised how Julia didn't figure out the type parameter in this case. Thank you @serenity4 for checking it.

The tests are failing could you please take a look?

serenity4:
I already took a look, see my comment above. The failures are not related to these changes and happen on master as well. Something going wrong when resizing circular arrays, I have no clue why.

juliohm:
Our Project.toml specifies CircularArrays 1.1 and yet Julia is updating it to 1.3, I think it is a problem of semantic versioning? When the versions are above 1.x they are considered not breaking? Not many packages are 1.x in Julia.

juliohm:
I opened an issue in CircularArrays.jl, I think some breaking change wasn't communicated properly to downstream packages.

serenity4:
Indeed, I thought that compat bounds would keep version 1.1 to 1.2 (excluded), but it allows for more with semantic versioning. We can temporarily fix that by specifying a version of `~ 1.1` to include 1.1 releases until 1.2 (or track the version that works and allow from 1.1 to there, as I don't know where it broke exactly), and remove this spec when the bug is addressed.

juliohm:
That would be a good idea, I will try to fix the Project.toml locally now...

serenity4:
The project looks small enough that I can take a look later in the day and see if I can find a fix. I am not sure it's a breaking change, it might be just a simple bug.

juliohm:
That would be lovely. In the meantime I am following your suggestion to fix the Project.toml to `~1.1` with a patch release in Meshes.jl

serenity4:
Does the version fix work for you locally? If I'm not wrong if should not, pushing to a `CircularArray` (the source of the errors) has been disallowed for two years (or at least is known not to work) and is even tested on CircularArrays.jl. I don't know what introduced the bug, if you would have a working Meshes package locally it would be great if you could dump your project status to see if we're not missing something.

juliohm:
It worked for me locally. The tests pass when I fix the version of CircularArrays.jl in the Project.toml to ~1.1:

```julia
(Meshes) pkg> st -m
     Project Meshes v0.17.8
      Status `~/.julia/dev/Meshes/Manifest.toml`
  [79e6a3ab] Adapt v3.3.1
  [324d7699] CategoricalArrays v0.10.1
  [d360d2e6] ChainRulesCore v1.4.0
  [7a955b69] CircularArrays v1.1.2
  [34da2185] Compat v3.37.0
  [a8cc5b0e] Crayons v4.0.4
  [9a962f9c] DataAPI v1.8.0
  [864edb3b] DataStructures v0.18.10
  [e2d170a0] DataValueInterfaces v1.0.0
  [b4f34e82] Distances v0.10.4
  [ffbed154] DocStringExtensions v0.8.5
  [92d709cd] IrrationalConstants v0.1.0
  [c8e1da08] IterTools v1.3.0
  [82899510] IteratorInterfaceExtensions v1.0.0
  [692b3bcd] JLLWrappers v1.3.0
  [2ab3a3ac] LogExpFunctions v0.3.3
  [1914dd2f] MacroTools v0.5.8
  [e1d29d7a] Missings v1.0.2
  [b8a86587] NearestNeighbors v0.4.9
  [6fe1bfb0] OffsetArrays v1.10.7
  [bac558e1] OrderedCollections v1.4.1
  [21216c6a] Preferences v1.2.2
  [3cdcf5f2] RecipesBase v1.1.2
  [74f56ac7] ReferenceFrameRotations v1.0.1
  [ae029012] Requires v1.1.3
  [699a6c99] SimpleTraits v0.9.4
  [a2af1166] SortingAlgorithms v1.0.1
  [276daf66] SpecialFunctions v1.6.1
  [90137ffa] StaticArrays v1.2.12
  [82ae8749] StatsAPI v1.0.0
  [2913bbd2] StatsBase v0.33.10
  [3783bdb8] TableTraits v1.0.1
  [bd369af6] Tables v1.5.2
  [efe28fd5] OpenSpecFun_jll v0.5.5+0
  [0dad84c5] ArgTools
  [56f22d72] Artifacts
  [2a0f44e3] Base64
  [ade2ca70] Dates
  [8bb1440f] DelimitedFiles
  [8ba89e20] Distributed
  [f43a241f] Downloads
  [9fa8497b] Future
  [b77e0a4c] InteractiveUtils
  [b27032c2] LibCURL
  [76f85450] LibGit2
  [8f399da3] Libdl
  [37e2e46d] LinearAlgebra
  [56ddb016] Logging
  [d6f4376e] Markdown
  [a63ad114] Mmap
  [ca575930] NetworkOptions
  [44cfe95a] Pkg
  [de0858da] Printf
  [3fa0cd96] REPL
  [9a3f8284] Random
  [ea8e919c] SHA
  [9e88b42a] Serialization
  [1a1011a3] SharedArrays
  [6462fe0b] Sockets
  [2f01184e] SparseArrays
  [10745b16] Statistics
  [fa267f1f] TOML
  [a4e569a6] Tar
  [8dfed614] Test
  [cf7118a7] UUIDs
  [4ec0a83e] Unicode
  [e66e0078] CompilerSupportLibraries_jll
  [deac9b47] LibCURL_jll
  [29816b5a] LibSSH2_jll
  [c8ffd9c3] MbedTLS_jll
  [14a3606d] MozillaCACerts_jll
  [83775a58] Zlib_jll
  [8e850ede] nghttp2_jll
  [3f19e933] p7zip_jll
```

serenity4:
Found the exact issue. The problem was the result type of a broadcast operation over a `CircularArray`. In version 1.1 it yields a `Vector`, and now it yields the `CircularArray` type. This is actually an improvement, though I see how it can be breaking. Though this difference in behavior should have been accounted for and a breaking release should have been issued (or at least, fixes to downstream packages that might suffer from it), I think the error on our side was relying on having a vector after this broadcast operation which yielded an unnatural result.

TL;DR I don't think there is anything to fix in CircularArrays.jl, though it would have been great if they could have foreseen the problem and/or done something about it. I'll make a PR to suggest a fix and upgrade to CircularArrays@v1.3 and then we can take care of this one.

juliohm:
We should probably rebase this one to the latest master and then add tests to make sure we don't fall in this allocation trap in the future. It is tricky to write tests for that.

serenity4:
I think we can test whether `point in polyarea` allocates in `test/intersection.jl`. That's what led to this issue in the first place.

juliohm:
Maybe we just add a line here with the allocation test? https://github.com/JuliaGeometry/Meshes.jl/blob/d8b8bcf2d4dc52052a382eed8189d11c9c46d091/test/polytopes.jl#L309

PR #174, opened 2021-09-20
Fix CircularVector broadcasting with manual collect

serenity4:
Takes care of fixing the breakage introduced in CircularArrays.jl.

### DIFF
--- a/src/polytopes.jl
+++ b/src/polytopes.jl
@@ -193,7 +193,7 @@ function bridge(p::Polygon{Dim,T}; width=zero(T)) where {Dim,T}
   end
 
   # retrieve chains as vectors of coordinates
-  pchains = [coordinates.(vertices(c)) for c in chains(p)]
+  pchains = [coordinates.(collect(vertices(c))) for c in chains(p)]
 
   # sort vertices lexicographically
   coords  = [coord for pchain in pchains for coord in pchain]
--- a/src/polytopes.jl
+++ b/src/polytopes.jl
@@ -193,7 +193,7 @@ function bridge(p::Polygon{Dim,T}; width=zero(T)) where {Dim,T}
   end
 
   # retrieve chains as vectors of coordinates
-  pchains = [coordinates.(vertices(c).data) for c in chains(p)]
+  pchains = [coordinates.(parent(vertices(c))) for c in chains(p)]
 
   # sort vertices lexicographically
   coords  = [coord for pchain in pchains for coord in pchain]
--- a/src/polytopes.jl
+++ b/src/polytopes.jl
@@ -193,7 +193,7 @@ function bridge(p::Polygon{Dim,T}; width=zero(T)) where {Dim,T}
   end
 
   # retrieve chains as vectors of coordinates
-  pchains = [coordinates.(collect(vertices(c))) for c in chains(p)]
+  pchains = [coordinates.(vertices(c).data) for c in chains(p)]
 
   # sort vertices lexicographically
   coords  = [coord for pchain in pchains for coord in pchain]
--- a/src/polytopes.jl
+++ b/src/polytopes.jl
@@ -193,7 +193,7 @@ function bridge(p::Polygon{Dim,T}; width=zero(T)) where {Dim,T}
   end
 
   # retrieve chains as vectors of coordinates
-  pchains = [coordinates.(parent(vertices(c))) for c in chains(p)]
+  pchains = [coordinates.(vertices(open!(c))) for c in chains(p)]
 
   # sort vertices lexicographically
   coords  = [coord for pchain in pchains for coord in pchain]
--- a/src/polytopes.jl
+++ b/src/polytopes.jl
@@ -193,7 +193,7 @@ function bridge(p::Polygon{Dim,T}; width=zero(T)) where {Dim,T}
   end
 
   # retrieve chains as vectors of coordinates
-  pchains = [coordinates.(vertices(open!(c))) for c in chains(p)]
+  pchains = [coordinates.(vertices(open(c))) for c in chains(p)]
 
   # sort vertices lexicographically
   coords  = [coord for pchain in pchains for coord in pchain]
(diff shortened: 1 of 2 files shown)


juliohm:
So a vector implementing the `AbstractVector` interface is not guaranteed to work well with broadcast? I find it strange that we need to handle this that way.

serenity4:
I think it's more about: what should the resulting vector be? Would we expect to always return a `Vector` after a broadcast over an `AbstractVector`? No, and it's what changed. We now get back the original vector type which wraps the new contents, which seems more natural to me.

juliohm:
> I think it's more about: what should the resulting vector be? Would we expect to always return a `Vector` after a broadcast over an `AbstractVector`? No, and it's what changed. We now get back the original vector type which wraps the new contents, which seems more natural to me.

I fully agree we should get the CircularArray type as returned type. You mean that we are assuming `Vector` specifically in some function in our API? Maybe this function should be fixed instead?

src/polytopes.jl
```diff
@@ -193,7 +193,7 @@ function bridge(p::Polygon{Dim,T}; width=zero(T)) where {Dim,T}
   end
 
   # retrieve chains as vectors of coordinates
-  pchains = [coordinates.(vertices(c)) for c in chains(p)]
+  pchains = [coordinates.(vertices(c).data) for c in chains(p)]
```
juliohm: We can't access internal fields like this.

serenity4: I know it's risky. The alternative is copying everything into a vector with `collect` (which I did before this commit), which is slow but more stable. Unfortunately I don't know of any way to access the wrapped array present in a `CircularArray`. I can revert to `collect`ing the vertices instead if you prefer.

serenity4:
We were assuming getting a `Vector` at the location of the change. If we get a `CircularArray`, then things don't work because we can't push to it. And even if we allow pushing to it by upstreaming a change to CircularArrays, the result is not correct (a vertex is missing in the result of the function, I haven't debugged it further than that).

src/polytopes.jl
juliohm: I think we may be fixing the code in the wrong place? Shouldn't `coordinates.(::CircularArray{<:Point})` work? I didn't have time to read the code carefully because the week is crazy busy around here. I can try to take a look with more time on the weekend.

serenity4: Another alternative is to just use version 1.1 of CircularArrays and not worry about it. But I fear it's not stable on the long-term either. Or we can change to not using CircularArrays and handle wrapping indices ourselves.

juliohm:
So `AbstractArray` doesn't require `push` behavior?

src/polytopes.jl
serenity4: `coordinates.(::CircularArray{<:Point})` returns a `CircularArray` which doesn't work with the rest of the code.

serenity4:
Nope, `StaticArrays.SArray` for example does not have `push!`. It has  `push`, but an `AbstractArray` does not require mutability.

src/polytopes.jl
juliohm: Got it, because the length of the CircularArrays is off by 1? With the first point repeated? I will try to take a look, the `collect` may be the most reasonable solution in this case indeed.

serenity4: I am not sure exactly why. Maybe something like this. If you want to test the code with the `push!` functionality added to CircularArrays feel free to use [my fork](https://github.com/serenity4/CircularArrays.jl).

serenity4:
Using `parent` seems to work fine, I had missed this method which accesses the internal vector in a stable way. We can either make `vertices` on `Chain` return the internal vector (but then we don't have the nice indexing functionality - probably not what we want), or stick with this fix for just this specific case. We could also (if we need it) add a method that gives back a list of coordinates from a `Chain` directly (basically what the generator function inside the affected line does, refactored into its own function for clarity), but since it touches only one case at the moment I see little point in complexifying the thing.

src/polytopes.jl
```diff
@@ -193,7 +193,7 @@ function bridge(p::Polygon{Dim,T}; width=zero(T)) where {Dim,T}
   end
 
   # retrieve chains as vectors of coordinates
-  pchains = [coordinates.(vertices(c)) for c in chains(p)]
+  pchains = [coordinates.(parent(vertices(c))) for c in chains(p)]
```
```suggestion
  pchains = [coordinates.(vertices(open!(c))) for c in chains(p)]
```
What about this? I think the intention becomes more clear? I find `parent` a bit hacky because it has nothing to do with the algorithm itself. What I really wanted in this line was the list of coordinates of the corresponding open chain.

I am assuming that calling `open!(c)` won't modify the original set of `chains(p)`. If it does, then we probably want to add `open(c) = copy(c) |> open!` first in a separate PR and then rebase this work on top of it. What do you think?

serenity4: Looks better indeed! It looks like we are allocating a new array in `chains(p)`, so this modification should not have side effects on the polygon.

juliohm: I've added `open/close` to the master branch in any case. I will accept the suggestion here and see if tests pass 👌🏽

juliohm:
Tests are passing with `open!` 🎉 Do you think it is too dangerous to assume that `chains(p)` is returning copies and that modifying these copies won't cause side effects? Alternatively, we could use `open(c)` without mutation to be super safe. What do you think?

serenity4:
If we ever decide to make `chains` return views or uncopied arrays, that depends if it is the only place where we mutate the result. If so (or if we usually don't assume that we can safely mutate `chains` in existing uses), then it could introduce bugs that are very hard to get to, and I wouldn't mind being on the safer side with `open(c)`. Also it looks to me like using `open!` instead of `open` for performance reasons is the wrong place to get performance from, where it is more the role of `chains` not to allocate IMO. So I think I'd be in favor of `open(c)`. Any thoughts on your side?

juliohm:
Fully agree with you. I think we will have much more headache trying to fix bugs downstream if something internal like that changes. If you can help with rebasing this PR on top of the latest master, that would be great. We can then use `open` instead of `open!` and merge. 💯

juliohm:
Thanks for the quick fix! 🚀  now we can head over to #172 :)

PR #178, opened 2021-10-24
Add explicit annotation with Symmetric to Ellipsoid function

fnaghetini:
Add a `Symmetric` function around `R'*Λ*R` to fix [this issue](https://github.com/JuliaEarth/GeoStats.jl/issues/197).


juliohm:
Amazing @fnaghetini could you also add a test in this same PR with that problematic example? Basically you can add the example [here](https://github.com/JuliaGeometry/Meshes.jl/blob/6dc46d3c952f0733eded54a286fd447b12d46de6/test/neighborhoods.jl#L96)

juliohm:
I think GitHub allows you to update the PR in place using the online edit features. Please let me know if you have issues with this process. Happy to help.

PR #180, opened 2021-10-27
provides `Base.length(::Segment)` with a test

JeffreySarnoff:


JeffreySarnoff:
The length test had been failing because the `length` obtained in the test was Float32 precision... however, the test run in the REPL gave Float64 precision. I eased the test to work with either/mixed .. however I do not know what is behind this behavior.

src/polytopes/segment.jl
```diff
@@ -24,6 +24,7 @@ Base.minimum(s::Segment) = s.vertices[1]
 Base.maximum(s::Segment) = s.vertices[2]
 Base.extrema(s::Segment) = s.vertices[1], s.vertices[2]
 measure(s::Segment) = norm(s.vertices[2] - s.vertices[1])
+Base.length(s::Segment)  = measure(s)
```
```suggestion
Base.length(s::Segment) = measure(s)
```

test/polytopes.jl
```diff
@@ -7,6 +7,7 @@
     @test minimum(s) == P2(0,0)
     @test maximum(s) == P2(1,1)
     @test extrema(s) == (P2(0,0), P2(1,1))
+    @test isapprox(length(s), sqrt(2))
```
```suggestion
    @test isapprox(length(s), sqrt(T(2)) 
```

juliohm:
I've added a `T` to the test, let's see how it goes...

test/polytopes.jl
```diff
@@ -7,6 +7,7 @@
     @test minimum(s) == P2(0,0)
     @test maximum(s) == P2(1,1)
     @test extrema(s) == (P2(0,0), P2(1,1))
+    @test isapprox(T(length(s)), sqrt(T(2)))
```
juliohm: I think we only need T in the right hand side right? Was the test failing? We need to investigate why.

JeffreySarnoff: The test was not failing.  I will remove both the Ts, as the test seems to have resolved itself (it works in the REPL).

test/polytopes.jl
```suggestion
    @test isapprox(length(s), sqrt(T(2))) 
```

test/polytopes.jl
```diff
@@ -7,6 +7,7 @@
     @test minimum(s) == P2(0,0)
     @test maximum(s) == P2(1,1)
     @test extrema(s) == (P2(0,0), P2(1,1))
+    @test isapprox(length(s), sqrt(T(2)))
```
juliohm, line 10: We still need the one in the right-hand-side for consistency so that operations are performed in single precision.

JeffreySarnoff: if you really want these tests to be done in single precision, change
`    s = Segment(P2(0,0), P2(1,1))` to `  s = Segment(P2(0.0f0,0.0f0), P2(1.0f0,1.0f0))`
that way `length(s)::Float32` otherwise, `length(s)::Float64` .

juliohm: The P2 in the tests is both Float32 and Float64. We run the test suite for single and double precision. Check the runtests.jl for more details. We don't need to use single precision literals here.

juliohm:
Nice! Tests are passing! Merging this one as well. Thanks 🎉

Issue #187, opened 2021-11-03
SimpleMesh failed when types of inputs are AbstractVector{<:Point} and ::TP


deltaeecs:
When I construct a volume cell with view of points and then use function **_boundary_** to generate its boundaries
```
nodes = rand(Point3, 4)
tetra = Tetrahedron(view(nodes, 4:-1:1))
facets = boundary(tetra )
```
I get error message
```
ERROR: MethodError: no method matching SimpleMesh(::SubArray{Point3, 1, Vector{Point3}, Tuple{StepRange{Int64, Int64}}, true}, ::FullTopology{Connectivity{Triangle{Dim, T} where {Dim, T}, 3}})
Closest candidates are:
  SimpleMesh(::AbstractVector{<:Point}, ::AbstractVector{<:Connectivity}) at C:\Users\12253\.julia\packages\Meshes\GrGir\src\mesh\simplemesh.jl:25
  SimpleMesh(::Array{Point{Dim, T}, 1}, ::TP) where {Dim, T, TP<:Topology} at C:\Users\12253\.julia\packages\Meshes\GrGir\src\mesh\simplemesh.jl:20
Stacktrace:
 [1] SimpleMesh(points::SubArray{Point3, 1, Vector{Point3}, Tuple{StepRange{Int64, Int64}}, true}, connec::Vector{Connectivity{Triangle{Dim, T} where {Dim, T}, 3}})
   @ Meshes C:\Users\12253\.julia\packages\Meshes\GrGir\src\mesh\simplemesh.jl:25
 [2] boundary(t::Tetrahedron{3, Float64, SubArray{Point3, 1, Vector{Point3}, Tuple{StepRange{Int64, Int64}}, true}})
   @ Meshes C:\Users\12253\.julia\packages\Meshes\GrGir\src\polytopes\tetrahedron.jl:27
 [3] top-level scope
   @ REPL[21]:1
```

Clearly **SimpleMesh** does not accept view, I can fix it by modifying **SimpleMesh** to 
```
struct SimpleMesh{Dim,T,TP<:Topology} <: Mesh{Dim,T}
  points::AbstractVector{Point{Dim,T}}
  topology::TP
end
```
But will this lower the performance? Is there a better way?

juliohm:
I am surprised how you are finding these corner cases 💯 I don't know why the default constructor of SimpleMesh is not taking care of the conversion from view to actual vector. Asking our Julia expert @serenity4 for guidance =)

[...]

PR #188, opened 2021-11-04
Allow points of AbstractVector as inputs of SimpleMesh

deltaeecs:
Fix https://github.com/JuliaGeometry/Meshes.jl/issues/187 by employing  @https://github.com/juliohm ’s commits
https://github.com/JuliaGeometry/Meshes.jl/issues/187#issuecomment-959588059


deltaeecs:
I just modify the **measure** function for **Hexahedron** since the origin one outputs wrong results when the input hexahedron is not a unit cube
The correct one should be like this
image

deltaeecs:
Some conflicts happens between several merges, I'll check and resubmit this.

PR #194, opened 2021-11-04
Segment plane intersect

jackbeagley:
Implemented `intersecttype(s::Segment{3,T}, p::Plane{3,T}) where {T}`. Method taken from [https://en.wikipedia.org/wiki/Line-plane_intersection](url). Four tests written too.

### DIFF
--- a/src/Meshes.jl
+++ b/src/Meshes.jl
@@ -281,7 +281,7 @@ export
   SpatialPredicatePartition,
   ProductPartition,
   HierarchicalPartition,
-  partition, split, →,
+  partition, split,
 
   # intersections
   Intersection,
--- a/src/Meshes.jl
+++ b/src/Meshes.jl
@@ -297,6 +297,7 @@ export
   CornerTouchingBoxes,
   IntersectingSegmentTriangle,
   CrossingSegmentPlane,
+  TouchingSegmentPlane,
   OverlappingSegmentPlane,
   intersecttype,
   hasintersect,
--- a/src/Meshes.jl
+++ b/src/Meshes.jl
@@ -296,6 +296,8 @@ export
   FaceTouchingBoxes,
   CornerTouchingBoxes,
   IntersectingSegmentTriangle,
+  IntersectingSegmentPlane,
+  ContainedSegmentPlane,
   intersecttype,
   hasintersect,
 
--- a/src/Meshes.jl
+++ b/src/Meshes.jl
@@ -144,7 +144,7 @@ export
   Box, Ball, Sphere, Cylinder,
   ncontrols, degree, Horner, DeCasteljau,
   radius, height, sides,
-  measure, diagonal,
+  measure, diagonal, origin,
 
   # polytopes
   Polytope, Polygon, Polyhedron,
--- a/src/Meshes.jl
+++ b/src/Meshes.jl
@@ -295,6 +295,7 @@ export
   OverlappingBoxes,
   FaceTouchingBoxes,
   CornerTouchingBoxes,
+  IntersectingSegmentTriangle,
   intersecttype,
   hasintersect,
 
--- a/src/Meshes.jl
+++ b/src/Meshes.jl
@@ -296,8 +296,8 @@ export
   FaceTouchingBoxes,
   CornerTouchingBoxes,
   IntersectingSegmentTriangle,
-  IntersectingSegmentPlane,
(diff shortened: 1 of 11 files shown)


src/Meshes.jl
```diff
@@ -296,6 +296,8 @@ export
   FaceTouchingBoxes,
   CornerTouchingBoxes,
   IntersectingSegmentTriangle,
+  IntersectingSegmentPlane,
+  ContainedSegmentPlane,
```
juliohm: Maybe we should try to follow the naming convention of other intersections? Would it make sense to call these `CrossingSegmentPlane` and `OverlappingSegmentPlane` instead?

src/intersections/planes.jl
```diff
@@ -0,0 +1,45 @@
...
+    
+    # Calculate components
+    ln = (sᵥ[2] - sᵥ[1]) ⋅ n
+    pₒn = pₒ ⋅ n
+  
+    # If ln is zero, the segment is parallel to the plane
+    if isapprox(ln, zero(T))
+        # If the numerator is zero, the segment is coincident
+        if isapprox((pₒ - sᵥ[1]) ⋅ n, zero(T), atol=atol(T))
+            return ContainedSegmentPlane(s)
```
juliohm: We need to be very precise with these terms, coincident, contained, overlapping, etc to avoid proliferation of names in the intersection algorithms. If you can double check the exact concept that is being used here, then we can try to follow the naming convention everywhere in the function.

src/intersections/planes.jl
```diff
@@ -0,0 +1,45 @@
...
+        else
+            return NoIntersection()
+        end
+    else
+        # Calculate the segment parameter
```
juliohm: Can you please follow the convention for comments in the code and use all lowercase? I think I missed that in the other PR with segment triangle intersection.

src/intersections/planes.jl
```diff
@@ -0,0 +1,45 @@
...
+        λ = ((pₒ - sᵥ[1]) ⋅ n) / ln
+
+        # If λ is approximately 0 or 1, set as so to prevent any domain errors
+        λ = isapprox(λ, zero(T), atol=atol(T)) ? zero(T) : (isapprox(λ, one(T), atol=atol(T)) ? one(T) : λ)
```
juliohm: Please split this line into two separate lines, one for 0 and one for 1.

juliohm (changes requested):
Thank you @jackbeagley for another great PR 🚀 Attached is a first round of reviews.

jackbeagley:
@juliohm with the change in terminology in mind, perhaps it makes sense to distinguish between three types of intersection? `CrossingSegmentPlane`, `OverlappingSegmentPlane` and `TouchingSegmentPlane` (cases where `λ ≈ 0` or `λ ≈ 1`). I believe these names are consistent with the existing types. I will implement these and you can take a look.

jackbeagley:
@juliohm Implemented those changes and added the new type `TouchingSegmentPlane`, happy to remove it if you think thats best, but in making the naming more consistent with other intersections it occurred to me that it might make sense to do here too.

juliohm:
Amazing @jackbeagley I will try to review it tomorrow morning. Thanks for the updates 💯

src/intersections.jl
```diff
@@ -81,6 +81,22 @@ struct IntersectingSegmentTriangle{P<:Point} <: Intersection
   value::P
 end
 
+# ---------------------------
+# SEGMENT-PLANE INTERSECTIONS
+# ---------------------------
```
```suggestion
# ----------------------------
# SEGMENT-PLANE INTERSECTIONS
# ----------------------------
```

src/intersections/planes.jl
```diff
@@ -0,0 +1,49 @@
...
+
+    # get the normal of the plane
+    n = normal(p)
+    
+    # calculate components
+    ln = (sᵥ[2] - sᵥ[1]) ⋅ n
+    pₒn = pₒ ⋅ n
+  
+    # if ln is zero, the segment is parallel to the plane
+    if isapprox(ln, zero(T))
```
juliohm:
```suggestion
    if isapprox(ln, zero(T), atol=atol(T))
```

juliohm (approved):

juliohm:
@jackbeagley approved the changes, amazing PRs 🥇

PR #196, opened 2021-11-07
Make Triangle callable

### LINKED ISSUE #84: Make parametric objects callable

HaoxuanGuo:
Fix #84 . Made `Triangle` callable.


src/polytopes/ngon.jl
```diff
@@ -160,11 +160,22 @@ function normal(t::Triangle{3})
   n / norm(n)
 end
 
+function (t::Triangle)(u::T, v::T) where T
+    w = (one(T) - u - v)
+    if u < zero(T) || u > one(T) ||
+       v < zero(T) || v > one(T) ||
+       w < zero(T) || w > one(T)
+       throw(DomainError("barycentric coordinates out of range"))
+    end
```
juliohm, line 169: Can you please report how Julia optimizes  these checks? Do we loose any performance with this assertions?

src/polytopes/ngon.jl
```diff
@@ -160,11 +160,22 @@ function normal(t::Triangle{3})
...
+    v₁, v₂, v₃ = coordinates.(t.vertices)
+    Point(v₁*u + v₂*v + v₃*w)
```
juliohm: Now that I am seeing the behavior in the tests, I think it is more interesting to use the last two vertices as reference so that the coordinate `(u,v) = (0,0)` is in the "bottom left", the coordinate `(u,v) = (1,0)` is in the "bottom right" and `(u,v) = (0,1)` is in the "top". Can you please adjust this? It will be more intuitive compared to the quadrangle case.

src/sampling/homogeneous.jl
```diff
@@ -34,18 +34,16 @@ end
...
 function sample(rng::AbstractRNG, triangle::Triangle{Dim,T},
                 method::HomogeneousSampling) where {Dim,T}
-  A, B, C = coordinates.(vertices(triangle))
   function randpoint()
     # sample barycentric coordinates
     u₁, u₂ = rand(rng, T, 2)
     λ₁, λ₂ = 1 - √u₁, u₂ * √u₁
-    λ₃     = 1 - λ₁ - λ₂
-    Point(λ₁ .* A + λ₂ .* B + λ₃ .* C)
+    triangle(λ₁, λ₂)
```
juliohm, line 41: Awesome :)

test/polytopes.jl
```diff
@@ -43,6 +43,12 @@
     t = Triangle(P2(0.4,0.4), P2(0.6,0.4), P2(0.8,0.4))
     @test P2(0.2,0.4) ∉ t
     t = Triangle(P2(0,0), P2(1,0), P2(0,1))
+    @test t(T(0.0), T(0.0)) == P2(0,1)
+    @test t(T(0.0), T(1.0)) == P2(1,0)
+    @test t(T(1.0), T(0.0)) == P2(0,0)
+    @test t(T(0.5), T(0.5)) == P2(0.5,0.0)
```
juliohm: After adjusting the reference vertices, these tests will feel more intuitive with (u,v) --> (u,v)

juliohm (changes requested):
Awesome PR 💯

src/polytopes/ngon.jl
hx-guo: I think julia will calculate `zero(T)` and `one(T)` while it compile the code.

Some code part from `@code_warntype`:

```
Body::Point3
1 ──       Core.NewvarNode(:(@_4))
│          Core.NewvarNode(:(v₃))
│          Core.NewvarNode(:(v₂))
│          Core.NewvarNode(:(v₁))
│    %5  = Meshes.one($(Expr(:static_parameter, 1)))::Core.Const(1.0)
│    %6  = (%5 - u)::Float64
│          (w = %6 - v)
│    %8  = Meshes.zero($(Expr(:static_parameter, 1)))::Core.Const(0.0)
│    %9  = (u < %8)::Bool
└───       goto #3 if not %9
2 ──       goto #12
3 ── %12 = Meshes.one($(Expr(:static_parameter, 1)))::Core.Const(1.0)
│    %13 = (u > %12)::Bool
└───       goto #5 if not %13
4 ──       goto #12
5 ── %16 = Meshes.zero($(Expr(:static_parameter, 1)))::Core.Const(0.0)
│    %17 = (v < %16)::Bool
└───       goto #7 if not %17
6 ──       goto #12
7 ── %20 = Meshes.one($(Expr(:static_parameter, 1)))::Core.Const(1.0)
│    %21 = (v > %20)::Bool
└───       goto #9 if not %21
8 ──       goto #12
9 ── %24 = w::Float64
│    %25 = Meshes.zero($(Expr(:static_parameter, 1)))::Core.Const(0.0)
│    %26 = (%24 < %25)::Bool
└───       goto #11 if not %26
10 ─       goto #12
11 ─ %29 = w::Float64
│    %30 = Meshes.one($(Expr(:static_parameter, 1)))::Core.Const(1.0)
│    %31 = (%29 > %30)::Bool
└───       goto #13 if not %31
12 ┄ %33 = Meshes.DomainError("barycentric coordinates out of range")::Any
└───       Meshes.throw(%33)
```

We can notice that julia know their type is `Core.Const(1.0)` or `Core.Const(0.0)` which proofs that julia calculated it while it compiled.

The `@code_llvm` shows

```
; ┌ @ float.jl:371 within `<'
   %7 = fcmp uge double %2, 0.000000e+00
   %8 = fcmp ule double %2, 1.000000e+00
; └
  %or.cond = and i1 %7, %8
  %9 = fcmp uge double %3, 0.000000e+00
  %or.cond12 = and i1 %or.cond, %9
  %10 = fcmp ule double %3, 1.000000e+00
  %or.cond13 = and i1 %10, %or.cond12
  %11 = fcmp uge double %6, 0.000000e+00
  %or.cond14 = and i1 %11, %or.cond13
  %12 = fcmp ule double %6, 1.000000e+00
  %or.cond15 = and i1 %12, %or.cond14
```

It know that is a double 0.000000e+00 or 1.000000e+00.

You may ask why there is an `Any`

```
12 ┄ %33 = Meshes.DomainError("barycentric coordinates out of range")::Any
└───       Meshes.throw(%33)
```

Because, it is really should be checked.

Also, use `@inbound` may improve the performance. But I think it should not hint by us. (Maybe julia compiler will recognize it in the near future.)

src/polytopes/ngon.jl
hx-guo: You are right. I will change it.

GuoHaoxuan:
@juliohm Thank your for your advice. I changed the reference.

src/polytopes/ngon.jl
juliohm: Thanks for explaining it in detail ❤️

test/polytopes.jl
```diff
@@ -43,6 +43,12 @@
     t = Triangle(P2(0.4,0.4), P2(0.6,0.4), P2(0.8,0.4))
     @test P2(0.2,0.4) ∉ t
     t = Triangle(P2(0,0), P2(1,0), P2(0,1))
+    @test t(T(0.0), T(0.0)) == P2(0,0)
+    @test t(T(0.0), T(1.0)) == P2(0,1)
+    @test t(T(1.0), T(0.0)) == P2(1,0)
```
```suggestion
    @test t(T(1.0), T(0.0)) == P2(1,0)
    @test t(T(0.0), T(1.0)) == P2(0,1)
```

test/polytopes.jl
```diff
@@ -72,6 +78,12 @@
       @test p ∉ t
     end
     t = Triangle(P3(0,0,0), P3(0,1,0), P3(0,0,1))
+    @test t(T(0.0), T(0.0)) == P3(0,0,0)
+    @test t(T(0.0), T(1.0)) == P3(0,0,1)
+    @test t(T(1.0), T(0.0)) == P3(0,1,0)
```
```suggestion
    @test t(T(1.0), T(0.0)) == P3(0,1,0)
    @test t(T(0.0), T(1.0)) == P3(0,0,1)
```

juliohm (approved):

PR #199, opened 2021-11-08
force Float64 Point

### LINKED ISSUE #161: Constructors with Integer literals should convert to floating point coordinates

HaoxuanGuo:
Fix #161 .

### DIFF
--- a/src/points.jl
+++ b/src/points.jl
@@ -29,6 +29,7 @@ Point{Dim,T}(coords...) where {Dim,T} = Point{Dim,T}(SVector{Dim,T}(coords...))
 Point(coords::SVector{Dim,T}) where {Dim,T} = Point{Dim,T}(coords)
 Point(coords::AbstractVector{T}) where {T} = Point{length(coords),T}(coords)
 Point(coords...) = Point(SVector(coords...))
+Point(coords::SVector{Dim,T}) where {Dim,T<:Integer} = Point(SVector{Dim, Float64}(coords))
 
 # coordinate type conversions
 Base.convert(::Type{Point{Dim,T}}, coords) where {Dim,T} = Point{Dim,T}(coords)
--- a/src/points.jl
+++ b/src/points.jl
@@ -22,7 +22,7 @@ O = Point(0.0, 0.0) # origin of 2D Euclidean space
 struct Point{Dim,T}
   coords::SVector{Dim,T}
   Point{Dim,T}(coords::SVector) where {Dim,T} = new{Dim,T}(coords)
-  Point{Dim,T}(coords::SVector{Dim,T2}) where {Dim,T,T2<:Integer} = new{Dim,Float64}(coords)
+  Point{Dim,T}(coords::SVector{Dim,V}) where {Dim,T,V<:Integer} = new{Dim,Float64}(coords)
 end
 
 # convenience constructors
--- a/src/points.jl
+++ b/src/points.jl
@@ -22,6 +22,7 @@ O = Point(0.0, 0.0) # origin of 2D Euclidean space
 struct Point{Dim,T}
   coords::SVector{Dim,T}
   Point{Dim,T}(coords::SVector) where {Dim,T} = new{Dim,T}(coords)
+  Point{Dim,T}(coords::SVector{Dim,T2}) where {Dim,T,T2<:Integer} = new{Dim,Float64}(coords)
 end
 
 # convenience constructors
@@ -29,7 +30,6 @@ Point{Dim,T}(coords...) where {Dim,T} = Point{Dim,T}(SVector{Dim,T}(coords...))
 Point(coords::SVector{Dim,T}) where {Dim,T} = Point{Dim,T}(coords)
 Point(coords::AbstractVector{T}) where {T} = Point{length(coords),T}(coords)
 Point(coords...) = Point(SVector(coords...))
-Point(coords::SVector{Dim,T}) where {Dim,T<:Integer} = Point(SVector{Dim, Float64}(coords))
 
 # coordinate type conversions
 Base.convert(::Type{Point{Dim,T}}, coords) where {Dim,T} = Point{Dim,T}(coords)
--- a/test/points.jl
+++ b/test/points.jl
@@ -2,7 +2,7 @@
   @test embeddim(Point(1)) == 1
   @test embeddim(Point(1, 2)) == 2
   @test embeddim(Point(1, 2, 3)) == 3
-  @test coordtype(Point(1, 1)) == Int
+  @test coordtype(Point(1, 1)) == Float64
   @test coordtype(Point(1.,1.)) == Float64
   @test coordtype(Point(1f0, 1f0)) == Float32
   @test coordtype(Point1(1)) == Float64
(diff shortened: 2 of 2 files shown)


src/points.jl
```diff
@@ -29,6 +29,7 @@ Point{Dim,T}(coords...) where {Dim,T} = Point{Dim,T}(SVector{Dim,T}(coords...))
 Point(coords::SVector{Dim,T}) where {Dim,T} = Point{Dim,T}(coords)
 Point(coords::AbstractVector{T}) where {T} = Point{length(coords),T}(coords)
 Point(coords...) = Point(SVector(coords...))
+Point(coords::SVector{Dim,T}) where {Dim,T<:Integer} = Point(SVector{Dim, Float64}(coords))
```
juliohm: Maybe this conversion should happen inside our inner constructor whenever `T <: Integer`?

GuoHaoxuan:
@juliohm You are right. What about now?

src/points.jl
```diff
@@ -22,6 +22,7 @@ O = Point(0.0, 0.0) # origin of 2D Euclidean space
 struct Point{Dim,T}
   coords::SVector{Dim,T}
   Point{Dim,T}(coords::SVector) where {Dim,T} = new{Dim,T}(coords)
+  Point{Dim,T}(coords::SVector{Dim,T2}) where {Dim,T,T2<:Integer} = new{Dim,Float64}(coords)
```
juliohm: Maybe we need to convert the type differently?

```julia
struct Point{Dim,T}
  coords::SVector{Dim,T}
  function Point{Dim,T}(coords) where {Dim,T}
    eltype(coords) <: Integer ? new(Float64.(coords)) : new(coords)
  end
end
```

Does that work? I am not very familiar with the machinery behind inner constructors. We also need to double check that this is efficient.

hx-guo: Maybe we should use the multiple-dispatch to dispatch `SVector{Dim,T} where T<:Integer` separately instead of make it manual. I am not familiar that, too.

serenity4: I am not sure we need inner constructors at all. Maybe we can keep the default constructors, and add a method `Point(coords::SVector{Dim,<:Integer}) where {Dim} = Point(convert(SVector{Dim,Float64}, coords))`

juliohm: Yes, I find inner constructors a bit difficult to grasp, but remember someone added it to handle the ambiguities with the outer constructors somehow. If we can eliminate the inner constructor and still convert Integer coordinates do Float64 that would be a double win. Appreciate any help on this.

I think our test suite is already covering the past issues people had with Point constructors.

juliohm: @HaoxuanGuo did your previous attempt with an external constructor worked fine? All tests passed?

hx-guo: @juliohm Both the 2 inner or outer one works well. You may notice that there are some test cases not passed. Because they are

```julia
@test coordtype(Point(1, 1)) == Int
@test coordtype(Point([1])) == Int
@test coordtype(Point([1,2])) == Int
@test coordtype(Point([1,2,3])) == Int
```

Obviously, we want to convert them to `Float64` instead of `Int`. It not passed is just our expected.

So, both them are worked fine, but not all tests passed. While **`not all tests passed` is our expected**.

juliohm: Nice, feel free to update the tests to match the new behavior. I can review the PR again by the end of the day. Thanks for this must needed addition.

GuoHaoxuan:
@juliohm I changed the tests to adapt the behavior. Left the construction as inner one.

src/points.jl
```suggestion
  Point{Dim,T}(coords::SVector{Dim,V}) where {Dim,T,V<:Integer} = new{Dim,Float64}(coords)
```

juliohm (approved):

juliohm:
That is awesome @HaoxuanGuo 🚀 Approved the PR.

@serenity4 are we good to go? Any other comment before we merge this? It is a simple but sensitive change of behavior.

serenity4 (approved):
Looks good to me.

juliohm:
Amazing @HaoxuanGuo ! 👌🏽

PR #203, opened 2021-11-09
n-dimensional Point in Segment method

### LINKED ISSUE #185: \in method for Point in Segment
This method is missing. However, this can be easily achieved by defining a segment with the same point:

```julia
s1 = Segment(P2(0,0), P2(1,1))
p1 = Point(0.5,0.5)
!isnothing(intersect(s1, Segment(p1, p1)))
```
I am not sure if that it is the right approach, might be better to have methods for Point with Segment, and use it in Segment with Segment?

ErickChacon:
This is a n-dimensional Point in Segment method that fixes #185 and #197. Tests include 1d, 2d, 3d, and 4d.

### DIFF
--- a/src/polytopes/segment.jl
+++ b/src/polytopes/segment.jl
@@ -37,7 +37,7 @@ end
 function Base.in(p::Point{Dim,T}, s::Segment{Dim,T}) where {Dim,T}
   a, b = s.vertices
   ab, ap = b - a, p - a
-  # points a, b, p are collinear if and only if the cross-products (cp) for ab and ap
+  # points a, b, p are collinear if and only if the cross-products for ab and ap
   # with respect to all possible pairs of coordinates are zero
   iscollinear = true
   for i in 1:Dim, j in (i+1):Dim
--- a/src/polytopes/segment.jl
+++ b/src/polytopes/segment.jl
@@ -40,13 +40,11 @@ function Base.in(p::Point{Dim,T}, s::Segment{Dim,T}) where {Dim,T}
   # points a, b, p are collinear if and only if the cross-products (cp) for ab and ap
   # with respect to all possible pairs of coordinates are zero
   iscollinear = true
-  if Dim > 1
-    for i in 1:Dim, j in (i+1):Dim
-      cp = Vec{2,T}(ab[[i, j]]) × Vec{2,T}(ap[[i, j]])
-      if !isapprox(cp, zero(T), atol=atol(T)^2)
-        iscollinear = false
-        break
-      end
+  for i in 1:Dim, j in (i+1):Dim
+    cp = Vec{2,T}(ab[[i, j]]) × Vec{2,T}(ap[[i, j]])
+    if !isapprox(cp, zero(T), atol=atol(T)^2)
+      iscollinear = false
+      break
     end
   end
   # given collinear points (a, b, p), the point p intersects
--- a/src/polytopes/segment.jl
+++ b/src/polytopes/segment.jl
@@ -36,15 +36,16 @@ end
 
 function Base.in(p::Point{Dim,T}, s::Segment{Dim,T}) where {Dim,T}
   a, b = s.vertices
-  if Dim == 2
-    _0 = zero(T)
-  elseif Dim == 3
-    _0 = Vec{3,T}(0, 0, 0)
+  ab, ap = b - a, p - a
+  # points a, b, p are collinear if and only if the cross-products (cp) for ab and ap
+  # with respect to all possible pairs of coordinates are zero
+  if T == 1
+    iscollinear = true
   else
-    throw(ErrorException("not implemented"))
+    cp = [Vec{2,T}(ab[[i, j]]) × Vec{2,T}(ap[[i, j]]) for i in 1:Dim for j in (i+1):Dim]
+    iscollinear = all(isapprox.(cp, zero(T), atol=atol(T)^2))
   end
   # given collinear points (a, b, p), the point p intersects
   # segment ab if and only if vectors satisfy 0 ≤ ap ⋅ ab ≤ ||ab||²
-  iscollinear = isapprox((b - a) × (p - a), _0, atol=atol(T)^2)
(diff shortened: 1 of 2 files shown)


src/polytopes/segment.jl
```diff
@@ -36,15 +36,16 @@ end
...
-  elseif Dim == 3
-    _0 = Vec{3,T}(0, 0, 0)
+  ab, ap = b - a, p - a
+  # points a, b, p are collinear if and only if the cross-products (cp) for ab and ap
+  # with respect to all possible pairs of coordinates are zero
+  if T == 1
+    iscollinear = true
   else
-    throw(ErrorException("not implemented"))
+    cp = [Vec{2,T}(ab[[i, j]]) × Vec{2,T}(ap[[i, j]]) for i in 1:Dim for j in (i+1):Dim]
```
juliohm: We should probably split this into a for loop with premature termination whenever a pair fails the condition? Right now you are computing all possible cross-products to then check later if they are all zero. You could write a simple for loop and break the loop if any intermediate result is non-zero. Makes sense? This can also improve readibility.

I wonder if you have a reference to cite here to support this statement about cross-products for future readers?

juliohm (changes requested):
Review attached.

src/polytopes/segment.jl
juliohm: @ErickChacon can you refactor this into a for loop with early termination? You can do something like:

```julia
iscollinear = true
for i in 1:Dim, j in (i+1):Dim
  # do cross product with i and j
  if !iszero(cp)
    iscollinear = false
    break
  end
end
```

src/polytopes/segment.jl
```diff
@@ -36,15 +36,20 @@ end
...
-    _0 = zero(T)
-  elseif Dim == 3
-    _0 = Vec{3,T}(0, 0, 0)
-  else
-    throw(ErrorException("not implemented"))
+  ab, ap = b - a, p - a
+  # points a, b, p are collinear if and only if the cross-products (cp) for ab and ap
+  # with respect to all possible pairs of coordinates are zero
+  iscollinear = true
+  if Dim > 1
```
juliohm: I think this if statement is redudant? When Dim==1, we never enter in the loop.

src/polytopes/segment.jl
```diff
@@ -36,15 +36,20 @@ end
...
+    for i in 1:Dim, j in (i+1):Dim
+      cp = Vec{2,T}(ab[[i, j]]) × Vec{2,T}(ap[[i, j]])
```
```suggestion
     v = Vec{2,T}(ab[i], ab[j]) × Vec{2,T}(ap[i], ap[j])
```

src/polytopes/segment.jl
```diff
@@ -36,15 +36,20 @@ end
...
+      if !isapprox(cp, zero(T), atol=atol(T)^2)
```
```suggestion
      if !isapprox(v, zero(T), atol=atol(T)^2)
```

test/polytopes.jl
```diff
@@ -3,20 +3,28 @@
...
     s = Segment(P2(0,0), P2(1,1))
     @test minimum(s) == P2(0,0)
     @test maximum(s) == P2(1,1)
     @test extrema(s) == (P2(0,0), P2(1,1))
     @test isapprox(length(s), sqrt(T(2)))
     @test s(T(0)) == P2(0,0)
     @test s(T(1)) == P2(1,1)
-    @test all(p ∈ s for p in [P2(0.11, 0.11), P2(0.5, 0.5), P2(0.87, 0.87)])
-    @test all(p ∉ s for p in [P2(-0.1, -0.1), P2(1.1, 1.1), P2(1, 2)])
+    @test all(P2(x,x) ∈ s for x in 0:0.01:1)
```
juliohm, line 16: Can you please preserve the original version of the test? We try to leave tests untouched unless there is an actual change of behavior. The second modification is fine because you added a new point to the list.

juliohm (changes requested):

src/polytopes/segment.jl
```diff
@@ -36,15 +36,18 @@ end
...
-  elseif Dim == 3
-    _0 = Vec{3,T}(0, 0, 0)
-  else
-    throw(ErrorException("not implemented"))
+  ab, ap = b - a, p - a
+  # points a, b, p are collinear if and only if the cross-products (cp) for ab and ap
+  # with respect to all possible pairs of coordinates are zero
+  iscollinear = true
+  for i in 1:Dim, j in (i+1):Dim
+    cp = Vec{2,T}(ab[[i, j]]) × Vec{2,T}(ap[[i, j]])
```
juliohm: You rejected the suggestion here, but I think we need to avoid this allocation with [i,j]. Can you please explain why the suggestion was not applied?

ErickChacon:
Sorry @juliohm, I think the suggestion was rejected when I did a push when you were reviewing.

juliohm:
Oh got it, you can still see the suggested changes in the history above. Can you please incorporate these changes in the next set of commits? I think we are almost ready to merge it 👌🏽

ErickChacon:
They are already included. I left some comments about the test, that is the only pending issue. Could you confirm you want me to change it to the previous one?

src/polytopes/segment.jl
```diff
@@ -36,15 +36,18 @@ end
...
 function Base.in(p::Point{Dim,T}, s::Segment{Dim,T}) where {Dim,T}
   a, b = s.vertices
-  if Dim == 2
-    _0 = zero(T)
-  elseif Dim == 3
-    _0 = Vec{3,T}(0, 0, 0)
-  else
-    throw(ErrorException("not implemented"))
+  ab, ap = b - a, p - a
+  # points a, b, p are collinear if and only if the cross-products (cp) for ab and ap
```
```suggestion
  # points a, b, p are collinear if and only if the cross-products for ab and ap
```

juliohm:
Can't find the comments, but I think the change is fine, let's wait for the test suite to finish running 👍🏽

juliohm (approved):

juliohm:
Thanks for another great contribution! 💯

PR #205, opened 2021-11-11
Ray triangle intersection

HaoxuanGuo:
Implemented intersection of `Ray`s and `Triangle`s. I think this is a step towards #125 .

I used moller's algorithm instead of @jackbeagley 's chosen which is specified for `Segment`.

You may saw there is a test not passed. That is the floating point error. I don't know how to fix it. Sorry for that.

### DIFF
--- a/src/intersections.jl
+++ b/src/intersections.jl
@@ -103,7 +103,7 @@ struct IntersectingSegmentTriangle{P<:Point} <: Intersection
 end
 
 # ---------------------------
-# Ray-TRIANGLE INTERSECTIONS
+# RAY-TRIANGLE INTERSECTIONS
 # ---------------------------
 
 struct IntersectingRayTriangle{P<:Point} <: Intersection
--- a/src/intersections.jl
+++ b/src/intersections.jl
@@ -102,6 +102,14 @@ struct IntersectingSegmentTriangle{P<:Point} <: Intersection
   value::P
 end
 
+# ---------------------------
+# Ray-TRIANGLE INTERSECTIONS
+# ---------------------------
+
+struct IntersectingRayTriangle{P<:Point} <: Intersection
+  value::P
+end
+
 # ----------------------------
 # SEGMENT-PLANE INTERSECTIONS
 # ----------------------------
@@ -166,7 +174,7 @@ function hasintersect(g1::Geometry{Dim,T}, g2::Geometry{Dim,T}) where {Dim,T}
 
   # first point in Minkowski difference
   P = minkowskipoint(g1, g2, d)
-  
+
   # origin of coordinate system
   O = minkowskiorigin(Dim, T)
 
--- a/src/intersections/triangles.jl
+++ b/src/intersections/triangles.jl
@@ -158,7 +158,7 @@ function intersecttype(r::Ray{3,Type}, t::Triangle{3,Type}) where {Type}
     det = -det
   end
 
-  if det < -atol(Type)
+  if det < atol(Type)
       # This ray is parallel to the plane of the triangle.
       return NoIntersection()
   end
--- a/src/intersections/triangles.jl
+++ b/src/intersections/triangles.jl
@@ -142,20 +142,20 @@ and apply function `f` to it.
   (https://www.tandfonline.com/doi/abs/10.1080/10867651.1997.10487468)
 """
 function intersecttype(f::Function, r::Ray{3,T}, t::Triangle{3,T}) where {T}
-  vₜ = vertices(t)
-  origin = r.p
-  D = r.v
+  vs = vertices(t)
+  o = r.p
+  d = r.v
 
-  E₁ = vₜ[3] - vₜ[1]
-  E₂ = vₜ[2] - vₜ[1]
-  P = D × E₂
-  det = E₁ ⋅ P
+  e₁ = vs[3] - vs[1]
+  e₂ = vs[2] - vs[1]
+  p = d × e₂
+  det = e₁ ⋅ p
 
   # keep det > 0, modify T accordingly
   if det > atol(T)
-    TT = origin - vₜ[1]
+    t_ = o - vs[1]
   else
-    TT = vₜ[1] - origin
+    t_ = vs[1] - o
     det = -det
   end
 
@@ -165,20 +165,20 @@ function intersecttype(f::Function, r::Ray{3,T}, t::Triangle{3,T}) where {T}
   end
 
   # calculate u parameter and test bounds
-  u = TT ⋅ P
+  u = t_ ⋅ p
   if u < -atol(T) || u > det
     return NoIntersection() |> f
   end
 
-  Q = TT × E₁
+  q = t_ × e₁
 
   # calculate v parameter and test bounds
-  v = D ⋅ Q
+  v = d ⋅ q
   if v < -atol(T) || u + v > det
     return NoIntersection() |> f
   end
 
-  t = E₂ ⋅ Q
+  t = e₂ ⋅ q
 
   invdet = one(T) / det
   t *= invdet
--- a/src/intersections/triangles.jl
+++ b/src/intersections/triangles.jl
@@ -132,7 +132,8 @@ end
 """
     intersecttype(s, t)
 
-Calculate the intersection of a ray and triangle in 3D.
+Calculate the intersection of a ray and triangle in 3D
+and apply function `f` to it.
 
 ## References
 
@@ -140,7 +141,7 @@ Calculate the intersection of a ray and triangle in 3D.
   intersection. Journal of graphics tools]
   (https://www.tandfonline.com/doi/abs/10.1080/10867651.1997.10487468)
 """
-function intersecttype(r::Ray{3,T}, t::Triangle{3,T}) where {T}
+function intersecttype(f::Function, r::Ray{3,T}, t::Triangle{3,T}) where {T}
   vₜ = vertices(t)
   origin = r.p
   D = r.v
@@ -160,13 +161,13 @@ function intersecttype(r::Ray{3,T}, t::Triangle{3,T}) where {T}
 
   if det < atol(T)
       # This ray is parallel to the plane of the triangle.
-      return NoIntersection()
+      return NoIntersection() |> f
   end
 
   # calculate u parameter and test bounds
   u = TT ⋅ P
   if u < -atol(T) || u > det
-      return NoIntersection()
+      return NoIntersection() |> f
   end
 
   Q = TT × E₁
@@ -174,7 +175,7 @@ function intersecttype(r::Ray{3,T}, t::Triangle{3,T}) where {T}
   # calculate v parameter and test bounds
   v = D ⋅ Q
   if v < -atol(T) || u + v > det
-      return NoIntersection()
+      return NoIntersection() |> f
   end
 
   t = E₂ ⋅ Q
@@ -183,8 +184,8 @@ function intersecttype(r::Ray{3,T}, t::Triangle{3,T}) where {T}
   t *= invdet
 
   if t < -atol(T)
-      return NoIntersection()
+      return NoIntersection() |> f
   end
 
-  return IntersectingRayTriangle(r(t))
+  return IntersectingRayTriangle(r(t)) |> f
 end
--- a/src/intersections/triangles.jl
+++ b/src/intersections/triangles.jl
@@ -17,7 +17,7 @@ and apply function `f` to it.
 function intersecttype(f::Function, s::Segment{3,T}, t::Triangle{3,T}) where {T}
   vₛ = vertices(s)
   vₜ = vertices(t)
-  
+
   A = vₛ[1] - vₜ[3]
   B = vₜ[1] - vₜ[3]
   C = vₜ[2] - vₜ[3]
@@ -36,7 +36,7 @@ function intersecttype(f::Function, s::Segment{3,T}, t::Triangle{3,T}) where {T}
 
     W₂ = A × D
     tᵥ = W₂ ⋅ C
-    
+
     # rejection 3
     if tᵥ < -atol(T)
       return NoIntersection() |> f
@@ -61,7 +61,7 @@ function intersecttype(f::Function, s::Segment{3,T}, t::Triangle{3,T}) where {T}
 
     W₂ = A × D
     tᵥ = W₂ ⋅ C
-    
+
     # rejection 3
     if tᵥ > atol(T)
       return NoIntersection() |> f
@@ -122,9 +122,69 @@ function intersecttype(f::Function, s::Segment{3,T}, t::Triangle{3,T}) where {T}
       # rejection 1, coplanar segment
       return NoIntersection() |> f
     end
-  end      
+  end
 
   λ = clamp(wᵥ / (wᵥ - sᵥ), zero(T), one(T))
 
   return IntersectingSegmentTriangle(s(λ)) |> f
+end
+
+"""
+    intersecttype(s, t)
+
+Calculate the intersection of a ray and triangle in 3D.
+
+## References
+
+* Möller, T. and Trumbore, B., 1997. [Fast, minimum storage ray-triangle
+  intersection. Journal of graphics tools]
+  (https://www.tandfonline.com/doi/abs/10.1080/10867651.1997.10487468)
+"""
+function intersecttype(r::Ray{3,Type}, t::Triangle{3,Type}) where {Type}
+  vₜ = vertices(t)
+  origin = r.p
+  D = r.v
+
+  E₁ = vₜ[3] - vₜ[1]
+  E₂ = vₜ[2] - vₜ[1]
+  P = D × E₂
+  det = E₁ ⋅ P
+
+  # keep det > 0, modify T accordingly
+  if det > atol(Type)
+    T = origin - vₜ[1]
+  else
+    T = vₜ[1] - origin
+    det = -det
+  end
+
+  if det < -atol(Type)
+      # This ray is parallel to the plane of the triangle.
+      return NoIntersection()
+  end
+
+  # calculate u parameter and test bounds
+  u = T ⋅ P
+  if u < -atol(Type) || u > det
+      return NoIntersection()
+  end
+
+  Q = T × E₁
+
+  # calculate v parameter and test bounds
+  v = D ⋅ Q
+  if v < -atol(Type) || u + v > det
+      return NoIntersection()
+  end
+
+  t = E₂ ⋅ Q
+
+  invdet = one(Type) / det
+  t *= invdet
+
+  if t < -atol(Type)
+      return NoIntersection()
+  end
+
+  return IntersectingRayTriangle(r(t))
 end
--- a/src/intersections/triangles.jl
+++ b/src/intersections/triangles.jl
@@ -178,14 +178,14 @@ function intersecttype(f::Function, r::Ray{3,T}, t::Triangle{3,T}) where {T}
     return NoIntersection() |> f
   end
 
-  t = e₂ ⋅ q
+  λ = e₂ ⋅ q
 
   invdet = one(T) / det
-  t *= invdet
+  λ *= invdet
 
-  if t < -atol(T)
+  if λ < -atol(T)
     return NoIntersection() |> f
   end
 
-  return IntersectingRayTriangle(r(t)) |> f
+  return IntersectingRayTriangle(r(λ)) |> f
 end
--- a/src/intersections/triangles.jl
+++ b/src/intersections/triangles.jl
@@ -140,7 +140,7 @@ Calculate the intersection of a ray and triangle in 3D.
   intersection. Journal of graphics tools]
   (https://www.tandfonline.com/doi/abs/10.1080/10867651.1997.10487468)
 """
-function intersecttype(r::Ray{3,Type}, t::Triangle{3,Type}) where {Type}
+function intersecttype(r::Ray{3,T}, t::Triangle{3,T}) where {T}
   vₜ = vertices(t)
   origin = r.p
   D = r.v
@@ -151,38 +151,38 @@ function intersecttype(r::Ray{3,Type}, t::Triangle{3,Type}) where {Type}
   det = E₁ ⋅ P
 
   # keep det > 0, modify T accordingly
-  if det > atol(Type)
-    T = origin - vₜ[1]
+  if det > atol(T)
+    TT = origin - vₜ[1]
   else
-    T = vₜ[1] - origin
+    TT = vₜ[1] - origin
     det = -det
   end
 
-  if det < atol(Type)
+  if det < atol(T)
       # This ray is parallel to the plane of the triangle.
       return NoIntersection()
   end
 
   # calculate u parameter and test bounds
-  u = T ⋅ P
-  if u < -atol(Type) || u > det
+  u = TT ⋅ P
+  if u < -atol(T) || u > det
       return NoIntersection()
   end
 
-  Q = T × E₁
+  Q = TT × E₁
 
   # calculate v parameter and test bounds
   v = D ⋅ Q
-  if v < -atol(Type) || u + v > det
+  if v < -atol(T) || u + v > det
       return NoIntersection()
   end
 
   t = E₂ ⋅ Q
 
-  invdet = one(Type) / det
+  invdet = one(T) / det
   t *= invdet
 
-  if t < -atol(Type)
+  if t < -atol(T)
       return NoIntersection()
   end
 
--- a/src/intersections/triangles.jl
+++ b/src/intersections/triangles.jl
@@ -153,9 +153,9 @@ function intersecttype(f::Function, r::Ray{3,T}, t::Triangle{3,T}) where {T}
 
   # keep det > 0, modify T accordingly
   if det > atol(T)
-    t_ = o - vs[1]
+    τ = o - vs[1]
   else
-    t_ = vs[1] - o
+    τ = vs[1] - o
     det = -det
   end
 
@@ -165,12 +165,12 @@ function intersecttype(f::Function, r::Ray{3,T}, t::Triangle{3,T}) where {T}
   end
 
   # calculate u parameter and test bounds
-  u = t_ ⋅ p
+  u = τ ⋅ p
   if u < -atol(T) || u > det
     return NoIntersection() |> f
   end
 
-  q = t_ × e₁
+  q = τ × e₁
 
   # calculate v parameter and test bounds
   v = d ⋅ q
--- a/src/intersections/triangles.jl
+++ b/src/intersections/triangles.jl
@@ -178,9 +178,7 @@ function intersecttype(f::Function, r::Ray{3,T}, t::Triangle{3,T}) where {T}
     return NoIntersection() |> f
   end
 
-  λ = e₂ ⋅ q
-
-  λ *= one(T) / det
+  λ = (e₂ ⋅ q) * (one(T) / det)
 
   if λ < -atol(T)
     return NoIntersection() |> f
--- a/src/intersections/triangles.jl
+++ b/src/intersections/triangles.jl
@@ -180,8 +180,7 @@ function intersecttype(f::Function, r::Ray{3,T}, t::Triangle{3,T}) where {T}
 
   λ = e₂ ⋅ q
 
-  invdet = one(T) / det
-  λ *= invdet
+  λ *= one(T) / det
 
   if λ < -atol(T)
     return NoIntersection() |> f
--- a/src/intersections/triangles.jl
+++ b/src/intersections/triangles.jl
@@ -160,14 +160,14 @@ function intersecttype(f::Function, r::Ray{3,T}, t::Triangle{3,T}) where {T}
   end
 
   if det < atol(T)
-      # This ray is parallel to the plane of the triangle.
-      return NoIntersection() |> f
+    # This ray is parallel to the plane of the triangle.
+    return NoIntersection() |> f
   end
 
   # calculate u parameter and test bounds
   u = TT ⋅ P
   if u < -atol(T) || u > det
-      return NoIntersection() |> f
+    return NoIntersection() |> f
   end
 
   Q = TT × E₁
@@ -175,7 +175,7 @@ function intersecttype(f::Function, r::Ray{3,T}, t::Triangle{3,T}) where {T}
   # calculate v parameter and test bounds
   v = D ⋅ Q
   if v < -atol(T) || u + v > det
-      return NoIntersection() |> f
+    return NoIntersection() |> f
   end
 
   t = E₂ ⋅ Q
@@ -184,7 +184,7 @@ function intersecttype(f::Function, r::Ray{3,T}, t::Triangle{3,T}) where {T}
   t *= invdet
 
   if t < -atol(T)
-      return NoIntersection() |> f
+    return NoIntersection() |> f
   end
 
   return IntersectingRayTriangle(r(t)) |> f
--- a/test/intersections.jl
+++ b/test/intersections.jl
@@ -162,6 +162,127 @@
     s = Segment(P3(0.2, 0.2, 1.0), P3(0.2, 0.2, -1.0))
     @test intersecttype(s, t) isa IntersectingSegmentTriangle
     @test s ∩ t == P3(0.2, 0.2, 0.0)
+
+    # intersects at a vertex of t
+    s = Segment(P3(0.0, 0.0, 1.0), P3(0.0, 0.0, -1.0))
+    @test intersecttype(s, t) isa IntersectingSegmentTriangle
+    @test s ∩ t == P3(0.0, 0.0, 0.0)
+
+    # normal to, doesn't intersect with t
+    s = Segment(P3(0.9, 0.9, 1.0), P3(0.9, 0.9, -1.0))
+    @test intersecttype(s, t) isa NoIntersection
+    @test isnothing(s ∩ t)
+
+    # coplanar, intersects with t (but should return NoIntersection)
+    s = Segment(P3(-0.2, 0.2, 0.0), P3(1.2, 0.2, 0.0))
+    @test intersecttype(s, t) isa NoIntersection
+    @test isnothing(s ∩ t)
+
+    # coplanar, doesn't intersect with t
+    s = Segment(P3(-0.2, -0.2, 0.0), P3(1.2, -0.2, 0.0))
+    @test intersecttype(s, t) isa NoIntersection
+    @test isnothing(s ∩ t)
+
+    # parallel, above, doesn't intersect with t
+    s = Segment(P3(-0.2, 0.2, 1.0), P3(1.2, 0.2, 1.0))
+    @test intersecttype(s, t) isa NoIntersection
+    @test isnothing(s ∩ t)
+
+    # parallel, below, doesn't intersect with t
+    s = Segment(P3(-0.2, 0.2, -1.0), P3(1.2, 0.2, -1.0))
+    @test intersecttype(s, t) isa NoIntersection
+    @test isnothing(s ∩ t)
+
+    # segment colinear with edge of t (but should return NoIntersection)
+    s = Segment(P3(-1.0, 0.0, 0.0), P3(1.0, 0.0, 0.0))
+    @test intersecttype(s, t) isa NoIntersection
+    @test isnothing(s ∩ t)
+
+    # coplanar, within bounding box of t, no intersection
+    s = Segment(P3(0.7, 0.8, 0.0), P3(0.8, 0.7, 0.0))
+    @test intersecttype(s, t) isa NoIntersection
+    @test isnothing(s ∩ t)
+
+    # segment above and to right of t, no intersection
+    s = Segment(P3(1.0, 1.0, 0.0), P3(1.0, 1.0, 1.0))
+    @test intersecttype(s, t) isa NoIntersection
+    @test isnothing(s ∩ t)
+
+    # segment below t, no intersection
+    s = Segment(P3(0.5, -1.0, 0.0), P3(0.5, -1.0, 1.0))
+    @test intersecttype(s, t) isa NoIntersection
+    @test isnothing(s ∩ t)
+
+    # segment left of t, no intersection
+    s = Segment(P3(-1.0, 0.5, 0.0), P3(-1.0, 0.5, 1.0))
+    @test intersecttype(s, t) isa NoIntersection
+    @test isnothing(s ∩ t)
+
+    # segment above and to right of t, no intersection
+    s = Segment(P3(1.0, 1.0, 0.0), P3(1.0, 1.0, -1.0))
+    @test intersecttype(s, t) isa NoIntersection
+    @test isnothing(s ∩ t)
+    @test intersecttype(reverse_segment(s), t) isa NoIntersection
+    @test isnothing(reverse_segment(s) ∩ t)
+
+    # segment below t, no intersection
+    s = Segment(P3(0.5, -1.0, 0.0), P3(0.5, -1.0, -1.0))
+    @test intersecttype(s, t) isa NoIntersection
+    @test isnothing(s ∩ t)
+    @test intersecttype(reverse_segment(s), t) isa NoIntersection
+    @test isnothing(reverse_segment(s) ∩ t)
+
+    # segment left of t, no intersection
+    s = Segment(P3(-1.0, 0.5, 0.0), P3(-1.0, 0.5, -1.0))
+    @test intersecttype(s, t) isa NoIntersection
+    @test isnothing(s ∩ t)
+    @test intersecttype(reverse_segment(s), t) isa NoIntersection
+    @test isnothing(reverse_segment(s) ∩ t)
+
+    # segment above and to right of t, no intersection
+    s = Segment(P3(1.0, 1.0, 1.0), P3(1.0, 1.0, 0.0))
+    @test intersecttype(s, t) isa NoIntersection
+    @test isnothing(s ∩ t)
+
+    # segment below t, no intersection
+    s = Segment(P3(0.5, -1.0, 1.0), P3(0.5, -1.0, 0.0))
+    @test intersecttype(s, t) isa NoIntersection
+    @test isnothing(s ∩ t)
+
+    # segment left of t, no intersection
+    s = Segment(P3(-1.0, 0.5, 1.0), P3(-1.0, 0.5, 0.0))
+    @test intersecttype(s, t) isa NoIntersection
+    @test isnothing(s ∩ t)
+
+    # intersections with an inclined inclined triangle t
+    t = Triangle(P3(0, 0, 0), P3(2, 0, 0), P3(0, 2, 2))
+
+    # doesn't reach t, no intersection
+    s = Segment(P3(0.5, 0.5, 1.9), P3(0.5, 0.5, 1.8))
+    @test intersecttype(s, t) isa NoIntersection
+    @test isnothing(s ∩ t)
+
+    # parallel, offset from t, no intersection
+    s = Segment(P3(0.0, 0.5, 1.0), P3(1.0, 0.5, 1.0))
+    @test intersecttype(s, t) isa NoIntersection
+    @test isnothing(s ∩ t)
+
+    # triangle as first argument
+    t = Triangle(P3(0, 0, 0), P3(1, 0, 0), P3(0, 1, 0))
+    s = Segment(P3(0.2, 0.2, 1.0), P3(0.2, 0.2, -1.0))
+    @test intersecttype(t, s) isa IntersectingSegmentTriangle
+    @test s ∩ t == t ∩ s == P3(0.2, 0.2, 0.0)
+
+    # type stability tests
+    s = Segment(P3(0.2, 0.2, 1.0), P3(0.2, 0.2, -1.0))
+    t = Triangle(P3(0, 0, 0), P3(1, 0, 0), P3(0, 1, 0))
+    @inferred someornone(s, t)
+
+
+    # Intersection for a triangle and a ray
+    t = Triangle(P3(0, 0, 0), P3(1, 0, 0), P3(0, 1, 0))
+
+    # intersects through t
     r = Ray(P3(0.2, 0.2, 1.0), V3(0.0, 0.0, -1.0))
     @test intersecttype(r, t) isa IntersectingRayTriangle
     @test r ∩ t == P3(0.2, 0.2, 0.0)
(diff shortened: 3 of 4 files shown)


src/intersections.jl
```diff
@@ -81,6 +81,14 @@ struct IntersectingSegmentTriangle{P<:Point} <: Intersection
   value::P
 end
 
+# ---------------------------
+# Ray-TRIANGLE INTERSECTIONS
```
```suggestion
# RAY-TRIANGLE INTERSECTIONS
```

src/intersections/triangles.jl
```diff
@@ -121,12 +121,73 @@ function intersecttype(s::Segment{3,T}, t::Triangle{3,T}) where {T}
...
+
+Calculate the intersection of a ray and triangle in 3D.
+
+## References
+
+* Möller, T. and Trumbore, B., 1997. [Fast, minimum storage ray-triangle
+  intersection. Journal of graphics tools]
+  (https://www.tandfonline.com/doi/abs/10.1080/10867651.1997.10487468)
+"""
+function intersecttype(r::Ray{3,Type}, t::Triangle{3,Type}) where {Type}
```
juliohm: Please follow the code style `T` for coordinate type.

juliohm (changes requested):
Thank you for another great contribution @HaoxuanGuo. How does this algorithm compare with https://github.com/JuliaGeometry/TriangleIntersect.jl?

GuoHaoxuan:
@juliohm That code is a little hard to read. I guess that code do

1. Calculate the intersection point for ray and the plane where triangle lies in.
2. Check if the intersection point in the triangle.

That is the plain idea to deal with that question. Moller's algorithms improves the performance than it with some linear algebra trick. That is why moller's algorithm come out.

Besides, I changed the code style `T` for coordinate type.

src/intersections/triangles.jl
```diff
@@ -121,12 +121,73 @@ function intersecttype(s::Segment{3,T}, t::Triangle{3,T}) where {T}
...
+
+Calculate the intersection of a ray and triangle in 3D.
+
+## References
+
+* Möller, T. and Trumbore, B., 1997. [Fast, minimum storage ray-triangle
+  intersection. Journal of graphics tools]
+  (https://www.tandfonline.com/doi/abs/10.1080/10867651.1997.10487468)
+"""
+function intersecttype(r::Ray{3,T}, t::Triangle{3,T}) where {T}
```
juliohm: Can you please rebase the PR to the latest master? We have recently added a function as the first argument of `intersecttype` to escape type instability, see the latest comments here: https://github.com/JuliaGeometry/Meshes.jl/issues/54

juliohm:
Nice, so you are suggesting that this Moller's algorithm is more efficient than TriangleIntersect.jl? It would be nice to see some `@btime` results for future reference. Can you run two implementations with a simple example just to have an idea? :)

GuoHaoxuan:
I rebased it.

A little strange. My code is slower than `TriangleIntersect.jl`. Why?

Moller's algorithm:

* code:

```julia
using Meshes
using BenchmarkTools

t = Triangle((1.0, 0.0, 0.0), (0.0, 1.0, 0.0), (0.0, 0.0, 1.0))
s = [Ray((rand(), rand(), rand()), (rand(), rand(), rand())) for i in 1:10000]

@btime [intersect(si, t) for si in s]
```

* result: `613.700 μs (33324 allocations: 1.71 MiB)`

---

TriangleIntersect.jl:

* code:

```julia
using TriangleIntersect
using BenchmarkTools

t = Triangle(Point(1.0, 0.0, 0.0), Point(0.0, 1.0, 0.0), Point(0.0, 0.0, 1.0))
s = [Ray(Point(rand(), rand(), rand()), Point(rand(), rand(), rand())) for i in 1:10000]
@btime [intersect(si, t) for si in s]
```

* result: `167.300 μs (3 allocations: 390.72 KiB)`

juliohm:
I would start finding out where these allocations are coming from. Does it also happen when you call `intersecttype` directly? What if you call `intersecttype` with a function as first argument, does the time change?

GuoHaoxuan:
My code: `47.273 ns (1 allocation: 64 bytes)` while TriangleIntersect.jl `36.556 ns (2 allocations: 112 bytes)`.

The allocations is because TriangleIntersect.jl use a is-bits stable type. The Vector could be allocated before computing.

GuoHaoxuan:
Oh. I know the reason. Maybe because It cached too much info on the triangle.

```julia
struct Triangle
    a::Point
    v1::Point
    v2::Point
    normal::Point
    v1v1::Float64
    v2v2::Float64
    v1v2::Float64
    denom::Float64
    function Triangle(a::Point, b::Point, c::Point)
        v1 = b-a
        v2 = c-a
        normal = cross(v1, v2)
        v1v1 = v1*v1
        v2v2 = v2*v2
        v1v2 = v1*v2
        denom = v1v2*v1v2 - v1v1*v2v2
        new(a, v1, v2, normal, v1v1, v2v2, v1v2, denom)
    end
end
```

normal. denom, and some dots.

GuoHaoxuan:
I found that the cache doesn't cost too much time.

```julia
@btime Triangle(Point(1.0, 0.0, 0.0), Point(0.0, 1.0, 0.0), Point(0.0, 0.0, 1.0))
```

will get `0.001 ns (0 allocations: 0 bytes)`.

Confused. 😭

juliohm:
If you can write your algorithm in a new function using TriangleIntersect.jl Triangle and Point types, you may be able to find the source of the issue. I am asking about performance here because it may be a good time to nail it down using another reference implementation. Also, notice that some tests are not passing in the PR (didn't have time to look).

GuoHaoxuan:
@juliohm I followed your suggestion. Re-write the two algorithm in the same Point, Ray and Intersection definition and disabled the cache in triangle.

<details>
<summary>test code</summary>

```julia
import Base: -,*,+,/, intersect

struct Point
    x::Float64
    y::Float64
    z::Float64
end

*(p::Point, n::Number) = Point(p.x*n, p.y*n, p.z*n)
*(n::Number, p::Point) = p*n
*(p1::Point, p2::Point) = p1.x*p2.x+p1.y*p2.y+p1.z*p2.z
-(p1::Point, p2::Point) = Point(p1.x-p2.x, p1.y-p2.y, p1.z-p2.z)
+(p1::Point, p2::Point) = Point(p1.x+p2.x, p1.y+p2.y, p1.z+p2.z)
cross(p1::Point, p2::Point) = Point(p1.y*p2.z-p1.z*p2.y, -p1.x*p2.z+p1.z*p2.x, p1.x*p2.y-p1.y*p2.x)
dot(p1::Point, p2::Point) = p1.x*p2.x+p1.y*p2.y+p1.z*p2.z

/(p::Point, n::Number) = Point(p.x/n, p.y/n, p.z/n)
unitize(p::Point) = p/(p*p)

struct Triangle
    a::Point
    b::Point
    c::Point
end

struct Ray
    origin::Point
    direction::Point
    Ray(a,b) = new(a,unitize(b))
end

struct Intersection
    ip::Point # intersecting point
    id::Float64 # intersecting distance
    is_intersection::Bool
end

const no_intersection = Intersection(Point(0,0,0), 0.0, false)

function intersect_old(r::Ray, t::Triangle)
    # Prepare cache
    t_v1 = b-a
    t_v2 = c-a
    t_normal = cross(t_v1, t_v2)
    t_v1v1 = t_v1*t_v1
    t_v2v2 = t_v2*t_v2
    t_v1v2 = t_v1*t_v2
    t_denom = t_v1v2*t_v1v2 - t_v1v1*t_v2v2
    # Calculation
    denom = t_normal*r.direction
    denom == 0 && return no_intersection
    ri = t_normal*(t.a - r.origin) / denom
    ri <= 0 && return no_intersection
    plane_intersection =  ri * r.direction + r.origin
    w = plane_intersection - t.a
    wv1 = w*t_v1
    wv2 = w*t_v2
    s_intersection = (t_v1v2*wv2 - t_v2v2*wv1) / t_denom
    s_intersection <= 0 && return no_intersection
    s_intersection >= 1 && return no_intersection
    t_intersection = (t_v1v2*wv1 - t_v1v1*wv2) / t_denom
    t_intersection <= 0 && return no_intersection
    t_intersection >= 1 && return no_intersection
    s_intersection + t_intersection >= 1 && return no_intersection
    Intersection(t.a + s_intersection*t_v1+t_intersection*t_v2, ri, true)
end

function intersect_moller(r::Ray, t::Triangle)
    origin = r.origin
    D = r.direction

    E₁ = t.c - t.a
    E₂ = t.b - t.a
    P = cross(D, E₂)
    det = dot(E₁, P)

    if det >= 0
        TT = origin - t.a
    else
        TT = t.a - origin
        det = -det
    end

    if det <= 0
        return no_intersection
    end

    u = dot(TT, P)
    if u <= 0 || u >= det
        return no_intersection
    end

    Q = cross(TT, E₁)

    v = dot(D, Q)
    if v <= 0 || u + v >= det
        return no_intersection
    end

    t = dot(E₂, Q)
    invdet = 1 / det
    t *= invdet

    if t <= 0
        return no_intersection
    end

    return Intersection(origin + t * D, t, true)
end



using Test
using BenchmarkTools

a = Point(0,0,0)
b = Point(1,0,0)
c = Point(0,1,0)
t = Triangle(a,b,c)
r = Ray(Point(0.1,0.1,1), Point(0.1,0.1,-1))

@test intersect_old(r, t) == Intersection(Point(0.2,0.2,0.0), 1.02, true)
@test intersect_moller(r, t) == Intersection(Point(0.2,0.2,0.0), 1.02, true)
@btime intersect_old(r, t)
@btime intersect_moller(r, t)

r_reverse = Ray(Point(0.1,0.1,1), Point(-0.1,-0.1,1))
@btime intersect_old(r_reverse, t)
@btime intersect_moller(r_reverse, t)
```
</details>

In intersection condition, TriangleIntersect.jl's algorithm costs `360.748 ns (23 allocations: 432 bytes)` while moller's algorithm costs `16.817 ns (0 allocations: 0 bytes)`.

In no intersection condition (the function will end in the middle), TriangleIntersect.jl's algorithm costs `134.155 ns (8 allocations: 160 bytes)` while moller's algotithm costs `9.500 ns (0 allocations: 0 bytes)`.

That proofs moller's algorithm is more efficient.

---

You mentioned there is a test not passed. That test is I added and want to discuss.

```
Triangles: Test Failed at C:\Users\skyair\Desktop\Meshes.jl\test\intersections.jl:171
  Expression: r ∩ t == P3(0.2, 0.2, 0.0)
   Evaluated: Point(0.2, 0.2, 1.1102230246251565e-16) == Point(0.2, 0.2, 0.0)
```

Here `1.1102230246251565e-16 != 0.0`, but they are approx. That is because of a floating point number error. Is is acceptable?

juliohm:
These are impressive numbers @HaoxuanGuo 👏🏽 I am getting different results on my machine though:

```
  19.002 ns (0 allocations: 0 bytes)
  14.168 ns (0 allocations: 0 bytes)
  7.230 ns (0 allocations: 0 bytes)
  9.821 ns (0 allocations: 0 bytes)
```

Also, notice I had to change the beginning of intersect_old to reference `t.a`, `t.b`, `t.c` instead of the global variables. In the non-intersection case the old algorithm is a bit faster, but 2 ns is pretty much irrelevant :)

I think the next step consists of refactoring these two functions in terms of Meshes.jl Point, Triangle and Ray to see if the performance results stay similar to the results above. If they change a lot, we need to understand why and try to fix that. I think our main challenge here is try to beat the performance of an implementation where a `struct Point x; y; z; end` has fixed dimension like above.

Regarding the test that failed, you can use `\approx` like we do in other intersection algorithms. There is also the longer version of the function with `isapprox(x, y, atol=atol(T))`.

GuoHaoxuan:
Sorry for the global variables. I have not noticed that. Now it runs on my machine

```
24.197 ns (0 allocations: 0 bytes)
15.816 ns (0 allocations: 0 bytes)
9.000 ns (0 allocations: 0 bytes)
9.800 ns (0 allocations: 0 bytes)
```

GuoHaoxuan:
@juliohm  For Meshes.jl defination, the result is below:

method | intersection | no intersection
--|--|--
old|80.579 ns (1 allocation: 32 bytes)|53.394 ns (0 allocations: 0 bytes)
moller|71.692 ns (1 allocation: 32 bytes)|58.494 ns (0 allocations: 0 bytes)

<details>
<summary>compared_code</summary>

```julia
function intersecttype_compare(f::Function, r::Ray{3,T}, t::Triangle{3,T}) where {T}
    # Prepare
    vₜ = vertices(t)
    t_v1 = vₜ[3] - vₜ[1]
    t_v2 = vₜ[2] - vₜ[1]
    t_normal = t_v1 × t_v2
    t_v1v1 = t_v1 ⋅ t_v1
    t_v2v2 = t_v2 ⋅ t_v2
    t_v1v2 = t_v1 ⋅ t_v2
    t_denom = t_v1v2 ⋅ t_v1v2 - t_v1v1 ⋅ t_v2v2
    # Calculation
    denom = t_normal ⋅ r.v
    denom == zero(T) && return NoIntersection()
    ri = t_normal ⋅ (vₜ[1] - r.p) / denom
    ri < zero(T) && return NoIntersection()
    plane_intersection = ri * r.v + r.p
    w = plane_intersection - vₜ[1]
    wv1 = w ⋅ t_v1
    wv2 = w ⋅ t_v2
    s_intersection = (t_v1v2 ⋅ wv2 - t_v2v2 ⋅ wv1) / t_denom
    s_intersection < zero(T) && return NoIntersection()
    s_intersection > one(T) && return NoIntersection()
    t_intersection = (t_v1v2 ⋅ wv1 - t_v1v1 ⋅ wv2) / t_denom
    t_intersection < zero(T) && return NoIntersection()
    t_intersection > one(T) && return NoIntersection()
    s_intersection + t_intersection > one(T) && return NoIntersection()
    IntersectingRayTriangle(r(ri)) |> f
end
```

Moller's algorithm code is which in this PR.
</details>

juliohm:
Thank you @HaoxuanGuo, can you provide the full script so that I can reproduce the results? Can you also remove all the intersection types and always return a tuple with the same type for each entry? That way we can be sure that the performance difference is coming from the dot/cross/etc products and types in Meshes.jl and not a function of our current intersecttype design.

juliohm:
The main thing we want to make sure here is that if we use the Meshes.jl Point, Triangle, Ray in the algorithm, the performance will be similar to the hand-crafted Point, Triangle, Ray in TriangleIntersect.jl. After that we can merge the PR, and think about refactoring the intersercttype to avoid this type instability in the return as discussed in #54

GuoHaoxuan:
@juliohm  I tested it by

```julia
using Meshes
a = Point(0,0,0)
b = Point(1,0,0)
c = Point(0,1,0)
t = Triangle(a,b,c)
r = Ray((0.1,0.1,1), (0.1,0.1,-1))
using BenchmarkTools
@btime intersecttype(identity, r ,t)
@btime Meshes.intersecttype_compare(identity, r ,t)
r_reverse = Ray((0.1,0.1,1), (-0.1,-0.1,1))
@btime intersecttype(identity, r_reverse ,t)
@btime Meshes.intersecttype_compare(identity, r_reverse ,t)
```

GuoHaoxuan:
@juliohm Return tuple version.

functions:

```julia
function special_test_moller(r::Ray{3,T}, t::Triangle{3,T}) where {T}
  vₜ = vertices(t)
  origin = r.p
  D = r.v

  E₁ = vₜ[3] - vₜ[1]
  E₂ = vₜ[2] - vₜ[1]
  P = D × E₂
  det = E₁ ⋅ P

  if det > atol(T)
    TT = origin - vₜ[1]
  else
    TT = vₜ[1] - origin
    det = -det
  end

  if det < atol(T)
    # This ray is parallel to the plane of the triangle.
    return (0.0,0.0,0.0,false)
  end

  # calculate u parameter and test bounds
  u = TT ⋅ P
  if u < -atol(T) || u > det
    return (0.0,0.0,0.0,false)
  end

  Q = TT × E₁

  # calculate v parameter and test bounds
  v = D ⋅ Q
  if v < -atol(T) || u + v > det
    return (0.0,0.0,0.0,false)
  end

  t = E₂ ⋅ Q

  invdet = one(T) / det
  t *= invdet

  if t < -atol(T)
    return (0.0,0.0,0.0,false)
  end

  return (r(t).coords..., true)
end

function special_test_compare(r::Ray{3,T}, t::Triangle{3,T}) where {T}
    # Prepare
    vₜ = vertices(t)
    t_v1 = vₜ[3] - vₜ[1]
    t_v2 = vₜ[2] - vₜ[1]
    t_normal = t_v1 × t_v2
    t_v1v1 = t_v1 ⋅ t_v1
    t_v2v2 = t_v2 ⋅ t_v2
    t_v1v2 = t_v1 ⋅ t_v2
    t_denom = t_v1v2 ⋅ t_v1v2 - t_v1v1 ⋅ t_v2v2
    # Calculation
    denom = t_normal ⋅ r.v
    denom == zero(T) && return (0.0,0.0,0.0,false)
    ri = t_normal ⋅ (vₜ[1] - r.p) / denom
    ri < zero(T) && return (0.0,0.0,0.0,false)
    plane_intersection = ri * r.v + r.p
    w = plane_intersection - vₜ[1]
    wv1 = w ⋅ t_v1
    wv2 = w ⋅ t_v2
    s_intersection = (t_v1v2 ⋅ wv2 - t_v2v2 ⋅ wv1) / t_denom
    s_intersection < zero(T) && return (0.0,0.0,0.0,false)
    s_intersection > one(T) && return (0.0,0.0,0.0,false)
    t_intersection = (t_v1v2 ⋅ wv1 - t_v1v1 ⋅ wv2) / t_denom
    t_intersection < zero(T) && return (0.0,0.0,0.0,false)
    t_intersection > one(T) && return (0.0,0.0,0.0,false)
    s_intersection + t_intersection > one(T) && return (0.0,0.0,0.0,false)
    (r(ri).coords..., true)
end
```

tests:

```julia
using Meshes
a = Point(0,0,0)
b = Point(1,0,0)
c = Point(0,1,0)
t = Triangle(a,b,c)
r = Ray((0.1,0.1,1), (0.1,0.1,-1))
using BenchmarkTools
@btime Meshes.special_test_moller(r ,t)  # 26.707 ns (1 allocation: 48 bytes)
@btime Meshes.special_test_compare(r ,t)  # 34.070 ns (1 allocation: 48 bytes)
r_reverse = Ray((0.1,0.1,1), (-0.1,-0.1,1))
@btime Meshes.special_test_moller(r_reverse ,t)  # 25.100 ns (1 allocation: 48 bytes)
@btime Meshes.special_test_compare(r_reverse ,t)  # 21.700 ns (1 allocation: 48 bytes)
```

juliohm:
# Script

I had trouble trying to run the script. Here is a script that runs with the master branch:

```julia
using Pkg
Pkg.activate(@__DIR__)
using Meshes
using LinearAlgebra

function special_test_compare(r::Ray{3,T}, t::Triangle{3,T}) where {T}
    # Prepare
    vₜ = vertices(t)
    t_v1 = vₜ[3] - vₜ[1]
    t_v2 = vₜ[2] - vₜ[1]
    t_normal = t_v1 × t_v2
    t_v1v1 = t_v1 ⋅ t_v1
    t_v2v2 = t_v2 ⋅ t_v2
    t_v1v2 = t_v1 ⋅ t_v2
    t_denom = t_v1v2 ⋅ t_v1v2 - t_v1v1 ⋅ t_v2v2
    # Calculation
    denom = t_normal ⋅ r.v
    denom == zero(T) && return (0.0,0.0,0.0,false)
    ri = t_normal ⋅ (vₜ[1] - r.p) / denom
    ri < zero(T) && return (0.0,0.0,0.0,false)
    plane_intersection = ri * r.v + r.p
    w = plane_intersection - vₜ[1]
    wv1 = w ⋅ t_v1
    wv2 = w ⋅ t_v2
    s_intersection = (t_v1v2 ⋅ wv2 - t_v2v2 ⋅ wv1) / t_denom
    s_intersection < zero(T) && return (0.0,0.0,0.0,false)
    s_intersection > one(T) && return (0.0,0.0,0.0,false)
    t_intersection = (t_v1v2 ⋅ wv1 - t_v1v1 ⋅ wv2) / t_denom
    t_intersection < zero(T) && return (0.0,0.0,0.0,false)
    t_intersection > one(T) && return (0.0,0.0,0.0,false)
    s_intersection + t_intersection > one(T) && return (0.0,0.0,0.0,false)
    (r(ri).coords..., true)
end

function special_test_moller(r::Ray{3,T}, t::Triangle{3,T}) where {T}
  vₜ = vertices(t)
  origin = r.p
  D = r.v

  E₁ = vₜ[3] - vₜ[1]
  E₂ = vₜ[2] - vₜ[1]
  P = D × E₂
  det = E₁ ⋅ P

  if det > atol(T)
    TT = origin - vₜ[1]
  else
    TT = vₜ[1] - origin
    det = -det
  end

  if det < atol(T)
    # This ray is parallel to the plane of the triangle.
    return (0.0,0.0,0.0,false)
  end

  # calculate u parameter and test bounds
  u = TT ⋅ P
  if u < -atol(T) || u > det
    return (0.0,0.0,0.0,false)
  end

  Q = TT × E₁

  # calculate v parameter and test bounds
  v = D ⋅ Q
  if v < -atol(T) || u + v > det
    return (0.0,0.0,0.0,false)
  end

  t = E₂ ⋅ Q

  invdet = one(T) / det
  t *= invdet

  if t < -atol(T)
    return (0.0,0.0,0.0,false)
  end

  return (r(t).coords..., true)
end

a = Point(0,0,0)
b = Point(1,0,0)
c = Point(0,1,0)
t = Triangle(a,b,c)
r = Ray((0.1,0.1,1), (0.1,0.1,-1))
using BenchmarkTools
@btime special_test_compare(r ,t)
@btime special_test_moller(r ,t)
r_reverse = Ray((0.1,0.1,1), (-0.1,-0.1,1))
@btime special_test_compare(r_reverse ,t)
@btime special_test_moller(r_reverse ,t)
```

I ran the above script with `julia --track-allocation main.jl` assuming that a `Project.toml` exists in the same folder with `] dev Meshes` in it.

# Allocations

The run showed the lines in the source code that are still allocating.

## Point constructor

```julia
        - struct Point{Dim,T}
        -   coords::SVector{Dim,T}
        -   Point{Dim,T}(coords::SVector) where {Dim,T} = new{Dim,T}(coords)
        -   Point{Dim,T}(coords::SVector{Dim,V}) where {Dim,T,V<:Integer} = new{Dim,Float64}(coords)
        - end
        - 
        - # convenience constructors
        - Point{Dim,T}(coords...) where {Dim,T} = Point{Dim,T}(SVector{Dim,T}(coords...))
        0 Point(coords::SVector{Dim,T}) where {Dim,T} = Point{Dim,T}(coords)
        - Point(coords::AbstractVector{T}) where {T} = Point{length(coords),T}(coords)
      192 Point(coords...) = Point(SVector(coords...))
```

## Polytope constructor

```julia
        - """
        -     Polytope{K,Dim,T}
        - 
        - We say that a geometry is a K-polytope when it is a collection of "flat" sides
        - that constitue a `K`-dimensional subspace. They are called polygon and polyhedron
        - respectively for 2D (`K=2`) and 3D (`K=3`) subspaces, embedded in a `Dim`-dimensional
        - space. The parameter `K` is also known as the rank or parametric dimension of the
        - polytope: https://en.wikipedia.org/wiki/Abstract_polytope.
        - 
        - The term polytope expresses a particular combinatorial structure. A polyhedron,
        - for example, can be decomposed into faces. Each face can then be decomposed into
        - edges, and edges into vertices. Some conventions act as a mapping between vertices
        - and higher dimensional features (edges, faces, cells...), removing the need to
        - store all features.
        - 
        - Additionally, the following property must hold in order for a geometry to be considered
        - a polytope: the boundary of a (K+1)-polytope is a collection of K-polytopes, which may
        - have (K-1)-polytopes in common. See https://en.wikipedia.org/wiki/Polytope.
        - 
        - ### Notes
        - 
        - - Type aliases are `Polygon`, `Polyhedron`.
        - """
        - abstract type Polytope{K,Dim,T} <: Geometry{Dim,T} end
        - 
      160 (::Type{PL})(vertices::Vararg{P}) where {PL<:Polytope,P<:Point} = PL(SVector(vertices))
```

Do you think it is possible to remove these allocations? Either by using different constructors in the intersection algorithm or by fixing some of our constructors for Point and Polytope.

juliohm:
Nevermind, the allocations are happening outside the functions in the initalization of the inputs. I am adjusting it locally here now.

juliohm:
Here is the updated script:

```julia
using Pkg
Pkg.activate(@__DIR__)
using Meshes
using LinearAlgebra
using BenchmarkTools

function intersect_old(r::Ray{3,T}, t::Triangle{3,T}) where {T}
  # Prepare
  vₜ = vertices(t)
  t_v1 = vₜ[3] - vₜ[1]
  t_v2 = vₜ[2] - vₜ[1]
  t_normal = t_v1 × t_v2
  t_v1v1 = t_v1 ⋅ t_v1
  t_v2v2 = t_v2 ⋅ t_v2
  t_v1v2 = t_v1 ⋅ t_v2
  t_denom = t_v1v2 ⋅ t_v1v2 - t_v1v1 ⋅ t_v2v2
  # Calculation
  denom = t_normal ⋅ r.v
  denom == zero(T) && return (0.0,0.0,0.0,false)
  ri = t_normal ⋅ (vₜ[1] - r.p) / denom
  ri < zero(T) && return (0.0,0.0,0.0,false)
  plane_intersection = ri * r.v + r.p
  w = plane_intersection - vₜ[1]
  wv1 = w ⋅ t_v1
  wv2 = w ⋅ t_v2
  s_intersection = (t_v1v2 ⋅ wv2 - t_v2v2 ⋅ wv1) / t_denom
  s_intersection < zero(T) && return (0.0,0.0,0.0,false)
  s_intersection > one(T) && return (0.0,0.0,0.0,false)
  t_intersection = (t_v1v2 ⋅ wv1 - t_v1v1 ⋅ wv2) / t_denom
  t_intersection < zero(T) && return (0.0,0.0,0.0,false)
  t_intersection > one(T) && return (0.0,0.0,0.0,false)
  s_intersection + t_intersection > one(T) && return (0.0,0.0,0.0,false)
  (r(ri).coords..., true)
end

function intersect_new(r::Ray{3,T}, t::Triangle{3,T}) where {T}
  vₜ = vertices(t)
  origin = r.p
  D = r.v

  E₁ = vₜ[3] - vₜ[1]
  E₂ = vₜ[2] - vₜ[1]
  P = D × E₂
  det = E₁ ⋅ P

  if det > atol(T)
    TT = origin - vₜ[1]
  else
    TT = vₜ[1] - origin
    det = -det
  end

  if det < atol(T)
    # This ray is parallel to the plane of the triangle.
    return (0.0,0.0,0.0,false)
  end

  # calculate u parameter and test bounds
  u = TT ⋅ P
  if u < -atol(T) || u > det
    return (0.0,0.0,0.0,false)
  end

  Q = TT × E₁

  # calculate v parameter and test bounds
  v = D ⋅ Q
  if v < -atol(T) || u + v > det
    return (0.0,0.0,0.0,false)
  end

  t = E₂ ⋅ Q

  invdet = one(T) / det
  t *= invdet

  if t < -atol(T)
    return (0.0,0.0,0.0,false)
  end

  return (r(t).coords..., true)
end

t  = Triangle(Point3(0.,0.,0.),Point3(1.,0.,0.),Point3(0.,1.,0.))
r1 = Ray((0.1,0.1,1.), (0.1,0.1,-1.))
r2 = Ray((0.1,0.1,1.), (-0.1,-0.1,1.))
@btime intersect_old(r1, t)
@btime intersect_new(r1 , t)
@btime intersect_old(r2, t)
@btime intersect_new(r2 , t)
```

It is currently producing

```
  29.179 ns (1 allocation: 48 bytes)
  23.766 ns (1 allocation: 48 bytes)
  17.747 ns (1 allocation: 48 bytes)
  21.533 ns (1 allocation: 48 bytes)
```

juliohm:
I will double check our Point constructors to see if we can avoid some of these allocations in the inputs anyways. Meanwhile, it would be nice to understand what is causing the difference between the low-level version:

```
  19.002 ns (0 allocations: 0 bytes)
  14.168 ns (0 allocations: 0 bytes)
  7.230 ns (0 allocations: 0 bytes)
  9.821 ns (0 allocations: 0 bytes)
```

and the high-level version with Meshes.jl types:

```
  29.179 ns (1 allocation: 48 bytes)
  23.766 ns (1 allocation: 48 bytes)
  17.747 ns (1 allocation: 48 bytes)
  21.533 ns (1 allocation: 48 bytes)
```

juliohm:
@HaoxuanGuo, we forgot to use interpolation in the `@btime` calls.

# Scripts

Below are the two updated scripts with low-level and high-level implementations:

## Low-level

```julia
using BenchmarkTools

import Base: -,*,+,/, intersect

struct Point
    x::Float64
    y::Float64
    z::Float64
end

*(p::Point, n::Number) = Point(p.x*n, p.y*n, p.z*n)
*(n::Number, p::Point) = p*n
*(p1::Point, p2::Point) = p1.x*p2.x+p1.y*p2.y+p1.z*p2.z
-(p1::Point, p2::Point) = Point(p1.x-p2.x, p1.y-p2.y, p1.z-p2.z)
+(p1::Point, p2::Point) = Point(p1.x+p2.x, p1.y+p2.y, p1.z+p2.z)
cross(p1::Point, p2::Point) = Point(p1.y*p2.z-p1.z*p2.y, -p1.x*p2.z+p1.z*p2.x, p1.x*p2.y-p1.y*p2.x)
dot(p1::Point, p2::Point) = p1.x*p2.x+p1.y*p2.y+p1.z*p2.z

/(p::Point, n::Number) = Point(p.x/n, p.y/n, p.z/n)
unitize(p::Point) = p/(p*p)

struct Triangle
    a::Point
    b::Point
    c::Point
end

struct Ray
    origin::Point
    direction::Point
    Ray(a,b) = new(a,unitize(b))
end

struct Intersection
    ip::Point # intersecting point
    id::Float64 # intersecting distance
    is_intersection::Bool
end

const no_intersection = Intersection(Point(0,0,0), 0.0, false)

function intersect_old(r::Ray, t::Triangle)
    # Prepare cache
    t_v1 = t.b-t.a
    t_v2 = t.c-t.a
    t_normal = cross(t_v1, t_v2)
    t_v1v1 = t_v1*t_v1
    t_v2v2 = t_v2*t_v2
    t_v1v2 = t_v1*t_v2
    t_denom = t_v1v2*t_v1v2 - t_v1v1*t_v2v2
    # Calculation
    denom = t_normal*r.direction
    denom == 0 && return no_intersection
    ri = t_normal*(t.a - r.origin) / denom
    ri <= 0 && return no_intersection
    plane_intersection =  ri * r.direction + r.origin
    w = plane_intersection - t.a
    wv1 = w*t_v1
    wv2 = w*t_v2
    s_intersection = (t_v1v2*wv2 - t_v2v2*wv1) / t_denom
    s_intersection <= 0 && return no_intersection
    s_intersection >= 1 && return no_intersection
    t_intersection = (t_v1v2*wv1 - t_v1v1*wv2) / t_denom
    t_intersection <= 0 && return no_intersection
    t_intersection >= 1 && return no_intersection
    s_intersection + t_intersection >= 1 && return no_intersection
    Intersection(t.a + s_intersection*t_v1+t_intersection*t_v2, ri, true)
end

function intersect_new(r::Ray, t::Triangle)
    origin = r.origin
    D = r.direction

    E₁ = t.c - t.a
    E₂ = t.b - t.a
    P = cross(D, E₂)
    det = dot(E₁, P)

    if det >= 0
        TT = origin - t.a
    else
        TT = t.a - origin
        det = -det
    end

    if det <= 0
        return no_intersection
    end

    u = dot(TT, P)
    if u <= 0 || u >= det
        return no_intersection
    end

    Q = cross(TT, E₁)

    v = dot(D, Q)
    if v <= 0 || u + v >= det
        return no_intersection
    end

    t = dot(E₂, Q)
    invdet = 1 / det
    t *= invdet

    if t <= 0
        return no_intersection
    end

    return Intersection(origin + t * D, t, true)
end


t  = Triangle(Point(0.,0.,0.), Point(1.,0.,0.), Point(0.,1.,0.))
r1 = Ray(Point(0.1,0.1,1.), Point(0.1,0.1,-1.))
r2 = Ray(Point(0.1,0.1,1.), Point(-0.1,-0.1,1.))
@btime intersect_old($r1, $t)
@btime intersect_new($r1, $t)
@btime intersect_old($r2, $t)
@btime intersect_new($r2, $t)
```

## High-level

```julia
using Pkg
Pkg.activate(@__DIR__)
using Meshes
using LinearAlgebra
using BenchmarkTools

function intersect_old(r::Ray{3,T}, t::Triangle{3,T}) where {T}
  # Prepare
  vₜ = vertices(t)
  t_v1 = vₜ[3] - vₜ[1]
  t_v2 = vₜ[2] - vₜ[1]
  t_normal = t_v1 × t_v2
  t_v1v1 = t_v1 ⋅ t_v1
  t_v2v2 = t_v2 ⋅ t_v2
  t_v1v2 = t_v1 ⋅ t_v2
  t_denom = t_v1v2 ⋅ t_v1v2 - t_v1v1 ⋅ t_v2v2
  # Calculation
  denom = t_normal ⋅ r.v
  denom == zero(T) && return (0.0,0.0,0.0,false)
  ri = t_normal ⋅ (vₜ[1] - r.p) / denom
  ri < zero(T) && return (0.0,0.0,0.0,false)
  plane_intersection = ri * r.v + r.p
  w = plane_intersection - vₜ[1]
  wv1 = w ⋅ t_v1
  wv2 = w ⋅ t_v2
  s_intersection = (t_v1v2 ⋅ wv2 - t_v2v2 ⋅ wv1) / t_denom
  s_intersection < zero(T) && return (0.0,0.0,0.0,false)
  s_intersection > one(T) && return (0.0,0.0,0.0,false)
  t_intersection = (t_v1v2 ⋅ wv1 - t_v1v1 ⋅ wv2) / t_denom
  t_intersection < zero(T) && return (0.0,0.0,0.0,false)
  t_intersection > one(T) && return (0.0,0.0,0.0,false)
  s_intersection + t_intersection > one(T) && return (0.0,0.0,0.0,false)
  (r(ri).coords..., true)
end

function intersect_new(r::Ray{3,T}, t::Triangle{3,T}) where {T}
  vₜ = vertices(t)
  origin = r.p
  D = r.v

  E₁ = vₜ[3] - vₜ[1]
  E₂ = vₜ[2] - vₜ[1]
  P = D × E₂
  det = E₁ ⋅ P

  if det > atol(T)
    TT = origin - vₜ[1]
  else
    TT = vₜ[1] - origin
    det = -det
  end

  if det < atol(T)
    # This ray is parallel to the plane of the triangle.
    return (0.0,0.0,0.0,false)
  end

  # calculate u parameter and test bounds
  u = TT ⋅ P
  if u < -atol(T) || u > det
    return (0.0,0.0,0.0,false)
  end

  Q = TT × E₁

  # calculate v parameter and test bounds
  v = D ⋅ Q
  if v < -atol(T) || u + v > det
    return (0.0,0.0,0.0,false)
  end

  t = E₂ ⋅ Q

  invdet = one(T) / det
  t *= invdet

  if t < -atol(T)
    return (0.0,0.0,0.0,false)
  end

  return (r(t).coords..., true)
end

t  = Triangle(Point(0.,0.,0.), Point(1.,0.,0.), Point(0.,1.,0.))
r1 = Ray(Point(0.1,0.1,1.), Vec(0.1,0.1,-1.))
r2 = Ray(Point(0.1,0.1,1.), Vec(-0.1,-0.1,1.))
@btime intersect_old($r1, $t)
@btime intersect_new($r1, $t)
@btime intersect_old($r2, $t)
@btime intersect_new($r2, $t)
```

# Timings

```bash
$ julia lowlevel.jl
  0.013 ns (0 allocations: 0 bytes)
  0.014 ns (0 allocations: 0 bytes)
  0.014 ns (0 allocations: 0 bytes)
  0.013 ns (0 allocations: 0 bytes)

$ julia highlevel.jl
  12.624 ns (0 allocations: 0 bytes)
  12.203 ns (0 allocations: 0 bytes)
  4.566 ns (0 allocations: 0 bytes)
  7.709 ns (0 allocations: 0 bytes)
```

Can you try to investigate what is happening now for the time difference?

GuoHaoxuan:
@juliohm It's hard to find. But I find something strange in the profiling.

In low level, about 2/3 of time was used to calculation. Others looks like function calling?

image

In high level, lots of time used to idling. And a `typeinf_ext_toplevel` strangle function appeared. I don't know what that is. But I think it looks bad.

image
image

juliohm:
I am wondering where this macro expansion is coming from. I can try to find the time to look into each implementation more carefully, comment line by line to see what is adding extra time. Maybe `StaticArrays.jl` is doing something magical, we need to check.

juliohm:
Another interesting fact is that the Moller implementation shows lower variance of time compared to the TriangleIntersect.jl one:

image

juliohm:
But of course these results are very hard to interpret given that we are in nanoseconds. I will investigate the code a bit further now.

src/intersections/triangles.jl
```diff
@@ -122,9 +122,70 @@ function intersecttype(f::Function, s::Segment{3,T}, t::Triangle{3,T}) where {T}
...
+## References
+
+* Möller, T. and Trumbore, B., 1997. [Fast, minimum storage ray-triangle
+  intersection. Journal of graphics tools]
+  (https://www.tandfonline.com/doi/abs/10.1080/10867651.1997.10487468)
+"""
+function intersecttype(f::Function, r::Ray{3,T}, t::Triangle{3,T}) where {T}
+  vₜ = vertices(t)
+  origin = r.p
+  D = r.v
```
juliohm: Let's try to call these vectors `o` and `d` in lowercase. We try to reserve uppercase for points.

src/intersections/triangles.jl
```diff
@@ -122,9 +122,70 @@ function intersecttype(f::Function, s::Segment{3,T}, t::Triangle{3,T}) where {T}
...
+and apply function `f` to it.
+
+## References
+
+* Möller, T. and Trumbore, B., 1997. [Fast, minimum storage ray-triangle
+  intersection. Journal of graphics tools]
+  (https://www.tandfonline.com/doi/abs/10.1080/10867651.1997.10487468)
+"""
+function intersecttype(f::Function, r::Ray{3,T}, t::Triangle{3,T}) where {T}
+  vₜ = vertices(t)
```
juliohm: Let's call this vector of vertices as `vs`

src/intersections/triangles.jl
```diff
@@ -122,9 +122,70 @@ function intersecttype(f::Function, s::Segment{3,T}, t::Triangle{3,T}) where {T}
...
+  origin = r.p
+  D = r.v
+
+  E₁ = vₜ[3] - vₜ[1]
+  E₂ = vₜ[2] - vₜ[1]
+  P = D × E₂
```
juliohm: vectors in lowercase

src/intersections/triangles.jl
```diff
@@ -122,9 +122,70 @@ function intersecttype(f::Function, s::Segment{3,T}, t::Triangle{3,T}) where {T}
...
+  det = E₁ ⋅ P
+
+  # keep det > 0, modify T accordingly
+  if det > atol(T)
+    TT = origin - vₜ[1]
```
juliohm: are you following the notation in the paper? we only use these syngle letters in algorithms when they follow the paper.

test/intersections.jl
```diff
@@ -149,110 +149,173 @@
...
     t = Triangle(P3(0, 0, 0), P3(1, 0, 0), P3(0, 1, 0))
 
     # intersects through t
     s = Segment(P3(0.2, 0.2, 1.0), P3(0.2, 0.2, -1.0))
     @test intersecttype(s, t) isa IntersectingSegmentTriangle
     @test s ∩ t == P3(0.2, 0.2, 0.0)
-    
+    r = Ray(P3(0.2, 0.2, 1.0), V3(0.0, 0.0, -1.0))
+    @test intersecttype(r, t) isa IntersectingRayTriangle
+    @test r ∩ t == P3(0.2, 0.2, 0.0)
```
juliohm: we should use \approx here

test/intersections.jl
```diff
@@ -149,110 +149,173 @@
...
+
+
     # intersects at a vertex of t
     s = Segment(P3(0.0, 0.0, 1.0), P3(0.0, 0.0, -1.0))
     @test intersecttype(s, t) isa IntersectingSegmentTriangle
     @test s ∩ t == P3(0.0, 0.0, 0.0)
-    
+    r = Ray(P3(0.0, 0.0, 1.0), V3(0.0, 0.0, -1.0))
+    @test intersecttype(r, t) isa IntersectingRayTriangle
+    @test r ∩ t == P3(0.0, 0.0, 0.0)
```
juliohm: \approx

juliohm (changes requested):
Please move the new tests with Ray to after the tests with Segment. That will help us adjust things in a single place. I left a few other suggestions of improvement.

src/intersections/triangles.jl
hx-guo: It should be called `T`. But `T` used for type, so.. I will try to change to another letter.

juliohm: The paper uses `T` for this variable then? Interesting :) We just need to clear things up for future readers. We either choose single letters when the notation follows the notation of the paper or more explicit names like `dir` and `orig` when the paper doesn't provide notation.

GuoHaoxuan:
@juliohm What about now? I doesn't how to call that vector. I used `t_` because `T` for type and `t` for triangle. Any ideas?

juliohm:
It is perfect @HaoxuanGuo , thanks for the adjustments. They pay off in the long run :)

Let me give another round of review, and then we wait a bit to see if anyone has any idea why using our types slows the algorithm a bit in https://github.com/JuliaGeometry/Meshes.jl/pull/205#issuecomment-968068666 (cc: @serenity4)

src/intersections/triangles.jl
```diff
@@ -36,7 +36,7 @@ function intersecttype(f::Function, s::Segment{3,T}, t::Triangle{3,T}) where {T}
 
     W₂ = A × D
     tᵥ = W₂ ⋅ C
-
```
juliohm, line 39: Please disable whitespace changes in your editor in future PRs so that we can concentrate on the actual meaningful changes. 👍🏽 🙂

test/intersections.jl
```diff
@@ -277,6 +277,121 @@
     s = Segment(P3(0.2, 0.2, 1.0), P3(0.2, 0.2, -1.0))
     t = Triangle(P3(0, 0, 0), P3(1, 0, 0), P3(0, 1, 0))
     @inferred someornone(s, t)
+
```
juliohm, line 280:
```suggestion
```

juliohm (approved):

src/intersections/triangles.jl
hx-guo: OK. Sorry for that.

juliohm:
Is it possible to construct an example of intersection between a segment and a triangle and another similar example between a ray and a triangle where the results are slightly different? It seems like the possible intersection types between these two are very similar? Basically, I am asking if this Moller's algorithm works with segments too, and if we could reuse a single algorithm for both cases.

GuoHaoxuan:
@juliohm In the tests, there are 2 test with same parameters for segment/triangle. But got different result. The only different is a ray has an infinite length but segment has a finite one. A segment may not reach the triangle while a ray will.

Obviously, moller's algorithm can be used on segment/triangle intersection. But @jackbeagley 's PR #163 is a more efficient algorithm for segment/triangle. Sadly, that code doesn't work well in ray. (because a ray is infinite, that PR uses another end point's info.)

juliohm:
Perfect @HaoxuanGuo 💯 that is what I wanted to hear. I will just wait a bit to see if anyone else wants to suggest an improvement and then I will merge. I am still curious why we are loosing performance when we switch to our Point, Ray, Triangle types.

If you still have time, it would be nice to randomize that benchmark above with some hundreds or thousands of triangles to see if Moller's vs TriangleIntersect.jl behave the same.

GuoHaoxuan:
@juliohm moller's algorithm is faster a little bit.

method | low level | high level
--|--|--
moller | 3.243 ms (2 allocations: 3.81 MiB) | 3.159 ms (2 allocations: 3.05 MiB)
TriangleIntersect | 3.431 ms (2 allocations: 3.81 MiB) | 3.352 ms (2 allocations: 3.05 MiB)

Another interesting things is there is no difference between our definitions and hand crafts in large amount calculation numbers.

## Scripts

### low level

```julia
using BenchmarkTools

import Base: -,*,+,/, intersect

struct Point
    x::Float64
    y::Float64
    z::Float64
end

*(p::Point, n::Number) = Point(p.x*n, p.y*n, p.z*n)
*(n::Number, p::Point) = p*n
*(p1::Point, p2::Point) = p1.x*p2.x+p1.y*p2.y+p1.z*p2.z
-(p1::Point, p2::Point) = Point(p1.x-p2.x, p1.y-p2.y, p1.z-p2.z)
+(p1::Point, p2::Point) = Point(p1.x+p2.x, p1.y+p2.y, p1.z+p2.z)
cross(p1::Point, p2::Point) = Point(p1.y*p2.z-p1.z*p2.y, -p1.x*p2.z+p1.z*p2.x, p1.x*p2.y-p1.y*p2.x)
dot(p1::Point, p2::Point) = p1.x*p2.x+p1.y*p2.y+p1.z*p2.z

/(p::Point, n::Number) = Point(p.x/n, p.y/n, p.z/n)
unitize(p::Point) = p/(p*p)

struct Triangle
    a::Point
    b::Point
    c::Point
end

struct Ray
    origin::Point
    direction::Point
    Ray(a,b) = new(a,unitize(b))
end

struct Intersection
    ip::Point # intersecting point
    id::Float64 # intersecting distance
    is_intersection::Bool
end

const no_intersection = Intersection(Point(0,0,0), 0.0, false)

function intersect_old(r::Ray, t::Triangle)
    # Prepare cache
    t_v1 = t.b-t.a
    t_v2 = t.c-t.a
    t_normal = cross(t_v1, t_v2)
    t_v1v1 = t_v1*t_v1
    t_v2v2 = t_v2*t_v2
    t_v1v2 = t_v1*t_v2
    t_denom = t_v1v2*t_v1v2 - t_v1v1*t_v2v2
    # Calculation
    denom = t_normal*r.direction
    denom == 0 && return no_intersection
    ri = t_normal*(t.a - r.origin) / denom
    ri <= 0 && return no_intersection
    plane_intersection =  ri * r.direction + r.origin
    w = plane_intersection - t.a
    wv1 = w*t_v1
    wv2 = w*t_v2
    s_intersection = (t_v1v2*wv2 - t_v2v2*wv1) / t_denom
    s_intersection <= 0 && return no_intersection
    s_intersection >= 1 && return no_intersection
    t_intersection = (t_v1v2*wv1 - t_v1v1*wv2) / t_denom
    t_intersection <= 0 && return no_intersection
    t_intersection >= 1 && return no_intersection
    s_intersection + t_intersection >= 1 && return no_intersection
    Intersection(t.a + s_intersection*t_v1+t_intersection*t_v2, ri, true)
end

function intersect_new(r::Ray, t::Triangle)
    origin = r.origin
    D = r.direction

    E₁ = t.c - t.a
    E₂ = t.b - t.a
    P = cross(D, E₂)
    det = dot(E₁, P)

    if det >= 0
        TT = origin - t.a
    else
        TT = t.a - origin
        det = -det
    end

    if det <= 0
        return no_intersection
    end

    u = dot(TT, P)
    if u <= 0 || u >= det
        return no_intersection
    end

    Q = cross(TT, E₁)

    v = dot(D, Q)
    if v <= 0 || u + v >= det
        return no_intersection
    end

    t = dot(E₂, Q)
    invdet = 1 / det
    t *= invdet

    if t <= 0
        return no_intersection
    end

    return Intersection(origin + t * D, t, true)
end


t  = [Triangle(Point(rand(),rand(),rand()), Point(rand(),rand(),rand()), Point(rand(),rand(),rand())) for i in 1:100000]
r  = [Ray(Point(rand(),rand(),rand()), Point(rand(),rand(),rand())) for i in 1:100000]

@btime intersect_old.($r, $t)
@btime intersect_new.($r, $t)
```

### high level

```julia
using Meshes
using Meshes: atol
using LinearAlgebra
using BenchmarkTools

function intersect_old(r::Ray{3,T}, t::Triangle{3,T}) where {T}
  # Prepare
  vₜ = vertices(t)
  t_v1 = vₜ[3] - vₜ[1]
  t_v2 = vₜ[2] - vₜ[1]
  t_normal = t_v1 × t_v2
  t_v1v1 = t_v1 ⋅ t_v1
  t_v2v2 = t_v2 ⋅ t_v2
  t_v1v2 = t_v1 ⋅ t_v2
  t_denom = t_v1v2 ⋅ t_v1v2 - t_v1v1 ⋅ t_v2v2
  # Calculation
  denom = t_normal ⋅ r.v
  denom == zero(T) && return (0.0,0.0,0.0,false)
  ri = t_normal ⋅ (vₜ[1] - r.p) / denom
  ri < zero(T) && return (0.0,0.0,0.0,false)
  plane_intersection = ri * r.v + r.p
  w = plane_intersection - vₜ[1]
  wv1 = w ⋅ t_v1
  wv2 = w ⋅ t_v2
  s_intersection = (t_v1v2 ⋅ wv2 - t_v2v2 ⋅ wv1) / t_denom
  s_intersection < zero(T) && return (0.0,0.0,0.0,false)
  s_intersection > one(T) && return (0.0,0.0,0.0,false)
  t_intersection = (t_v1v2 ⋅ wv1 - t_v1v1 ⋅ wv2) / t_denom
  t_intersection < zero(T) && return (0.0,0.0,0.0,false)
  t_intersection > one(T) && return (0.0,0.0,0.0,false)
  s_intersection + t_intersection > one(T) && return (0.0,0.0,0.0,false)
  (r(ri).coords..., true)
end

function intersect_new(r::Ray{3,T}, t::Triangle{3,T}) where {T}
  vₜ = vertices(t)
  origin = r.p
  D = r.v

  E₁ = vₜ[3] - vₜ[1]
  E₂ = vₜ[2] - vₜ[1]
  P = D × E₂
  det = E₁ ⋅ P

  if det > atol(T)
    TT = origin - vₜ[1]
  else
    TT = vₜ[1] - origin
    det = -det
  end

  if det < atol(T)
    # This ray is parallel to the plane of the triangle.
    return (0.0,0.0,0.0,false)
  end

  # calculate u parameter and test bounds
  u = TT ⋅ P
  if u < -atol(T) || u > det
    return (0.0,0.0,0.0,false)
  end

  Q = TT × E₁

  # calculate v parameter and test bounds
  v = D ⋅ Q
  if v < -atol(T) || u + v > det
    return (0.0,0.0,0.0,false)
  end

  t = E₂ ⋅ Q

  invdet = one(T) / det
  t *= invdet

  if t < -atol(T)
    return (0.0,0.0,0.0,false)
  end

  return (r(t).coords..., true)
end

t  = [Triangle(Point(rand(),rand(),rand()), Point(rand(),rand(),rand()), Point(rand(),rand(),rand())) for i in 1:100000]
r  = [Ray((rand(),rand(),rand()), (rand(),rand(),rand())) for i in 1:100000]

@btime intersect_old.($r, $t)
@btime intersect_new.($r, $t)

```

juliohm:
Very nice @HaoxuanGuo , I will just rename the variable `t_` by the Greek letter `\tau`, wait for the tests and merge 👌🏽 🙂

src/intersections/triangles.jl
```diff
@@ -122,9 +122,70 @@ function intersecttype(f::Function, s::Segment{3,T}, t::Triangle{3,T}) where {T}
...
+  t = e₂ ⋅ q
+
+  invdet = one(T) / det
+  t *= invdet
+
+  if t < -atol(T)
+    return NoIntersection() |> f
+  end
+
+  return IntersectingRayTriangle(r(t)) |> f
```
```suggestion
  if det > atol(T)
    τ = o - vs[1]
  else
    τ = vs[1] - o
    det = -det
  end

  if det < atol(T)
    # This ray is parallel to the plane of the triangle.
    return NoIntersection() |> f
  end

  # calculate u parameter and test bounds
  u = τ ⋅ p
  if u < -atol(T) || u > det
    return NoIntersection() |> f
  end

  q = τ × e₁

  # calculate v parameter and test bounds
  v = d ⋅ q
  if v < -atol(T) || u + v > det
    return NoIntersection() |> f
  end

  t = e₂ ⋅ q

  invdet = one(T) / det
  t *= invdet

  if t < -atol(T)
    return NoIntersection() |> f
  end

  return IntersectingRayTriangle(r(t)) |> f
```

juliohm:
@HaoxuanGuo I just noticed that you assign a variable `t` by the end of the function and that conflicts with the argument of the function. Can you please choose a different variable name?

GuoHaoxuan:
@juliohm `\tau` is a good name. It looks and pronounces similar with `t`. I rename the interpolation coefficient to `\lambda`. `\lambda` is a commonly used symbol to express interpolation coefficient in math.

src/intersections/triangles.jl
```diff
@@ -122,9 +122,70 @@ function intersecttype(f::Function, s::Segment{3,T}, t::Triangle{3,T}) where {T}
...
+  # calculate v parameter and test bounds
+  v = d ⋅ q
+  if v < -atol(T) || u + v > det
+    return NoIntersection() |> f
+  end
+
+  λ = e₂ ⋅ q
+
+  invdet = one(T) / det
+  λ *= invdet
```
```suggestion
  λ *= one(T) / det
```

src/intersections/triangles.jl
```diff
@@ -122,9 +122,69 @@ function intersecttype(f::Function, s::Segment{3,T}, t::Triangle{3,T}) where {T}
...
+
+  # calculate v parameter and test bounds
+  v = d ⋅ q
+  if v < -atol(T) || u + v > det
+    return NoIntersection() |> f
+  end
+
+  λ = e₂ ⋅ q
+
+  λ *= one(T) / det
```
```suggestion
  λ = (e₂ ⋅ q) * (one(T) / det)
```

juliohm:
Amazing work @HaoxuanGuo ! Thanks for another awesome improvement 💯 🎉

Issue #201, opened 2021-11-09
Can we expand a CartesianGrid adding vertices to the left?


ErickChacon:
Imagine that you define a CartesianGrid.
```julia
using Meshes
g = CartesianGrid(Point(10.0,), Point(20.0,), dims = (200,))
```
You do some operations, but now you need to do another operation with one additional vertex to the left. I can create the new grid as follows:
```julia
lag = 1
g2 = CartesianGrid(Point(10.0,) - spacing(g), Point(20.0,), dims = (200 + lag,))
```
In theory, there are 201 vertices that are common in both grids, but in practice they are not because of computational precision.
```julia
collect(vertices(g)) == collect(vertices(g2))[2:end]
# false
```
For my use case, this leads to problems. There is no problem if you add vertices to the right because you do not change the origin. One way to solve that problem is by not modifying the origin and defining an offset such as `origin = vertices[offset]` with a default value equal to `1` as defined in `StepRangeLen`. Here is an example:

```julia
# original range
r = StepRangeLen(10.0, 10/200, 201, 1)
# range with additional vertex value
lag = 1
r2 = StepRangeLen(10.0, 10/200, 201 + lag, 1 + lag)
# compare common vertices
collect(r) == collect(r2)[2:end]
# true
```
Currently, is there a way to expand a `CartesianGrid` ensuring the common vertices are computationally the same? If not, I think adding an `offset` argument and adding a `expand` method would be quite useful. This could be applied to the n-dimensional grid.

In case there is no interest in this feature, just close the issue, please.

juliohm:
You probably have good reasons for this feature in your use case. I am not familiar myself with the StepRangeLen mechanism that leads to robust numerical comparisons. If you can sketch a PR, we can see how much of the code would need to change in order to accommodate the feature.

Before starting the PR, can you please share examples of this API you have in mind for constructing grids with offsets? What changes compared to the current constructors?

ErickChacon:
Thanks for kindly receiving feature requests. The idea behind `StepRangeLen(ref::R, step::S, len, [offset=1])` is to return a range `r` such as the reference value `ref` is not necessarily the origin but it is equals to `r[offset]` and is used as a reference value to compute any other element of the range.  For example:

Using an `offset = 1` is the same as the usual range.
```julia
StepRangeLen(ref, step, len, [offset=1])
# ref, ref + step, ref + 2step, ref + step, ..., ref + (len - 1)step
```
[...]

PR #208, opened 2021-11-16
CartesianGrid with offset methods

ErickChacon:
This adds the feature requested in #201. It basically extends the `CartesianGrid` with fields `dims`, `reference`,
`spacing`, and `offset`. The reason we include `reference` and `offset` is to be able to compute all vertices with respect to `reference` such as the n-dimensional vertex `i` is computed as:
```
reference .+ (i .- offset) .* spacing
```

## What has changed?

Most of previous functionality have not changed for the default value `offset = ntuple(i->1, Dim)`. However, it differs from how subsetting is done. In this PR, subsetting is done using the same `reference`  for the subgrid, this ensures that the vertices are numerically the same between the original grid and the subgrid.

The ideas behind this extension come from `StepRangeLen`; however, we allow the offset to be lower than 1 (this is necessary for the functionality of subgrids). See the image below, for example, where the `offset` corresponding to the subgrid has elements lower than 1.

```julia
using Meshes
using Plots
g = CartesianGrid((10, 10), Point(0.0, 0.0), Vec(1.0, 1.0), Vec(0, 0))
plot(g, label = "grid")
plot!(g[2:5, 3:6], lc = :blue, label = "subgrid")
plot!(Point(0.0, 0.0), color = :red, label = "reference", legend = true)
(g[2:5, 3:6]).offset
# 2-element SVector{2, Int64} with indices SOneTo(2):
#   0
#   -1
```

subgrid

One current funny (maybe useful?) behavior of subgrids is that it is not restricted to the dimensions of the grid. I am not sure if this worked like that previously (it probably did). The interesting part is that way you are creating grids where the common vertices are exactly the same. Not sure if it is a desirable behavior, but happy to put restrictions on the `getindex` function.

```julia
using Meshes
using Plots
g = CartesianGrid((10, 10), Point(0.0, 0.0), Vec(1.0, 1.0), Vec(1, 1))
plot(g, label = "grid")
plot!(g[-1:12, 3:9], lc = :blue, label = "subgrid")
plot!(Point(0.0, 0.0), color = :red, label = "reference", legend = true)
```

subgrid2

## Additional notes:
- Using the term `origin` instead of `reference` might be also fine. Happy to put it back.
- I was not sure if we need to change the `show` method. For now, I was also printing `reference` to check the functionality of subgrids.

### DIFF
--- a/src/mesh/cartesiangrid.jl
+++ b/src/mesh/cartesiangrid.jl
@@ -54,7 +54,6 @@ struct CartesianGrid{Dim,T} <: Mesh{Dim,T}
   function CartesianGrid{Dim,T}(dims, reference, spacing, offset) where {Dim,T}
     @assert all(dims .> 0) "dimensions must be positive"
     @assert all(spacing .> 0) "spacing must be positive"
-    # @assert all(offset .> 0) "offset must be positive"
     new(dims, reference, spacing, offset)
   end
 end
@@ -229,6 +228,5 @@ function Base.show(io::IO, ::MIME"text/plain", g::CartesianGrid)
   println(io, g)
   println(io, "  minimum: ", minimum(g))
   println(io, "  maximum: ", maximum(g))
-  println(io, "  reference: ", g.reference)
   print(  io, "  spacing: ", Tuple(spacing(g)))
 end
--- a/src/mesh/cartesiangrid.jl
+++ b/src/mesh/cartesiangrid.jl
@@ -121,6 +121,7 @@ Base.minimum(g::CartesianGrid) = g.origin - (g.offset .- 1) .* g.spacing
 Base.maximum(g::CartesianGrid) = g.origin + (g.dims .- g.offset .+ 1) .* g.spacing
 Base.extrema(g::CartesianGrid) = minimum(g), maximum(g)
 spacing(g::CartesianGrid) = g.spacing
+offset(g::CartesianGrid) = g.offset
 
 function vertices(g::CartesianGrid)
   inds = CartesianIndices(g.dims .+ 1)
--- a/src/mesh/cartesiangrid.jl
+++ b/src/mesh/cartesiangrid.jl
@@ -60,18 +60,18 @@ end
 
 CartesianGrid(dims::Dims{Dim}, reference::Point{Dim,T},
               spacing::SVector{Dim,T},
-              offset::SVector{Dim,Int}=SVector(ntuple(i->1, Dim))) where {Dim,T} =
+              offset::SVector{Dim,Int}=SVector(ntuple(i->one(T), Dim))) where {Dim,T} =
   CartesianGrid{Dim,T}(dims, reference, spacing, offset)
 
 CartesianGrid(dims::Dims{Dim}, reference::NTuple{Dim,T},
               spacing::NTuple{Dim,T},
-              offset::NTuple{Dim,Int}=ntuple(i->1, Dim)) where {Dim,T} =
+              offset::NTuple{Dim,Int}=ntuple(i->one(T), Dim)) where {Dim,T} =
   CartesianGrid{Dim,T}(dims, Point(reference), SVector(spacing), SVector(offset))
 
 function CartesianGrid(start::Point{Dim,T}, finish::Point{Dim,T},
                        spacing::SVector{Dim,T}) where {Dim,T}
   dims = Tuple(ceil.(Int, (finish - start) ./ spacing))
-  CartesianGrid{Dim,T}(dims, start, spacing)
+  CartesianGrid{Dim,T}(dims, start, spacing, SVector(ntuple(i->one(T), Dim)))
 end
 
 CartesianGrid(start::NTuple{Dim,T}, finish::NTuple{Dim,T},
@@ -80,14 +80,14 @@ CartesianGrid(start::NTuple{Dim,T}, finish::NTuple{Dim,T},
 
 CartesianGrid(start::Point{Dim,T}, finish::Point{Dim,T};
               dims::Dims{Dim}=ntuple(i->100, Dim)) where {Dim,T} =
-  CartesianGrid{Dim,T}(dims, start, (finish - start) ./ dims)
+CartesianGrid{Dim,T}(dims, start, (finish - start) ./ dims, SVector(ntuple(i->one(T), Dim)))
 
 CartesianGrid(start::NTuple{Dim,T}, finish::NTuple{Dim,T};
               dims::Dims{Dim}=ntuple(i->100, Dim)) where {Dim,T} =
   CartesianGrid(Point(start), Point(finish); dims=dims)
 
 CartesianGrid{T}(dims::Dims{Dim}) where {Dim,T} =
-  CartesianGrid{Dim,T}(dims, ntuple(i->zero(T), Dim), ntuple(i->one(T), Dim))
+CartesianGrid{Dim,T}(dims, ntuple(i->zero(T), Dim), ntuple(i->one(T), Dim), ntuple(i->one(T), Dim))
 
 CartesianGrid{T}(dims::Vararg{Int,Dim}) where {Dim,T} = CartesianGrid{T}(dims)
 
--- a/src/mesh/cartesiangrid.jl
+++ b/src/mesh/cartesiangrid.jl
@@ -3,9 +3,9 @@
 # ------------------------------------------------------------------
 
 """
-    CartesianGrid(dims, origin, spacing)
+    CartesianGrid(dims, reference, spacing)
 
-A Cartesian grid with dimensions `dims`, lower left corner at `origin`
+A Cartesian grid with dimensions `dims`, lower left corner at `reference`
 and cell spacing `spacing`. The three arguments must have the same length.
 
     CartesianGrid(start, finish, dims=dims)
@@ -23,7 +23,7 @@ point using a given `spacing`.
 
 Finally, a Cartesian grid can be constructed by only passing the dimensions
 `dims` as a tuple, or by passing each dimension `dim1`, `dim2`, ... separately.
-In this case, the origin and spacing default to (0,0,...) and (1,1,...).
+In this case, the reference and spacing default to (0,0,...) and (1,1,...).
 
 ## Examples
 
@@ -33,7 +33,7 @@ Create a 3D grid with 100x100x50 locations:
 julia> CartesianGrid(100,100,50)
 ```
 
-Create a 2D grid with 100x100 locations and origin at (10.,20.) units:
+Create a 2D grid with 100x100 locations and reference at (10.,20.) units:
 
 ```julia
 julia> CartesianGrid((100,100),(10.,20.),(1.,1.))
@@ -47,44 +47,52 @@ julia> CartesianGrid((-1.,),(1.,), dims=(100,))
 """
 struct CartesianGrid{Dim,T} <: Mesh{Dim,T}
   dims::Dims{Dim}
-  origin::Point{Dim,T}
+  reference::Point{Dim,T}
   spacing::SVector{Dim,T}
+  offset::SVector{Dim,Int}
 
-  function CartesianGrid{Dim,T}(dims, origin, spacing) where {Dim,T}
+  function CartesianGrid{Dim,T}(dims, reference, spacing, offset) where {Dim,T}
     @assert all(dims .> 0) "dimensions must be positive"
     @assert all(spacing .> 0) "spacing must be positive"
-    new(dims, origin, spacing)
+    @assert all(offset .> 0) "offset must be positive"
+    new(dims, reference, spacing, offset)
   end
 end
 
-CartesianGrid(dims::Dims{Dim}, origin::Point{Dim,T},
-              spacing::SVector{Dim,T}) where {Dim,T} =
-  CartesianGrid{Dim,T}(dims, origin, spacing)
+CartesianGrid(dims::Dims{Dim}, reference::Point{Dim,T},
+              spacing::SVector{Dim,T},
+              offset::SVector{Dim,Int}=SVector(ntuple(i->1, Dim))) where {Dim,T} =
+  CartesianGrid{Dim,T}(dims, reference, spacing, offset)
 
-CartesianGrid(dims::Dims{Dim}, origin::NTuple{Dim,T},
-              spacing::NTuple{Dim,T}) where {Dim,T} =
-  CartesianGrid{Dim,T}(dims, Point(origin), SVector(spacing))
+CartesianGrid(dims::Dims{Dim}, reference::NTuple{Dim,T},
+              spacing::NTuple{Dim,T},
+              offset::NTuple{Dim,Int}=ntuple(i->1, Dim)) where {Dim,T} =
+  CartesianGrid{Dim,T}(dims, Point(reference), SVector(spacing), SVector(offset))
 
 function CartesianGrid(start::Point{Dim,T}, finish::Point{Dim,T},
-                       spacing::SVector{Dim,T}) where {Dim,T}
+                       spacing::SVector{Dim,T},
+                       offset::SVector{Dim,Int}=SVector(ntuple(i->1, Dim))) where {Dim,T}
   dims = Tuple(ceil.(Int, (finish - start) ./ spacing))
-  CartesianGrid{Dim,T}(dims, start, spacing)
+  reference = start + (offset .- 1) .* spacing
+  CartesianGrid{Dim,T}(dims, reference, spacing, offset)
 end
 
 CartesianGrid(start::NTuple{Dim,T}, finish::NTuple{Dim,T},
-              spacing::NTuple{Dim,T}) where {Dim,T} =
-  CartesianGrid(Point(start), Point(finish), SVector(spacing))
+              spacing::NTuple{Dim,T},
+              offset::NTuple{Dim,Int}=ntuple(i->1, Dim)) where {Dim,T} =
+  CartesianGrid(Point(start), Point(finish), SVector(spacing), SVector(offset))
 
 CartesianGrid(start::Point{Dim,T}, finish::Point{Dim,T};
               dims::Dims{Dim}=ntuple(i->100, Dim)) where {Dim,T} =
-  CartesianGrid{Dim,T}(dims, start, (finish - start) ./ dims)
+  CartesianGrid{Dim,T}(dims, start, (finish - start) ./ dims, SVector(ntuple(i->1, Dim)))
 
 CartesianGrid(start::NTuple{Dim,T}, finish::NTuple{Dim,T};
               dims::Dims{Dim}=ntuple(i->100, Dim)) where {Dim,T} =
   CartesianGrid(Point(start), Point(finish); dims=dims)
 
 CartesianGrid{T}(dims::Dims{Dim}) where {Dim,T} =
-  CartesianGrid{Dim,T}(dims, ntuple(i->zero(T), Dim), ntuple(i->one(T), Dim))
+  CartesianGrid{Dim,T}(dims, ntuple(i->zero(T), Dim), ntuple(i->one(T), Dim),
+                       ntuple(i->1, Dim))
 
 CartesianGrid{T}(dims::Vararg{Int,Dim}) where {Dim,T} = CartesianGrid{T}(dims)
 
@@ -94,18 +102,20 @@ CartesianGrid(dims::Vararg{Int,Dim}) where {Dim} = CartesianGrid{Float64}(dims)
 
 ==(g1::CartesianGrid, g2::CartesianGrid) =
   g1.dims    == g2.dims    &&
-  g1.origin  == g2.origin  &&
-  g1.spacing == g2.spacing
+  g1.reference  == g2.reference  &&
+  g1.spacing == g2.spacing &&
+  g1.offset == g2.offset
 
 Base.size(g::CartesianGrid) = g.dims
-Base.minimum(g::CartesianGrid) = g.origin
-Base.maximum(g::CartesianGrid) = g.origin + g.dims .* g.spacing
+Base.minimum(g::CartesianGrid) = g.reference - (g.offset .- 1) .* g.spacing
+Base.maximum(g::CartesianGrid) = g.reference + (g.dims .+ 1 .- g.offset) .* g.spacing
 Base.extrema(g::CartesianGrid) = minimum(g), maximum(g)
 spacing(g::CartesianGrid) = g.spacing
 
 function vertices(g::CartesianGrid)
-  inds = CartesianIndices(g.dims .+ 1)
-  vec([g.origin + (ind.I .- 1) .* g.spacing for ind in inds])
+  inds_ranges = map((dims,offset) -> (0:dims) .+ 1 .- offset, g.dims, g.offset.data)
+  inds = CartesianIndices(inds_ranges)
+  vec([g.reference + ind.I .* g.spacing for ind in inds])
 end
 
 elements(g::CartesianGrid) = (g[i] for i in 1:nelements(g))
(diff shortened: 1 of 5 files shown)


juliohm:
Thanks for preparing these examples with plots in the PR, they help a lot the review 🙏🏽 

Please let me know when the PR is ready for review and I will take a look. Currently it is failing some tests, and the code seems to contain commented lines you used during development.

ErickChacon:
The plots are useful for me too. There was only one commented line which is already removed. I will check the tests that are failing, it mainly seems to be related to subgrids (as I mentioned the way it works has changed). I will let you know when this PR is ready for review.

ErickChacon:
Hi @juliohm. I think this PR is ready for review. Please take into consideration the following notes.

As mentioned above, the main idea behind extending the `CartesianGrid` type is to be able to build grids where the vertices are comparable. We can achieve this by using a `reference` and `offset` argument such as the vertex associated to the `CartesianIndex` `i` is computed as:
```
reference .+ (i .- offset) .* spacing.
```

Before accepting this PR, there are some considerations we should agree on:

## Constructors

Right now, the more basic constructor uses the arguments `dims`, `origin` and `spacing`. Our extension uses `dims`, `reference`, `spacing` and `offset`. Initially, I was adding the `offset` argument to all type of constructors but I think that the user interested on this feature will want to work with the basic constructor because it allows control over `reference` and `offset`. Other constructors using `offset` are possible but that will be rarely used, so I am defining `offset=SVector(ntuple(i->1, Dim)` for other constructors.

## Offset values

The function `StepRangeLen` have the following restriction `1 ≤ offset ≤ n`. Putting a similar restriction for grids might not be adequate because it does not allow subsetting a grid keeping the same `reference` value.

See below some examples to understand the `offset` behaviour.

```julia
# CartesianGrid with default offset
g = CartesianGrid((10, 10), Point(0.0, 0.0), Vec(1.0, 1.0))
plot(g, label = "grid")
plot!(g.reference, c = :red, label = "reference", legend = true)
```

offset-default

```julia
# CartesianGrid with positive offset
g = CartesianGrid((10, 10), Point(0.0, 0.0), Vec(1.0, 1.0), Vec(4, 3))
plot(g, label = "grid")
plot!(g.reference, c = :red, label = "reference", legend = true)
```

offset-positive

```julia
# CartesianGrid with negative offset
g = CartesianGrid((10, 10), Point(0.0, 0.0), Vec(1.0, 1.0), Vec(-2, -1))
plot(g, label = "grid")
plot!(g.reference, c = :red, label = "reference", legend = true)
```

offset-negative

## Subsetting

As mentioned previously, I modified the subsetting behaviour in order to have subgrids where the vertices are comparable with the original behaviour. This basically means than the `reference` value should be the same for the original grid and the subgrid. The only arguments that should change are the `offset` and `dims` attributes. Is this a desirable approach or do we prefer the previous implementation approach where, in practice, the common vertices are not numerically the same?

The subsetting approach of this PR allows the new grid to not be limited to the domain of the original grid. Should we put a restriction? I did not put any restriction and the behaviour can be understood from the initial description of the PR.

## Equivalence

Currently, we consider two grids to be the same if the `dims`, `origin` and `spacing` attributes are the same. In theory, this is right but not in practice because the vertices will not be numerically the same because we are using different `origins`. For now, I am considering that two grids are equivalent if the `dims`, `reference`, `spacing` and `offset` are the same, this means that two grids will be equivalent if the vertices created will also be numerically the same. For this reason, I had to modify some tests in order to define the right `CartesianGrid` object to compare with. Again, is this the desired behaviour? Or do we prefer the previous behaviour where grids are still equivalent even if the vertices are not numerically the same?

See below an example of theoretically same grids, but in practice, those that use the same offset have the same vertices.

```julia
g = CartesianGrid((10,10), P2(0.0,0.0), V2(1.2,1.2))
sg1 = g[2:4, 5:7] 
sg2 = CartesianGrid((3,3), P2(0.0,0.0), V2(1.2,1.2), Vec(0,-3))
sg3 = CartesianGrid(P2(1.2,4.8), P2(4.8,8.4), dims = (3,3))
sg4 = CartesianGrid((3,3), P2(1.2,4.8), V2(1.2,1.2))

ind = vcat(reshape(1:121, 11, 11)[2:5, 5:8]...)
collect(vertices(g))[ind] == collect(vertices(sg1)) # true
collect(vertices(g))[ind] == collect(vertices(sg2)) # true
collect(vertices(g))[ind] == collect(vertices(sg3)) # false
collect(vertices(g))[ind] == collect(vertices(sg4)) # false

function myplot(gg; c = :black)
    pp = plot(gg, c = c)
    plot!(pp, collect(vertices(g))[ind], c = :red)
end

plot(
    myplot(sg1),
    myplot(sg2, c = 1),
    myplot(sg3, c = 2),
    myplot(sg4, c = 3)
   )
```

vertices-comparison

juliohm:
@ErickChacon I will try to find some time next week to review this. Really busy around here finishing up some projects, but I will come back to this at some point. 🙏🏽

juliohm:
@ErickChacon sorry for not having had the time to review your PR yet. I've released a minor version of the project with a few breaking changes and improvements and the plan is to incorporate your PR and fixes to #54 in the next minor release.

I will ping you in January so that we can work together on updating the PR to the latest master branch, and the latest release of the language.

ErickChacon:
Hi @juliohm. no problem at all. Happy to work on this in January.

juliohm:
Hi @ErickChacon any chance you can revive this PR in the following weeks? We are planning a next minor release and it would be nice to have this generalization you proposed.

juliohm:
Closing it as inactive. Feel free to start a fresh PR based on the latest master branch.

ErickChacon:
Hi @juliohm. Sorry for the late reply, I was a bit busy these days. Sure, I would be happy to revive this PR.  Can I get your feedback about the comments I did above with respect to *constructors*, *offset values*, *subsetting* and *equivalence* so I am sure how to proceed? Are you sure you want me to create another PR or just update this one? Cheers.

juliohm:
Hi @ErickChacon we can update this PR, no worries. I will take a look at the comments now.

src/mesh/cartesiangrid.jl
```diff
@@ -3,9 +3,9 @@
 # ------------------------------------------------------------------
 
 """
-    CartesianGrid(dims, origin, spacing)
+    CartesianGrid(dims, reference, spacing)
```
juliohm: Can we stick to the name origin to reduce the amount of changes in the docstrings and code? I believe the meaning of the variable is still the same right? You are only adding a new offset argument to the algorithms.

src/mesh/cartesiangrid.jl
```diff
@@ -47,28 +47,31 @@ julia> CartesianGrid((-1.,),(1.,), dims=(100,))
...
+CartesianGrid(dims::Dims{Dim}, reference::NTuple{Dim,T},
+              spacing::NTuple{Dim,T},
+              offset::NTuple{Dim,Int}=ntuple(i->1, Dim)) where {Dim,T} =
+  CartesianGrid{Dim,T}(dims, Point(reference), SVector(spacing), SVector(offset))
 
 function CartesianGrid(start::Point{Dim,T}, finish::Point{Dim,T},
                        spacing::SVector{Dim,T}) where {Dim,T}
   dims = Tuple(ceil.(Int, (finish - start) ./ spacing))
-  CartesianGrid{Dim,T}(dims, start, spacing)
+  CartesianGrid{Dim,T}(dims, start, spacing, SVector(ntuple(i->1, Dim)))
```
juliohm: Please define the offset variable outside the constructor for improved readability.

src/mesh/cartesiangrid.jl
```diff
@@ -77,14 +80,14 @@ CartesianGrid(start::NTuple{Dim,T}, finish::NTuple{Dim,T},
 
 CartesianGrid(start::Point{Dim,T}, finish::Point{Dim,T};
               dims::Dims{Dim}=ntuple(i->100, Dim)) where {Dim,T} =
-  CartesianGrid{Dim,T}(dims, start, (finish - start) ./ dims)
+  CartesianGrid{Dim,T}(dims, start, (finish - start) ./ dims, SVector(ntuple(i->1, Dim)))
```
juliohm: Please define offset variable outside constructor.

src/mesh/cartesiangrid.jl
```diff
@@ -77,14 +80,14 @@ CartesianGrid(start::NTuple{Dim,T}, finish::NTuple{Dim,T},
...
 
 CartesianGrid(start::NTuple{Dim,T}, finish::NTuple{Dim,T};
               dims::Dims{Dim}=ntuple(i->100, Dim)) where {Dim,T} =
   CartesianGrid(Point(start), Point(finish); dims=dims)
 
 CartesianGrid{T}(dims::Dims{Dim}) where {Dim,T} =
-  CartesianGrid{Dim,T}(dims, ntuple(i->zero(T), Dim), ntuple(i->one(T), Dim))
+  CartesianGrid{Dim,T}(dims, ntuple(i->zero(T), Dim), ntuple(i->one(T), Dim), ntuple(i->1, Dim))
```
juliohm: Same here.

juliohm (changes requested):
Hi @ErickChacon I appreciate if you can undo the `origin` -> `reference` change. This will clean up the PR and will highlight only the meaningful changes. The word `origin` is still appropriate if I understand your proposal correctly.

ErickChacon:
This is still work in progress. @juliohm, I will let you know when I think this is ready for review.

ErickChacon:
Hi @juliohm. This pull request is ready to be reviewed, I already made the modifications you asked for. Cheers.

test/partitioning.jl
```diff
@@ -390,6 +390,6 @@
 
     d = CartesianGrid{T}(10,10)
     s = slice(d, T(0.5):T(11), T(0.5):T(11))
-    @test s == CartesianGrid(P2(1,1), P2(10,10), dims=(9,9))
+    @test s == CartesianGrid((9,9), P2(0,0), V2(1,1), (0,0))
```
juliohm: Hi @ErickChacon before I start reviewing the PR, could you please explain why some of these old tests are being touched? I had understood that the addition of the new `offset` field should be backward compatible?

src/mesh/cartesiangrid.jl
```diff
@@ -49,42 +54,50 @@ struct CartesianGrid{Dim,T} <: Mesh{Dim,T}
   dims::Dims{Dim}
   origin::Point{Dim,T}
   spacing::SVector{Dim,T}
+  offset::NTuple{Dim,Int}
```
juliohm: Should this be offset::Dims{Dim} ?

juliohm:
Hi @ErickChacon appreciate if you can clarify some of the questions I left above. Also, could you please add new tests for the expected usage of the new offset field?

test/partitioning.jl
ErickChacon: Hi @juliohm. I forgot to change these tests back . Previously, they were failing because of the definition of `==`. However, I also extended this more accordingly with the use of offset. I will change back these old tests.

src/mesh/cartesiangrid.jl
ErickChacon: Agree with that.

src/mesh/cartesiangrid.jl
```diff
@@ -49,42 +54,50 @@ struct CartesianGrid{Dim,T} <: Mesh{Dim,T}
...
               spacing::NTuple{Dim,T}) where {Dim,T} =
   CartesianGrid(Point(start), Point(finish), SVector(spacing))
 
-CartesianGrid(start::Point{Dim,T}, finish::Point{Dim,T};
-              dims::Dims{Dim}=ntuple(i->100, Dim)) where {Dim,T} =
-  CartesianGrid{Dim,T}(dims, start, (finish - start) ./ dims)
+function CartesianGrid(start::Point{Dim,T}, finish::Point{Dim,T};
+                       dims::Dims{Dim}=ntuple(i->100, Dim)) where {Dim,T}
+  offset = ntuple(i->1, Dim)
+  CartesianGrid{Dim,T}(dims, start, (finish - start) ./ dims, offset)
```
juliohm:
```suggestion
  origin  = start
  spacing = (finish - start) ./ dims
  offset  = ntuple(i->1, Dim)
  CartesianGrid{Dim,T}(dims, origin, spacing, offset)
```

src/mesh/cartesiangrid.jl
```diff
@@ -49,42 +54,50 @@ struct CartesianGrid{Dim,T} <: Mesh{Dim,T}
...
 
 CartesianGrid(start::NTuple{Dim,T}, finish::NTuple{Dim,T};
               dims::Dims{Dim}=ntuple(i->100, Dim)) where {Dim,T} =
   CartesianGrid(Point(start), Point(finish); dims=dims)
 
-CartesianGrid{T}(dims::Dims{Dim}) where {Dim,T} =
-  CartesianGrid{Dim,T}(dims, ntuple(i->zero(T), Dim), ntuple(i->one(T), Dim))
+function CartesianGrid{T}(dims::Dims{Dim}) where {Dim,T}
+  offset = ntuple(i->1, Dim)
+  CartesianGrid{Dim,T}(dims, ntuple(i->zero(T), Dim), ntuple(i->one(T), Dim), offset)
```
juliohm:
```suggestion
  origin  = ntuple(i->zero(T), Dim)
  spacing = ntuple(i->one(T), Dim)
  offset  = ntuple(i->1, Dim)
  CartesianGrid{Dim,T}(dims, origin, spacing, offset)
```

src/mesh/cartesiangrid.jl
```diff
@@ -94,18 +107,19 @@ CartesianGrid(dims::Vararg{Int,Dim}) where {Dim} = CartesianGrid{Float64}(dims)
 
 ==(g1::CartesianGrid, g2::CartesianGrid) =
   g1.dims    == g2.dims    &&
-  g1.origin  == g2.origin  &&
-  g1.spacing == g2.spacing
+  g1.spacing == g2.spacing &&
+  g1.origin - g2.origin == (g1.offset .- g2.offset) .* g1.spacing
```
juliohm, line 116: Nice generalization 👍🏽

juliohm (changes requested):
Nice @ErickChacon , can you please add tests now for the new constructor?

Also, did you have a chance to benchmark this change against the master branch? For example, constructing a `grid = CartesianGrid(1000,1000)` and collecting the elements with collect(grid) would call the new expressions involving grid.offset

ErickChacon:
> Nice @ErickChacon , can you please add tests now for the new constructor?


I added tests for (i) the constructor,  (ii) subgrids with comparable vertices with respect to the original grid (which is a new feature thank to the offset behaviour), and (iii) expanding a CartesianGrid where the vertices are comparable. Feel free to suggest to remove the test (iii). In general, the inclusion of the `offset` allow us to create grids, subgrids or intersecting grids where the vertices and elements are numerically comparable.

> Also, did you have a chance to benchmark this change against the master branch? For example, constructing a `grid = CartesianGrid(1000,1000)` and collecting the elements with collect(grid) would call the new expressions involving grid.offset

I think this PR should not create additional computational cost because the master branch is using implicitly an `offset = (1,1)`.  In some cases, I just had to change the expression `.- 1` to `.- offset`. See below an benchmark example.

```julia
using BenchmarkTools
using Meshes

grid = CartesianGrid(1000,1000)
@benchmark a = collect(grid)

# CartessianGrid branch:
# BenchmarkTools.Trial: 16 samples with 1 evaluation.
#  Range (min … max):  161.654 ms … 404.918 ms  ┊ GC (min … max):  0.00% … 58.44%
#  Time  (median):     323.978 ms               ┊ GC (median):    49.38%
#  Time  (mean ± σ):   316.144 ms ±  50.887 ms  ┊ GC (mean ± σ):  47.98% ± 12.82%
#
#                                      █     ██
#   ▇▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▇▇█▇▁▁▇▁██▁▁▁▁▁▁▁▁▁▁▁▁▁▇▁▁▁▇ ▁
#   162 ms           Histogram: frequency by time          405 ms <
#
#  Memory estimate: 389.10 MiB, allocs estimate: 7000012.

# Master branch:
# BenchmarkTools.Trial: 16 samples with 1 evaluation.
#  Range (min … max):  161.936 ms … 418.074 ms  ┊ GC (min … max):  0.00% … 57.95%
#  Time  (median):     330.737 ms               ┊ GC (median):    50.60%
#  Time  (mean ± σ):   333.634 ms ±  57.050 ms  ┊ GC (mean ± σ):  49.27% ± 13.42%
#
#                                          █▃
#   ▇▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▇▁▁▁▁▇▁▁▁▇██▁▇▁▁▇▇▁▇▁▁▁▁▇▁▁▇▁▁▁▇ ▁
#   162 ms           Histogram: frequency by time          418 ms <
#
#  Memory estimate: 389.10 MiB, allocs estimate: 7000012.
#

```

src/mesh/cartesiangrid.jl
```diff
@@ -49,42 +54,54 @@ struct CartesianGrid{Dim,T} <: Mesh{Dim,T}
...
+              spacing::NTuple{Dim,T},
+              offset::Dims{Dim}=ntuple(i->1, Dim)) where {Dim,T} =
+  CartesianGrid{Dim,T}(dims, Point(origin), SVector(spacing), offset)
 
 function CartesianGrid(start::Point{Dim,T}, finish::Point{Dim,T},
                        spacing::SVector{Dim,T}) where {Dim,T}
   dims = Tuple(ceil.(Int, (finish - start) ./ spacing))
-  CartesianGrid{Dim,T}(dims, start, spacing)
+  offset = ntuple(i->1, Dim)
+  CartesianGrid{Dim,T}(dims, start, spacing, offset)
```
juliohm:
```suggestion
  origin = start
  offset = ntuple(i->1, Dim)
  CartesianGrid{Dim,T}(dims, origin, spacing, offset)
```

src/mesh/cartesiangrid.jl
```diff
@@ -197,10 +216,9 @@ Base.getindex(g::CartesianGrid{Dim}, r::Vararg{UnitRange{Int},Dim}) where {Dim}
   getindex(g, CartesianIndex(first.(r)):CartesianIndex(last.(r)))
 
 function Base.getindex(g::CartesianGrid{Dim}, I::CartesianIndices{Dim}) where {Dim}
-  start  = coordinates(g.origin) .+ (first(I).I .- 1) .* g.spacing
-  finish = coordinates(g.origin) .+ (last(I).I      ) .* g.spacing
+  newoffset = g.offset .- first(I).I .+ 1
```
juliohm:
```suggestion
  offset = g.offset .- first(I).I .+ 1
```

src/mesh/cartesiangrid.jl
```diff
@@ -197,10 +216,9 @@ Base.getindex(g::CartesianGrid{Dim}, r::Vararg{UnitRange{Int},Dim}) where {Dim}
...
   dims   = size(I)
-  CartesianGrid(Point(start), Point(finish), dims=dims)
+  CartesianGrid(dims, g.origin, g.spacing, newoffset)
```
juliohm:
```suggestion
  CartesianGrid(dims, g.origin, g.spacing, offset)
```

src/mesh/cartesiangrid.jl
```diff
@@ -197,10 +216,9 @@ Base.getindex(g::CartesianGrid{Dim}, r::Vararg{UnitRange{Int},Dim}) where {Dim}
   getindex(g, CartesianIndex(first.(r)):CartesianIndex(last.(r)))
 
 function Base.getindex(g::CartesianGrid{Dim}, I::CartesianIndices{Dim}) where {Dim}
-  start  = coordinates(g.origin) .+ (first(I).I .- 1) .* g.spacing
-  finish = coordinates(g.origin) .+ (last(I).I      ) .* g.spacing
+  offset = g.offset .- first(I).I .+ 1
   dims   = size(I)
```
juliohm:
```suggestion
  dims   = size(I)
  offset = g.offset .- first(I).I .+ 1
```

test/mesh.jl
```diff
@@ -110,6 +128,15 @@
     @test grid[1] == Quadrangle(P2[(0,0), (1,0), (1,1), (0,1)])
     @test grid[2] == Quadrangle(P2[(1,0), (2,0), (2,1), (1,1)])
 
+    # expand CartesianGrid with comparable vertices
+    grid = CartesianGrid((10,10), P2(0.0,0.0), V2(1.,1.))
+    left = (1,1)
+    right = (1,1)
+    newdim = grid.dims .+ left .+ right
+    newoffset = grid.offset .+ left
+    grid2 = CartesianGrid(newdim, grid.origin, grid.spacing, newoffset)
```
juliohm:
```suggestion
    grid2 = CartesianGrid(newdim, origin(grid), spacing(grid), newoffset)
```

test/mesh.jl
```diff
@@ -110,6 +128,15 @@
     @test grid[1] == Quadrangle(P2[(0,0), (1,0), (1,1), (0,1)])
     @test grid[2] == Quadrangle(P2[(1,0), (2,0), (2,1), (1,1)])
 
+    # expand CartesianGrid with comparable vertices
+    grid = CartesianGrid((10,10), P2(0.0,0.0), V2(1.,1.))
+    left = (1,1)
+    right = (1,1)
```
juliohm:
```suggestion
    left, right = (1,1), (1,1)
```

juliohm (approved):
That is really nice work @ErickChacon , thank you for the generalization. I did minor adjustments to the PR and will just wait the tests to pass.

test/mesh.jl
```diff
@@ -110,6 +128,14 @@
     @test grid[1] == Quadrangle(P2[(0,0), (1,0), (1,1), (0,1)])
     @test grid[2] == Quadrangle(P2[(1,0), (2,0), (2,1), (1,1)])
 
+    # expand CartesianGrid with comparable vertices
+    grid = CartesianGrid((10,10), P2(0.0,0.0), V2(1.,1.))
+    left, right = (1,1), (1,1)
+    newdim = grid.dims .+ left .+ right
+    newoffset = grid.offset .+ left
+    grid2 = CartesianGrid(newdim, origin(grid), spacing(grid), newoffset)
```
juliohm:
```suggestion
    grid2 = CartesianGrid(newdim, minimum(grid), spacing(grid), newoffset)
```

test/mesh.jl
```diff
@@ -110,6 +128,14 @@
     @test grid[1] == Quadrangle(P2[(0,0), (1,0), (1,1), (0,1)])
     @test grid[2] == Quadrangle(P2[(1,0), (2,0), (2,1), (1,1)])
 
+    # expand CartesianGrid with comparable vertices
+    grid = CartesianGrid((10,10), P2(0.0,0.0), V2(1.,1.))
+    left, right = (1,1), (1,1)
+    newdim = grid.dims .+ left .+ right
+    newoffset = grid.offset .+ left
```
juliohm:
```suggestion
    newdim = size(grid) .+ left .+ right
    newoffset = offset(grid) .+ left
```

src/mesh/cartesiangrid.jl
```diff
@@ -94,18 +112,19 @@ CartesianGrid(dims::Vararg{Int,Dim}) where {Dim} = CartesianGrid{Float64}(dims)
...
+  g1.origin - g2.origin == (g1.offset .- g2.offset) .* g1.spacing
+
 
 Base.size(g::CartesianGrid) = g.dims
-Base.minimum(g::CartesianGrid) = g.origin
-Base.maximum(g::CartesianGrid) = g.origin + g.dims .* g.spacing
+Base.minimum(g::CartesianGrid) = g.origin - (g.offset .- 1) .* g.spacing
+Base.maximum(g::CartesianGrid) = g.origin + (g.dims .- g.offset .+ 1) .* g.spacing
 Base.extrema(g::CartesianGrid) = minimum(g), maximum(g)
 spacing(g::CartesianGrid) = g.spacing
```
juliohm, line 123:
```suggestion
spacing(g::CartesianGrid) = g.spacing
offset(g::CartesianGrid) = g.offset
```

ErickChacon:
Hi @juliohm. Happy with the changes you did! Let me know in case we need something else.

PR #213, opened 2021-11-19
Add intersection between Quadrangles and rays/segments

HaoxuanGuo:
An extra step towards #125 . I tried to write `[intersect(ray, triangle) for triangle in mesh]` but it failed because there were some `Quadrangle`s. So, this PR is to fix that.


GuoHaoxuan:
Do not merge this! Here is a bug when the `Ngon` is a concave polygon. Sorry about that.

GuoHaoxuan:
At least `Quadrangle` should be considered. I wrote this condition and considered concave quadrangle.

src/intersections/quadrangle.jl
```diff
@@ -0,0 +1,33 @@
...
+# ------------------------------------------------------------------
+
+"""
+    intersecttype(f, segment_or_ray, quadrangle)
+
+Calculate the intersection type of `Segment` or `Ray` and `Quadrangle` and apply function
+`f` to it.
+"""
+function intersecttype(
+  f::Function, segment_or_ray::Union{Segment{3,T},Ray{3,T}}, n::Quadrangle{3,T}
```
juliohm, line 12: I think we need to organize this better. We don't want to write it in terms of a Union type. Also, I think the internal implementation should be expressed in terms of intersecttype so that the function `f` can be forwarded. More generally, I think we should consider and `Ngon` as input, triangulate this ngon with the existing triangulation methods, and check the intersection.

juliohm (changes requested):
Hi @HaoxuanGuo I think we need to write this down more carefully and more generically in terms of `Polygon` instead. Given a polygon, you can triangulate it, perform intersections and return the appropriate value. This could even be done in parallel with multiple threads in the sense that each triangle making up the polygon is checked indenpendently in a separate thread.

GuoHaoxuan:
Close due to #214 .

PR #215, opened 2021-11-20
intersection for ray and box

HaoxuanGuo:
Intersection for ray and box


src/Meshes.jl
```diff
@@ -305,6 +305,7 @@ export
   OverlappingBoxes,
   FaceTouchingBoxes,
   CornerTouchingBoxes,
+  RayCrossingBox,
```
juliohm: Can we rename this to `IntersectingRayBox` for consistency with other names?

src/intersections/boxes.jl
```diff
@@ -34,3 +34,30 @@ function intersecttype(f::Function, b1::Box{Dim,T}, b2::Box{Dim,T}) where {Dim,T
...
+Calculate the intersection type of a `ray`` and a `box`` and apply function `f` to it.
+
+## References
+
+* Williams A, Barrus S, Morley R K, et al., 2005. [An efficient and robust ray-box
+  intersection algorithm]
+  (https://dl.acm.org/doi/abs/10.1145/1198555.1198748)
+"""
+function intersecttype(f::Function, r::Ray{3,T}, b::Box{3,T}) where {T}
+  sign = (x -> x < 0 ? 2 : 1).(r.v)
```
juliohm: We cannot access internal fields of geometries in methods. We need to always use the public API to avoid breaking algorithms downstream. This also helps when we are refactoring things in general. Here you are assuming that Ray will have a field named `v`, and that may not be true in the future.

src/intersections/boxes.jl
```diff
@@ -34,3 +34,30 @@ function intersecttype(f::Function, b1::Box{Dim,T}, b2::Box{Dim,T}) where {Dim,T
...
+  invdir = one(T) ./ r.v
+  bounds = [b.min, b.max]
```
juliohm: If you can use tuple, they are statically allocated.

juliohm (changes requested):

juliohm:
Really cool additions 🚀

src/Meshes.jl
hx-guo: @juliohm  In fact, there may be another condition. `RayTouchingBox` contains a `Point` instead of a `Segment`. Use this name to reserve space for that possibility.

You may ask why have not write that condition. Because that is not stable catches due to floating point error. But maybe we got a great idea in the future?

src/intersections/boxes.jl
hx-guo: @juliohm  But where is no public API for that at all. 😭

src/Meshes.jl
juliohm: Nice, knowing that there are other cases not covered yet, it now makes more sense to choose a more specific name. What about `CrossingRayBox`? We place the verb first to match the other cases with segments an boxes.

Regarding the floating point issue, I was thinking we could at least leave the branch in the algorithm if that is possible. In the future we can come back with interval arithmetic to fix it.

GuoHaoxuan:
@juliohm Renamed `RayTouchingBox`. left the un-stable catches branch. But there is no public API for `r.v` for my changing.

src/intersections/boxes.jl
juliohm: Wow we added intersection algorithms for Ray intersection accessing the internal fields then? 😨 We need to fix that 🙏🏽 

Can you please open a PR with a public API for Ray, then fix the intersection algorithms to use the public API instead? I missed that during the reviews of the previous PRs.

hx-guo: @juliohm Added #217 .

GuoHaoxuan:
@juliohm Fixed tuple. Changed to public Ray APIs.

src/intersections/boxes.jl
```diff
@@ -34,3 +34,32 @@ function intersecttype(f::Function, b1::Box{Dim,T}, b2::Box{Dim,T}) where {Dim,T
...
+Calculate the intersection type of a `ray`` and a `box`` and apply function `f` to it.
+
+## References
+
+* Williams A, Barrus S, Morley R K, et al., 2005. [An efficient and robust ray-box
+  intersection algorithm]
+  (https://dl.acm.org/doi/abs/10.1145/1198555.1198748)
+"""
+function intersecttype(f::Function, r::Ray{3,T}, b::Box{3,T}) where {T}
+  sign = (x -> x < 0 ? 2 : 1).(direction(r))
```
juliohm: Should we use `Base.sign` instead? This lambda function isn't very easy to read at a first time.

src/intersections/boxes.jl
```diff
@@ -34,3 +34,32 @@ function intersecttype(f::Function, b1::Box{Dim,T}, b2::Box{Dim,T}) where {Dim,T
...
+  invdir = one(T) ./ direction(r)
+  bounds = (b.min, b.max)
+
+  tmin = maximum((bounds[sign[1]] - origin(r)) .* invdir)
+  tmin = max(tmin, zero(T))
+  tmax = minimum((bounds[3 - sign[1]] - origin(r)) .* invdir)
```
juliohm: Can you please add a comment explaining the idea behind these lines? They look a bit more complicated than they coul be I guess. If you can describe in words we can think in a better code to describe the idea.

src/intersections/boxes.jl
```diff
@@ -34,3 +34,32 @@ function intersecttype(f::Function, b1::Box{Dim,T}, b2::Box{Dim,T}) where {Dim,T
...
+
+  if tmin < tmax
+    return CrossingRayBox(Segment(r(tmin), r(tmax))) |> f
+  elseif tmin ≈ tmax
+    return TouchingRayBox(r(tmin)) |> f
```
juliohm:
```suggestion
    segment = Segment(r(tmin), r(tmax))
    return CrossingRayBox(segment) |> f
  elseif tmin ≈ tmax
    point = r(tmin)
    return TouchingRayBox(point) |> f
```

src/intersections/boxes.jl
hx-guo: @juliohm I guess you are confused on `bounds[sign]` and `bounds[3 .- sign]`. Someone told me this magic can get 1.5x faster than use comparation.

The original version looks like

```julia
if (invdir[1] >= 0)
  tmin = (max[1] - origin(r)[1]) * invdir[1]
  tmax = (max[1] - origin(r)[1]) * invdir[1]
else
  tmax = (max[1] - origin(r)[1]) * invdir[1]
  tmin = (max[1] - origin(r)[1]) * invdir[1]
end
```

Use the array access to avoid this comparation. This trick may hard to added in the code via comments.

Besides, I found I wrote a wrong code previously. I am trying to fix this.

GuoHaoxuan:
@juliohm I changed to a easy to read format and fix some problems.

You mentioned `Base.sign`. That is not fitted.
1. `Base.sign` defined on Reals. Here we want to use the sign as index, Integer preferred.
2. This sign is a little different. The value is 1 or 2 while `Base.sign` is 1 or -1.

GuoHaoxuan:
I found another bug when the ray is parallel a face of box. It will return a NaN.

GuoHaoxuan:
@juliohm I just returned a `NoIntersection` when the ray is on a face of box. (This condition may got a NaN.)

src/intersections/boxes.jl
```diff
@@ -34,3 +34,59 @@ function intersecttype(f::Function, b1::Box{Dim,T}, b2::Box{Dim,T}) where {Dim,T
...
+
+* Williams A, Barrus S, Morley R K, et al., 2005. [An efficient and robust ray-box
+  intersection algorithm]
+  (https://dl.acm.org/doi/abs/10.1145/1198555.1198748)
+"""
+function intersecttype(f::Function, r::Ray{3,T}, b::Box{3,T}) where {T}
+  sign = (x -> x < 0 ? 2 : 1).(direction(r))
+  invdir = one(T) ./ direction(r)
+  bounds = (b.min.coords, b.max.coords)
+  ori = origin(r).coords
```
juliohm: You are using internal fields again. Please use the `coordinates` function from the public api.

GuoHaoxuan:
@juliohm Fixed public APIs.

src/intersections/boxes.jl
```diff
@@ -34,3 +34,59 @@ function intersecttype(f::Function, b1::Box{Dim,T}, b2::Box{Dim,T}) where {Dim,T
...
+## References
+
+* Williams A, Barrus S, Morley R K, et al., 2005. [An efficient and robust ray-box
+  intersection algorithm]
+  (https://dl.acm.org/doi/abs/10.1145/1198555.1198748)
+"""
+function intersecttype(f::Function, r::Ray{3,T}, b::Box{3,T}) where {T}
+  sign = (x -> x < 0 ? 2 : 1).(direction(r))
+  invdir = one(T) ./ direction(r)
+  bounds = (coordinates(b.min), coordinates(b.max))
```
juliohm: Still using internal fields of structs. You want to call `extrema(b)` instead.

src/intersections/boxes.jl
```diff
@@ -34,3 +34,59 @@ function intersecttype(f::Function, b1::Box{Dim,T}, b2::Box{Dim,T}) where {Dim,T
...
+Calculate the intersection type of a `ray`` and a `box`` and apply function `f` to it.
+
+## References
+
+* Williams A, Barrus S, Morley R K, et al., 2005. [An efficient and robust ray-box
+  intersection algorithm]
+  (https://dl.acm.org/doi/abs/10.1145/1198555.1198748)
+"""
+function intersecttype(f::Function, r::Ray{3,T}, b::Box{3,T}) where {T}
+  sign = (x -> x < 0 ? 2 : 1).(direction(r))
```
juliohm: This variable is only used later. Please define variables near where they are used. Also, this syntax you are using is not aligned with our code style, please consider removing the lambda. Let's start with the version of the algorithm that is more natural, and then we can consider performance optimizations. You are mixing these two things.

src/intersections/boxes.jl
```diff
@@ -34,3 +34,59 @@ function intersecttype(f::Function, b1::Box{Dim,T}, b2::Box{Dim,T}) where {Dim,T
...
+
+  # Use index access to avoid unnecessary comparitions.
+  # Avoid the following code:
+  # if (invdir[1] >= 0)
+  #   tmin = (coordinates(b.min)[1] - ori[1]) * invdir[1]
+  #   tmax = (coordinates(b.max)[1] - ori[1]) * invdir[1]
+  # else
+  #   tmax = (coordinates(b.min)[1] - ori[1]) * invdir[1]
+  #   tmin = (coordinates(b.max)[1] - ori[1]) * invdir[1]
+  # end
```
juliohm: If the commented code is more intuitive, please use it instead. We can always come back to the optimization with sign, which may not be relevant after all.

juliohm (changes requested):
Review attached.

GuoHaoxuan:
@juliohm OK. I dropped all the magic optimization. Left the original intuitive code.

GuoHaoxuan:
@juliohm Sorry for the tests error. I forgot to `]activate .` before test. 😭 Now fixed.

src/intersections/boxes.jl
```diff
@@ -34,3 +34,71 @@ function intersecttype(f::Function, b1::Box{Dim,T}, b2::Box{Dim,T}) where {Dim,T
...
+  intersection algorithm]
+  (https://dl.acm.org/doi/abs/10.1145/1198555.1198748)
+"""
+function intersecttype(f::Function, r::Ray{3,T}, b::Box{3,T}) where {T}
+  invdir = one(T) ./ direction(r)
+  bounds = coordinates.(extrema(b))
+  orig = coordinates(origin(r))
+
+  # X slab
+  if invdir[1] >= 0
```
juliohm:
```suggestion
  if invdir[1] ≥ 0
```

src/intersections/boxes.jl
```diff
@@ -34,3 +34,71 @@ function intersecttype(f::Function, b1::Box{Dim,T}, b2::Box{Dim,T}) where {Dim,T
...
+    tmin = (bounds[1][1] - orig[1]) * invdir[1]
+    tmax = (bounds[2][1] - orig[1]) * invdir[1]
+  else
+    tmin = (bounds[2][1] - orig[1]) * invdir[1]
+    tmax = (bounds[1][1] - orig[1]) * invdir[1]
+  end
+  tmin = max.(tmin, zero(T))
+
+  # Y slab
+  if invdir[2] >= 0
```
juliohm:
```suggestion
  if invdir[2] ≥ 0
```

src/intersections/boxes.jl
```diff
@@ -34,3 +34,71 @@ function intersecttype(f::Function, b1::Box{Dim,T}, b2::Box{Dim,T}) where {Dim,T
...
+"""
+function intersecttype(f::Function, r::Ray{3,T}, b::Box{3,T}) where {T}
+  invdir = one(T) ./ direction(r)
+  bounds = coordinates.(extrema(b))
+  orig = coordinates(origin(r))
+
+  # X slab
+  if invdir[1] ≥ 0
+    tmin = (bounds[1][1] - orig[1]) * invdir[1]
+    tmax = (bounds[2][1] - orig[1]) * invdir[1]
```
juliohm: Instead of using `bounds[1]` and `bounds[2]` in various places, can you please define two separate variables with the bottom left and top right coordinates? For example, the following definitions in the beginning of the function will improve the readability by a lot:

```julia
lo, up = coordinates.(extrema(b))
```

src/intersections/boxes.jl
```diff
@@ -34,3 +34,71 @@ function intersecttype(f::Function, b1::Box{Dim,T}, b2::Box{Dim,T}) where {Dim,T
...
+  else
+    tmin = (bounds[2][1] - orig[1]) * invdir[1]
+    tmax = (bounds[1][1] - orig[1]) * invdir[1]
+  end
+  tmin = max.(tmin, zero(T))
```
juliohm: I don't think we need broadcasting here?

src/intersections/boxes.jl
```diff
@@ -34,3 +34,71 @@ function intersecttype(f::Function, b1::Box{Dim,T}, b2::Box{Dim,T}) where {Dim,T
...
+
+  # Y slab
+  if invdir[2] ≥ 0
+    tymin = (bounds[1][2] - orig[2]) * invdir[2]
+    tymax = (bounds[2][2] - orig[2]) * invdir[2]
+  else
+    tymin = (bounds[2][2] - orig[2]) * invdir[2]
+    tymax = (bounds[1][2] - orig[2]) * invdir[2]
+  end
+  if tmin > tymax || tymin > tmax
```
juliohm: So you called the x variable `tmin` and the y variable `tymin`? See the inconsistency? It makes reviewing code a lot harder. Can you please follow the same naming convention for x and y?

juliohm (changes requested):
Review attached.

src/intersections/boxes.jl
hx-guo: Yes. We need not. That's a mistake.

src/intersections/boxes.jl
hx-guo: @juliohm Here `tmin` is the maximum of `txmin`, `tymin` and `tzmin`. It is not similar with `tymin`. I use `tmin` because it covers all 3 dimensions. One in X coordination is calculated first. There is nothing before that to be compared. It is need not to write `tmin = txmin = ...`; only `tmin = ...` is enough. So here is `tmin` instead of `txmin`.

juliohm: In any case, we need to make this logic more explicit by introducing variable names that help the reader of the code understand the logic. I understand there is a variable that you compute in each dimension and a global variable that you update as you process the dimensions. These 3 variables + 1 global variable could be more explicit. You can initialize the global variable with dummy values such as `-Inf, Inf` and then update based on the results per dimension.

GuoHaoxuan:
@juliohm I changed it to a more logical format.

GuoHaoxuan:
Maybe typemax could got an Inf? I will try it tomorrow.

Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

GuoHaoxuan:
@juliohm `typemax` works well. Now we get some more natural code.

src/intersections/boxes.jl
```diff
@@ -34,3 +34,51 @@ function intersecttype(f::Function, b1::Box{Dim,T}, b2::Box{Dim,T}) where {Dim,T
...
+function intersecttype(f::Function, r::Ray{3,T}, b::Box{3,T}) where {T}
+  invdir = one(T) ./ direction(r)
+  lo, up = coordinates.(extrema(b))
+  orig = coordinates(origin(r))
+
+  tmin = zero(T)
+  tmax = typemax(T)
+
+  for axis in 1:3
+    if invdir[axis] ≥ 0
```
juliohm:
```suggestion
    if invdir[axis] ≥ zero(T)
```

src/intersections/boxes.jl
```diff
@@ -34,3 +34,51 @@ function intersecttype(f::Function, b1::Box{Dim,T}, b2::Box{Dim,T}) where {Dim,T
...
+    else
+      taxismin = (up[axis] - orig[axis]) * invdir[axis]
+      taxismax = (lo[axis] - orig[axis]) * invdir[axis]
+    end
+    if tmin > taxismax || taxismin > tmax
+      return NoIntersection() |> f
+    end
+    tmin = max(tmin, taxismin)
+    tmax = min(tmax, taxismax)
+  end
```
juliohm:
```suggestion
  for i in 1:3
    if invdir[i] ≥ zero(T)
      taxismin = (lo[i] - orig[i]) * invdir[i]
      taxismax = (up[i] - orig[i]) * invdir[i]
    else
      taxismin = (up[i] - orig[i]) * invdir[i]
      taxismax = (lo[i] - orig[i]) * invdir[i]
    end
    if tmin > taxismax || taxismin > tmax
      return NoIntersection() |> f
    end
    tmin = max(tmin, taxismin)
    tmax = min(tmax, taxismax)
  end
```

src/intersections/boxes.jl
```diff
@@ -34,3 +34,51 @@ function intersecttype(f::Function, b1::Box{Dim,T}, b2::Box{Dim,T}) where {Dim,T
...
+"""
+function intersecttype(f::Function, r::Ray{3,T}, b::Box{3,T}) where {T}
+  invdir = one(T) ./ direction(r)
+  lo, up = coordinates.(extrema(b))
+  orig = coordinates(origin(r))
+
+  tmin = zero(T)
+  tmax = typemax(T)
+
+  for i in 1:3
```
juliohm: Ok, now that the algorithm is becoming clear, do we have anything special about Dim=3? Isn't this algorithm N-dimensional?

juliohm:
@HaoxuanGuo can you please describe in simple words what this algorithm is doing? It seems very similar to the algorithms we have for Box intersections.

src/intersections/boxes.jl
hx-guo: @juliohm Yes. You are right. It is adapt for any N-dimensional.

GuoHaoxuan:
@juliohm Calculate the intersection for a box is difficult. But a box can be seems as the intersection for serval slabs. It is easy to calculate the intersection for ray and a pair of slab. Thus, compute all the intersections for slabs and get lots of segments. Then, get the intersection for these segments which is the final results.

Looks like:
```julia
slab_pairs::Vector{SlabPair} = box::Box
segments::Vector{Segment} = ray::Ray .\cap slab_pairs
final_result = \cap(segments...)
```

src/intersections/boxes.jl
```diff
@@ -34,3 +34,51 @@ function intersecttype(f::Function, b1::Box{Dim,T}, b2::Box{Dim,T}) where {Dim,T
...
+    else
+      taxismin = (up[i] - orig[i]) * invdir[i]
+      taxismax = (lo[i] - orig[i]) * invdir[i]
+    end
+    if tmin > taxismax || taxismin > tmax
+      return NoIntersection() |> f
+    end
+    tmin = max(tmin, taxismin)
+    tmax = min(tmax, taxismax)
+  end
```
juliohm:
```suggestion
  for i in 1:Dim
    if invdir[i] ≥ zero(T)
      imin = (lo[i] - orig[i]) * invdir[i]
      imax = (up[i] - orig[i]) * invdir[i]
    else
      imin = (up[i] - orig[i]) * invdir[i]
      imax = (lo[i] - orig[i]) * invdir[i]
    end
    if tmin > imax || imin > tmax
      return NoIntersection() |> f
    end
    tmin = max(tmin, imin)
    tmax = min(tmax, imax)
  end
```

src/intersections/boxes.jl
```diff
@@ -34,3 +34,51 @@ function intersecttype(f::Function, b1::Box{Dim,T}, b2::Box{Dim,T}) where {Dim,T
...
+  intersection algorithm]
+  (https://dl.acm.org/doi/abs/10.1145/1198555.1198748)
+"""
+function intersecttype(f::Function, r::Ray{Dim,T}, b::Box{Dim,T}) where {Dim,T}
+  invdir = one(T) ./ direction(r)
+  lo, up = coordinates.(extrema(b))
+  orig = coordinates(origin(r))
+
+  tmin = zero(T)
+  tmax = typemax(T)
```
juliohm, line 55: I find this asymmetry mathematically unpleasant. Are you sure that the minimum should be initialized as zero and the maximum as typemax? Shouldn't the minimum be initialized as typemin? Can you please justify this discrepancy?

src/intersections/boxes.jl
```diff
@@ -34,3 +34,51 @@ function intersecttype(f::Function, b1::Box{Dim,T}, b2::Box{Dim,T}) where {Dim,T
...
+
+  for i in 1:Dim
```
juliohm, line 58: Now that you explained the algorithm in simple words, this same explanation should be added here before the loop. A simple short sentence explaining that you are taking intersection with slabs would help the readers tremendously.

src/intersections/boxes.jl
```diff
@@ -34,3 +34,51 @@ function intersecttype(f::Function, b1::Box{Dim,T}, b2::Box{Dim,T}) where {Dim,T
...
+    if invdir[i] ≥ zero(T)
+      imin = (lo[i] - orig[i]) * invdir[i]
+      imax = (up[i] - orig[i]) * invdir[i]
+    else
+      imin = (up[i] - orig[i]) * invdir[i]
+      imax = (lo[i] - orig[i]) * invdir[i]
+    end
```
juliohm: Notice that now that the code is clean we can see the logic here as "swap imin and imax if the sign of invdir is negative". I would erase this code repetition and would write it as:

```julia
for i in 1:Dim
  imin = (lo[i] - orig[i]) * invdir[i]
  imax = (up[i] - orig[i]) * invdir[i]
  
  # swap if necessary
  invdir[i] < zero(T) && ((imin, imax) = (imax, imin))

  # continue the code...
```

juliohm (changes requested):
More suggestions attached.

src/intersections/boxes.jl
hx-guo: @juliohm Because a ray is asymmetry. A ray can not extend on one direction (follows it's direction vector). Here t describes the distance between the intersection and the starting point. The minimum distance is zero instead of -Inf. If I use typemin, it means allowing the ray to extend in the reverse direction.

GuoHaoxuan:
@juliohm Changing applied.

src/intersections/boxes.jl
```diff
@@ -34,3 +34,49 @@ function intersecttype(f::Function, b1::Box{Dim,T}, b2::Box{Dim,T}) where {Dim,T
...
+
+  # check for intersection with slabs along with each axis
+  for i in 1:Dim
+    imin = (lo[i] - orig[i]) * invdir[i]
+    imax = (up[i] - orig[i]) * invdir[i]
+    # swap if necessary
+    invdir[i] < zero(T) && ((imin, imax) = (imax, imin))
+    if tmin > imax || imin > tmax
+      return NoIntersection() |> f
+    end
```
juliohm:
```suggestion

    # swap if necessary
    invdir[i] < zero(T) && ((imin, imax) = (imax, imin))
    
    if tmin > imax || imin > tmax
      return NoIntersection() |> f
    end
```

src/intersections/boxes.jl
```diff
@@ -34,3 +34,51 @@ function intersecttype(f::Function, b1::Box{Dim,T}, b2::Box{Dim,T}) where {Dim,T
...
+  for i in 1:Dim
+    imin = (lo[i] - orig[i]) * invdir[i]
+    imax = (up[i] - orig[i]) * invdir[i]
+
+    # swap if necessary
+    invdir[i] < zero(T) && ((imin, imax) = (imax, imin))
+    
+    if tmin > imax || imin > tmax
+      return NoIntersection() |> f
+    end
```
juliohm:
```suggestion
    (tmin > imax || imin > tmax) && return NoIntersection() |> f
    
```

src/intersections/boxes.jl
```diff
@@ -34,3 +34,50 @@ function intersecttype(f::Function, b1::Box{Dim,T}, b2::Box{Dim,T}) where {Dim,T
...
+  # Touching
+  if tmin ≈ tmax
+    point = r(tmin)
+    return TouchingRayBox(point) |> f
+  # Crossing
+  elseif tmin < tmax
+    segment = Segment(r(tmin), r(tmax))
+    return CrossingRayBox(segment) |> f
+  # May got NaN
+  else
```
juliohm:
```suggestion
  if tmin ≈ tmax
    point = r(tmin)
    return TouchingRayBox(point) |> f
  elseif tmin < tmax
    segment = Segment(r(tmin), r(tmax))
    return CrossingRayBox(segment) |> f
  else
```

src/intersections/boxes.jl
```diff
@@ -34,3 +34,47 @@ function intersecttype(f::Function, b1::Box{Dim,T}, b2::Box{Dim,T}) where {Dim,T
...
+    tmin = max(tmin, imin)
+    tmax = min(tmax, imax)
+  end
+
+  if tmin ≈ tmax
+    point = r(tmin)
+    return TouchingRayBox(point) |> f
+  elseif tmin < tmax
+    segment = Segment(r(tmin), r(tmax))
+    return CrossingRayBox(segment) |> f
```
juliohm:
```suggestion
    return TouchingRayBox(r(tmin)) |> f
  elseif tmin < tmax
    return CrossingRayBox(Segment(r(tmin), r(tmax))) |> f
```

src/intersections/boxes.jl
```diff
@@ -34,3 +34,47 @@ function intersecttype(f::Function, b1::Box{Dim,T}, b2::Box{Dim,T}) where {Dim,T
...
+  else
```
juliohm: Hi @HaoxuanGuo can you please comment on this last branch? You had a comment that said "May got NaN", why is that? Also what happens when the `invdir` explodes to `Inf` because one of the coordinates is zero?

juliohm: Found this comment in the cited paper:

image

juliohm: BTW very interesting to find out a paper specifically for this intersection type.

hx-guo: @juliohm This is a very special situation - the ray is on a face of the box. In fact, it is just no intersection but with a special branch condition.

For example, this code will hit the last branch.

```julia
using Meshes
r = Ray((1.5,0,0), (-1,1,0.0))
b = Box((0,0,0), (1,1,1))
b ∩ r
```

juliohm: Isn't this case an intersection? I mean strictly speaking the set of points is non empty in this case and it forms a new ray, right?

GuoHaoxuan:
@juliohm You are right. That is a intersection. Now, it is considered.

src/intersections/boxes.jl
```diff
@@ -34,3 +34,48 @@ function intersecttype(f::Function, b1::Box{Dim,T}, b2::Box{Dim,T}) where {Dim,T
...
+    imin = (lo[i] - orig[i]) * invdir[i]
+    imax = (up[i] - orig[i]) * invdir[i]
+
+    # swap if necessary
+    invdir[i] < zero(T) && ((imin, imax) = (imax, imin))
+
+    # the ray is on a face of the box, avoid NaN
+    if isnan(imin) || isnan(imax)
+      continue
+    end
```
juliohm:
```suggestion
    (isnan(imin) || isnan(imax)) && continue
```

src/intersections/boxes.jl
```diff
@@ -34,3 +34,46 @@ function intersecttype(f::Function, b1::Box{Dim,T}, b2::Box{Dim,T}) where {Dim,T
...
+
+  tmin = zero(T)
+  tmax = typemax(T)
+
+  # check for intersection with slabs along with each axis
+  for i in 1:Dim
+    imin = (lo[i] - orig[i]) * invdir[i]
+    imax = (up[i] - orig[i]) * invdir[i]
+
+    # swap if necessary
```
juliohm:
```suggestion
    # swap variables if necessary
```

src/intersections/boxes.jl
```diff
@@ -34,3 +34,46 @@ function intersecttype(f::Function, b1::Box{Dim,T}, b2::Box{Dim,T}) where {Dim,T
...
+    
+    tmin = max(tmin, imin)
+    tmax = min(tmax, imax)
+  end
+
+  if tmin ≈ tmax
+    return TouchingRayBox(r(tmin)) |> f
+  elseif tmin < tmax
+    return CrossingRayBox(Segment(r(tmin), r(tmax))) |> f
+  end
```
juliohm: Can you please double check the return type of the function with `@code_warntype`? Now that we don't have a final `else` clause, Julia may be returning a `Union{Intersection,Nothing}` type. We should probably add a dummy `else` clause with `NoIntersection` as before even after the special intersection fix.

hx-guo: @juliohm It is impossible to mis-hit on the elseif because we checked NaN in previous code. Now, I changed the code format to avoid this elseif.

juliohm: And the result of `@code_warntype` is ok? I think we can merge after that 👍🏽

src/intersections/boxes.jl
```diff
@@ -34,3 +34,43 @@ function intersecttype(f::Function, b1::Box{Dim,T}, b2::Box{Dim,T}) where {Dim,T
...
+    # the ray is on a face of the box, avoid NaN
+    (isnan(imin) || isnan(imax)) && continue
+
+    (tmin > imax || imin > tmax) && return NoIntersection() |> f
+    
+    tmin = max(tmin, imin)
+    tmax = min(tmax, imax)
+  end
+
+  tmin ≈ tmax && return TouchingRayBox(r(tmin)) |> f
```
juliohm, line 74:
```suggestion
  tmin ≈ tmax && return TouchingRayBox(r(tmin)) |> f
  
```

test/intersections.jl
```diff
@@ -477,6 +477,29 @@
     b1 = Box(P2(0,0), P2(1,1))
     b2 = Box(P2(0.5,0.5), P2(2,2))
     @inferred someornone(b1, b2)
+
+    # Ray-Box intersection
+    b = Box(P3(0,0,0), P3(1,1,1))
+
+    r = Ray(P3(0,0,0), V3(1.0,1.0,1.0))
```
```suggestion
    r = Ray(P3(0,0,0), V3(1,1,1))
```

src/intersections/boxes.jl
hx-guo: @juliohm 

1 red line:

```julia
Body::Union{NoIntersection, CrossingRayBox{Segment{3, Float64, StaticArrays.SVector{2, Point3}}}, TouchingRayBox{Point3}}
```

1 yellow line:

```julia
@_5::Union{Nothing, Tuple{Int64, Int64}}
```

full code:

```julia
Variables
  #self#::Core.Const(Meshes.intersecttype)
  f::Core.Const(identity)
  r::Ray{3, Float64}
  b::Box{3, Float64}
  @_5::Union{Nothing, Tuple{Int64, Int64}}
  @_6::Int64
  tmax::Float64
  tmin::Float64
  orig::StaticArrays.SVector{3, Float64}
  up::StaticArrays.SVector{3, Float64}
  lo::StaticArrays.SVector{3, Float64}
  invdir::StaticArrays.SVector{3, Float64}
  i::Int64
  imax::Float64
  imin::Float64

Body::Union{NoIntersection, CrossingRayBox{Segment{3, Float64, StaticArrays.SVector{2, Point3}}}, TouchingRayBox{Point3}}
1 ── %1  = Meshes.one($(Expr(:static_parameter, 2)))::Core.Const(1.0)
│    %2  = Meshes.direction(r)::StaticArrays.SVector{3, Float64}
│    %3  = Base.broadcasted(Meshes.:/, %1, %2)::Core.PartialStruct(Base.Broadcast.Broadcasted{StaticArrays.StaticArrayStyle{1}, Nothing, typeof(/), Tuple{Float64, StaticArrays.SVector{3, Float64}}}, Any[Core.Const(/), Core.PartialStruct(Tuple{Float64, StaticArrays.SVector{3, Float64}}, Any[Core.Const(1.0), StaticArrays.SVector{3, Float64}]), Core.Const(nothing)])
│          (invdir = Base.materialize(%3))
│    %5  = Meshes.extrema(b)::Tuple{Point3, Point3}
│    %6  = Base.broadcasted(Meshes.coordinates, %5)::Base.Broadcast.Broadcasted{Base.Broadcast.Style{Tuple}, Nothing, typeof(coordinates), Tuple{Tuple{Point3, Point3}}}
│    %7  = Base.materialize(%6)::Tuple{StaticArrays.SVector{3, Float64}, StaticArrays.SVector{3, Float64}}
│    %8  = Base.indexed_iterate(%7, 1)::Core.PartialStruct(Tuple{StaticArrays.SVector{3, Float64}, Int64}, Any[StaticArrays.SVector{3, Float64}, Core.Const(2)])
│          (lo = Core.getfield(%8, 1))
│          (@_6 = Core.getfield(%8, 2))
│    %11 = Base.indexed_iterate(%7, 2, @_6::Core.Const(2))::Core.PartialStruct(Tuple{StaticArrays.SVector{3, Float64}, Int64}, Any[StaticArrays.SVector{3, Float64}, Core.Const(3)])
│          (up = Core.getfield(%11, 1))
│    %13 = Meshes.origin(r)::Point3
│          (orig = Meshes.coordinates(%13))
│          (tmin = Meshes.zero($(Expr(:static_parameter, 2))))
│          (tmax = Meshes.typemax($(Expr(:static_parameter, 2))))
│    %17 = (1:$(Expr(:static_parameter, 1)))::Core.Const(1:3)
│          (@_5 = Base.iterate(%17))
│    %19 = (@_5::Core.Const((1, 1)) === nothing)::Core.Const(false)
│    %20 = Base.not_int(%19)::Core.Const(true)
└───       goto #15 if not %20
2 ┄─ %22 = @_5::Tuple{Int64, Int64}::Tuple{Int64, Int64}
│          (i = Core.getfield(%22, 1))
│    %24 = Core.getfield(%22, 2)::Int64
│    %25 = Base.getindex(lo, i)::Float64
│    %26 = Base.getindex(orig, i)::Float64
│    %27 = (%25 - %26)::Float64
│    %28 = Base.getindex(invdir, i)::Float64
│          (imin = %27 * %28)
│    %30 = Base.getindex(up, i)::Float64
│    %31 = Base.getindex(orig, i)::Float64
│    %32 = (%30 - %31)::Float64
│    %33 = Base.getindex(invdir, i)::Float64
│          (imax = %32 * %33)
│    %35 = Base.getindex(invdir, i)::Float64
│    %36 = Meshes.zero($(Expr(:static_parameter, 2)))::Core.Const(0.0)
│    %37 = (%35 < %36)::Bool
└───       goto #4 if not %37
3 ── %39 = imax::Float64
│    %40 = imin::Float64
│          (imin = %39)
│          (imax = %40)
└───       goto #4
4 ┄─ %44 = Meshes.isnan(imin)::Bool
└───       goto #6 if not %44
5 ──       goto #7
6 ── %47 = Meshes.isnan(imax)::Bool
└───       goto #8 if not %47
7 ┄─       goto #13
8 ── %50 = (tmin > imax)::Bool
└───       goto #10 if not %50
9 ──       goto #11
10 ─ %53 = (imin > tmax)::Bool
└───       goto #12 if not %53
11 ┄ %55 = Meshes.NoIntersection()::Core.Const(NoIntersection())
│    %56 = (%55 |> f)::Core.Const(NoIntersection())
└───       return %56
12 ─       (tmin = Meshes.max(tmin, imin))
└───       (tmax = Meshes.min(tmax, imax))
13 ┄       (@_5 = Base.iterate(%17, %24))
│    %61 = (@_5 === nothing)::Bool
│    %62 = Base.not_int(%61)::Bool
└───       goto #15 if not %62
14 ─       goto #2
15 ┄ %65 = (tmin ≈ tmax)::Bool
└───       goto #17 if not %65
16 ─ %67 = (r)(tmin)::Point3
│    %68 = Meshes.TouchingRayBox(%67)::TouchingRayBox{Point3}
│    %69 = (%68 |> f)::TouchingRayBox{Point3}
└───       return %69
17 ─ %71 = (r)(tmin)::Point3
│    %72 = (r)(tmax)::Point3
│    %73 = Meshes.Segment(%71, %72)::Segment{3, Float64, StaticArrays.SVector{2, Point3}}
│    %74 = Meshes.CrossingRayBox(%73)::CrossingRayBox{Segment{3, Float64, StaticArrays.SVector{2, Point3}}}
│    %75 = (%74 |> f)::CrossingRayBox{Segment{3, Float64, StaticArrays.SVector{2, Point3}}}
└───       return %75
```

juliohm: I think it is ok, it is returning an union of intersection types. We will fix that in the next release by refactoring the internal intersection API.

juliohm (approved):

PR #217, opened 2021-11-20
Add public API for Rays

HaoxuanGuo:
Add public API for Rays

### DIFF
--- a/src/Meshes.jl
+++ b/src/Meshes.jl
@@ -155,7 +155,7 @@ export
   Box, Ball, Sphere, Cylinder,
   ncontrols, degree, Horner, DeCasteljau,
   radius, height, sides,
-  measure, diagonal, origin,
+  measure, diagonal, origin, direction,
 
   # polytopes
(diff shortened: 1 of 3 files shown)


src/primitives/ray.jl
```diff
@@ -26,3 +26,17 @@ function (r::Ray)(t)
   end
   r.p + t * r.v
 end
+
+"""
+    origin(ray)
+
+The endpoint of the ray.
```
juliohm:
```suggestion
The starting point of the ray.
```

juliohm:
You are on fire! 🔥 Thanks, just waiting for the build to finish.

src/Meshes.jl
```diff
@@ -155,7 +155,7 @@ export
   Box, Ball, Sphere, Cylinder,
   ncontrols, degree, Horner, DeCasteljau,
   radius, height, sides,
-  measure, diagonal, origin,
+  measure, diagonal, origin, direction,
```
```suggestion
  measure, diagonal,
  origin, direction,
```

PR #219, opened 2021-11-20
ray triangle intersection public APIs

HaoxuanGuo:
ray triangle intersection public APIs

### DIFF
--- a/src/intersections/triangles.jl
+++ b/src/intersections/triangles.jl
@@ -143,8 +143,8 @@ and apply function `f` to it.
 """
 function intersecttype(f::Function, r::Ray{3,T}, t::Triangle{3,T}) where {T}
   vs = vertices(t)
-  o = r.p
-  d = r.v
+  o = origin(r)
+  d = direction(r)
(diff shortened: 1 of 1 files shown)


src/intersections/triangles.jl
```diff
@@ -143,19 +143,17 @@ and apply function `f` to it.
 """
 function intersecttype(f::Function, r::Ray{3,T}, t::Triangle{3,T}) where {T}
   vs = vertices(t)
-  o = r.p
-  d = r.v
 
   e₁ = vs[3] - vs[1]
   e₂ = vs[2] - vs[1]
-  p = d × e₂
+  p = direction(r) × e₂
```
juliohm: You don't need to change the code here, you can just redefine the variables `o` and `d` using the public API now.

juliohm (changes requested):
No need to change the algorithm, just the first lines of it.

GuoHaoxuan:
@juliohm Modified. Just the first line.

PR #228, opened 2021-12-14
Intersection between `Polygon`s and `Sengment`s and `Ray`s.

### LINKED ISSUE #214: Intersection between `Polygon`s and `Sengment`s and `Ray`s.
This is an extra step towards #125 . Moved from #213 to here. Make `[intersect(ray, triangle) for triangle in mesh]` works.

This may require 2 tiny steps. 

- [x] Triangulate `Polygon`. Make Discretization to 3D.
- [ ] Check intersections to every triangle.

HaoxuanGuo:
Fixes #214

### DIFF
--- a/src/intersections/triangles.jl
+++ b/src/intersections/triangles.jl
@@ -196,7 +196,7 @@ Calculate the intersection of a ray and polygon in 3D
 and apply function `f` to it.
 """
 function intersecttype(f::Function, r::Ray{3,T}, p::Ngon{N,3,T}) where {N,T}
-  for t in collect(triangulate(p))
+  for t in triangulate(p)
     res = intersecttype(identity, r, t)
     # A valid polygon has only one intersection with a ray.
     if res != NoIntersection()
--- a/src/intersections/triangles.jl
+++ b/src/intersections/triangles.jl
@@ -188,3 +188,20 @@ function intersecttype(f::Function, r::Ray{3,T}, t::Triangle{3,T}) where {T}
 
   return IntersectingRayTriangle(r(λ)) |> f
 end
+
+"""
+    intersecttype(s, p)
+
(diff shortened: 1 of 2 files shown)


src/intersections/triangles.jl
```diff
@@ -188,3 +188,20 @@ function intersecttype(f::Function, r::Ray{3,T}, t::Triangle{3,T}) where {T}
...
 end
+
+"""
+    intersecttype(s, p)
+
+Calculate the intersection of a ray and polygon in 3D
+and apply function `f` to it.
+"""
+function intersecttype(f::Function, r::Ray{3,T}, p::Ngon{N,3,T}) where {N,T}
+  for t in collect(triangulate(p))
```
juliohm: Do we really need to collect the resulting mesh?

hx-guo: @juliohm I currently have no better approach.

juliohm: What I mean is that `Mesh` types are iterable, you can omit the `collect` call. `for t in triangulate(p)` should work and it will materialize one triangle at a time.

hx-guo: @juliohm Thank you for your suggestion. Now this have been changed.

src/intersections/triangles.jl
```diff
@@ -188,3 +188,20 @@ function intersecttype(f::Function, r::Ray{3,T}, t::Triangle{3,T}) where {T}
...
+"""
+    intersecttype(s, p)
+
+Calculate the intersection of a ray and polygon in 3D
+and apply function `f` to it.
+"""
+function intersecttype(f::Function, r::Ray{3,T}, p::Ngon{N,3,T}) where {N,T}
+  for t in triangulate(p)
+    res = intersecttype(identity, r, t)
+    # A valid polygon has only one intersection with a ray.
```
```suggestion
```

Issue #137, opened 2021-05-26
Implement algorithm for fixing the orientation of cell complexes


juliohm:
Currently we assume that the list of connectivities is given with coherent orientation, i.e. all n-gons are CCW. At some point we will need to fix orientation for the end user. An example method is described in Borodin et al 2004:

https://ieeexplore.ieee.org/document/1309188

juliohm:
This is no longer an issue! 🎉 We can fix the orientation of polygons at construction time in our HalfEdgeTopology constructor: https://github.com/JuliaGeometry/Meshes.jl/commit/f4b1fd1162d50140a3029f5462c5e2a62f35febb

PR #229, opened 2021-12-14
reverse Box/Hexahedron boundary normal

HaoxuanGuo:
The normals of the geometry should be facing outwards, but the normals generated by calling the boundary method on Box before are facing inwards. Fix this problem now.


juliohm:
Awesome, that explains why our vizualization was a bit dark. Can you also fix the orientation of the Hexahedron boundary?

GuoHaoxuan:
@juliohm done for Hexahedron

juliohm:
Awesome, thank you for the quick updates. Can we also add some tests so that we don't reintroduce the issue in future adjustments to the code?

GuoHaoxuan:
@juliohm Although we can add a test to generate a ray along the normal with an origin on the surface from each face of the boundary, then the intersection of the ray and the mesh should be nothing. But this is only a mechanical guarantee that the current geometry does not have this problem. We may forget this test when adding new geometry. A better way may be to solve this problem completely by dealing with #137 .

juliohm:
I don't know if I followed the suggestion. Can't we add a test that computes the inner product between all normals and makes sure that the value is never negative? Meaning the normals are pointing outwards?

GuoHaoxuan:
@juliohm I'm afraid not. Because \cdot(-\vec{b})).  The sign of the normal point multiplication has nothing to do with their direction.

juliohm:
Oh sorry, my mistake. If there are no tests that could be written easily here we can go ahead and merge it.

GuoHaoxuan:
@juliohm If we really do a similar test, perhaps we need to use the method of ray judgment. If a point is inside the geometry, then the ray starting from it will have an odd number of intersections with the geometry, otherwise there will be an even number (0 is also an even number). If the normal is facing outward, the point on the surface should be outside the geometry if it moves along the normal, otherwise it should be inside the geometry. Then we can test the number of intersections. Unfortunately, writing such a test is not a simple process. And more importantly, people who add new geometry may forget this test.

juliohm:
Thank you @HaoxuanGuo for explaining the issue. I will go ahead and merge the PR with the improvements. 💯

PR #231, opened 2021-12-15
more mathful intersection

HaoxuanGuo:
Although the original code behaves correctly on integers and floating point numbers, its handling of units is imperfect. For a number without a unit, comparing its reciprocal with itself is nothing. Occasionally dropping a unit will not have any effect. Anyway, there is no unit.

But for the quantity containing units, the performance of one and oneunit is different. We should ensure that the unit in the calculation process is always correct. Although it seems that dividing by a oneunit is a useless operation, it ensures that t as an interpolation ratio is a dimensionless number, and it also ensures that there will be no dimension mismatch when comparing with invdir.

The improved code passed the test in Unitful's Dimensional Number, but because it did not want to introduce additional dependencies, it was not included in the unit test.

### DIFF
--- a/src/intersections/boxes.jl
+++ b/src/intersections/boxes.jl
@@ -51,8 +51,8 @@ function intersecttype(f::Function, r::Ray{Dim,T}, b::Box{Dim,T}) where {Dim,T}
   lo, up = coordinates.(extrema(b))
   orig = coordinates(origin(r))
 
-  tmin = zero(T)
-  tmax = typemax(T)
+  tmin = zero(T) / oneunit(T)
+  tmax = typemax(T) / oneunit(T)
 
   # check for intersection with slabs along with each axis
   for i in 1:Dim
@@ -60,7 +60,7 @@ function intersecttype(f::Function, r::Ray{Dim,T}, b::Box{Dim,T}) where {Dim,T}
     imax = (up[i] - orig[i]) * invdir[i]
 
     # swap variables if necessary
-    invdir[i] < zero(T) && ((imin, imax) = (imax, imin))
+    invdir[i] < (zero(T) / oneunit(T)^2) && ((imin, imax) = (imax, imin))
 
     # the ray is on a face of the box, avoid NaN
     (isnan(imin) || isnan(imax)) && continue
--- a/src/intersections/boxes.jl
+++ b/src/intersections/boxes.jl
@@ -60,7 +60,7 @@ function intersecttype(f::Function, r::Ray{Dim,T}, b::Box{Dim,T}) where {Dim,T}
     imax = (up[i] - orig[i]) * invdir[i]
 
     # swap variables if necessary
-    invdir[i] < zero(T)^3 && ((imin, imax) = (imax, imin))
+    invdir[i] < zero(eltype(invdir)) && ((imin, imax) = (imax, imin))
 
     # the ray is on a face of the box, avoid NaN
     (isnan(imin) || isnan(imax)) && continue
--- a/src/intersections/boxes.jl
+++ b/src/intersections/boxes.jl
@@ -60,7 +60,7 @@ function intersecttype(f::Function, r::Ray{Dim,T}, b::Box{Dim,T}) where {Dim,T}
     imax = (up[i] - orig[i]) * invdir[i]
 
     # swap variables if necessary
-    invdir[i] < (zero(T) / oneunit(T)^2) && ((imin, imax) = (imax, imin))
+    invdir[i] < zero(T)^3 && ((imin, imax) = (imax, imin))
 
     # the ray is on a face of the box, avoid NaN
     (isnan(imin) || isnan(imax)) && continue
--- a/src/intersections/triangles.jl
+++ b/src/intersections/triangles.jl
@@ -152,21 +152,21 @@ function intersecttype(f::Function, r::Ray{3,T}, t::Triangle{3,T}) where {T}
   det = e₁ ⋅ p
 
   # keep det > 0, modify T accordingly
-  if det > atol(T)
+  if det > atol(T) * oneunit(T)^2
     τ = o - vs[1]
   else
     τ = vs[1] - o
     det = -det
   end
 
-  if det < atol(T)
+  if det < atol(T) * oneunit(T)^2
     # This ray is parallel to the plane of the triangle.
     return NoIntersection() |> f
   end
 
   # calculate u parameter and test bounds
   u = τ ⋅ p
-  if u < -atol(T) || u > det
+  if u < -atol(T) * oneunit(T)^2 || u > det
     return NoIntersection() |> f
   end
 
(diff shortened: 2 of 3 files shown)


src/intersections/boxes.jl
```diff
@@ -51,16 +51,16 @@ function intersecttype(f::Function, r::Ray{Dim,T}, b::Box{Dim,T}) where {Dim,T}
...
+  tmax = typemax(T) / oneunit(T)
 
   # check for intersection with slabs along with each axis
   for i in 1:Dim
     imin = (lo[i] - orig[i]) * invdir[i]
     imax = (up[i] - orig[i]) * invdir[i]
 
     # swap variables if necessary
-    invdir[i] < zero(T) && ((imin, imax) = (imax, imin))
+    invdir[i] < zero(T)^3 && ((imin, imax) = (imax, imin))
```
juliohm: I am not sure this is the right way to do this. We are setting zero(T)^3 arbitrary here where we should be using atol(T) or atol(T)^2 if comparing areas.

src/intersections/triangles.jl
```diff
@@ -152,39 +152,39 @@ function intersecttype(f::Function, r::Ray{3,T}, t::Triangle{3,T}) where {T}
   det = e₁ ⋅ p
 
   # keep det > 0, modify T accordingly
-  if det > atol(T)
+  if det > atol(T) * oneunit(T)^2
```
juliohm: You seem to be manually fixing all occurrences of atol(T) with a oneunit(T)^2 factor. This is not good for maintenance. If a fix is necessary in multiple lines of  the code, we need to define a new variable like threshold = atol(T)^2 and use it everywhere. Consistent and clean code is super important in a package that large.

src/polytopes/ngon.jl
```diff
@@ -129,10 +129,10 @@ function Base.in(p::Point{3,T}, t::Triangle{3,T}) where {T}
   (λ₂ ≥ zero(T)) && (λ₃ ≥ zero(T)) && ((λ₂ + λ₃) ≤ one(T))
 end
 
-function normal(t::Triangle{3})
+function normal(t::Triangle{3,T}) where {T}
   a, b, c = t.vertices
   n = (b - a) × (c - a)
-  n / norm(n)
+  n / norm(n) * oneunit(T)
```
juliohm, line 135: I am confused with the usage of oneunit. It seems like what we really want is to propagate units correctly and  that oneunit is just a hack to achieve that. Is there any recommendation in the Unitful.jl docs about propagation of units in generic code? I don' t see why our code with generic T should fail with units. If the coordinates are unitful everything should work no?

src/intersections/boxes.jl
hx-guo: @juliohm atol(T)^2 does not seem to work. For example, if atol(T) is 1e-10 mm, then atol(T)^2 will make 1e-20 mm^2. However, the purpose of using atol is to avoid floating-point errors. Although the unit of atol(T)^2 is correct, it also changes the value, which is not what we want.

src/intersections/triangles.jl
hx-guo: @juliohm Ammmm. You are right. Patches scattered in the package are not conducive to maintenance. Maybe we should think of a way on the atol function. For example, `det> atol(det)` or `det> atol(typeof(det))` is better than `det> atol(T) * oneunit(T)^2`. If we write this way, we don't have to multiply by a factor manually.

src/polytopes/ngon.jl
hx-guo: @juliohm In fact, your understanding is correct. As long as we use the generic T correctly and the correct mathematical description, we don't have to worry about the problem of unit propagation. For this line of code, it needs to be fixed because it is mathematically wrong. Although we really like to get a unit vector by dividing by its modulus length, we rarely notice that the modulus length divided by here should be a number without a unit. Otherwise, dividing two quantities of the same dimension will lose the unit.

In this line, `normalize(n)` may seems better.  Unfortunately, the definition of normalize does not take this issue.  `normalize(n) == n / norm(n)` in current. I think this is a bug in upstream. Before it is realized by Base or Unitful, we have to manually add a oneunit(T) here.

GuoHaoxuan:
@juliohm You may have discovered that all the changes I made above are when comparing with 0, with inf and defining constants by myself. At these times we did not choose the correct unit of 0, infinity, or constant. Now I changed their types to be correct as expected.

As you can imagine, the unit of the vector is automatic when doing calculations, so we don't need to worry about it.

Now look at the modified code, there is almost no explicit oneunit factor. Is it what you want to see?

src/polytopes/ngon.jl
hx-guo: In the long run, n / norm(n) is more secure than normalize(n), because it will not be affected by how upstream treats normalize units.

GuoHaoxuan:
In addition, I even think that a Unitful unit test can be added in addition to Float32 and Float64 to find out that although the value is correct, the unit is incorrect. Of course, this is another long-term topic, and a separate issue should be opened for discussion after the end of this PR.

src/intersections/boxes.jl
```diff
@@ -51,16 +51,16 @@ function intersecttype(f::Function, r::Ray{Dim,T}, b::Box{Dim,T}) where {Dim,T}
...
+  tmax = typemax(T) / oneunit(T)
 
   # check for intersection with slabs along with each axis
   for i in 1:Dim
     imin = (lo[i] - orig[i]) * invdir[i]
     imax = (up[i] - orig[i]) * invdir[i]
 
     # swap variables if necessary
-    invdir[i] < zero(T) && ((imin, imax) = (imax, imin))
+    invdir[i] < zero(eltype(invdir)) && ((imin, imax) = (imax, imin))
```
juliohm, line 63: So invdir has the correct eltype even though we didn't use oneunit(T) right? I am still confused how units propagate. Before we can address this issue here, we need to discuss units in more depth in an issue I think. Mind opening an issue with a proposal that addresses the problem of unit propagation consistently across the codebase?

hx-guo: @juliohm In this sample, invdir is the inverse of direction. If the direction is a dimensionless number (the dimensionless dimension is 1, such as Float64), the dimension of invdir is the reciprocal of 1, which is still 1, then the type of invdir is the same as the direction. But the agreement here is because of coincidence, because the reciprocal of 1 is still 1.

If the dimension of direction is millimeters, then the dimension of invdir should be the reciprocal of millimeters, which is mm^-1. At this time, the units of direction and invdir are different, so T can not be used to represent. To correctly indicate the type of invdir, typeof(invdir) should be used.

In all the previous codes, because only Float32 and Float64 are considered, this issue has not been carefully considered. And because 1 is a unit of Float, too many operations related to 1 are ignored and simplified. Therefore, the code before the constant with the unit will expose these problems.

I agree with you that we should open an issue to discuss this matter in detail and carefully before making a decision.

juliohm:
I will close this PR because as we discussed, we need a more general approach to handle units that applies to the entire codebase. It is a major project that we could try to address next year as we evolve the treatment of coordinates.

Issue #224, opened 2021-12-05
julia 1.7 compatible


GuoHaoxuan:
Test in julia 1.7 may not passed.

1. Wrapping `Vararg` directly in UnionAll is deprecated.
2. Random engine changes (Xorshift for now)
3. Some Colors type deprecated?

```julia
     Testing Running tests...
WARNING: Wrapping `Vararg` directly in UnionAll is deprecated (wrap the tuple instead).
WARNING: Wrapping `Vararg` directly in UnionAll is deprecated (wrap the tuple instead).
WARNING: Wrapping `Vararg` directly in UnionAll is deprecated (wrap the tuple instead).
WARNING: Wrapping `Vararg` directly in UnionAll is deprecated (wrap the tuple instead).
WARNING: Wrapping `Vararg` directly in UnionAll is deprecated (wrap the tuple instead).
WARNING: Wrapping `Vararg` directly in UnionAll is deprecated (wrap the tuple instead).
WARNING: Wrapping `Vararg` directly in UnionAll is deprecated (wrap the tuple instead).
WARNING: Wrapping `Vararg` directly in UnionAll is deprecated (wrap the tuple instead).
WARNING: Wrapping `Vararg` directly in UnionAll is deprecated (wrap the tuple instead).
WARNING: Wrapping `Vararg` directly in UnionAll is deprecated (wrap the tuple instead).
WARNING: Wrapping `Vararg` directly in UnionAll is deprecated (wrap the tuple instead).
WARNING: Wrapping `Vararg` directly in UnionAll is deprecated (wrap the tuple instead).
WARNING: Wrapping `Vararg` directly in UnionAll is deprecated (wrap the tuple instead).
WARNING: Wrapping `Vararg` directly in UnionAll is deprecated (wrap the tuple instead).
WARNING: Wrapping `Vararg` directly in UnionAll is deprecated (wrap the tuple instead).
WARNING: Wrapping `Vararg` directly in UnionAll is deprecated (wrap the tuple instead).
WARNING: Wrapping `Vararg` directly in UnionAll is deprecated (wrap the tuple instead).
WARNING: Wrapping `Vararg` directly in UnionAll is deprecated (wrap the tuple instead).
WARNING: Wrapping `Vararg` directly in UnionAll is deprecated (wrap the tuple instead).
WARNING: Wrapping `Vararg` directly in UnionAll is deprecated (wrap the tuple instead).
[...]

PR #232, opened 2021-12-15
Fix tests for Julia v1.7

hyrodium:
This PR fixes https://github.com/JuliaGeometry/Meshes.jl/issues/224.


.github/workflows/CI.yml
```diff
@@ -17,6 +17,7 @@ jobs:
       fail-fast: false
       matrix:
         version:
+          - '1'
```
juliohm: Awesome, can you please use 1.7 explicitly here so that we officially support 1.6 and 1.7 from now on?

test/collections.jl
```diff
@@ -58,13 +58,18 @@
...
+        ver = ""
+      else
+        ver = "_1.7"
+      end
+      @test_reference "data/pset-1D-$T$ver.png" plot(PointSet(rand(P1,10)))
+      @test_reference "data/pset-2D-$T$ver.png" plot(PointSet(rand(P2,10)))
+      @test_reference "data/pset-3D-$T$ver.png" plot(PointSet(rand(P3,10)))
+      @test_reference "data/pset-1D-$T-data$ver.png" plot(PointSet(rand(P1,10)),1:10)
+      @test_reference "data/pset-2D-$T-data$ver.png" plot(PointSet(rand(P2,10)),1:10)
+      @test_reference "data/pset-3D-$T-data$ver.png" plot(PointSet(rand(P3,10)),1:10)
```
juliohm: I think what we really want is to create a set of points deterministically and update the tests. We used random points because we wrote these tests quickly. Can you create a new set of points that is not a function of the RNG and update the figures? They should match both in Julia v1.6 and v1.7. Thank you for the help 💯

juliohm (changes requested):
Review attached.

.github/workflows/CI.yml
hyrodium: Okay, I'll change it that way. (I was just thinking `'1'` would be better because we can notice test fails before getting a new issue like #224.)

test/collections.jl
hyrodium: Thank you for the review! I've changed it that way.

test/partitioning.jl
```diff
@@ -13,7 +13,11 @@
     p = partition(grid, UniformPartition(3, false))
     @test setify(indices(p)) == setify([[1,2,3], [4,5,6], [7,8,9]])
     p = partition(grid, UniformPartition(3))
-    @test setify(indices(p)) == setify([[8,6,9], [4,1,7], [2,3,5]])
+    if VERSION < v"1.7"
```
juliohm, line 16: Can we compare with v1.6 somehow? A comparison that honors semantic versioning.

.github/workflows/CI.yml
juliohm: Yes, I think we want this kind of issue open. It is better to be explicit about the versions we support than assume things are supported in future versions of the language.

test/partitioning.jl
juliohm: I think you can do something like `major(VERSION) == 1 && minor(VERSION) == 6`

hyrodium: It seems we are supporting `v1.5` (see [Project.toml](https://github.com/JuliaGeometry/Meshes.jl/blob/d35561d21301039f142c0448e7f2eacd8c368d02/Project.toml#L40)), so I think it's better to keep `VERSION < v"1.7"` here.

hyrodium: Maybe we will drop support Julia v1.5 in the future, but I think It's better to avoid unintentional errors with `< v"1.7"`.

juliohm: Yes, I think we should only support LTS and the latest release. It is too much on us maintainers to keep the project working across multiple versions that none of us are using anymore. I am planning to do a major cleanup in all Project.toml and CI of a bunch of packages to only support LTS and latest, that is why I think we should drop v1.5 from here as well.

juliohm: It is a fresh restart to officially state what versions are supported. We've been lazy about it but as the package matures towards a v1.0 we will need to adopt some rule like that.

hyrodium: >Yes, I think we should only support LTS and the latest release. It is too much on us maintainers to keep the project working across multiple versions that none of us are using anymore. I am planning to do a major cleanup in all Project.toml and CI of a bunch of packages to only support LTS and latest, that is why I think we should drop v1.5 from here as well.

I agree with that:+1:

>A comparison that honors semantic versioning.

Julia is semantically versioned, so I'm not sure why we should avoid `VERSION < v"1.7"`.
It seems `VERSION < v"1.7"` is easier to read, fewer characters, and more meaningful than `major(VERSION) == 1 && minor(VERSION) == 6`, I guess.

Is it okay to just adding a comment like `# The random number generation algorithm has been changed since Julia 1.7.` ?

juliohm: Sounds good, let's merge it and then we can sort out a more explicit documentation about the supported versions of the language.

juliohm (approved):

PR #235, opened 2021-12-16
added methods for vectors for ∠

kylebeggs:
Added the `clamp(x, -1, 1)` function because there were corner cases where `(A ⋅ B) / (norm(A) * norm(B))` would be slightly over |1| (such as -1.0000000000000002) due to round-off error and so then `acos()` would be undefined.

Also, support should be added to always convert Vectors from Int to Float like how Points are constructed. Perhaps I should discuss in Tulip or open an issue..


src/angles.jl
```diff
@@ -26,3 +26,21 @@ function ∠(A::P, B::P, C::P) where {P<:Point{3}}
   BA, BC = A - B, C - B
   atan(norm(BA × BC), BA ⋅ BC) # discard sign
 end
+
+"""
+    ∠(A, B)
+
+Angle between vectors A and B.
```
juliohm: We use lowercase letters for vectors usually. Can you rename to u and v?

src/angles.jl
```diff
@@ -26,3 +26,21 @@ function ∠(A::P, B::P, C::P) where {P<:Point{3}}
...
+See https://en.wikipedia.org/wiki/Angle.
+
+Uses `acos` returning a value in the range [0, π].
```
juliohm:
```suggestion
Uses [`acos`](@ref) returning a value in the range [0, π].
```

src/angles.jl
```diff
@@ -26,3 +26,21 @@ function ∠(A::P, B::P, C::P) where {P<:Point{3}}
...
+
+## Examples
+
+```julia
+∠(Vec(1,0), Vec(0,1)) == π/2
+```
+"""
+function ∠(A::V, B::V) where {V<:Vec}
```
juliohm:
```suggestion
function ∠(u::Vec{Dim,T}, v::Vec{Dim,T}) where {Dim,T}
```

src/angles.jl
```diff
@@ -26,3 +26,21 @@ function ∠(A::P, B::P, C::P) where {P<:Point{3}}
...
+  acos(clamp((A ⋅ B) / (norm(A) * norm(B)), -1, 1))
```
juliohm: Please split this line into multiple lines for clarity and use -one(T) and one(T) instead of the literals -1 and 1.

juliohm (changes requested):
Thank you @kylebeggs for the addition 💯 Can you please incorporate the suggestions before we enable the tests on GitHub?

kylebeggs:
Ok, I think I fixed everything you suggested.

src/angles.jl
kylebeggs: I wasn't sure if you wanted just to use hanging ( to go to new lines or use variables. I used variables and checked that it does compile out so the resulting code is the same.

src/angles.jl
```diff
@@ -26,3 +26,23 @@ function ∠(A::P, B::P, C::P) where {P<:Point{3}}
...
+## Examples
+
+```julia
+∠(Vec(1,0), Vec(0,1)) == π/2
+```
+"""
+function ∠(u::Vec{Dim,T}, v::Vec{Dim,T}) where {Dim,T}
+  x = (u ⋅ v) / (norm(u) * norm(v))
+  x = clamp(x, -one(T), one(T))  # restrict domain in case of round-off error.
+  acos(x)
```
juliohm:
```suggestion
  cosθ = (u ⋅ v) / (norm(u) * norm(v))
  acos(clamp(cosθ, -one(T), one(T)))
```

juliohm (approved):

juliohm:
@kylebeggs , I think we can do even better if we fallback to the methods defined for points. That way we can preserve the sign of the angle and avoid numerical issues. Can you rewrite the method to compute the angle between 3 points (u, origin, v)?

kylebeggs:
I rewrote to match same method as the points.

src/angles.jl
```diff
@@ -26,3 +26,26 @@ function ∠(A::P, B::P, C::P) where {P<:Point{3}}
...
+∠(Vec(1,0), Vec(0,1)) == π/2
+```
+"""
+function ∠(u::Vec{2,T}, v::Vec{2,T}) where T
+  atan(u × v, u ⋅ v)  # preserve sign
+end
+
+function ∠(u::Vec{3,T}, v::Vec{3,T}) where T
+  atan(norm(u × v), u ⋅ v)  # discard sign
+end
```
juliohm:
```suggestion
function ∠(u::Vec{2}, v::Vec{2})
  atan(u × v, u ⋅ v)  # preserve sign
end

function ∠(u::Vec{3}, v::Vec{3})
  atan(norm(u × v), u ⋅ v)  # discard sign
end
```

src/angles.jl
```diff
@@ -26,3 +26,26 @@ function ∠(A::P, B::P, C::P) where {P<:Point{3}}
...
+∠(Vec(1,0), Vec(0,1)) == π/2
+```
+"""
+function ∠(u::Vec{2}, v::Vec{2})
+  atan(u × v, u ⋅ v)  # preserve sign
+end
+
+function ∠(u::Vec{3}, v::Vec{3})
+  atan(norm(u × v), u ⋅ v)  # discard sign
+end
```
juliohm:
```suggestion
∠(u::Vec{2}, v::Vec{2}) = atan(u × v, u ⋅ v)  # preserve sign
∠(u::Vec{3}, v::Vec{3}) = atan(norm(u × v), u ⋅ v)  # discard sign
```

juliohm:
Awesome @kylebeggs . Now we just need to re-express the methods with points in terms of the new methods with Vectors. Can you help with that?

kylebeggs:
Sure. I think this is what you meant?

src/angles.jl
```diff
@@ -17,12 +17,23 @@ See https://en.wikipedia.org/wiki/Atan2.
...
+See https://en.wikipedia.org/wiki/Atan2.
+
+## Examples
+
+```julia
+∠(Vec(1,0), Vec(0,1)) == π/2
+```
+"""
+∠(u::Vec{2}, v::Vec{2}) = atan(u × v, u ⋅ v)  # preserve sign
+∠(u::Vec{3}, v::Vec{3}) = atan(norm(u × v), u ⋅ v)  # discard sign
```
juliohm:
```suggestion
∠(u::V, v::V) where {V<:Vec{2}} = atan(u × v, u ⋅ v)  # preserve sign
∠(u::V, v::V) where {V<:Vec{3}} = atan(norm(u × v), u ⋅ v)  # discard sign
```

juliohm:
Thanks @kylebeggs ! Awesome addition. 👍🏽

kylebeggs:
@juliohm thank you for being patient with me, this was my first PR!! Looking forward to more!

Issue #239, opened 2021-12-22
Specify broadcast behavior for Point


juliohm:
This code should work fine but it doesn't currently:

```julia
julia> using Meshes

julia> A = rand(Point2)
Point(0.6573997766238846, 0.6666740453491755)

julia> Bs = rand(Point2, 10)
10-element Vector{Point2}:
 Point(0.7417085674977997, 0.8951757340967124)
 Point(0.30445377004236973, 0.6820654287086994)
 Point(0.7260784357032839, 0.9640053918106126)
 Point(0.473098333874463, 0.20506498511943105)
 Point(0.015199853998886903, 0.30178497490417966)
 Point(0.642732034127067, 0.7274587877348253)
 Point(0.23427887771588662, 0.20945800330058972)
 Point(0.8173283918758907, 0.2631886201839835)
 Point(0.7116938366721508, 0.7271525284821094)
 Point(0.6025325520661857, 0.5612801787882418)

julia> A .- Bs
ERROR: MethodError: no method matching length(::Point2)
```

We need to add a method for broadcast with Point so that it works. I think we want to add a method for this function: https://docs.julialang.org/en/v1/base/arrays/#Base.Broadcast.broadcastable that returns a Ref. But it is worth double checking.

juliohm:
Fixed by #241

PR #241, opened 2021-12-22
Specify broadcast behavior for Point #239

kylebeggs:
Previously, you could not subtract a single Point from an array of Points using broadcasting. This PR fixes that (Issue #239).

### DIFF
--- a/src/points.jl
+++ b/src/points.jl
@@ -60,8 +60,8 @@ const Point1f = Point{1,Float32}
 const Point2f = Point{2,Float32}
 const Point3f = Point{3,Float32}
 
-# Enable broadcasting behavior for Point
-Base.Broadcast.broadcastable(p::Point) = Ref(p)
+# enable broadcasting behavior for Point
+Broadcast.broadcastable(p::Point) = Ref(p)
 
 """
     embeddim(point)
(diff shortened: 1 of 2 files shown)


src/points.jl
```diff
@@ -60,6 +60,9 @@ const Point1f = Point{1,Float32}
 const Point2f = Point{2,Float32}
 const Point3f = Point{3,Float32}
 
+# Enable broadcasting behavior for Point
+Base.Broadcast.broadcastable(p::Point) = Ref(p)
```
```suggestion
# enable broadcasting behavior for Point
Broadcast.broadcastable(p::Point) = Ref(p)
```

juliohm:
Thank you @kylebeggs , can you please add some tests?

kylebeggs:
Ok, should be good now. Still trying to get a full grip on PRs and merging, etc. with git..

src/points.jl
```diff
@@ -60,6 +60,9 @@ const Point1f = Point{1,Float32}
 const Point2f = Point{2,Float32}
 const Point3f = Point{3,Float32}
 
+# enable broadcasting behavior for Point
```
```suggestion
# broadcast behavior
```

Issue #243, opened 2022-01-11
Missing cylinder triangulation


VEZY:
Hello, I think there is no implementation of `triangulate` for the `Cylinder` primitive. 
I found the issue when trying to plot a simple cylinder with MeshViz:

```julia
using Meshes, MeshViz, GLMakie
viz(Cylinder(0.1))
```

juliohm:
We have the `boundary(cylinder)` implemented as a `CylinderSurface` already. We just need to add a RegularDiscretization method for this surface. Do you think you can help with that? There are examples with spheres to serve as a base.

I can try to work on that after some deadlines this month in case you don't know how to proceed.

VEZY:
Just to be sure, you mean we need to implement:

```julia
function discretize(cyl::Cylinder{T}, method::RegularDiscretization) where {T}
[...]
end
```

But also a RegularSampling right ? 

and then:

```julia
triangulate(cyl::Cylinder{T}) =
  discretize(cyl, RegularDiscretization(100)) |> triangulate
```

I would be happy to help but I'm not sure to understand the package design correctly. I must say I have very little time and I'm not sure to have the skills to do it (though I'm sure there are some resources somewhere online?).

I'll give it a try if I find a little bit of time. Thank you for considering this issue.

juliohm:
Yes. Everything is very composible and we just need to know how to sample
points on the surface of the cylinder and how to connect these points into
quadrangles + triangles. The rest follows from the other implementations.

What about first submitting a PR for RegularSampling with CylinderSurfsce?
After that we can work on the connectivities.

VEZY:
OK, I just did a PR (#244). I didn't have time to check the validity domain of the implementation yet. Hope this helps.

juliohm:
Fixed on master. Now we can visualize cylinders and cylinder surfaces that are oblique, for example:

![image](https://user-images.githubusercontent.com/3345261/183258627-614e2201-f63a-44ee-8224-f0e9dc711944.png)

As usual, the code composes really nicely, you can simply do:

```julia
using Meshes, MeshViz
import GLMakie as Mke

c = Cylinder(1.0) # aligned with z axis

viz(c)
```
![image](https://user-images.githubusercontent.com/3345261/183258671-77c22a65-f019-41a1-9c0e-bde9485a9b66.png)

```julia
# top plane has a normal (1,1,1)
[...]

PR #244, opened 2022-01-14
Add RegularSampling method for a CylinderSurface + Cylinder

VEZY:
As discussed in #243, this is a PR to add a RegularSampling method for a CylinderSurface and for Cylinder (not sure this one is useful though?)

It needs a review before merging, I'm not sure of the implementation. 

What the implementation does is sampling a cylinder starting from the bottom to the top. It uses the cylinder radius and a segment between the top center and bottom center. We can sample with two values: the first one gives the number of sectors to divide the cylinder (nb of points around each circle) and the other one gives the number of circles along the segment. 

Hope this helps. Let me know if I can be of any further help.

### DIFF
--- a/src/sampling/regular.jl
+++ b/src/sampling/regular.jl
@@ -129,8 +129,8 @@ function sample(::AbstractRNG, cyl::CylinderSurface{T},
   θrange = range(θmin, stop = θmax, length = sz[1])
 
   # Sampling points along the bottom and top circles
-  top_circle = [origin(cyl.top) + r * cos(θ) * cyl.top.u + r * sin(θ) * cyl.top.v for θ in θrange]
-  bot_circle = [origin(cyl.bot) + r * cos(θ) * cyl.bot.u + r * sin(θ) * cyl.bot.v for θ in θrange]
+  top_circle = [cyl.top(cos(θ), sin(θ)) for θ in θrange]
+  bot_circle = [cyl.bot(cos(θ), sin(θ)) for θ in θrange]
 
   # Iterator for sampling each point of each circle
   ivec(Segment(bot_circle[i], top_circle[i])(t) for i in 1:length(top_circle) for t in c_range)
--- a/src/sampling/regular.jl
+++ b/src/sampling/regular.jl
@@ -110,3 +110,32 @@ function sample(::AbstractRNG, hex::Hexahedron{Dim,T},
   wrange = range(T(0), T(1), length=sz[3])
   ivec(hex(u, v, w) for u in urange, v in vrange, w in wrange)
 end
+
+
+function sample(::AbstractRNG, cyl::CylinderSurface{T}, 
+                method::RegularSampling) where {T}
+  # See http://www.songho.ca/opengl/gl_cylinder.html for a good introduction to this
+  # computation.
+
+  V = T <: AbstractFloat ? T : Float64
+  sz = fitdims(method.sizes, paramdim(cyl))
+  r = radius(cyl)
+  θmin, θmax = V(0), V(2π)
+
+  # Number of circles used for sampling along the cylinder:
+  cyl_segment = Segment(origin(cyl.bot), origin(cyl.top))
+  c_range = sample(cyl_segment, RegularSampling(sz[2]))
+
+  # Number of points sampling each circle:
+  θrange = range(θmin, stop = θmax, length = sz[1])
+
+  # Function to sample points along a circle
+  sample_point(θ) = Vec{3,V}(r * cos(θ), r * sin(θ), 0)
+
+  # Iterator for sampling each point of each circle
+Meshes.ivec(c + sample_point(θ) for θ in θrange, c in c_range)
+end
+
+function sample(::AbstractRNG, cyl::Cylinder{T}, method::RegularSampling) where {T}
+  sample(Random.GLOBAL_RNG, boundary(cyl), method)
+end
--- a/src/sampling/regular.jl
+++ b/src/sampling/regular.jl
@@ -112,7 +112,7 @@ function sample(::AbstractRNG, hex::Hexahedron{Dim,T},
 end
 
 
-function sample(::AbstractRNG, cyl::CylinderSurface{T}, 
+function sample(::AbstractRNG, cyl::CylinderSurface{T},
                 method::RegularSampling) where {T}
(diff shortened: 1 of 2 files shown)


src/sampling/regular.jl
```diff
@@ -110,3 +110,32 @@ function sample(::AbstractRNG, hex::Hexahedron{Dim,T},
...
+  # Function to sample points along a circle
+  sample_point(θ) = Vec{3,V}(r * cos(θ), r * sin(θ), 0)
+
+  # Iterator for sampling each point of each circle
+Meshes.ivec(c + sample_point(θ) for θ in θrange, c in c_range)
+end
+
+function sample(::AbstractRNG, cyl::Cylinder{T}, method::RegularSampling) where {T}
+  sample(Random.GLOBAL_RNG, boundary(cyl), method)
+end
```
juliohm:
```suggestion
```

src/sampling/regular.jl
```diff
@@ -110,3 +110,32 @@ function sample(::AbstractRNG, hex::Hexahedron{Dim,T},
...
+  c_range = sample(cyl_segment, RegularSampling(sz[2]))
+
+  # Number of points sampling each circle:
+  θrange = range(θmin, stop = θmax, length = sz[1])
+
+  # Function to sample points along a circle
+  sample_point(θ) = Vec{3,V}(r * cos(θ), r * sin(θ), 0)
+
+  # Iterator for sampling each point of each circle
+Meshes.ivec(c + sample_point(θ) for θ in θrange, c in c_range)
```
juliohm:
```suggestion
  ivec(c + sample_point(θ) for θ in θrange, c in c_range)
```

juliohm:
Thank you @VEZY for the addition. Can you please share some previews in the PR with the points sampled? Are you assuming that the top and bottom circles of the cylinders are parallel? We have a definition that allows non-parallel bases.

I would first consider the angle theta around the axis of the cylinder and would then lookup the corresponding points in the two circles on bottom and top to make a segment.

VEZY:
> Thank you @VEZY for the addition. Can you please share some previews in the PR with the points sampled? Are you assuming that the top and bottom circles of the cylinders are parallel? We have a definition that allows non-parallel bases.
> 
> I would first consider the angle theta around the axis of the cylinder and would then lookup the corresponding points in the two circles on bottom and top to make a segment.

Ha yes indeed I assumed both planes were parallel. OK I'll try something following your recommendation. 
Not sure what you mean by "Can you please share some previews in the PR with the points sampled?". You mean sharing a 3d plot of the vertices ?

juliohm:
> You mean sharing a 3d plot of the vertices ?

Exactly :) and after we are positive things are working as expected we could even add them as visual tests. Before committing the images to the PR we can just check here with on the thread.

VEZY:
OK will do! 
But first can you help me fix the issue that makes the test not passing ? 
I found out it is coming from [here](https://github.com/JuliaGeometry/Meshes.jl/blob/73bc118aa646bd9f8ebf650409f496100cc05aa2/test/neighborhoods.jl#L41):

```julia
m = metric(MetricBall(T.((1.0,0.5,0.5)), TaitBryanAngles(T(π/4),T(),T(0))))
```

The `T()` is not valid, maybe you meant `T(0)` ? Tests are passing with this.

juliohm:
Interesting! I don't know how the master branch is passing the tests. You are correct, it should be T(0).

VEZY:
OK I pushed a new commit. This time we use the u and v vectors of the planes to orient the circles in the right directions, and as you suggested we use the segments between similar points of the circles to sample points in-between to create the circles between the top and bottom. 

Here is a little bit of code to plot the results when needed:

```julia
using Meshes, MeshViz, GLMakie
right_cyl = Cylinder(Float32(0.1))
sampled_right_cyl = collect(sample(boundary(right_cyl), RegularSampling(100, 100)))
viz(sampled_right_cyl)
```

Which gives the following plot:

image


Now this is fine for a right cylinder, but there is an issue: we use the radius as the radius of the top and bottom circles, which is **not** the radius of the resulting cylinder if the planes are not parallel to the cylinder axis:

```julia
cyl = Cylinder(
    0.1,
    Plane(Meshes.Point3(0, 0, 0), Meshes.Vec3(0, 1, 1)),
    Plane(Meshes.Point3(0, 0, 1), Meshes.Vec3(0, 1, 1))
)
viz(collect(sample(boundary(cyl), RegularSampling(100, 100))))
```

image

Here we should transform the circle to an ellipse so the resulting cylinder stays circular and of radius r. 

PS: I used some resources from [this website](https://www.paulbourke.net/geometry/circlesphere/), there's a readily usable implementation of a tapered cylinder, which could be interesting to implement in the future.

juliohm:
Awesome. It is starting to look more promising. I think what we really need is to simply define a single polar angle around the cylinder axis and then use this angle to compute the corresponding value at the top and bottom circles. Notice that our Plane objects are already parameterized as plane(s, t) so you can simply call plane(cos(theta), sin(theta)) or something to get the corresponding point on the plane circle. Can you try to follow this logic?

I would also reduce the number of samples in the next visualization so that we can more easily debug things :) It is looking really nice, and will be super useful feature! Thanks!

juliohm:
P.S.: the link to the website is broken for me.

VEZY:
OK I'll look into it. I updated the code to use your suggestion (`plane(...)`). I edited the website link, it should work now.

VEZY:
OK I tried to understand what you said and find some information on the web, but my knowledge on geometry is dating from lessons I took 15 years ago ^^.
 
I don't know how a polar angle is computed or how we can use it to project our points back onto our cylinder.

juliohm:
If you define the angle theta as in the attached sketch and use the plane
parametrizations with this theta it should work assuming that the
projection of the basis of the two planes are aligned. Makes sense? Sorry
for the delay to reply. Finishing up a draft for a new paper this month.

juliohm:
Basically, given the axis of the cylinder which is independent of the top and bottom circles, you can define an angle of rotation. This angle can then be passed to plane(cos(theta), sin(theta)) to pick points on the two planes and connect as segments right? Let me know if I am missing something.

juliohm:
Closing the PR as inactive. Feel free to start a fresh PR in the future based on the current master branch.

PR #254, opened 2022-04-15
Fix type error for custom type discretization

HaoxuanGuo:
The following code can not run before fixing.

```julia
using Meshes
using Uniftul

sphere = Sphere((0.0u"cm", 0.0u"cm", 0.0u"cm"), 0.5u"cm")
mesh = triangulate(discretize(sphere, RegularDiscretization(100)))
```


GuoHaoxuan:
Sorry for blank characters.

PR #255, opened 2022-04-15
Fix type error for custom type discretization

HaoxuanGuo:
The following code can not run before fixing.

```julia
using Meshes
using Uniftul

sphere = Sphere((0.0u"cm", 0.0u"cm", 0.0u"cm"), 0.5u"cm")
mesh = triangulate(discretize(sphere, RegularDiscretization(100)))
```

### DIFF
--- a/src/discretization/regular.jl
+++ b/src/discretization/regular.jl
@@ -57,7 +57,8 @@ function discretize(sphere::Sphere{3,T},
   # add north and south poles
   c = center(sphere)
   r = radius(sphere)
-  e⃗ = Vec{3,T}(0, 0, 1)
+  V = T <: AbstractFloat ? T : Float64
+  e⃗ = Vec{3,V}(0, 0, 1)
   push!(points, c + r*e⃗)
   push!(points, c - r*e⃗)
 
--- a/src/sampling/regular.jl
+++ b/src/sampling/regular.jl
@@ -64,7 +64,7 @@ function sample(::AbstractRNG, sphere::Sphere{3,T},
   θrange = range(θmin+δθ, stop=θmax-δθ, length=sz[1])
   φrange = range(φmin, stop=φmax-δφ, length=sz[2])
 
-  r⃗(θ, φ) = Vec{3,V}(r*sin(θ)*cos(φ), r*sin(θ)*sin(φ), r*cos(θ))
+  r⃗(θ, φ) = Vec{3,T}(r*sin(θ)*cos(φ), r*sin(θ)*sin(φ), r*cos(θ))
 
   ivec(c + r⃗(θ, φ) for θ in θrange, φ in φrange)
 end


src/discretization/regular.jl
```diff
@@ -57,7 +57,8 @@ function discretize(sphere::Sphere{3,T},
   # add north and south poles
   c = center(sphere)
   r = radius(sphere)
-  e⃗ = Vec{3,T}(0, 0, 1)
+  V = T <: AbstractFloat ? T : Float64
+  e⃗ = Vec{3,V}(0, 0, 1)
```
juliohm: This feels like a hack. Can you please explain why we need to hard-code Float64 here? You raised this Unitful issue before and we need to take it seriously.

hx-guo: OK.

I admit it looks a little unnatural. But this line of code was inspired by your code. You have wrote it in `src/sampling/regular.jl`. I didn't think much about it as I was writing it down.

Now that you brought this up, after thinking about it, I think it would be more appropriate to write

```julia
V = typeof(one(T))
```

Since `push!(points, c + r*e⃗)` follows immediately below, `e⃗` should be a dimensionless number. The `one` function is used to obtain the dimensionless unit quantity corresponding to a variable.

GuoHaoxuan:
@juliohm Now I changed the hard-coded. What about now?

src/discretization/regular.jl
juliohm: That is a good point. I wonder if we can write down a general recipe for dealing with unitful quantities in these types of calculations. Can you propose a fix using `oneunit` or `one`, whichever is appropriate?

More generally, say we have coordinates x = (x1, x2, x3). These coordinates are written with respect to a basis and an origin point x = x1*E1 + x2*E2 + x3*E3 + O. Should these coordinates have units? Can we say that unitful quantities form a field? When we think more generally about vector spaces over a field, we assume specific properties for the scalars x1, x2, x3.

hx-guo: The Unitful developers said [`one` is the multiplicative identity, i.e., `x * one(x) == x`. There is also `oneunit(3u"mm") == 1u"mm"`.](https://github.com/PainterQubits/Unitful.jl/issues/508#issuecomment-994529657)

Thus, we know. `one` is multiplicative identity, could act as a factor. `oneunit` is one with the same unit. With the question x' = Ex + O, x and O should have the same unit like meter. E as a factor should not have any unit but have the same integer or float expression. Which means `typeof(one(E)) == typeof(one(x))`.

What we tend to confuse is whether this is a coefficient or a real physical quantity. If it is a coefficient, i.e. a scaling factor, direction vector, etc., it should not carry units. Whereas if it is a real physical quantity, e.g. position (coordinates), displacement (difference of coordinates) should have units.

juliohm: That is an interesting summary, thanks for sharing. So the rule of thumb is to always use units with coordinates and avoid units with scaling factors.

src/discretization/regular.jl
```diff
@@ -57,7 +57,8 @@ function discretize(sphere::Sphere{3,T},
   # add north and south poles
   c = center(sphere)
   r = radius(sphere)
-  e⃗ = Vec{3,T}(0, 0, 1)
+  V = typeof(one(T))
```
juliohm: Could you please confirm that this is an alternative method for unit cancellation? https://painterqubits.github.io/Unitful.jl/stable/conversion/#Unit-cancellation

hx-guo: Good question.

This is indeed a good way to eliminate units. But there are two problems with this approach.

1. This method can only be used to eliminate units that appear to be units for quantities that do not already have them. Which means m kN J^-1 equals to 1000, NoUnits just convert it to 1000. `1u"mm" |> NoUnits` will throw an error because millimeter could not convert to number. We want to drop the unit instead of convert it into numbers. We need a scaling factor instead of actual quantities.

2. `NoUnits` will make Unitful as our dependences which is not expected. Our package should be general. Not only Unitful but also other Unitful like packages should work well. Thus, we should not use any Unitful like packages as our dependences only approach it with `one`, `oneunit` and other methods which has already included by Base.

juliohm: Fully agree. Thank you for checking the nuances. There is also ustrip to "get the number in front" of the units but that would also make Unitful.jl a dependency. I think we can introduce a helper function in our utils.jl file if it already exists to drop units, something like `dropunits(T) = typeof(one(T))` and then use it consistently in places of the code where we find scaling factors. Can you adapt this PR accordingly?

GuoHaoxuan:
I changed it for the method this issue mentioned.

The same problem may appears in lots of places in this packages which  have to fix in the future.

src/utils.jl
```diff
@@ -85,4 +85,12 @@ function proj2D(points::AbstractVector{Point{3,T}}) where {T}
   u = U[:,1]
   v = U[:,2]
   [Point(z⋅u, z⋅v) for z in eachcol(Z)]
-end
\ No newline at end of file
+end
+
+"""
+    dropunits(T)
```
juliohm, line 91: Thinking better, we need a different name because this dropunits function is acting on values and returning the type of the unitless type. It is not just dropping units. What about we introduce two helper functions dropunits(T) that takes a type and returns a type and utypeof(value) that takes a value and returns the unitless type? We could also take this opportunity and introduce a third function ueltype for the unitless eltype. What do you think? I think with these three helper functions we will be able to gradually fix implementations throughout the project.

juliohm: Also, it would be nice if we could add tests for these functions by adding Unitful.jl to the test dependencies if it is not already there. Maybe we should do this is in a separate PR and then rebase this PR here with the fix to the specific implementation.

hx-guo: `dropunits`, `utypeof` and `ueltype` sound good, but they just sound. We're probably making simple things more complicated, after all, even without any helper functions we're only going to do what we can do in 14 characters. Wrapping something that short once is meant to be readable and easy to modify, while wrapping it 3 times somewhat complicates a simple problem. In addition, the u prefix actually implies unsigned, `ueltype` rather than `dropunits` brings more ambiguity. So, to sum up, I do not recommend this continued encapsulation of the modification.

Adding Unitful to the test is a big job. Rome was not built in a day. This package was not designed well at the beginning with the case with units in mind, for example even the definition of Ray is something that needs to be modified. We have to fix the current PR first, and then improve such a big project little by little. It is not practical to expect to fix them all in one PR.

juliohm: > `dropunits`, `utypeof` and `ueltype` sound good, but they just sound. We're probably making simple things more complicated, after all, even without any helper functions we're only going to do what we can do in 14 characters.

I don't understand how splitting different functions with different use cases complicates things, can you elaborate? You mean that we should stick to a single function name `dropunits` and write various methods for types, values, and collections of values? I think we can stick to `dropunits` for the first two use cases, but for vectors of values we need a separate function if that is actually necessary.

> Adding Unitful to the test is a big job. Rome was not built in a day. This package was not designed well at the beginning with the case with units in mind, for example even the definition of Ray is something that needs to be modified. We have to fix the current PR first, and then improve such a big project little by little. It is not practical to expect to fix them all in one PR.

I think you misunderstood my suggestion. The suggestion is to write a self-contained PR with the dropunits function and a few tests for this function, nothing more. Fixing units in the entire codebase in a single PR is not reasonable. After this dropunits function is in master with its functionality tested and documented, we can then come back to this PR here and use the dropunits function on it. Does it make sense?

hx-guo: > I don't understand how splitting different functions with different use cases complicates things, can you elaborate? You mean that we should stick to a single function name dropunits and write various methods for types, values, and collections of values? I think we can stick to dropunits for the first two use cases, but for vectors of values we need a separate function if that is actually necessary.

We may have misunderstood each other's meaning. Let me rephrase my point.

First, it is true that the treatment of scalars and the treatment of vectors should be separated. Using a function that acts on both scalars and vectors can cause confusion.

In the case of scalars, we find that one and oneunit can be used on both values and types. Since Base has already done this, there is nothing wrong with designing dropunits that can be used on both types and values, just like Base. dropunits themselves do not need to be split up.

In the case of vectors, it is not possible to continue using dropunits. But we can nest and use a structure like dropunits(eltype(v)). It is not a bad idea to give this nested function another name. But ueltype is reminiscent of unsigned eltype, which is not a good name.

hx-guo: > I think you misunderstood my suggestion. The suggestion is to write a self-contained PR with the dropunits function and a few tests for this function, nothing more. Fixing units in the entire codebase in a single PR is not reasonable. After this dropunits function is in master with its functionality tested and documented, we can then come back to this PR here and use the dropunits function on it. Does it make sense?

Here is where I misunderstood, I thought you wanted to fix the whole project in one PR. It's a good idea to open a PR just for adding these few functions.

juliohm: @HaoxuanGuo would you like to work on this PR for the dropunits utility function with tests? We can then come back to this PR later and rebase if necessary.

hx-guo: @juliohm Of course. I would like to. Sorry for the PR closed due to my wrong operation.

PR #256, opened 2022-04-15
Refactor intersection API

juliohm:
Try to address #54 .

This PR replaces the intersection types by an enum type and replaces the function intersecttype by the function intersection. It passes all tests except for those tests related to type inference. The next step consists of converting the geometries to a buffer with `reinterpret` and then reconstructing the buffer at a later moment.


juliohm:
I will merge this as an improvement to the status quo. This PR replaces the large number of subtypes of intersection to a simple Enum and keeps the first-argument function to avoid type inference issues. A next PR could try to address this further by converting the geometries to buffers.

PR #258, opened 2022-04-22
Add missing case to intersection of lines

dorn-gerhard:
Add Base.in function to ray.jl

### DIFF
--- a/src/intersections/lines.jl
+++ b/src/intersections/lines.jl
@@ -9,7 +9,7 @@ The intersection type can be one of three types:
 2. overlap at more than one point
 3. do not overlap nor intersect
 =#
-function intersection(f::Function, l1::Line{3,T}, l2::Line{3,T}) where {T}
+function intersection(f, l1::Line{3,T}, l2::Line{3,T}) where {T}
   a, b = l1(0), l1(1)
   c, d = l2(0), l2(1)
 
@@ -40,7 +40,7 @@ function intersection(f::Function, l1::Line{2,T}, l2::Line{2,T}) where {T}
 end
 
 # compute the intersection of two lines assuming that it is a point
-function intersectpoint(l1::Line{2}, l2::Line{2})
+function intersectpoint(l1::Line, l2::Line)
   a, b = l1(0), l1(1)
   c, d = l2(0), l2(1)
 
@@ -55,20 +55,3 @@ function intersectpoint(l1::Line{2}, l2::Line{2})
   a - s*v1
 end
 
-# compute the intersection of two lines assuming that it is a point 
-# same code as for 2D case should also work in 3D by ignoring third component
-# TODO: find a better method
-function intersectpoint(l1::Line{3}, l2::Line{3})
-  a, b = l1(0), l1(1)
-  c, d = l2(0), l2(1)
-
-  v1  = a - b
-  v2  = c - d
-  v12 = a - c
-
-  # the intersection point lies in between a and b at a fraction s
-  # (https://en.wikipedia.org/wiki/Line-line_intersection#Formulas)
-  s = (v12[1] * v2[2] - v12[2] * v2[1]) / (v1[1] * v2[2] - v1[2] * v2[1])
-
-  a - s*v1
-end
--- a/src/intersections/lines.jl
+++ b/src/intersections/lines.jl
@@ -26,7 +26,7 @@ function intersection(f, l1::Line{3,T}, l2::Line{3,T}) where {T}
   end
 end
 
-function intersection(f::Function, l1::Line{2,T}, l2::Line{2,T}) where {T}
+function intersection(f, l1::Line{2,T}, l2::Line{2,T}) where {T}
   a, b = l1(0), l1(1)
   c, d = l2(0), l2(1)
 
--- a/src/intersections/lines.jl
+++ b/src/intersections/lines.jl
@@ -9,20 +9,24 @@ The intersection type can be one of three types:
 2. overlap at more than one point
 3. do not overlap nor intersect
 =#
-function intersection(f, l1::Line{3,T}, l2::Line{3,T}) where {T}
+function intersection(f::Function, l1::Line{3,T}, l2::Line{3,T}) where {T}
   a, b = l1(0), l1(1)
   c, d = l2(0), l2(1)
 
-  if measure(Tetrahedron(a, b, c, d)) > 0
+  if measure(Tetrahedron(a, b, c, d)) > 0 # not in same plane
     return @IT NoIntersection nothing f
-  elseif isapprox(abs((b - a) × (c - d)), zero(T), atol=atol(T)^2)
-    return @IT OverlappingLines l1 f
+  elseif isapprox(norm((b - a) × (c - d)), zero(T), atol=atol(T)^2)
+    if a in l2
+      return @IT OverlappingLines l1 f
+    else # parallel lines
+      return @IT NoIntersection nothing f
+    end
   else
     return @IT CrossingLines intersectpoint(l1, l2) f
   end
 end
 
-function intersection(f, l1::Line{2,T}, l2::Line{2,T}) where {T}
+function intersection(f::Function, l1::Line{2,T}, l2::Line{2,T}) where {T}
   a, b = l1(0), l1(1)
   c, d = l2(0), l2(1)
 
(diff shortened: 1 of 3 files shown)


src/intersections/lines.jl
```diff
@@ -9,20 +9,24 @@ The intersection type can be one of three types:
...
+function intersection(f::Function, l1::Line{3,T}, l2::Line{3,T}) where {T}
   a, b = l1(0), l1(1)
   c, d = l2(0), l2(1)
 
-  if measure(Tetrahedron(a, b, c, d)) > 0
+  if measure(Tetrahedron(a, b, c, d)) > 0 # not in same plane
     return @IT NoIntersection nothing f
-  elseif isapprox(abs((b - a) × (c - d)), zero(T), atol=atol(T)^2)
-    return @IT OverlappingLines l1 f
+  elseif isapprox(norm((b - a) × (c - d)), zero(T), atol=atol(T)^2)
```
dorn-gerhard, line 18: in 3D the cross product returns a vector so `abs` yielded an error here

src/intersections/lines.jl
```diff
@@ -9,20 +9,24 @@ The intersection type can be one of three types:
...
+    if a in l2
```
dorn-gerhard, line 19: at this point we know the two lines are in the same plane and parallel,
now we have to check if the lines are collinear

src/intersections/lines.jl
```diff
@@ -9,20 +9,24 @@ The intersection type can be one of three types:
...
+    else # parallel lines
+      return @IT NoIntersection nothing f
+    end
   else
     return @IT CrossingLines intersectpoint(l1, l2) f
   end
 end
 
-function intersection(f, l1::Line{2,T}, l2::Line{2,T}) where {T}
+function intersection(f::Function, l1::Line{2,T}, l2::Line{2,T}) where {T}
```
dorn-gerhard: I added the type of `f`. 
Not sure if it was intended to keep it unspecified

src/intersections/lines.jl
```diff
@@ -50,3 +54,21 @@ function intersectpoint(l1::Line{2}, l2::Line{2})
 
   a - s*v1
 end
+
```
dorn-gerhard: `intersectpoint` for 3D was missing, it's the same code as for 2D just taking into account the first two coordinates.
since the **purpose of intersectpoint** is not to check whether 2 objects intersect but to return the crossing point

Applying intersectpoint to objects that do not intersect would yield a Point which lies on the first line

src/primitives/ray.jl
```diff
@@ -42,3 +42,13 @@ origin(r::Ray) = r.p
 The direction of the ray.
 """
 direction(r::Ray) = r.v
+
+function Base.in(p::Point, r::Ray)
+	w = norm(r.v)
+	d = evaluate(Euclidean(), p, Line(r(0), r(1)))
+	if d+w ≈ w
+		return sum((p - r.p) .* r.v) >= 0 # return true if point is in ray
```
dorn-gerhard: The only thing one has to take into account here, compared to `Base.in(p::Point, l::Line)` is to check whether the ray is hitting the point. `p - r.p` yields the vector pointing from the origin to the point (could be zero)
Multiplication with `r.v` yields positive numbers if it is in the direction otherwise negative

src/intersections/lines.jl
```diff
@@ -9,20 +9,24 @@ The intersection type can be one of three types:
 2. overlap at more than one point
 3. do not overlap nor intersect
 =#
-function intersection(f, l1::Line{3,T}, l2::Line{3,T}) where {T}
+function intersection(f::Function, l1::Line{3,T}, l2::Line{3,T}) where {T}
```
juliohm: We don't need to explicitly request a `Funcion`, we can accept any `Base.Callable`. We only enforce types in the arguments when it helps with multiple dispatch.

src/intersections/lines.jl
juliohm: Interesting. Our tests are not covering this case then? We should add tests to make sure that this branch is covered.

src/intersections/lines.jl
juliohm: I am surprised that the algorithm was missing important branches. Let's wait for the tests to finish running to see if all previous tests are still passing.

src/intersections/lines.jl
juliohm: Yes, it was intended. We should remove type annotations when they don't help with multiple dispatch. The code is more flexible that way.

src/intersections/lines.jl
juliohm: Can you please clarify the exact differences between the 2D and 3D implementations? By reading them quickly with the naked eyes I am seeing no differences. If they are the same implementation, maybe we just need to relax the type of the arguments instead of duplicating code?

src/primitives/ray.jl
```diff
@@ -42,3 +42,13 @@ origin(r::Ray) = r.p
...
+	else
+		return false
+	end
+end
```
juliohm:
```suggestion
```

This function is out of the scope of this PR. We can add it in a separate PR after we fix the bugs in the intersection you mentioned. The more self-contained the better. In the future we can come back to this PR and understand the fix.

juliohm (changes requested):

juliohm:
ping @dorn-gerhard

src/intersections/lines.jl
dorn-gerhard: add the following test case for 3D line intersections: `intersect(Line((1., 0, 0), (0,0,0)), Line((0., 1, 0), (0,0,0)))`

src/intersections/lines.jl
dorn-gerhard: Add the test `intersect(Line((1,0,0), (0,0,0)), Line((1,1,0), (0,1,0)))`

src/intersections/lines.jl
juliohm: This issue is not resolved, can you please clarify why you requested a new round of reviews when the code remains unchanged?

src/intersections/lines.jl
juliohm: You need to add the tests in the PR so that GitHub will run them and check correctness. We cannot merge this without tests included. Can you please add the tests to the PR?

dorn-gerhard: ok, I will do, so I just add test cases in the `/test/intersections.jl` file?
sorry for the inconvenience - first PR

src/intersections/lines.jl
dorn-gerhard: Now I removed the type specification

src/intersections/lines.jl
juliohm: No worries, the first PR is always a learning curve :) Yes, you can add tests to the corresponding file in the test folder, in this case intersections.jl. There is a specific section for line-line intersection I think. If not, feel free to add one.

juliohm (changes requested):
Changes requested in the comments.

src/intersections/lines.jl
dorn-gerhard: Cool :)
I added 4 tests for 3d lines at the correct place in the test/intersections.jl

dorn-gerhard (commented):
I added the tests for intersecting 3d lines in the test/intersections.jl file

test/intersections.jl
```diff
@@ -509,6 +510,29 @@
...
+    l2 = Line(P3(2,0,0), P3(3,0,0))
+    @test l1 ∩ l2 == l2 ∩ l1 == l1
+
+    # crossing in one point
+    l1 = Line(P3(1,2,3), P3(2,1,0))
+    l2 = Line(P3(1,2,3), P3(1,1,1))
+    @test l1 ∩ l2 == l2 ∩ l1 == P3(1,2,3)
+
+
+
```
```suggestion
```

src/intersections/lines.jl
```diff
@@ -50,3 +54,4 @@ function intersectpoint(l1::Line{2}, l2::Line{2})
 
   a - s*v1
 end
+
```
```suggestion
```

juliohm (approved):

juliohm:
Awesome first PR @dorn-gerhard 👏🏽 all tests are passing. Merging it now, thank you for the fix!

PR #259, opened 2022-04-23
Add function dropunits

HaoxuanGuo:
I added `dropunits` function and its tests as [we discussed](https://github.com/JuliaGeometry/Meshes.jl/pull/255).

We had some differences of opinion before. But in general, it is consistent to construct a dropunits function for getting unitless scalars. Hence this PR.

Sorry for [the PR](https://github.com/JuliaGeometry/Meshes.jl/pull/255) that was turned off by mistake.

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -90,7 +90,7 @@ end
 """
     dropunits(T)
 
-Gets the dimensionless memory type representation of a variable or type.
+Returns the unitless type of a (unitful) type or value. See Unitful.jl.
 i.e. `dropunits(1u"mm") == Int`
 """
 dropunits(T) = typeof(one(T))
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -86,3 +86,11 @@ function proj2D(points::AbstractVector{Point{3,T}}) where {T}
   v = U[:,2]
   [Point(z⋅u, z⋅v) for z in eachcol(Z)]
 end
+
+"""
+    dropunits(T)
+
+Gets the dimensionless memory type representation of a variable or type.
+i.e. `dropunits(1u"mm") == Int`
+"""
+dropunits(T) = typeof(one(T))
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -93,4 +93,4 @@ end
 Returns the unitless type of a (unitful) type or value. See Unitful.jl.
 i.e. `dropunits(1u"mm") == Int`
 """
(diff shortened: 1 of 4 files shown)


src/utils.jl
```diff
@@ -85,4 +85,12 @@ function proj2D(points::AbstractVector{Point{3,T}}) where {T}
...
   v = U[:,2]
   [Point(z⋅u, z⋅v) for z in eachcol(Z)]
-end
\ No newline at end of file
+end
+
+"""
+    dropunits(T)
+
+Gets the dimensionless memory type representation of a variable or type.
```
```suggestion
Returns the unitless type of a (unitful) type or value. See Unitful.jl.
```

test/runtests.jl
```diff
@@ -11,6 +11,7 @@ using SparseArrays
 using PlyIO
 using Test, Random, Plots
 using ReferenceTests, ImageIO
+using Unitful
```
juliohm: Please move this to the line before the line with `using Test`, Random, Plots`

src/utils.jl
```diff
@@ -85,4 +85,12 @@ function proj2D(points::AbstractVector{Point{3,T}}) where {T}
...
+i.e. `dropunits(1u"mm") == Int`
+"""
+dropunits(T) = typeof(one(T))
```
```suggestion
dropunits(v) = typeof(one(v))
```

test/utils.jl
```diff
@@ -15,4 +15,7 @@
   @test sideof(p1, c) == :INSIDE
   @test sideof(p2, c) == :OUTSIDE
   @test sideof(p3, c) == :INSIDE
+
+  @test Meshes.dropunits(1.0u"mm") === Float64
+  @test Meshes.dropunits(typeof(1.0u"mm")) === Float64
```
```suggestion
  # drop units from unitful value and type
  @test Meshes.dropunits(1.0u"mm") == Float64
  @test Meshes.dropunits(typeof(1.0u"mm")) == Float64
  
  # return the same type in case of no units
  @test Meshes.dropunits(1.0) == Float64
  @test Meshes.dropunits(Float64) == Float64
```

juliohm (changes requested):

test/runtests.jl
hx-guo: OK.

juliohm (approved):

juliohm:
Thanks for the addition @HaoxuanGuo , feel free to open PRs fixing specific places with the new dropunits 💯

Issue #267, opened 2022-05-27
Out of memory with 8 triangles


ederag:
Uncommenting the last line eats about 1Gb per second, with no end (it froze my desktop once),
so be prepared to interrupt.

```julia
using Meshes

points = [
	Point(-0.5, -0.8660254037844386)
    Point(0.5, -0.8660254037844386)
    Point(1.5, -0.8660254037844386)
    Point(-1.0, 0.0)
    Point(0.0, 0.0)
    Point(1.0, 0.0)
    Point(-0.5, 0.8660254037844386)
    Point(0.5, 0.8660254037844386)
    Point(1.5, 0.8660254037844386)
]

triangles = [
	(1, 2, 5)
    (1, 4, 5)
    (2, 3, 6)
    (2, 5, 6)
    (4, 5, 7)
    (5, 6, 8)
    (5, 7, 8)
    (6, 8, 9)
]

mesh = Meshes.SimpleMesh(points, connect.(triangles))

# Beware, this eats all the RAM quickly (few 10s of seconds) !
# Meshes.refine(mesh, Meshes.TriRefinement())
```

Another clue:
`MeshViz.viz(mesh)` works fine, but `viz(mesh, showfacets=true)` eats all the RAM as well.

Tested with `Meshes v0.22.5` and `julia-1.6.5`.

juliohm:
You are not ordering the connectivities coherently. Please make sure that they are all counter-clockwise, otherwise the underlying half-edge structure will break in undefined ways. See also #137.

ederag:
Thanks, I had read the manual (nicely written, by the way) but overlooked
https://juliageometry.github.io/Meshes.jl/stable/geometries/polytopes.html#Meshes.Ngon
and
https://juliageometry.github.io/Meshes.jl/stable/geometries/polytopes.html#Meshes.PolyArea
and did not find #137.

But as a feedback, I'm surprised that this is not considered as a bug
and kept open to warn other users.
Crashing a whole desktop with a toy example should be avoided, IMO.

At the very least, that limitation should be mentioned in
https://juliageometry.github.io/Meshes.jl/stable/meshes.html#Meshes.SimpleMesh
(which I did read before usage)

juliohm:
> But as a feedback, I'm surprised that this is not considered as a bug
> and kept open to warn other users.
> Crashing a whole desktop with a toy example should be avoided, IMO.

[...]

PR #268, opened 2022-05-27
Mention orientation in SimpleMesh doc

ederag:
Follow-up to https://github.com/JuliaGeometry/Meshes.jl/issues/267.

### DIFF
--- a/src/mesh/simplemesh.jl
+++ b/src/mesh/simplemesh.jl
@@ -6,6 +6,9 @@
     SimpleMesh(points, connec)
 
 A simple mesh with `points` and connectivities `connec`.
+Note: please make sure all the connectivity polygons have
+`orientation(polygon) === :CCW`.
+
 The i-th face of the mesh is lazily built based on
 the connectivity list `connec[i]`.
 


src/mesh/simplemesh.jl
```diff
@@ -6,6 +6,9 @@
     SimpleMesh(points, connec)
 
 A simple mesh with `points` and connectivities `connec`.
+Note: please make sure all the connectivity polygons have
+`orientation(polygon) === :CCW`.
+
```
juliohm, line 11: Please follow the convention for notes in docstring in other docstrings. We have notes at the end of the docstring in a separate section. Also notes usually don't use code, they just describe the situation in simple English words.

juliohm (changes requested):
Review attached.

src/mesh/simplemesh.jl
ederag: This note has nothing to do with the next part, that's why I put it there.

And now that made some progress, I'm not sure how to phrase this.
CCW makes no sense in 3D, and the limitation might just be for 2D, but I'm not sure.
Giving up.

juliohm: > This note has nothing to do with the next part, that's why I put it there.

As I explained all notes in our docstrings appear as the last section. You didn't understand that.

> CCW makes no sense in 3D, and the limitation might just be for 2D, but I'm not sure.

Coherent orientation makes sense in any dimension. The fact that we use the term "CCW" to refer to it is irrelevant.

> Giving up.

We have high standards, sorry.

ederag: > Coherent orientation makes sense in any dimension. The fact that we use the term "CCW" to refer to it is irrelevant.

"CCW" has a very precise meaning, and in 3D it depends on some convention, that would need to be expressed.

> We have high standards, sorry.

That's good. It just became obvious that this domain is too foreign for me to contribute with quality.
Please keep on the good work as you see fit.

PR #269, opened 2022-05-31
Intersection of rays #2

dorn-gerhard:
This PR adds all possible intersections of 2D and 3D rays using the new IntersectionTypes:
CrossingRays
MidTouchingRays
CornerTouchingRays
OverlappingAlignedRays
OverlappingCollidingRays


test/intersections.jl
```diff
@@ -150,6 +150,38 @@
     @inferred someornone(s1, s2)
   end
 
+  @testset "Rays" begin
+    # rays in 2D
```
dorn-gerhard, line 154: tests for 2D

Please give feedback for 2D tests, then I will add tests for 3D

What else has to be tested?

dorn-gerhard (commented):
I redid the PR on intersection of rays starting from the latest version (sorry for the delay - was a week off)

src/intersections/rays.jl
```diff
@@ -0,0 +1,124 @@
...
+# ------------------------------------------------------------------
+
+#=
+The intersection type can be one of six types:
+
+1. intersect at one inner point (CrossingRays -> Point)
+2. intersect at origin of one ray (MidTouchingRays -> Point)
+3. intersect at origin of both rays (CornerTouchingRays -> Point)
+4. overlap with aligned vectors (OverlappingAlignedRays -> Ray)
+5. overlap with colliding vectors (OverlappingCollidingRays -> Segment)
```
juliohm: I understand that (4) is |---> |---> and (5) is |---> <---|. Can you think of alternative names? I think Aligned and Colliding in English have a slightly different meaning. I am trying to think of a better name...

src/intersections/rays.jl
```diff
@@ -0,0 +1,124 @@
...
+      end
+    else
+      if isapprox(p2, zero(T), atol=atol(T))
+        return @IT MidTouchingRays c f # CASE 2: origin of r2
+      else
+        return @IT CrossingRays (a-p1*v1) f #CASE 1: equal to c - p2 * v2
+      end
+    end
+  end
+end
```
juliohm, line 59:
```suggestion
end

```

src/intersections/rays.jl
dorn-gerhard: I thought the context is clear from Overlapping and Aligned. I suggest to keep overlapping as it also appears in the context of other intersections. What about
(4) OverlappingEquidirectionalRays
(5) OverlappingOpposingRays

dorn-gerhard:
> Can you please comment on the differences between this implementation and the other current implementations? I am feeling like we are duplicating code, but I didn't have time to read all methods carefully.
> 
> If the algorithms for segments, rays, lines can be reused or refactored to use a common core algorithm that would be ideal.

I guess this is rather hard, since we use IntersectionTypes that depend on the intersecting objects.

What we could do is a function `intersectparameters` that returns the parameters `p_1` and `p_2` of two pairs of points (a,b), (c,d) that lie in the same plane and form lines to describe the intersection point (so to put the core of the function `intersectpoint(::Line, ::Line)` in an own function)
``` julia
function intersectparameters(a::Point, b::Point, c::Point, d::Point) 
  v1  = a - b
  v2  = c - d
  v12 = a - c
  #QUESTION: is there a better way to calculate parameters in 3D or higher dimension including other components 
  # or does it not matter?
  p1 = (v12[1] * v2[2] - v12[2] * v2[1]) / (v1[1] * v2[2] - v1[2] * v2[1])
  p2 = (v12[1] * v1[2] - v12[2] * v1[1]) / (v1[1] * v2[2] - v1[2] * v2[1])
  p1, p2
end
```
and a second version:
``` julia
function intersectparameters(l1::Line, l2::Line)
  a = l1(0), b = l1(1)
  c = l2(0), d = l2(1)
  intersectparameters(a,b,c,d)
end
```

to be included in `intersectpoint`:
``` julia
function intersectpoint(l1::Line, l2::Line)
 p1, p2 = intersectparameters(l1, l2)
 l1(0) + p1 * (l1(1) - l1(0))
end
```

What do you think?

dorn-gerhard (commented):
In the comments I proposed two new names for the two types of OverlappingRays. 
In the comments I proposed to introduce a new function intersectparameters to return the parameters p_1 and p_2 of two intersecting lines (those parameters will be needed frequently when intersecting segments with rays or lines.

src/intersections/rays.jl
juliohm: OverlappingAgreeingRays and OverlappingOpposingRays is great 💯

juliohm:
I like the suggestion of introducing a new helper function to return the two points used to compute the final intersection point. Maybe twopoints(a, b, c, d) would be a good name. Can you please try to update the PR with these suggestions so that we see if the redundancy can be reduced? The more redundancy we eliminate the less bugs we will need to fix.

src/intersections/rays.jl
dorn-gerhard: implemented

dorn-gerhard:
I added the helper function with the name `intersectparameters`, since 
it does not calculate points but parameters. The name also fits the convention of `intersectpoint`

src/intersections/lines.jl
```diff
@@ -41,16 +41,23 @@ end
 
 # compute the intersection of two lines assuming that it is a point
 function intersectpoint(l1::Line, l2::Line)
-  a, b = l1(0), l1(1)
-  c, d = l2(0), l2(1)
+  p1 = first(intersectparameters(l1, l2))
+  l1(0) + p1 * (l1(1) - l1(0))
```
juliohm: Please restructure this code so that the function call is in its own line as

```julia
p1, p2 = intersectparameters(a, b, c, d)
```

and then the final expression becomes:

```julia
a + p1 * (b - a)
```

I think we only need a single function `intersectparameters(a, b, c, d)`. I saw that you added methods with `intersectparameters(l1, l2)` but that is going to become messy with Ray and Segment. We can always resort to the low-level method with 4 points instead.

juliohm (changes requested):
Hi @dorn-gerhard , thanks again for improving the code. There are too many changes here that are difficult to review safely and I would like to propose the start a simpler PR where you can introduce the `intersectparameters` function and refactor the existing intersections between lines and segments in terms of it.

After that we can see if something else needs to be refactored in order to accommodate the rays case with minimum code repetition.

src/intersections/lines.jl
dorn-gerhard: Sure, I'll implement it, good suggestion

dorn-gerhard:
> 

So your suggestion is to postpone the rays intersection pr and do a intersectparameters pr first?

src/intersections/lines.jl
dorn-gerhard: By the way, looking at segments.jl (intersection of segments)
the variant working with matrix `A`, the rank of augmented / simple matrix also returns the parameters and gives information about whether the lines / segments intersect or not

``` julia
 # get coordinates
  p₁, p₂ = coordinates.(vertices(s1))
  q₁, q₂ = coordinates.(vertices(s2))

  # create matrices and vectors to allow for evaluation of intersection
  A = [(p₂ - p₁) (q₁ - q₂)]
  b = q₁ - p₁

  λ = A \ b
```
I will think about how to use this aspect in the `intersectparameters` function
One idea could be to return the two ranks
``` julia
  # calculate the rank of the augmented matrix
  rₐ = rank([A b])
  # calculate the rank of the rectangular matrix
  r = rank(A)
```

to have 
`p1, p2, r_a, r = intersectparameters(a, b, c, d)`

What do you think?

juliohm: We can discuss these details in the new PR that will be introducing the intersectparameters function. To help future contributors understand our choices.

juliohm:
As we discussed, let's close this PR and start a new one with just the intersectparameters function.

Issue #270, opened 2022-06-03
Unclear documentation: subtracting Points


nick-adamowicz:
https://github.com/JuliaGeometry/Meshes.jl/blob/11855a2aaf0688fffae4678635bef08b377fb4d4/src/points.jl#L98-L104

The current documentation lists:

"Return the Vec associated with the direction from point A to point B."

wouldn't it be more accurate to say:

"Return the Vec associated with the direction from point B to point A."

since, e.g.:

```
pa = Point(1,1) # point a
pb = Point(5,8) # point b
dp = pa - pb # difference vector between points a & b
pb + dp == pa # => true
```

... or am I just confused :) ...

juliohm:
Good catch @fujiwara-tofu-tm , could you please submit a PR with the fix?

nick-adamowicz:
this is the first time I've ever made a pull request, so please bear with me. Did I do it correctly?

juliohm:
You did it perfectly 💯 Merged and should be included in the next release. Thank you!

PR #271, opened 2022-06-03
fix Point subtraction docstring

fujiwara-tofu-tm:
fix for issue #270


juliohm:
Thank you for the help ❤️

Issue #272, opened 2022-06-05
documentation issue, RegularSampling


nick-adamowicz:
https://github.com/JuliaGeometry/Meshes.jl/blob/4be3b986e7988cd61157f81b193964b0204f44fe/src/sampling/regular.jl#L5-L19

slight typo in ``... `np` poitns along ...`` -> ``... `np` points along ...``

also, the provided example:

```
sample(Sphere((0,0,0), 1), RegularSampling(360, 180))
```

throws an error since the following constructor doesn't convert the provided radius to a Float32 or Float64 (depending on the form of the constructed Point):

https://github.com/JuliaGeometry/Meshes.jl/blob/4be3b986e7988cd61157f81b193964b0204f44fe/src/primitives/sphere.jl#L17

suggested change to RegularSampling documentation:

```
sample(Sphere((0,0,0), 1), RegularSampling(360, 180))
```
->
```
sample(Sphere((0,0,0), 1.0), RegularSampling(360, 180))
```

... or I guess we could modify the Sphere constructor ....

juliohm:
Thank you @fujiwara-tofu-tm for catching these small typos and issues ❤️ These are super helpful contributions. Would you mind submitting a PR to fix the Sphere constructor? We need to fix many constructors that are still using integer arguments to convert to floating point.

Also, feel free to submit a separate PR with the typo fix in the RegularSampling docstring to replace poitns -> points

juliohm:
Fixed on master.

PR #273, opened 2022-06-05
fix typo in RegularSampling docstring

fujiwara-tofu-tm:
issue #272, typo fix


nick-adamowicz:
How does this code coverage check work? Not sure why it's failing since I just flipped around two letters in the documentation?

juliohm:
I think it is a bug of the codecov counts. Sometimes these tests fail, we need to investigate deeper why.

PR #274, opened 2022-06-05
add sphere constructor given Point

fujiwara-tofu-tm:
Adding another constructor to Sphere so that it can accept integer arguments.

When given a Point as the first argument to the constructor, this version tries to convert the provided radius to the element type associated to the given Point.

### DIFF
--- a/src/primitives/sphere.jl
+++ b/src/primitives/sphere.jl
@@ -15,6 +15,7 @@ struct Sphere{Dim,T} <: Primitive{Dim,T}
 end
 
 Sphere(center::Point{Dim,T}, radius) where {Dim,T} = Sphere(center, T(radius))
+
 Sphere(center::Tuple, radius) = Sphere(Point(center), radius)
 
 """
--- a/src/primitives/sphere.jl
+++ b/src/primitives/sphere.jl
@@ -14,6 +14,7 @@ struct Sphere{Dim,T} <: Primitive{Dim,T}
   radius::T
 end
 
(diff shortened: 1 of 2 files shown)


nick-adamowicz:
issue #272

juliohm:
That is awesome @fujiwara-tofu-tm , can you also add some new tests to capture the fact that you fixed the bug? That is, just add a few more tests in the Sphere section of our test suite constructing a Sphere with integer radius and making sure that the end result is converted correctly.

You can update a PR easily here on GitHub by pressing the "." on your keyboard. This will open VSCode on the browser and you can keep editing it. Alternatively, you can push more commits to the same branch of your fork and the PR will be updated.

src/primitives/sphere.jl
```diff
@@ -14,6 +14,7 @@ struct Sphere{Dim,T} <: Primitive{Dim,T}
   radius::T
 end
 
+Sphere(center::Point{Dim,T}, radius) where {Dim,T} = Sphere(center, T(radius))
```
juliohm, line 17:
```suggestion
Sphere(center::Point{Dim,T}, radius) where {Dim,T} = Sphere(center, T(radius))

```

juliohm (approved):

juliohm:
Thank you for another fix @fujiwara-tofu-tm ❤️ There are various other constructors that need similar fixes. It would be super helpful if you could submit PRs for them as well.

PR #276, opened 2022-06-30
add helper function intersectparameters

dorn-gerhard:


src/intersections/lines.jl
```diff
@@ -9,48 +9,77 @@ The intersection type can be one of three types:
 2. overlap at more than one point
 3. do not overlap nor intersect
 =#
-function intersection(f, l1::Line{3,T}, l2::Line{3,T}) where {T}
+function intersection(f, l1::Line{N,T}, l2::Line{N,T}) where {N,T}
```
dorn-gerhard, line 12: one can define one function for both 2d, 3d or in general Nd

src/intersections/lines.jl
```diff
@@ -9,48 +9,77 @@ The intersection type can be one of three types:
...
+  elseif r == rₐ == 1
+    return @IT OverlappingLines l1 f
+  else
+    @IT NoIntersection nothing f
+  end
+
+  #= 
+  if !isapprox(measure(Tetrahedron(a, b, c, d)), zero(T), atol = atol(T)) # not in same plane
     return @IT NoIntersection nothing f
-  elseif isapprox(norm((b - a) × (c - d)), zero(T), atol=atol(T)^2)
```
dorn-gerhard, line 18: not sure why here ´atol(T)´ was squared, maybe this has to be checked also in other intersection functions.
but anyway this code part can be removed.

src/intersections/lines.jl
```diff
@@ -9,48 +9,77 @@ The intersection type can be one of three types:
...
 # compute the intersection of two lines assuming that it is a point
 function intersectpoint(l1::Line, l2::Line)
   a, b = l1(0), l1(1)
   c, d = l2(0), l2(1)
+  p1, _ = intersectparameters(a,b,c,d)
+  a + p1 * (b - a)
+ end
 
+# compute the intersection parameters of the lines defined by the points a -> b and c -> d 
+function intersectparameters_fast(a::Point, b::Point, c::Point, d::Point)
```
dorn-gerhard, line 50: This is the old version of getting the parameters, 
my suggestion is to remove it. What do you think?

src/intersections/lines.jl
```diff
@@ -9,48 +9,77 @@ The intersection type can be one of three types:
...
+  p2 = (v12[1] * v1[2] - v12[2] * v1[1]) / (v1[1] * v2[2] - v1[2] * v2[1])
+  p1, p2
+end
+
+# compute the parameters of the lines defined by the points a -> b and c -> d 
+# so that the resulting points have minimal distance
+# the ranks help to identify the different types of intersection
+
+function intersectparameters(a::Point{N,T}, b::Point{N,T}, c::Point{N,T}, d::Point{N,T}) where {N,T}
+  # solves the equation (approximately): a + λ₁ ⋅ v⃗₁ = c + λ₂ ⋅ v⃗₂, v⃗₁ = b - a, v⃗₂ = d - c
```
dorn-gerhard, line 68: the only thing for me here to discuss is whether there should be a "fast" version of this function not calculating the ranks.
But I think this rank calculation (including a SVD of a 3 x N matrix) is similar expansive as solving the small linear (over determined) linear system `A \ b`

dorn-gerhard (commented):
Main open question for me is:
- adding a fast `ìntersectparameters` function without calculating the ranks?
Another question maybe related to other intersection functions is
- Why is there a `atol = atol(T)^2` in some functions?

To be done:
- adding `intersectparameters` to `segments.jl`
- some more test cases dealing with tolerances

src/intersections/lines.jl
```diff
@@ -9,48 +9,77 @@ The intersection type can be one of three types:
...
+  if r == rₐ == 2
+    return @IT CrossingLines a + p1 * (b-a) f
+  elseif r == rₐ == 1
+    return @IT OverlappingLines l1 f
+  else
+    @IT NoIntersection nothing f
+  end
+
+  #= 
+  if !isapprox(measure(Tetrahedron(a, b, c, d)), zero(T), atol = atol(T)) # not in same plane
```
juliohm, line 27: PRs with dead code are not ok.

src/intersections/lines.jl
juliohm: Here we are comparing areas (length ^ 2) so we square the tolerance as well. Please only modify the code that is strictly necessary.

src/intersections/lines.jl
```diff
@@ -9,48 +9,77 @@ The intersection type can be one of three types:
...
-  a, b = l1(0), l1(1)
-  c, d = l2(0), l2(1)
-
-  if !isapprox(abs((b - a) × (c - d)), zero(T), atol=atol(T)^2)
-    return @IT CrossingLines intersectpoint(l1, l2) f
-  elseif isapprox(measure(Triangle(a, b, c)), zero(T), atol=atol(T)^2)
-    return @IT OverlappingLines l1 f
-  else
-    return @IT NoIntersection nothing f
-  end
```
juliohm, line 39: Please make sure that the 2D and 3D algorithms are equivalent.

src/intersections/lines.jl
```diff
@@ -9,48 +9,77 @@ The intersection type can be one of three types:
...
+  =#
 end
 
 # compute the intersection of two lines assuming that it is a point
 function intersectpoint(l1::Line, l2::Line)
   a, b = l1(0), l1(1)
   c, d = l2(0), l2(1)
+  p1, _ = intersectparameters(a,b,c,d)
+  a + p1 * (b - a)
```
juliohm, line 46: You are not following the code style with spaces between commas, you are calling variable names p1 when the code uses p\_1 in other places etc.

src/intersections/lines.jl
juliohm: I think this PR is really difficult to review with code that is commented, with unrelated changes, without appropriate code style. I will close it so that you start an appropriate PR for review.

juliohm (changes requested):

PR #277, opened 2022-07-01
Helper function intersectparameters for intersection of lines

dorn-gerhard:


dorn-gerhard:
This PR adds a helper function `intersectparameters` independent of dimension `Dims`
that returns the parameters of two lines that describe the point of intersection or the closest points.
Further it returns ranks that ease the classification of intersection types for lines and which can
further be used for intersection of segments (`segments.jl`) and in the future for the intersection of rays, segment+ray, segment+line, ray+line

src/intersections/lines.jl
```diff
@@ -43,14 +28,30 @@ end
...
-  s = (v12[1] * v2[2] - v12[2] * v2[1]) / (v1[1] * v2[2] - v1[2] * v2[1])
-
-  a - s*v1
+  p₁, _ = intersectparameters(a, b, c, d)
+  a + p₁ * (b - a)
+ end
+
+# compute the parameters of the lines defined by the points a -> b and c -> d 
+# so that the resulting points have minimal distance
+# the ranks help to identify the different types of intersection
```
```suggestion
# compute the intersection parameters of the lines a--b and c--d 
# returned ranks help to identify the different types of intersection
```

src/intersections/lines.jl
```diff
@@ -43,14 +28,30 @@ end
...
+function intersectparameters(a::Point{N,T}, b::Point{N,T}, c::Point{N,T}, d::Point{N,T}) where {N,T}
+  # solves the equation (approximately): a + λ₁ ⋅ v⃗₁ = c + λ₂ ⋅ v⃗₂, v⃗₁ = b - a, v⃗₂ = d - c
```
juliohm: Please add a breakline after the second argument of the function so that a and b show in one line and c and d in a second line.

Similarly, split the long comment into multiple lines.

src/intersections/lines.jl
```diff
@@ -43,14 +28,30 @@ end
...
+  A = [(b - a) (c - d)]
+  b = c - a
+
+  λ = A \ b
+
+  # calculate the rank of the augmented matrix
+  rₐ = rank([A b], atol = atol(T))
+  # calculate the rank of the rectangular matrix
+  r = rank(A, atol = atol(T))
```
juliohm: Can we avoid doing Gaussian elimination of the system A \ b twice? Maybe we can obtain the rank from some matrix factorization such as LU? Worth checking if this would lead to any considerable performance boost.

src/intersections/lines.jl
```diff
@@ -43,14 +28,30 @@ end
...
+
+  # Intersection r == rₐ == 2
+  # Collinear: r == rₐ == 1
+  # no intersection: r != rₐ
+  #   no intersection and parallel:  r == 1, rₐ == 2
+  #   no intersection, skew lines: r == 2, rₐ == 3
+
+  λ[1], λ[2], r, rₐ
```
juliohm, line 77: What do you think of just returning \lambda, r, r_a?

src/intersections/lines.jl
```diff
@@ -43,14 +28,30 @@ end
...
-  v1  = a - b
-  v2  = c - d
-  v12 = a - c
-
-  # the intersection point lies in between a and b at a fraction s
-  # (https://en.wikipedia.org/wiki/Line-line_intersection#Formulas)
-  s = (v12[1] * v2[2] - v12[2] * v2[1]) / (v1[1] * v2[2] - v1[2] * v2[1])
-
-  a - s*v1
+  p₁, _ = intersectparameters(a, b, c, d)
```
juliohm: We need to be consistent with notation. In this case you should name this variable \lambda right? This is what is returned by the intersectparams function.

src/intersections/lines.jl
```diff
@@ -9,30 +9,15 @@ The intersection type can be one of three types:
...
 
-  if !isapprox(abs((b - a) × (c - d)), zero(T), atol=atol(T)^2)
-    return @IT CrossingLines intersectpoint(l1, l2) f
-  elseif isapprox(measure(Triangle(a, b, c)), zero(T), atol=atol(T)^2)
+  if r == rₐ == 2
+    return @IT CrossingLines (a + p₁ * (b - a)) f
+  elseif r == rₐ == 1
     return @IT OverlappingLines l1 f
   else
     return @IT NoIntersection nothing f
```
juliohm: Can we move this logic with intersection types to inside intersectparams? Why do we need to return ranks? Maybe we can just return the actual intersection type directly?

src/intersections/lines.jl
```diff
@@ -43,14 +28,30 @@ end
...
+  p₁, _ = intersectparameters(a, b, c, d)
+  a + p₁ * (b - a)
+ end
+
+# compute the parameters of the lines defined by the points a -> b and c -> d 
+# so that the resulting points have minimal distance
+# the ranks help to identify the different types of intersection
+function intersectparameters(a::Point{N,T}, b::Point{N,T}, c::Point{N,T}, d::Point{N,T}) where {N,T}
+  # solves the equation (approximately): a + λ₁ ⋅ v⃗₁ = c + λ₂ ⋅ v⃗₂, v⃗₁ = b - a, v⃗₂ = d - c
+  A = [(b - a) (c - d)]
```
juliohm: Notice that A is a heap-allocated matrix. We can use SMatrix from StaticArrays.jl instead to increase performance.

juliohm (changes requested):
Thank you @dorn-gerhard , attached a first round of review.

juliohm:
@dorn-gerhard did you have a chance to take a look into this?

src/intersections/lines.jl
dorn-gerhard: looking ahead for the cases of intersecting segments and rays (with each other or any combination with lines) it might be useful to return ranks instead of the fixed IntersectionTypes for lines since other criteria such as the actual value of the parameter will be needed to find the intersectionType.
And there might be more applications to cases where it is important why two lines are not crossing (e.g. parallel vs. skew lines), we have this information still in the ranks but not in the intersection types.

src/intersections/lines.jl
dorn-gerhard: I agree, I will change it when I am back at my computer.

src/intersections/lines.jl
dorn-gerhard: i don't know a situation where both values are needed in a vector, so when applying this function one would always have to access the individual elements. It will make code using this helper function more readable if we keep lambda[1] and lambda[2] seperated

src/intersections/lines.jl
dorn-gerhard: I will think about this, a benchmark might be helpful to decide.

dorn-gerhard:
@juliohm thank you for the review, I will implement the changes next week as I am travelling this week.

src/intersections/lines.jl
dorn-gerhard: Since `b - a` and `c - d` are SVectors already, `A` is an SMatrix by default.

dorn-gerhard:
``` code
let
  a = Point(0.2, 0.4, -1.0)
  b = Point(1.2, 1.4, 2.0)
  c = Point(0.2, 0.4, -1.0)
  d = Point(3.2, 2.4, 5.0)

  A = [(b - a) (c - d)]
end
```
image

src/intersections/lines.jl
dorn-gerhard: So I tested the approach to calculate the QR decomposition (LU is instable if matrix is singular).
Gaussian elimination (of an upper triangular matrix) is only needed if lines are not parallel or colinear (`r == 2`), (otherwise paramters are set to `zero(T)`).

For the evaluation i need only submatrices of the `Q` and `R` matrices. When accessing them, they are converted back from SArray to normal arrays which makes to evaluation slow (`intersectparameters_3`)
Therefore I implemented an additional QR decomposition on the static matrix A alone, which in total seems to be faster and commited that version  (`intersectparameters_2`) - maybe you have a better solution

image

intersectparameters_2:
``` code
function intersectparameters_2(a::Point{N,T}, b::Point{N,T}, 
                             c::Point{N,T}, d::Point{N,T}) where {N,T}
  # solves the equation (approximately): 
  # a + λ₁ ⋅ v⃗₁ = c + λ₂ ⋅ v⃗₂, 
  # with v⃗₁ = b - a, v⃗₂ = d - c
  A = [(b - a) (c - d)]
  y = c - a
  QR = qr([A y])

  # calculate the rank of the augmented matrix
  rₐ = sum(abs.(diag(QR.R)) .> atol(T))
  # calculate the rank of the rectangular matrix
  r = sum(abs.(diag(QR.R)[1:2]) .> atol(T))

  if r >= 2 # calculate parameters of intersection or closest point
    QR = qr(A)
    λ = QR.R \ QR.Q' * y
    return λ[1], λ[2], r, rₐ
  else # parallel or collinear
    return zero(T), zero(T), r, rₐ
  end
  
  # Intersection r == rₐ == 2
  # Collinear: r == rₐ == 1
  # no intersection: r != rₐ
  #   no intersection and parallel:  r == 1, rₐ == 2
  #   no intersection, skew lines: r == 2, rₐ == 3
end
```

intersectparameters_3:
```
function intersectparameters_3(a::Point{N,T}, b::Point{N,T}, 
                             c::Point{N,T}, d::Point{N,T}) where {N,T}
  # solves the equation (approximately): 
  # a + λ₁ ⋅ v⃗₁ = c + λ₂ ⋅ v⃗₂, 
  # with v⃗₁ = b - a, v⃗₂ = d - c
  A = [(b - a) (c - d)]
  y = c - a
  QR = qr([A y])

  # calculate the rank of the augmented matrix
  rₐ = sum(abs.(diag(QR.R)) .> atol(T))
  # calculate the rank of the rectangular matrix
  r = sum(abs.(diag(QR.R)[1:2]) .> atol(T))

  if r >= 2 # calculate parameters of intersection or closest point
    λ = QR.R[1:2,1:2] \ (QR.Q' * y)[1:2] 
    return λ[1], λ[2], r, rₐ
  else # parallel or collinear
    return zero(T), zero(T), r, rₐ
  end
  
  # Intersection r == rₐ == 2
  # Collinear: r == rₐ == 1
  # no intersection: r != rₐ
  #   no intersection and parallel:  r == 1, rₐ == 2
  #   no intersection, skew lines: r == 2, rₐ == 3
	
end
```

dorn-gerhard (commented):
I considered all your inputs, 
please give me feedback on the QR decomposition solution (done twice).

I did a QR decomposition two times (new line 53), since addressing parts of the Q and R matrix lead to a slowdown when benchmarking (addressing parts of a static array leads to a conversion back to normal array which might cause the slowdown)

See also the benchmarks in the comments using the 3 functions
`intersectparameters`, `intersectparameters_2` and `intersectparameters_3`,

juliohm:
@dorn-gerhard all tests are failing, can you please fix that before I give another round of review?

dorn-gerhard:
Sure ;)
There was a problem with rank calculation in the 2D case. I fixed that. 

@juliohm Now there is **another problem**, which I would like to discuss with you:
A rounding error issue. The tests implemented so far need use `==` for comparison. 
Shouldn't we use `isapprox(p_1, p_2, atol = tol(T))`?  

```code
Segments: Test Failed at /home/runner/work/Meshes.jl/Meshes.jl/test/intersections.jl:100
  Expression: s1 ∩ s2 == s2 ∩ s1 == P2(2.5, 1.5)
  Evaluated: Point(2.5, 1.5) == Point(2.5, 1.4999999999999998) == Point(2.5, 1.5)
```

juliohm:
As discussed on Zulip, I think we should start a separate PR to fix the `==` implementation for `Point` and possibly other geometries such as `Plane` to accept a tolerance according to our `atol(T)` api. The fix will make us feel more confident about the changes in this PR which are truly fundamental to many other algorithms downstream.

@dorn-gerhard can you work on this separate PR proposing a fix to == for Point? We can then rebase this PR here to see the tests passing without further changes.

dorn-gerhard:
Since changing the behaviour of `==` is very fundamental I just fixed the tests by using `isapprox` resp. `≈` whenever comparing points in the intersections test.
Tests now work out.

src/intersections/lines.jl
```diff
@@ -43,14 +28,38 @@ end
...
+  # with v⃗₁ = b - a, v⃗₂ = d - c
+  A = [(b - a) (c - d)]
+  y = c - a
+  QR = qr([A y])
 
-  # the intersection point lies in between a and b at a fraction s
-  # (https://en.wikipedia.org/wiki/Line-line_intersection#Formulas)
-  s = (v12[1] * v2[2] - v12[2] * v2[1]) / (v1[1] * v2[2] - v1[2] * v2[1])
+  # calculate the rank of the augmented matrix
+  rₐ = sum(sum(abs.(QR.R), dims = 2) .> atol(T))
```
```suggestion
  rₐ = sum(sum(abs, QR.R, dims = 2)) .> atol(T))
```

Perhaps you can express this in terms of the actual norm you are interested? The `norm` function isn't useful here?

src/intersections/lines.jl
```diff
@@ -43,14 +28,38 @@ end
...
+  # calculate the rank of the rectangular matrix
+  r = sum(abs.(diag(QR.R)[1:2]) .> atol(T))
```
```suggestion
  r = sum(abs, diag(QR.R)[1:2]) .> atol(T))
```

Same here, can we use `norm` to be more explicit? Performance should be the same right?

src/intersections/lines.jl
```diff
@@ -43,14 +28,38 @@ end
...
+
+  if r >= 2 # calculate parameters of intersection or closest point
+    QR = qr(A)
```
juliohm, line 72: Can you use a different variable name for this second QR factorization? Notice that overwriting variables in Julia is always something tricky for type stability. It would be nice to add a test in the test suite to show that this function is type stable. See `@inferred` tests.

src/intersections/lines.jl
```diff
@@ -43,14 +28,38 @@ end
...
+  else # parallel or collinear
+    return zero(T), zero(T), r, rₐ
+  end
 
-  a - s*v1
+  # Intersection r == rₐ == 2
+  # Collinear: r == rₐ == 1
+  # no intersection: r != rₐ
+  #   no intersection and parallel:  r == 1, rₐ == 2
+  #   no intersection, skew lines: r == 2, rₐ == 3
```
juliohm: I think this comment should come in the beginning of the function together with the big comment block.

juliohm (changes requested):

src/intersections/lines.jl
dorn-gerhard: To evaluate the rank of this rectangular matrix, I count how many rows are non-zero, so how many row norms are above the tolerance.
I don't see how norm can help me since there is no `dims` option. 
I will adopt your suggestion to avoid the syntax error:
`rₐ = sum(sum(abs, QR.R; dims = 2) .> atol(T))`

src/intersections/lines.jl
dorn-gerhard: Here the absolute value of the first diagonal entries is compared with the tolerance. 
The sum counts the `true` values. I don't see how to apply a norm here, except for replacing the sum
over logicals which is not necessary here. I would therefore stick to the suggested code

`r = sum(abs.(diag(QR.R)[1:2]) .> atol(T))`

Example where checking the diagonal (one summation less) does not work, which is the reason for the difference for line 48 and 50:

``` julia
[ -1.0  1.0  0.0
   0.0  0.0  1.0 ]
```

src/intersections/lines.jl
dorn-gerhard: I just figured out that my algorithm doesn't yield the right rank for all cases:

Here are the two examples I struggle with: `R` matrices from QR decompositions
``` julia
[ -3.31662   3.31662      -0.301511
   0.0      -4.44089e-16   0.904534
   0.0       0.0           0.301511 ]

[ -1.0  1.0  0.0
   0.0  0.0  1.0 ]
```
A solution to evaluate the rank correctly would be:
``` code
rₐ = N == 2 ? sum(sum(abs, QRₐ.R; dims = 2) .> atol(T)) :
              sum(abs.(diag(QRₐ.R)) .> atol(T))
```
Maybe you have a more elegant solution?

src/intersections/lines.jl
dorn-gerhard: Done

src/intersections/lines.jl
dorn-gerhard: Done

dorn-gerhard:
I added type stability tests.

Please have a look at my comments regarding the rank calculation and the two specific matrices

juliohm:
@dorn-gerhard tests are not passing again. Can you please take a look?

dorn-gerhard:
> 

Sure, I've only forgotten that `intersectparameters` is not an exported function, so I added `Meshes.` in the tests.
Now all tests work fine.

test/intersections.jl
```diff
@@ -529,12 +529,24 @@
     # crossing in one point
     l1 = Line(P3(1,2,3), P3(2,1,0))
     l2 = Line(P3(1,2,3), P3(1,1,1))
-    @test l1 ∩ l2 == l2 ∩ l1 == P3(1,2,3)
+    @test l1 ∩ l2 ≈ l2 ∩ l1 ≈ P3(1,2,3)
```
juliohm, line 542: Can you also add more tests for the cases where the previous code was failing? I see that you replaced the == by \approx and added a couple @inferred tests. Can we also add tests for the cases that motivated your proposed fix?

src/intersections/lines.jl
```diff
@@ -43,14 +28,43 @@ end
...
+  No intersection, skew lines: r == 2, rₐ == 3
+=#
+function intersectparameters(a::Point{N,T}, b::Point{N,T}, 
+                             c::Point{N,T}, d::Point{N,T}) where {N,T}
+  # solves the equation (approximately):
+  # a + λ₁ ⋅ v⃗₁ = c + λ₂ ⋅ v⃗₂,
+  # with v⃗₁ = b - a, v⃗₂ = d - c
+  A = [(b - a) (c - d)]
+  y = c - a
+  QRₐ  = qr([A y])
```
```suggestion
  QRₐ = qr([A y])
```

src/intersections/lines.jl
```diff
@@ -43,14 +28,43 @@ end
...
+
+  # calculate the rank of the augmented matrix
+  rₐ = N == 2 ? sum(sum(abs, QRₐ.R; dims = 2) .> atol(T)) :
+                sum(abs.(diag(QRₐ.R)) .> atol(T))
```
juliohm: Can you please double check that the QR factorization object doesn't have this information stored already? I find it strange that we need to compute the rank with these summations.

juliohm (changes requested):

test/intersections.jl
dorn-gerhard: sure, i will add a test, with varying the one point using a for loop

src/intersections/lines.jl
dorn-gerhard: I've tried and benchmarked a lot of thinks:

- using rank directly (version 2)
  - slows down significantly (involves svd decomposition)
- using pivoting in the QR decomposition (would make distinction between 2d and nd case obsolete) (version 3)
  - slows down by a factor of 20 (no matter whether static or normal array)
- converting static array to normal array (version 4 and 5)
  - makes things generally worse - that is the reason why recycling (implies conversion) 
    `λ = QRₐ.R[1:2,1:2] \ QRₐ.Q[:,1:2]' * y` 
  and back conversion `λ = SMatrix{2,2}(QRₐ.R[1:2,1:2]) \ SMatrix{N,2}(QRₐ.Q[:,1:2])' * y` slows down function and I used a second QR decomposition
- going directly for `A\y`
  - slows down significantly in `N > 2`

Regarding double checking:
rank for QR objects might be implemented in future:
see comment in [source code of qr.jl ](https://github.com/JuliaLang/julia/blob/master/stdlib/LinearAlgebra/src/qr.jl#L1058)

Attached the mentioned benchmarks (Pluto notebook)
[benchmark_intersectparameters.zip](https://github.com/JuliaGeometry/Meshes.jl/files/9249619/benchmark_intersectparameters.zip)

test/intersections.jl
```diff
@@ -510,6 +510,15 @@
     @test l1 == l2
     @test l1 ∩ l2 == l2 ∩ l1 == l1
 
+    # rounding errors
+    l1 = Line(P2(3.0,1.0), P2(2.0,2.0))
+    for k = 1:1000
+      Δ = k * atol(T)
+      l2 = Line(P2(1.5,1.5 + Δ), P2(3.0, 1.5 + Δ))
+      pᵢ = P2(2.5 - Δ, 1.5 + Δ)
+      @test l1 ∩ l2 ≈ l2 ∩ l1 ≈ pᵢ
```
```suggestion
      p = P2(2.5 - Δ, 1.5 + Δ)
      @test l1 ∩ l2 ≈ l2 ∩ l1 ≈ p
```

src/intersections/lines.jl
juliohm: Ok, now please refactor the summations as I had suggested before. These are all norms hidden as summations. Can you try expressing these expressions with `norm(x, p)`?

juliohm (changes requested):

src/intersections/lines.jl
juliohm: The sum of abs is the norm L1, in Julia you can `norm(x, 1)`

src/intersections/lines.jl
juliohm: After you refactor the code to use norms we can see if there is a mathematical identity to be explored.

dorn-gerhard:
Please have a closer look at the code, the sums cannot be refactored as norms (yet).
I need the l1 norm **in the rows** of the matrix, `norm` does not support a `dims` keyword yet:
Here is a [discussion](https://github.com/JuliaLang/LinearAlgebra.jl/issues/697) about that issue.
If `norm(A::Matrix, 1; dims = 2)` is supported I will open a new pull request.

The only solution to transfer `A` to `b` using `norm` with current julia versions is the following:
`[norm(A[i,:],1) for i=1:3]`

or using the mentioned mapslice method which I wouldn't prefer
`map(x->norm(x,1), eachslice(A, dims=1))`

That's the reason why I would like to stick to 
`sum(abs, A, dims = 2)`

``` code
A = [ -3.31662   3.31662      -0.301511
       0.0      -4.44089e-16   0.904534
       0.0       0.0           0.301511]
```
```code
 b = [ 6.934760925288563
       0.9045340337332914
       0.30151134457776363]
```

dorn-gerhard:
Regarding the second expression:
`sum(     abs.(diag(QRₐ.R)) .> atol(T)     )`

This is not a sum of absolute values but a sum of a logical vector, as mentioned before, norm would work here but sum is sufficient.

src/intersections/lines.jl
```diff
@@ -43,14 +28,43 @@ end
...
+  # calculate the rank of the rectangular matrix
+  r = sum(abs.(diag(QRₐ.R)[1:2]) .> atol(T))
 
-  # the intersection point lies in between a and b at a fraction s
-  # (https://en.wikipedia.org/wiki/Line-line_intersection#Formulas)
-  s = (v12[1] * v2[2] - v12[2] * v2[1]) / (v1[1] * v2[2] - v1[2] * v2[1])
+  if r >= 2 # calculate parameters of intersection or closest point
```
juliohm: please use unicode for inequalities

src/intersections/lines.jl
```diff
@@ -43,14 +28,43 @@ end
...
+    QR = qr(A)
+    λ = QR.R \ QR.Q' * y
+    return λ[1], λ[2], r, rₐ
+  else # parallel or collinear
+    return zero(T), zero(T), r, rₐ
```
juliohm: we don't add explicit return when the statement is the last statement of the function

src/intersections/lines.jl
```diff
@@ -43,14 +28,43 @@ end
...
+No intersection: r != rₐ
+  No intersection and parallel:  r == 1, rₐ == 2
+  No intersection, skew lines: r == 2, rₐ == 3
+=#
+function intersectparameters(a::Point{N,T}, b::Point{N,T}, 
+                             c::Point{N,T}, d::Point{N,T}) where {N,T}
+  # solves the equation (approximately):
+  # a + λ₁ ⋅ v⃗₁ = c + λ₂ ⋅ v⃗₂,
+  # with v⃗₁ = b - a, v⃗₂ = d - c
+  A = [(b - a) (c - d)]
```
juliohm: notice that the comment and the expression differ, the second column has the sign flipped, can you please double check and make it consistent?

juliohm (changes requested):
Two remarks to help clean up this QR factorization code that is a bit verbose still:

1. The QR factorization of an augmented matrix contains the QR factorization of the submatrix. So when you compute `qr(A)` and `qr([A y])`, you could instead try to call just the latter and use views to reason about the former.
2. The whole confusion with summations to identify the rank of the matrix can be improved if you add a comment explaining the intent of the code: "rank of A=QR is equal to non-zero diagonal entries of R". After that, you can try to simplify the code because it doesn't matter if we have 2D or 3D points, the rank of the matrix is defined as above. I don't understand the other expression you introduced in the 3D case. Can you elaborate?

src/intersections/lines.jl
dorn-gerhard: ah, I forgot, sure ;)

src/intersections/lines.jl
dorn-gerhard: everything is correct here, I thought this is clear from the context.
You have to bring the unknown lambdas on one side of the equation to solve them.
And you cannot redefine `v⃗₂` as it is the vector that points away from `c` with a length indicated by parameter `λ₂`
I'll add the comment line
``` code
# A = [v⃗₁ -v⃗₂]
```

juliohm: Yes, the more explicit the better. People reading this code for the first time will be more confident that things are the way they are for a reason and that the original author was super careful with minimal details.

dorn-gerhard:
ad 1) reusing `qr([A y])` for `qr(A)`
I discussed this previously, realizing that a reuse (view) is slower than doing a qr decomposition twice.
Here are four benchmarks:

1. My suggested version with using 2 QR decompositions: 270 - 340 ns
image
2. Reuse of QR decomposition (view): 460 - 510 ns
using 
``` code
λ = view(QRₐ.R, 1:2, 1:2) \ view(QRₐ.Q, :, 1:2)' * y
```
image

ad 2)
*"rank of A=QR is equal to non-zero diagonal entries of R"* - if it were that easy, problems would be solved, but there is one pathological case in 2D (which caused some tests to fail previously) where this statement for numerical evaluated QR decompositions is not true, namely when there are more columns than rows, like in 2D for the augmented matrix.
I already posted such a matrix before (see again below)
Therefore I came up with the exception for 2D.

So basically checking the diagonal works for
- the R part of the augmented matrix `[A y]` for `N ≥ 3`
- the R part of the matrix `A` for `N ≥ 2`

``` code
# 2D:
R = [ -1.0  1.0  0.0
       0.0  0.0  1.0 ]
```

I will add some lines of comments to explain that

src/intersections/lines.jl
```diff
@@ -43,14 +28,47 @@ end
...
 
-  v1  = a - b
-  v2  = c - d
-  v12 = a - c
+Intersection: r == rₐ == 2
+Collinear: r == rₐ == 1
+No intersection: r != rₐ
+  No intersection and parallel:  r == 1, rₐ == 2
+  No intersection, skew lines: r == 2, rₐ == 3
+=#
```
juliohm: Thank you @dorn-gerhard for checking all the comments and reviews here carefully. Keep in mind that the more comments and explanations you share upfront, the easier it will be to merge delicate fundamental changes to the master branch.

I still feel that this introductory comment is not self-contained and that anyone reading it for the first time will get lost. Can you try to rewrite it telling the full story upfront? Defining the variables, the matrices involved in the algorithm, etc? After you help the reader with this introductory comment, the function body can just contain small important explanations not already explained in the introductory comment.

I think that after that we will finally be able to squash and merge. Thanks for pushing it further, I am doing my best to review it in the middle of other activities.

src/intersections/lines.jl
```diff
@@ -43,14 +28,52 @@ end
...
-  # (https://en.wikipedia.org/wiki/Line-line_intersection#Formulas)
-  s = (v12[1] * v2[2] - v12[2] * v2[1]) / (v1[1] * v2[2] - v1[2] * v2[1])
+- Intersection: r == rₐ == 2
+- Collinear: r == rₐ == 1
+- No intersection: r != rₐ
+  - No intersection and parallel:  r == 1, rₐ == 2
+  - No intersection, skew lines: r == 2, rₐ == 3
+"""
+function intersectparameters(a::Point{N,T}, b::Point{N,T}, 
+                          c::Point{N,T}, d::Point{N,T}) where {N,T}
```
```suggestion
                             c::Point{N,T}, d::Point{N,T}) where {N,T}
```

src/intersections/lines.jl
```diff
@@ -43,14 +28,52 @@ end
...
+  rₐ = N == 2 ? sum(sum(abs, QRₐ.R; dims = 2) .> atol(T)) :
+                sum(abs.(diag(QRₐ.R)) .> atol(T))
+  # calculate the rank of the rectangular matrix
+  r = sum(sum(abs, QRₐ.R[:,1:2]; dims = 2) .> atol(T))
+
+  if r ≥ 2 # calculate parameters of intersection or closest point
+    QR = qr(A)
+    λ = QR.R \ QR.Q' * y
+  else # parallel or collinear
+    λ = [zero(T), zero(T)]
```
juliohm, line 75: can you please confirm that these two branches have the same return type? isn't the first returning a static vector?

src/intersections/lines.jl
```diff
@@ -43,14 +28,52 @@ end
...
+  # for N == 2 one has to check the L1 norm of rows as 
+  # there are more columns than rows
+  rₐ = N == 2 ? sum(sum(abs, QRₐ.R; dims = 2) .> atol(T)) :
+                sum(abs.(diag(QRₐ.R)) .> atol(T))
+  # calculate the rank of the rectangular matrix
+  r = sum(sum(abs, QRₐ.R[:,1:2]; dims = 2) .> atol(T))
+
+  if r ≥ 2 # calculate parameters of intersection or closest point
+    QR = qr(A)
+    λ = QR.R \ QR.Q' * y
```
juliohm, line 73: can't this expression be simplified to `QR \ y`?

src/intersections/lines.jl
```diff
@@ -43,14 +28,52 @@ end
...
+  QRₐ = qr([A y])
+
+  # calculate the rank of the augmented matrix
+  # by checking the zero entries of the diagonal of R
+  # for N == 2 one has to check the L1 norm of rows as 
+  # there are more columns than rows
+  rₐ = N == 2 ? sum(sum(abs, QRₐ.R; dims = 2) .> atol(T)) :
+                sum(abs.(diag(QRₐ.R)) .> atol(T))
+  # calculate the rank of the rectangular matrix
+  r = sum(sum(abs, QRₐ.R[:,1:2]; dims = 2) .> atol(T))
```
```suggestion
  Q, R = qr([A y])

  # calculate the rank of the augmented matrix
  # by checking the zero entries of the diagonal of R
  # for N == 2 one has to check the L1 norm of rows as 
  # there are more columns than rows
  rₐ = N == 2 ? sum(sum(abs, R; dims = 2) .> atol(T)) :
                sum(abs.(diag(R)) .> atol(T))
  # calculate the rank of the rectangular matrix
  r = sum(sum(abs, R[:,1:2]; dims = 2) .> atol(T))
```

juliohm (changes requested):
Left a few minor comments now. Almost there 🎉

src/intersections/lines.jl
dorn-gerhard: Yes you are right, line 73 returns a static vector. But the return value always are single numbers of type `T`.
StaticVector / Vector are just a canvas for two lines, not sure if this is a problem in the middle of a function. 
The return value will be type stable

src/intersections/lines.jl
dorn-gerhard: No that does not work, since QR is a static array (since A is a static Array).
If A were a normal array,  QR would be a so-called [QR object](https://docs.julialang.org/en/v1/stdlib/LinearAlgebra/#LinearAlgebra.QR) that offers that direct feature

src/intersections/lines.jl
juliohm: Yes, I think @inferred is enough to make sure that we have type stability. It should avoid red @code_warntype output.

juliohm (approved):

juliohm:
🎉 thanks for sticking with the review process! ❤️

PR #278, opened 2022-07-04
Visualize `Cylinder`

### LINKED ISSUE #243: Missing cylinder triangulation

jacobleft:
1. implement `discretize` and `sample` for `CylinderSurface`, so that `Cylinder` can be visualized.
2. use the Householder orthogonalization method to construct `Plane` with a normal.

This PR resolves #243.

### DIFF
--- a/src/discretization.jl
+++ b/src/discretization.jl
@@ -151,6 +151,8 @@ triangulate(sphere::Sphere{3}) =
 triangulate(ball::Ball{2}) =
   discretize(ball, RegularDiscretization(100)) |> triangulate
 
+triangulate(cylsurf::CylinderSurface) =
+  discretize(cylsurf, RegularDiscretization(10,3)) |> triangulate
 # ----------------
 # IMPLEMENTATIONS
 # ----------------
--- a/src/discretization/regular.jl
+++ b/src/discretization/regular.jl
@@ -16,6 +16,62 @@ end
 RegularDiscretization(sizes::Vararg{Int,N}) where {N} =
   RegularDiscretization(sizes)
 
+
+function discretize(cylsurf::CylinderSurface{T},
+                    method::RegularDiscretization) where {T}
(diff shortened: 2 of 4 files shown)


juliohm:
@jacobleft thank you for the awesome PR. Would you mind splitting it into multiple PRs? One PR for the sampling method with tests, and another PR with the discretization method with tests? That way we can review corner cases more easily together.

Looking forward to working more on this with you.

jacobleft:
Not at all. But can you be more specific about the Wheres and Hows of tests?

juliohm:
Of course. For every method we add, there is a corresponding section in the test folder. For instance, I would start with a PR that just adds the RegularSampling method for CylinderSurface, and then would add tests to the test/sampling.jl file, similar to the other tests that are already there. Tests in this case could check that the points lie on the cylinder surface given a radius and normals.

If we can work on this PR together, future PRs will follow the same process.

juliohm:
I will close this PR here so that we don't get confused. Looking forward to reviewing the PR on regular sampling over cylinder surface.

juliohm:
@jacobleft I don't know how familiar you are with testing in Julia, but you can just activate the folder of the project in package mode `] activate /path/to/Meshes.jl` and then type `] test` also in package mode. So you can edit the files in the test directory to include more tests and run the tests over and over.

juliohm:
Hi @jacobleft , now that we have the householderbasis function properly in place, do you think you can find time to implement the other missing features in this PR? First, implement the sampling method for the CylinderSurface with tests in a PR and then a final PR with the discretization method that simply calls the sampling and connects the vertices.

PR #279, opened 2022-07-05
Fix `Plane` with a normal.

jacobleft:
Previously, this give wrong tangent vectors:
```julia
julia> n = Vec3(1.0, 1.0, 0.0)
3-element StaticArraysCore.SVector{3, Float64} with indices SOneTo(3):
 1.0
 1.0
 0.0

julia> (;p,u,v) = Plane(Point(1, 0, 0), n)
Plane{Float64}(Point(1.0, 0.0, 0.0), [0.5, -0.5, 0.0], [-0.5, 0.5, 0.0])
```
This PR uses the [Householder orthogonalization method](https://doi.org/10.1016/j.cad.2012.11.003) to compute orthonormal tangent vectors from a `Plane`'s normal.

The tests are also changed to check for the orthogonal conditions instead of the computed vectors.

### DIFF
--- a/src/primitives/plane.jl
+++ b/src/primitives/plane.jl
@@ -19,24 +19,22 @@ struct Plane{T} <: Primitive{3,T}
   v::Vec{3,T}
 end
 
+# D. S. Lopes, M. T. Silva, and J. A. Ambrósio, “Tangent vectors to a 3-D surface normal: A geometric tool to find orthogonal vectors based on the Householder transformation,” Computer-Aided Design, vol. 45, no. 3, pp. 683–694, Mar. 2013, doi: 10.1016/j.cad.2012.11.003.
+function HouseholderOrthogonalization(n)
+    n̄ = norm(n)
+    h1 = max(n[1]-n̄,n[1]+n̄)
+    h2 = n[2]
+    h3 = n[3]
+    h = SVector(h1,h2,h3)
+    H = I - 2h*transpose(h)/(transpose(h)*h)
+    t = H[1:3,2]
+    b = H[1:3,3]
+    t,b
+end
+
 function Plane{T}(p::Point{3,T}, n::Vec{3,T}) where {T}
-  # origin of coordinate system
-  o = Vec{3,T}(zero(T), zero(T), zero(T))
-
-  uv = Vec{3,T}[]
-  for i in 1:3
-    # subtract projection of Euclidean basis onto normal
-    e = Vec{3,T}(ntuple(j -> j == i ? one(T) : zero(T), 3))
-    v = e - (e⋅n)/(n⋅n) * n
-
-    # check if it is a valid vector
-    isapprox(v, o, atol=atol(T)) || push!(uv, v)
-
-    # we need two vectors
-    length(uv) == 2 && break
-  end
-
-  Plane{T}(p, uv[1], uv[2])
+  u, v = HouseholderOrthogonalization(n)
+  Plane{T}(p, u, v)
 end
 
 Plane(p::Point{3,T}, n::Vec{3,T}) where {T} = Plane{T}(p, n)
--- a/test/primitives.jl
+++ b/test/primitives.jl
@@ -58,13 +58,16 @@
     @test normal(p) == Vec(0, 0, 1)
     @test isnothing(boundary(p))
 
-    p = Plane(P3(0, 0, 0), V3(0, 0, 1))
-    @test p(T(1), T(0)) == P3(1, 0, 0)
-    @test p(T(0), T(1)) == P3(0, 1, 0)
+    n = V3(1, 1, 0)
+    p = Plane(P3(0, 0, 0), n)
+    @test abs(p.u⋅p.v) < eps(T)
+    @test abs(p.u⋅n) < eps(T)
+    @test abs(p.v⋅n) < eps(T)
 
     p1 = Plane(P3(0, 0, 0), V3(1, 0, 0), V3(0, 1, 0))
     p2 = Plane(P3(0, 0, 0), V3(0, 0, 1))
-    @test p1 == p2
+    @test abs(p1.u⋅normal(p2)) < eps(T)
+    @test abs(p1.v⋅normal(p2)) < eps(T)
   end
 
   @testset "Bezier curves" begin


src/primitives/plane.jl
```diff
@@ -19,24 +19,22 @@ struct Plane{T} <: Primitive{3,T}
...
+    n̄ = norm(n)
+    h1 = max(n[1]-n̄,n[1]+n̄)
+    h2 = n[2]
+    h3 = n[3]
+    h = SVector(h1,h2,h3)
+    H = I - 2h*transpose(h)/(transpose(h)*h)
+    t = H[1:3,2]
+    b = H[1:3,3]
+    t,b
+end
```
juliohm: That is awesome. Can you move this function to our utils.jl file and add a docstring like with the other functions therein? Also, please rename the function to `householderbasis(normal)` or a similar name that clarifies the returned pair of vectors.

test/primitives.jl
```diff
@@ -58,13 +58,16 @@
     @test normal(p) == Vec(0, 0, 1)
     @test isnothing(boundary(p))
 
-    p = Plane(P3(0, 0, 0), V3(0, 0, 1))
-    @test p(T(1), T(0)) == P3(1, 0, 0)
-    @test p(T(0), T(1)) == P3(0, 1, 0)
```
juliohm: Can you please elaborate on the new results? The previous implementation wasn't correct? These tests that were erased seemed correct in behavior.

test/primitives.jl
```diff
@@ -58,13 +58,16 @@
...
+    n = V3(1, 1, 0)
+    p = Plane(P3(0, 0, 0), n)
+    @test abs(p.u⋅p.v) < eps(T)
+    @test abs(p.u⋅n) < eps(T)
+    @test abs(p.v⋅n) < eps(T)
```
juliohm: Notice that we cannot access fields of structs directly because that is not part of the public API. You can use methods of plane to recover the origin and normal vector instead.

juliohm (changes requested):
Attached a first review @jacobleft , thanks for looking into it.

juliohm:
@jacobleft did you have a chance to take a look into this?

jacobleft:
Sure. I was just too occupied lately.

The current implementation in Meshes.jl is indeed incorrect for some cases:

> ```julia
> julia> n = Vec3(1.0, 1.0, 0.0)
> 3-element StaticArraysCore.SVector{3, Float64} with indices SOneTo(3):
>  1.0
>  1.0
>  0.0
> 
> julia> (;p,u,v) = Plane(Point(1, 0, 0), n)
> Plane{Float64}(Point(1.0, 0.0, 0.0), [0.5, -0.5, 0.0], [-0.5, 0.5, 0.0])
> ```

We can see that the computed tangent vectors are parallel, thus incorrect. 

Actually, the implementation of the Householder method given by the [paper](https://doi.org/10.1016/j.cad.2012.11.003) is also slightly flawed. It fails for the normal vector `n = Vec3(x, 0.0, 0.0)` if `x` is negative.
I have resolved this problem with a little modification in the latest commit. 

Although I don't have time to give formal proof, the latest commit of this PR can guarantee to produce orthonormal tangent vectors `u` and `v` from any non-null normal `n`, such that `u,v,n` form a right-hand orthogonal system.

I have checked typical cases
```julia
julia> householderbasis([1,0,0])
([-0.0, 1.0, -0.0], [-0.0, -0.0, 1.0])

julia> householderbasis([0,1,0])
([-0.0, -0.0, 1.0], [1.0, -0.0, -0.0])

julia> householderbasis([0,0,1])
([1.0, -0.0, -0.0], [-0.0, 1.0, -0.0])

julia> householderbasis([-1,0,0])
([0.0, -0.0, 1.0], [0.0, 1.0, 0.0])

julia> householderbasis([0,-1,0])
([1.0, 0.0, 0.0], [-0.0, 0.0, 1.0])

julia> householderbasis([0,0,-1])
([-0.0, 1.0, 0.0], [1.0, 0.0, 0.0])
```
and also random normal vectors
```julia
julia> n = rand(3).-0.5; u,v = householderbasis(n); u×v-n./norm(n) |> norm
2.498001805406602e-16
```
These are all correct.

Regarding the tests, I think it is more reasonable to test the geometric conditions instead of the exact values of the computed vectors. For example, this fails
```julia
p1 = Plane(P3(0, 0, 0), V3(1, 0, 0), V3(0, 1, 0))
p2 = Plane(P3(0, 0, 0), V3(0, 0, 1))
@test p1 == p2
```
because of positive and negative zeros
```julia
Evaluated: Plane{Float32}(Point(0.0f0, 0.0f0, 0.0f0), Float32[1.0, 0.0, 0.0], Float32[0.0, 1.0, 0.0]) == Plane{Float32}(Point(0.0f0, 0.0f0, 0.0f0), Float32[1.0, -0.0, -0.0], Float32[-0.0, 1.0, -0.0])
```

Maybe we should provide a public API to retrieve tangent vectors of a plane?
Also, why is the normal not stored in a `Plane` struct?

src/utils.jl
```diff
@@ -94,3 +94,24 @@ Returns the unitless type of a (unitful) type or value. See Unitful.jl.
...
+"""
+    householderbasis(n)
+
+Returns a pair of orthonormal tangent vectors `u` and `v` from a normal `n`, 
+such that `u`, `v`, and `n` form a right-hand orthogonal system.
+
+Reference: ["Tangent vectors to a 3-D surface normal: A geometric tool to find orthogonal vectors based on the Householder transformation"](https://doi.org/10.1016/j.cad.2012.11.003)
+"""
+function householderbasis(n)
+  norm_n = norm(n)
```
juliohm: Please use n\bar or n\hat for the normalized vector.

src/utils.jl
```diff
@@ -94,3 +94,24 @@ Returns the unitless type of a (unitful) type or value. See Unitful.jl.
...
 """
 dropunits(v) = typeof(one(v))
+
+"""
+    householderbasis(n)
+
+Returns a pair of orthonormal tangent vectors `u` and `v` from a normal `n`, 
+such that `u`, `v`, and `n` form a right-hand orthogonal system.
+
+Reference: ["Tangent vectors to a 3-D surface normal: A geometric tool to find orthogonal vectors based on the Householder transformation"](https://doi.org/10.1016/j.cad.2012.11.003)
```
juliohm: Please use the reference format like in here: https://github.com/JuliaGeometry/Meshes.jl/blob/4d1d68cd43b0ac00c69395476053cc30296040a3/src/refinement/catmullclark.jl#L16-L20

juliohm (changes requested):
That is awesome @jacobleft. Could you please adjust these minor details and add tests for the new householderbasis function? I think we already have a test/utils.jl file. Try to be as comprehensive as possible in the tests to catch corner cases.

juliohm:
> Maybe we should provide a public API to retrieve tangent vectors of a plane?

That sounds like a good idea. Maybe we should introduce this `basis` function in a future separate PR.

> Also, why is the normal not stored in a `Plane` struct?

We had discussions in the past about the representation of planes. We opted for point + basis at that time. It turns out that most operations with planes lie on the plane. Computing the normal on the fly is already available as a function `normal(plane)`.

test/utils.jl
```diff
@@ -23,4 +23,18 @@
...
   @test Meshes.dropunits(Float64) == Float64
+
+  normals = [
+    [ 1, 0, 0],
+    [ 0, 1, 0],
+    [ 0, 0, 1],
+    [-1, 0, 0],
+    [ 0,-1, 0],
+    [ 0, 0,-1],
+    rand(3).-0.5
```
juliohm: Please use `V3` for the tests so that we can test both single and double precision.

test/utils.jl
```diff
@@ -23,4 +23,18 @@
...
+  ]
+  for n in normals    
+    u, v = Meshes.householderbasis(n)
```
juliohm: Let's export this utility function as well.

test/primitives.jl
```diff
@@ -61,10 +61,6 @@
     p = Plane(P3(0, 0, 0), V3(0, 0, 1))
     @test p(T(1), T(0)) == P3(1, 0, 0)
     @test p(T(0), T(1)) == P3(0, 1, 0)
-
-    p1 = Plane(P3(0, 0, 0), V3(1, 0, 0), V3(0, 1, 0))
-    p2 = Plane(P3(0, 0, 0), V3(0, 0, 1))
-    @test p1 == p2
```
juliohm, line 67: These are exactly the same plane. Are you suggesting that we should also update the equality operator maybe?

juliohm (changes requested):

test/primitives.jl
jacobleft: Yeah, I think so.

juliohm: Could you please propose a fix to the equality operator?

juliohm: In that case we shouldn't erase this test, it is a correct test. The equality is the problem.

jacobleft: Maybe in a separate PR? I don't think equality is a problem for this PR.
Firstly, because the tangent vectors are numerically computed to a certain accuracy, we shouldn't use equality to check for correctness. Using `isapprox` instead would make more sense **if** we want to verify the tangent vectors.
Secondly, how the tangent vectors are computed depends on the underlying method that is subject to change. Thus we shouldn't assume that the tangent vectors for a normal `V3(0, 0, 1)` always align along the Cartesian axes, even though that is true for the current implementation.

juliohm: I agree that the equality can be fixed in a separate PR. Appreciate if you can take a look at it after we merge this one. It seems like a serious bug in the current code.

test/utils.jl
```diff
@@ -23,4 +23,18 @@
...
+    V3( 0, 1, 0),
+    V3( 0, 0, 1),
+    V3(-1, 0, 0),
+    V3( 0,-1, 0),
+    V3( 0, 0,-1),
+    V3(rand(3).-0.5)
+  ]
+  for n in normals    
+    u, v = householderbasis(n)
+    @test u×v ≈ n./norm(n)
```
```suggestion
    @test u × v ≈ n ./ norm(n)
```

juliohm:
Thank you @jacobleft for the awesome improvement 💯 Appreciate if you can take a look at the equality and propose a fix. After that we can take a look at those other features you tried to add in a first PR.

juliohm:
@jacobleft did you have some time to take a look at the == implementation for Plane? Do you think you can propose a fix?

PR #281, opened 2022-08-10
fix rotation of cylinder axis.

jacobleft:
The rotation matrix of the cylinder axis, which is needed for `sample`, is wrong in some cases.
For example,
```julia
p1 = Meshes.Point3([0, 0.0, 0.7714704084729095])
p2 = Meshes.Point3([-0.7337119591070445, 1, 0.23839746687551142])
s = Meshes.Segment(p1,p2)
cyl = Meshes.Cylinder(0.03,s)
MeshViz.viz(cyl)
```
gives
image

This PR uses `householderbasis` to compute the rotation matrix, with which the above code now gives
image

### DIFF
--- a/src/sampling/regular.jl
+++ b/src/sampling/regular.jl
@@ -118,8 +118,8 @@ function sample(::AbstractRNG, cylsurf::CylinderSurface{T},
   l  = norm(d₃)
   d₃ /= l
   d₁, d₂ = householderbasis(d₃)
-  R = transpose([d₁;;d₂;;d₃])
-  
+  R = transpose([d₁ d₂ d₃])
+
   # new normals of planes in new rotated system
   nᵦ = R * normal(b)
   nₜ = R * normal(t)
--- a/src/sampling/regular.jl
+++ b/src/sampling/regular.jl
@@ -113,7 +113,6 @@ function sample(::AbstractRNG, cylsurf::CylinderSurface{T},
   zrange = range(zmin, stop=zmax,    length=sz[2])
 
   # rotation to align z axis with cylinder axis
-  e₃ = Vec{3,V}(0, 0, 1)
   d₃  = a(1) - a(0)
   l  = norm(d₃)
   d₃ /= l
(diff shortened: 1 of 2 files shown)


juliohm:
Thank you for the fix. Can you please avoid using ;; syntax? It is only
supported in Julia v1.8.

Em qua., 10 de ago. de 2022 05:59, codecov[bot] ***@***.***>
escreveu:

jacobleft:
Actually, it is already supported in [Julia v1.7](https://docs.julialang.org/en/v1/NEWS/#Language-changes)

juliohm:
In any case we support Julia LTS which ia currently Julia v1.6. We
shouldn't be using features like this unless strictly needed.

Em qua., 10 de ago. de 2022 07:58, jacobleft ***@***.***>
escreveu:

juliohm:
As you can see the tests are failing because you opted for a feature that is not available in LTS.

src/sampling/regular.jl
```diff
@@ -114,12 +114,12 @@ function sample(::AbstractRNG, cylsurf::CylinderSurface{T},
...
-  θ  = ∠(d, e₃)
-  w  = d × e₃
-  R  = convert(DCM, EulerAngleAxis(-θ, w))
-
+  d₃  = a(1) - a(0)
+  l  = norm(d₃)
+  d₃ /= l
+  d₁, d₂ = householderbasis(d₃)
+  R = transpose([d₁;;d₂;;d₃])
+
```
juliohm:
```suggestion
  R = transpose([d₁ d₂ d₃])

```

jacobleft:
Didn't know that. Thanks.

src/sampling/regular.jl
```diff
@@ -114,11 +114,11 @@ function sample(::AbstractRNG, cylsurf::CylinderSurface{T},
 
   # rotation to align z axis with cylinder axis
   e₃ = Vec{3,V}(0, 0, 1)
```
juliohm:
```suggestion
```

PR #283, opened 2022-08-11
WIP - Vec

eliascarv:


src/mesh/cartesiangrid.jl
```diff
@@ -53,28 +53,28 @@ julia> CartesianGrid((-1.,),(1.,), dims=(100,))
 struct CartesianGrid{Dim,T} <: Mesh{Dim,T}
   dims::Dims{Dim}
   origin::Point{Dim,T}
-  spacing::SVector{Dim,T}
+  spacing::Vec{Dim,T}
```
juliohm: You can leave this as SVector

eliascarv: I made this property a Vec because CartesianGrid uses operations between Vec and Points in its methods.

src/vectors.jl
```diff
@@ -15,14 +15,41 @@ A vector can be obtained by subtracting two [`Point`](@ref) objects.
...
+C = Vec(0, 0) # Integer is converted to Float64 by design
+D = Vec2(0, 1) # explicitly ask for double precision
+E = Vec2f(0, 1) # explicitly ask for single precision
+
+# 3D vectors
+F = Vec(1.0, 2.0, 3.0) # double precision as expected
+G = Vec(1f0, 2f0, 3f0) # single precision as expected
+H = Vec(1, 2, 3) # Integer is converted to Float64 by design
+I = Vec3(1, 2, 3) # explicitly ask for double precision
+J = Vec3f(1, 2, 3) # explicitly ask for single precision
```
juliohm:
```suggestion
Vec(0.0, 1.0) # double precision as expected
Vec(0f0, 1f0) # single precision as expected
Vec(0, 0) # Integer is converted to Float64 by design
Vec2(0, 1) # explicitly ask for double precision
Vec2f(0, 1) # explicitly ask for single precision

# 3D vectors
Vec(1.0, 2.0, 3.0) # double precision as expected
Vec(1f0, 2f0, 3f0) # single precision as expected
Vec(1, 2, 3) # Integer is converted to Float64 by design
Vec3(1, 2, 3) # explicitly ask for double precision
Vec3f(1, 2, 3) # explicitly ask for single precision
```

test/mesh.jl
```diff
@@ -103,9 +103,9 @@
     @test maximum(sub) == P2(5,8)
 
     # subgrid with comparable vertices of grid
-    grid = CartesianGrid((10,10), P2(0.0,0.0), V2(1.2,1.2))
+    grid = CartesianGrid((10,10), P2(0.0,0.0), SVector{2,T}(1.2,1.2))
```
juliohm: I think this is a good time to review the type of spacing in CartesianGrid. Perhaps it should just be a NTuple to be consistent with the dims field. Would that make the code much uglier? I think all constructors of CartesianGrid should only accept NTuple and Point variations so that end users never need to import StaticArrays here.

juliohm:
Awesome @eliascarv , thanks for the updates. Let's keep working on this PR until we fix all the incorrect usages of SVector. 💯

src/connectivities.jl
```diff
@@ -96,11 +92,11 @@ Materialize a face using the `connec` list and a global vector of `points`.
...
                      points::AbstractVector{P}) where {PL<:Polytope,P<:Point}
-  PL(view(points, SVector(connec.indices...)))
+  PL(view(points, connec.indices))
 end
 
 function Base.show(io::IO, c::Connectivity{PL}) where {PL}
   name = prettyname(PL)
   inds = c.indices
-  print(io, "$name$inds")
+  print(io, "$name($(join(inds, ", ")))")
```
juliohm: It would be cleaner to just define inds = Tuple(c.indices) and leave the print line unchanged.

src/connectivities.jl
```diff
@@ -11,7 +11,7 @@ of type `PL`. Indices are taken from a global vector of [`Point`](@ref).
 Connectivity objects are constructed with the [`connect`](@ref) function.
 """
 struct Connectivity{PL<:Polytope,N}
-  indices::NTuple{N,Int}
+  indices::SVector{N,Int}
```
juliohm: Thinking about this again, I am bit concerned that the constructor accepts a NTuple and the field is SVector. When users call indices(connec) they will get back a SVector even though they inserted a NTuple. Maybe we should preserve NTuple here? Let's discuss over Zulip what is the best way forward...

eliascarv:
This PR is getting too complex. I'm going to close it down and break it down into smaller PRs.

PR #289, opened 2022-09-01
New `Vec` implementation

### LINKED ISSUE #207: Make Vec a new type of StaticArrays.jl vector
Currently our `Vec` type is an alias to `SVector` and that helped us get off the ground and move fast with implementations. The only problem with `SVector` is that it doesn't force `AbstractFloat` coordinates. When the user types `Vec(1,2)` this will generate a vector with integer coordinates and many algorithms will break assuming that the vectors live in R^n.

We need a new definition for `Vec` that is simply a new type of `StaticVector` from StaticArrays.jl. They provide an interface and parent types to inherit most of the behavior of `SVector`. We just need to do that and add an inner constructor like we did for `Point` to force `AbstractFloat` coordinates.

eliascarv:
closes #207

### DIFF
--- a/src/intersections.jl
+++ b/src/intersections.jl
@@ -223,7 +223,7 @@ end
 
 # support point in Minkowski difference
 function minkowskipoint(g1::Geometry{Dim,T}, g2::Geometry{Dim,T}, d) where {Dim,T}
-  n = Vec{Dim}(d[1:Dim])
+  n = Vec{Dim,T}(d[1:Dim])
   v = supportfun(g1, n) - supportfun(g2, -n)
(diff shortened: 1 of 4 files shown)


juliohm (changes requested):
Awesome, can you also add tests for all the constructors, similar to the tests we wrote for Point? This is just to make sure that the behavior with Integer conversion, and other behaviors are never lost in future changes.

juliohm:
Amazing @eliascarv , what do you think of merging this first to make sure that we have all the constructors ready and working, and then start a separate PR to consider functions from the interface like embeddim, paramdim, etc? I think these functions will require some thought in terms of geometry.

juliohm (approved):

PR #293, opened 2022-09-02
Add Point ∈ Mesh method

kylebeggs:
This method checks whether a point is inside a mesh or not.

### DIFF
--- a/src/mesh/simplemesh.jl
+++ b/src/mesh/simplemesh.jl
@@ -44,3 +44,26 @@ list of points and [`FullTopology`](@ref).
 """
 Base.convert(::Type{<:SimpleMesh}, m::Mesh) =
   SimpleMesh(vertices(m), topology(m))
+
+
+"""
+    point ∈ mesh
+
+Tells whether a `point` is inside a surface `mesh`. Must be a surface mesh with only triangles as elements. Points that coincide with vertices of the mesh or lies on the face of a triangle return true.
+"""
+function Base.in(testpoint::Point{3,T}, mesh::Mesh{3,T}) where T
+  if !(eltype(mesh) <: Triangle)
+    error("This function only works for surface meshes with triangles as elements.")
+  end
+  ex = testpoint .- extrema(mesh)
+  direction = ex[argmax(norm.(ex))]
+  r = Ray(testpoint, Vec{3,T}(direction*2))
+  
+  intersects = false
+  for elem in mesh
+    if intersection(x -> x.type == NoIntersection ? false : true, r, elem)
+      intersects = !intersects
+    end
+  end
+  intersects
+end
--- a/test/mesh.jl
+++ b/test/mesh.jl
@@ -234,6 +234,16 @@
     @test vertices(mesh) == [vertices(mesh₁); vertices(mesh₂)]
     @test collect(elements(topology(mesh))) == connect.([(1,2,3),(4,5,6)])
 
+    # point in mesh
+    points = P3[(0,0,0), (1,0,0), (0,1,0), (0,0,1)]
+    connec = connect.([(1,3,2), (1,2,4), (1,4,3), (2,3,4)], Triangle)
+    mesh = SimpleMesh(points, connec)
+    @test (P3(0.2, 0.2, 0.1) ∈ mesh) == true
+    @test (P3(0.1, 0.1, 0.1) ∈ mesh) == true # ray goes through vertex
+    @test (P3(-0.1, 0.1, 0) ∈ mesh) == false
+    @test (P3(0.1, 0.1, 0) ∈ mesh) == true # on face of triangle
+    @test (P3(1, 0, 0) ∈ mesh) == true # on exisitng vertex
+
     # convert any mesh to SimpleMesh
     grid = CartesianGrid{T}(10,10)
     mesh = convert(SimpleMesh, grid)


kylebeggs:
Oh, I developed this using the old Vec, forgot to run tests after fetching that update...

juliohm:
Also it would be nice to introduce a new function name for this function.
For example isinside(point, mesh) if this name is already used with the
same idea of interior or ininterior(point, mesh).  The operation point \in
mesh means the point is in the elements of the mesh, I.e. the surface.

Em sex., 2 de set. de 2022 11:53, Kyle Beggs ***@***.***>
escreveu:

src/mesh/simplemesh.jl
```diff
@@ -44,3 +44,26 @@ list of points and [`FullTopology`](@ref).
 """
 Base.convert(::Type{<:SimpleMesh}, m::Mesh) =
   SimpleMesh(vertices(m), topology(m))
+
+
+"""
+    isinside(point, mesh)
+
+Tells whether a `point` is inside a surface `mesh`. Must be a surface mesh with only triangles as elements. Points that coincide with vertices of the mesh or lies on the face of a triangle return true.
```
juliohm:
```suggestion
Tells whether or not a `point` is in the interior or on the surface `mesh`.
```

src/mesh/simplemesh.jl
```diff
@@ -44,3 +44,26 @@ list of points and [`FullTopology`](@ref).
...
+"""
+function isinside(testpoint::Point{3,T}, mesh::Mesh{3,T}) where T
```
juliohm: Please rename `testpoint` to `point`. Also, do we need to specify the type parameters at this point?

What about:

```julia
function isinside(point::Point, mesh::Mesh)
```

src/mesh/simplemesh.jl
```diff
@@ -44,3 +44,26 @@ list of points and [`FullTopology`](@ref).
...
+  if !(eltype(mesh) <: Triangle)
+    error("This function only works for surface meshes with triangles as elements.")
```
juliohm: Please throw and `ErrorException` or a `ArgumentError`.

src/mesh/simplemesh.jl
```diff
@@ -44,3 +44,26 @@ list of points and [`FullTopology`](@ref).
...
+  end
+  ex = testpoint .- extrema(mesh)
+  direction = ex[argmax(norm.(ex))]
+  r = Ray(testpoint, direction*2)
```
juliohm: Can you work on these variable names? Some of them like `ex` are difficult to grasp and others like `direction` are too long. Try to stick to a single code style, and use comments if you need to clarify the meaning of variables. Usually, we prefer short variable names in these small functions.

src/mesh/simplemesh.jl
```diff
@@ -44,3 +44,26 @@ list of points and [`FullTopology`](@ref).
...
+  
+  intersects = false
+  for elem in mesh
+    if intersection(x -> x.type == NoIntersection ? false : true, r, elem)
+      intersects = !intersects
+    end
```
juliohm: Please try to use the do syntax:

```julia
if intersection(r, e) do I
  if type(I) == NoIntersection
    false
  else
    true
  end
end
```

Also, you probably can terminate the loop in the middle if the condition is met? You don't need to check all elements I guess? Can you please double check?

test/mesh.jl
```diff
@@ -202,6 +202,9 @@
     topo = topology(mesh)
     @test collect(faces(topo, 2)) == [connect((3,4,1), Triangle)]
     @test collect(faces(topo, 3)) == [connect((1,2,3,4), Tetrahedron)]
+
+    # use above mesh with tetrahedron for testing isinside() error
+    @test_throws ErrorException("This function only works for surface meshes with triangles as elements.") isinside(P3(0,0,0), mesh)
```
juliohm: Please more this test to be close to the other tests you introduced. If you need to create a different test mesh, that is fine. But each paragraph in our test suite is related to a specific PR usually.

juliohm (changes requested):
Thank you @kylebeggs for working on this feature. Attached you can find a few review comments to improve the code. Feel free to chat over Zulip if you have questions.

juliohm:
Hi @kylebeggs did you have a chance to take a look into this?

src/mesh/simplemesh.jl
kylebeggs: You have to check all the elements, but I will adopt the do syntax.

kylebeggs:
I've made most of the requested changes, but the issue is we need to also check if a ray intersects with an edge of the triangle. The issue right now is if this happens, it counts for both triangles that share the edge and gives the wrong answer. I have not had time to look into implementing this yet, but it seems @dorn-gerhard has been looking into implementing more intersections.

juliohm:
Thank you @kylebeggs for the update 👍🏽

dorn-gerhard:
> I've made most of the requested changes, but the issue is we need to also check if a ray intersects with an edge of the triangle. The issue right now is if this happens, it counts for both triangles that share the edge and gives the wrong answer. I have not had time to look into implementing this yet, but it seems @dorn-gerhard has been looking into implementing more intersections.

@kylebeggs, yes my next PR will treat rays - segments intersection (starting tomorrow or next week ;) )

kylebeggs:
@dorn-gerhard  Hi! First off, from all the users of Meshes.jl - thank you for all the recent work you've been doing. Is the rays-triangles / rays-segments work done?

dorn-gerhard:
@kylebeggs, you are welcome :)
Ray-Segment works, just intersect them^^
My roadmap for going to triangles includes the following steps (I was a bit busy lately, I hope I can finish those points soon)

- [ ] define an (algorithmic) naming convention for intersections of (almost) all geometries (is it RaySegment or SegmentRay) - discussed in #325 (although this is not so necessary for the oncoming two points, since it will be called anyway `RayTriangle`)
- [ ] adopting `intersectparameters` to multidimensions (such as planes, etc. to find out whether a line is in a plane and if not to get the right parameters)
- [ ] adding RayTriangle (identification of types already discussed here: #308 , names are not fixed yet - not sure where to put overlapping|touching, corner|edge) 

Regarding your PR I plan to introduce an intersection type called `EdgeTouchingRayTriangle` and `CornerTouchingRayTriangle` which indicate whether a ray goes (only) through one edge point or one corner point (here you might get multiple countings from your mesh) of your triangle.

A rather tricky case is when triangles of your mesh are in the same plane as the ray.
Here you have to check additional overlapping cases (all return a segment), such as 
- OverlappingRayTriangle
- OverlappingRayTriangleCorner (ray goes through triangle and also one corner point)
- OverlappingRayTriangleEdge (ray goes trough (whole or parts of) one edge of the triangle, so 1 corner point (if origin of ray is also on the edge) or two corner points of the triangle are also hit by the ray.

juliohm:
@dorn-gerhard did you have some time to update the roadmap you shared above?

juliohm:
@kylebeggs what is the status of this PR? Is it ready for final review? Should we change the status from draft to ready? Also, let's rename the function name `isinside` by the already existing name `sideof` and then we can return symbols like `:IN`, `:OUT` or `:ON` as valid answers. This function could be implemented in the utils.jl source file.

kylebeggs:
@juliohm Has @dorn-gerhard finished the functions I need? (ray intersecting triangle) I have not seen that. Once that is added this PR just needs a few small changes and it should be good to go

juliohm:
@kylebeggs by looking into master branch, I think we already have the ray-triangle intersection case. Can you double check if that is the missing algorithm? Looking forward to merging this PR :)

kylebeggs:
Yes, sorry I should have been more specific before. We have ray-triangle case, but the issue is this only returns `NoIntersection` or `IntersectingRayTriangle` and doesn't inform/return the scenario like if the ray intersects with a vertex or edge of the triangle and that is an issue for `isinside()`. Just need to input some returns whenever we hit these scenarios in the current ray-triangle algorithm, but I don't know how it works exactly. I'll see about reading that paper it is based on and see if I can figure it out.

kylebeggs:
@juliohm what is the easiest way to find the number of triangles in a mesh that a vertex belongs to? Once I have that, I can finish this.

juliohm:
> @juliohm what is the easiest way to find the number of triangles in a mesh that a vertex belongs to? Once I have that, I can finish this.

@kylebeggs , you can use a Coboundary{0,2} to map vertices to polygons in a polygonal mesh: https://juliageometry.github.io/Meshes.jl/stable/meshes.html#Relations

kylebeggs:
I needed to also add a case for RayTriangle intersection where the origin intersects the middle of the triangle - I called it `TouchingRayTriangle` which should keep with our naming convention in #377.

juliohm:
@kylebeggs I suggest starting a fresh PR with the minimum changes necessary for this new sideof function. And another PR before that to add the new intersection type to ray-triangle. That way we can easily trace back the history of the repository to understand the changes.

kylebeggs:
Closing this to break it up into smaller PRs. See #381

PR #294, opened 2022-09-02
[WIP] Point: Use `Vec` instead of `SVector`

eliascarv:


juliohm:
@eliascarv what about mimicking the same changes we made for Vec here? Use a NTuple as the coordinates? That way we can work with a struct definition that is completely independent of the vector definitions in the future.

eliascarv:
OK, I'll close this PR and open another one with this proposal

PR #296, opened 2022-09-10
Point: Use `Vec` instead of `SVector`

eliascarv:


src/points.jl
```diff
@@ -101,7 +120,7 @@ coordinates(A::Point) = A.coords
 Return the [`Vec`](@ref) associated with the direction
 from point `B` to point `A`.
 """
--(A::Point, B::Point) = Vec(A.coords - B.coords)
+-(A::Point, B::Point) = Vec(A.coords) - Vec(B.coords)
```
```suggestion
-(A::Point, B::Point) = Vec(A.coords .- B.coords)
```

test/points.jl
```diff
@@ -15,9 +15,9 @@
   @test coordtype(Point{2,T}((1, 1))) == T
   @test coordtype(Point{2,T}(1, 1)) == T
 
-  @test coordinates(P1(1)) == T[1]
-  @test coordinates(P2(1, 2)) == [1.0, 2.0]
-  @test coordinates(P3(1, 2, 3)) == [1.0, 2.0, 3.0]
+  @test coordinates(P1(1)) == T.((1.0,))
+  @test coordinates(P2(1, 2)) == T.((1.0, 2.0))
+  @test coordinates(P3(1, 2, 3)) == T.((1.0, 2.0, 3.0))
```
juliohm: These coordinates should always be StaticVector

test/points.jl
```diff
@@ -15,9 +15,9 @@
   @test coordtype(Point{2,T}((1, 1))) == T
   @test coordtype(Point{2,T}(1, 1)) == T
 
-  @test coordinates(P1(1)) == T[1]
-  @test coordinates(P2(1, 2)) == [1.0, 2.0]
-  @test coordinates(P3(1, 2, 3)) == [1.0, 2.0, 3.0]
+  @test coordinates(P1(1)) == T[1.0]
+  @test coordinates(P2(1, 2)) == T[1.0, 2.0]
+  @test coordinates(P3(1, 2, 3)) == T[1.0, 2.0, 3.0]
```
juliohm:
```suggestion
  @test coordinates(P1(1)) == T[1]
  @test coordinates(P2(1, 2)) == T[1, 2]
  @test coordinates(P3(1, 2, 3)) == T[1, 2, 3]
```

juliohm (approved):

PR #297, opened 2022-09-12
Intersection of rays #3

dorn-gerhard:
This PR adds all possible intersections of 2D and 3D rays using the new IntersectionTypes:
CrossingRays
MidTouchingRays
CornerTouchingRays
OverlappingAlignedRays
OverlappingCollidingRays

NEW: implemented `intersectparameters` and managed to fuse 2D and 3D case to a general ND case.

tests for 2D implemented (using approximative comparison for points due to possible rounding issues):

image

Please give feedback for 2D tests, then I will add tests for 3D


src/intersections/rays.jl
```diff
@@ -0,0 +1,61 @@
...
+#=
+The intersection type can be one of six types:
+1. intersect at one inner point (CrossingRays -> Point)
+2. intersect at origin of one ray (MidTouchingRays -> Point)
+3. intersect at origin of both rays (CornerTouchingRays -> Point)
+4. overlap with aligned vectors (OverlappingAgreeingRays -> Ray)
+5. overlap with colliding vectors (OverlappingOpposingRays -> Segment)
+6. do not overlap nor intersect (NoIntersection)
+=#
+function intersection(f, r1::Ray{N,T}, r2::Ray{N,T}) where {N,T}
```
juliohm, line 14: I think we can omit the N and T parameters given that they are not used in the body of the function.

test/intersections.jl
```diff
@@ -150,6 +150,38 @@
...
 
+  @testset "Rays" begin
+    # rays in 2D
+    r₁ = Ray(P2(1,0), V2(2.,1))
+    r₂ = Ray(P2(0,2), V2(2.,-3))
+    r₃ = Ray(P2(0.5,1), V2(1.,-2))
+    r₄ = Ray(P2(0,2), V2(1.,-3))
+    r₅ = Ray(P2(4,1.5), V2(4.,2))
+    r₆ = Ray(P2(2,0.5), V2(-0.5, -0.25))
+    r₇ = Ray(P2(4,0), V2(0.,1))
```
```suggestion
    r₁ = Ray(P2(1,0), V2(2,1))
    r₂ = Ray(P2(0,2), V2(2,-3))
    r₃ = Ray(P2(0.5,1), V2(1,-2))
    r₄ = Ray(P2(0,2), V2(1,-3))
    r₅ = Ray(P2(4,1.5), V2(4,2))
    r₆ = Ray(P2(2,0.5), V2(-0.5, -0.25))
    r₇ = Ray(P2(4,0), V2(0,1))
```

test/intersections.jl
```diff
@@ -150,6 +150,38 @@
...
+    @test intersection(r₅, r₁) |> type == OverlappingAgreeingRays #CASE 4
+    @test r₅ ∩ r₁ == r₅ #first ray
+    @test intersection(r₁, r₅) |> type == OverlappingAgreeingRays #CASE 4
+    @test r₁ ∩ r₅ == r₅ #second ray
+    @test intersection(r₁, r₆) |> type == OverlappingOpposingRays #CASE 5
+    @test r₁ ∩ r₆ == Segment(origin(r₁), origin(r₆))
+    @test intersection(r₁, r₄) |> type == NoIntersection #CASE 6
+    @test r₁ ∩ r₄ === r₄ ∩ r₁ === nothing
+
+    # rays in 3D
```
```suggestion
```

juliohm (changes requested):
Very nice PR with clean and organized code! 💯 Can you just comment on the possible current code repetitions between rays and segments? Your idea of encapsulating the core algorithm in intersectparameters was great, and will certainly facilitate maintenance in the long-term 🥇

dorn-gerhard:
> 

Thank you :)
Sure, regarding possible current code repetitions between segments and rays:
Current status with `segments.jl` is, that it uses two different methods for 2D and 3D where the 2D method uses 
*Balbes, R. and Siegel, J.* 1990.
[https://www.sciencedirect.com/science/article/abs/pii/0167839691900198](https://www.sciencedirect.com/science/article/abs/pii/0167839691900198)
I think the 2D method is fast and I would keep it for the moment.

The 3D method needs an adoption (I can make a PR) as it currently is based on the same idea of intersectparameters but not using that function. 
So I will implement it.

It will have the same overall structure using `r` and `r_a` to decide whether 
the lines obtained from the segments are **collinear**, **cross** or have **no crossing**.
Within the lines cases **collinear** and **crossing**, the parameters `\lambda_1` and `\lambda_2` will be used to 
decide on the segments case (`IntersectionType`) (same as in `rays.jl` but slightly different query)

src/intersections/rays.jl
dorn-gerhard: `T` is used in the code to decide whether the parameters are approximately zero, so would leave it as it is.

dorn-gerhard:
With the latest commit I fixed the one failing test which was due to rounding issues:
I added a query to check whether the parameters of `intersectparameters` are approximately `zero(T)`

src/intersections.jl
```diff
@@ -62,6 +62,13 @@ Type `IntersectionType` in a Julia session to see the full list.
   CrossingLines
   OverlappingLines
 
+  # ray-ray intersection
+  CrossingRays
+  MidTouchingRays
+  CornerTouchingRays
+  OverlappingAgreeingRays
+  OverlappingOpposingRays
+
```
juliohm: Please move this section to after the segment-segment section so that we track the difference over time

dorn-gerhard: Sure :)

juliohm (approved):

juliohm:
Thank you @dorn-gerhard ! 💯

Issue #299, opened 2022-09-21
Review internal functions intersectpoint, intersectcollinear, ...


juliohm:
Some of these internal functions are now superseded by the new `intersectparameters` by @dorn-gerhard . We should review and probably remove them from the code base.

juliohm:
@dorn-gerhard did you have a chance to take a look into this?

dorn-gerhard:
I was thinking of generalizing intersectparameters to define intersections of any two affine subvectorspaces, e.g. two planes in a 3, 4, or n-dimensional space. 
This could be used to prepare intersections of lines with planes (thinking of using this as basis for triangle-ray intersection) and so on.

It boils all down to defining the subvectorspace through a set of points (function input could be two vectors of points) and leads to the same linear systems to solve and the same rank analysis (using qr decomposition and so on).
function could look like:
``` math
λ⃗₁, λ⃗₂, r, rₐ = intersectparameters(p₁::Vector{Point{N,T}}, p₂::Vector{Point{N,T}})
```

With ``N`` the dimension of the space, ``n₁,n₂ = length(p₁)-1, length(p₂)-1`` the dimensions of the affine subvectorspaces, one has the following possibilities for the ranks:

$\max(n_1, n_2) \leq r \leq \max(N,n_1 + n_2)$
 $\max(n_1, n_2) \leq r \leq r_a \leq r+1 \leq \max(N,n_1 + n_2 + 1)$

If ranks are equal there is a solution, the dimension of the resulting intersection subvectorspace $n_i$ is given by
$n_i = N - r$ ($r =r_a$)

Example:
3D, 1 line given by p₁, 1 plane given by p₂, $N= 3, n_1 = 1, n_2 = 2$,
$r \in [2,3], r_a \in [2,3]$
Cases:
$r = 3, r_a = 3$: Point (lambdas are well defined),  (line p₁ hits plane p₂)
$r = 2, r_a = 3$: no intersection  (line p₁ parallel to plane p₂)
$r = 2, r_a = 2$: Line (lambda can be set to zero), (line p₁ lies within plane p₂)

Example:
6D,  affine 3D subspace given by p₁, plane given by p₂, $N= 6, n_1 = 3, n_2 = 2$,
$r \in [3,4,5], r_a \in [3,4,5,6]$
Cases:
$r = 3, r_a = 3$: Plane (lambdas can be set to zero),  (plane p₂ lies within 3D p₁)
[...]

PR #309, opened 2022-10-14
Implement intersectparameters also in segments.jl remove intersectcollinear and intersectpoint

dorn-gerhard:


dorn-gerhard:
A benchmark results for the new implementation using ``intersectparameters`` 
(CrossingSegments)
is better than the solution inspired by 
Balbes, R. and Siegel, J. 1990.
(https://www.sciencedirect.com/science/article/abs/pii/0167839691900198)

grafik

In addition intersectcollinear was wrong as mentioned in #299 
This is repaired now and a new test is added.

juliohm:
That is amazing @dorn-gerhard 👏🏽  I will wait for the tests to pass before starting the review.

juliohm:
@dorn-gerhard various tests are failing. This is a central key algorithm that we need to be extra careful to change. Can you please double check what is happening?

dorn-gerhard:
> @dorn-gerhard various tests are failing. This is a central key algorithm that we need to be extra careful to change. Can you please double check what is happening?

Sure, I checked the problem, was a little mistake.

A question arose regarding tests of Polytopes:
is the order of vertices important? 

e.g. currently ``Segment((1,0), (0,0)) == Segment((0,0), (1,0))`` returns ``false``

Reason why i am asking: 
The outcome of two overlapping segments could depend on the order or arguments so that
``s1 ∩ s2 == s2 ∩ s1`` returns ``false``

juliohm:
Yes, the segments are oriented in our design, A -- B is different than B -- A

juliohm:
Apparently the changes you proposed are causing the processes to hang @dorn-gerhard. I really appreciate it if you can be extra careful with all these intersection details as they can break a great deal of code downstream (like triangulation algorithms).

dorn-gerhard:
I see the problem and investigated the issue deeper.

A test in ``polytopes.jl`` fails, namely a ``issimiple(p::PolyArea)`` test which calls ``issimple(p.outer::Chain)`` in ``chains.jl`` 
which uses intersection of Segments and basically checks whether all Segment Intersections are ``CornerTouchingSegments`` or ``NoIntersections``.

The problem is of numerical nature and lies in the ``intersectparameters`` function when calculating the parameter for the intersections of two segments. It only occurs in two testfiles (``smooth$.jl``) and there only in five of 120 intersections.

here is one example
``` julia
P2 = Point{2, Float32}
s1 = Segment(P2(0.256951700559717, 0.382175329425997), P2(0.307299924914942, 0.272417978947013))
s2 = Segment(P2(0.307299924914942,0.272417978947013), P2(0.348147533748717, 0.184216111169297))
λ₁, λ₂ = Meshes.intersectparameters(s1(0), s1(1), s2(0), s2(1))
``` 
yields
``julia> λ₁ =  1.000007f0``, ``λ₂ =-3.548892f-6`` instead something close to ``one(T)``, ``zero(T)``.
``isapprox(λ₁, one(T), atol=atol(T))`` fails here.

So what is the reason for this numerical inaccuracy?

I checked the function ``intersectparameters`` where we used the qr decomposition to decide the rank AND calculate the parameters.

Now the following happens:
``` julia
a, b = s1(0), s1(1)
c, d = s2(0), s2(1)

A = [(b - a) (c - d)]
y = c - a

QQ = qr(A)

@info A \ y
@info QQ.R \ QQ.Q' * y
```
so the current implementation using the QR decomposition produces inaccurate results.

I therefore tried to solve the problem by replacing the second QR decomposition by ``A \ y``. 
Benchmarks are promising, function now yields:

``` julia
julia> @benchmark Meshes.intersectparameters(s1(0), s1(1), s2(0), s2(1))
BenchmarkTools.Trial: 10000 samples with 194 evaluations.
 Range (min … max):  495.876 ns … 66.443 μs  ┊ GC (min … max): 0.00% … 99.10%
 Time  (median):     516.495 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   611.109 ns ±  1.720 μs  ┊ GC (mean ± σ):  7.42% ±  2.62%

  █▆▄▆▄▄▂▂▄▃▂▁▂      ▂▁          ▁                             ▁
  ████████████████▇▇▇█████▇▇▇██▇███▇▇▆▆▆█▇▆▇▇▆▆▆▆▅▆▅▅▄▅▅▅▄▅▄▅▄ █
  496 ns        Histogram: log(frequency) by time      1.09 μs <

 Memory estimate: 432 bytes, allocs estimate: 13.
```
the new function (avoiding the second QR decomposition) yields:
```
julia> @benchmark intersectparameters3(s1(0), s1(1), s2(0), s2(1))
BenchmarkTools.Trial: 10000 samples with 195 evaluations.
 Range (min … max):  475.385 ns … 66.247 μs  ┊ GC (min … max): 0.00% … 99.15%
 Time  (median):     491.795 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   582.439 ns ±  1.692 μs  ┊ GC (mean ± σ):  7.65% ±  2.62%

  █▇▄▆▄▄▃▃▃▂▃▂▂▂ ▁   ▂▂                                        ▁
  █████████████████▇▇█████▇▇█▇▇▆▇▇▆▇▆▆▅▆▇▇▅▆▆▅▆▅▆▄▅▅▅▅▃▄▄▅▅▄▄▅ █
  475 ns        Histogram: log(frequency) by time      1.07 μs <

 Memory estimate: 432 bytes, allocs estimate: 13.
```

I also ran all tests locally, now it should work
``` julia
Testing polytopes.jl...
Test Summary:       | Pass  Total
Meshes.jl (Float32) |  371    371
```

src/intersections/segments.jl
```diff
@@ -3,159 +3,64 @@
...
-      # configuration (1)
-      return @IT CrossingSegments intersectpoint(s1, s2) f
+function intersection(f, s1::Segment{N,T}, s2::Segment{N,T}) where {N,T}
+  a, b = s1(0), s1(1)
+  c, d = s2(0), s2(1)
+
+  λ₁, λ₂, r, rₐ = intersectparameters(a, b, c, d)
+
+  # not in same plane or parallel
+  if r ≠ rₐ
```
```suggestion
  if r ≠ rₐ # not in same plane or parallel
```

src/intersections/segments.jl
```diff
@@ -3,159 +3,64 @@
...
+    return @IT NoIntersection nothing f #CASE 5
+  # collinear
+  elseif r == rₐ == 1
```
```suggestion
  elseif r == rₐ == 1 # collinear
```

src/intersections/segments.jl
```diff
@@ -3,159 +3,64 @@
...
+    # find parameters λc and λd for points c and d in s1
+    # use dimension with largest vector component to avoid division by zero
+    ix = argmax(abs.(b-a))
+    v₁ = (b-a)[ix]
```
juliohm: findmax?

if you also need the index, then try to compute b-a once, and then reuse the resulting vector in both lines

also, let's call this index `i` instead of `ix`

src/intersections/segments.jl
```diff
@@ -3,159 +3,64 @@
...
+    λc, λd = ((c-a)[ix], (d-a)[ix]) ./ v₁
+    λc = isapprox(λc, zero(T), atol=atol(T)) ? zero(T) : (isapprox(λc, one(T), atol=atol(T)) ? one(T) : λc)
+    λd = isapprox(λd, zero(T), atol=atol(T)) ? zero(T) : (isapprox(λd, one(T), atol=atol(T)) ? one(T) : λd)
```
juliohm: these two lines are calling for a function. Maybe you want to define a small function at the beginning of the function body explaning that you are rounding values near the extremes 0 and 1

src/intersections/segments.jl
```diff
@@ -3,159 +3,64 @@
...
+    if (λc > 1 && λd > 1) || (λc < 0 && λd < 0)
+      return @IT NoIntersection nothing f # CASE 5
+    elseif (λc == 0 && λd < 0) ||  (λd == 0 && λc < 0)
+      return @IT CornerTouchingSegments a f # CASE 3
+    elseif (λc == 1 && λd > 1) || (λd == 1 && λc > 1)
+      return @IT CornerTouchingSegments b f # CASE 3
```
juliohm, line 35: Notice that sometimes you use zero(T) and sometimes 0, same for one(T) versus 1. Let's try to stick to one of the two definitions in the entire function body.

src/intersections/segments.jl
```diff
@@ -3,159 +3,64 @@
...
     else
-      # configuration (5)
-      return @IT NoIntersection nothing f
+      p = sort([0,1,λc,λd])
+      return @IT OverlappingSegments Segment(s1(maximum(p[1:2])), s1(minimum(p[3:4]))) f # CASE 4
```
juliohm: these two lines are hard to read, try to give a more descriptive name for `p` , and then introduce new lines with new definitions for `maximum(perm[1:2])` and `minimum(p[3:4])`. I think you don't need to compute the maximum or minimum if the vector p is already sorted. You know already where the maximum and minimum are.

src/intersections/segments.jl
```diff
@@ -3,159 +3,64 @@
...
     end
-  elseif determinatex || determinatey # CASE (II)
-    if !(determinatex ? wxzero : wyzero)
-      if x1 ≈ y1 || x1 ≈ y2 || x2 ≈ y1 || x2 ≈ y2
-        # configuration (3)
-        return @IT CornerTouchingSegments intersectpoint(s1, s2) f
+  # in same plane, not parallel
+  else
+    λ₁ = isapprox(λ₁, zero(T), atol=atol(T)) ? zero(T) : (isapprox(λ₁, one(T), atol=atol(T)) ? one(T) : λ₁)
+    λ₂ = isapprox(λ₂, zero(T), atol=atol(T)) ? zero(T) : (isapprox(λ₂, one(T), atol=atol(T)) ? one(T) : λ₂)
```
juliohm: again, these two lines could be replaced by a function call, where the function is defined in the beginning of the function.

test/intersections.jl
```diff
@@ -17,6 +17,10 @@
     @test s1 ∩ s2 == Segment(P2(0.5,0.0), P2(1,0))
     @test s2 ∩ s1 == Segment(P2(0.5,0.0), P2(1,0))
 
+    s1 = Segment(P2(0,0), P2(1,-1))
+    s2 = Segment(P2(0.5,-0.5), P2(1.5,-1.5))
+    @test s1 ∩ s2 == s2 ∩ s1 == Segment(P2(0.5,-0.5), P2(1,-1))
```
juliohm, line 22: Aren't s1 and s2 parallel without intersection? Shouldn't the return be NoIntersection? Sorry if I misread the coordinates somehow, just plotted it on my head.

juliohm (changes requested):
Thank you @dorn-gerhard for working on this challenging PR. Can you please make sure that all the branches of the code are covered by our test suite? Do we already have tests for all 5 cases explained in the docstring?

src/intersections/segments.jl
dorn-gerhard: yes i need the index, I implement your second suggestion
``` julia
v₁ = (b-a)
i = argmax(abs.(v₁))
```

src/intersections/segments.jl
dorn-gerhard: Yes, acutally I thought so too. 
first version:
``approx(λ, x) = isapprox(λ, x, atol=atol(T)) ? x : λ``
could also be
``approx!(λ, x) = isapprox(λ, x, atol=atol(T)) ? x : λ``

src/intersections/segments.jl
dorn-gerhard: 1 and 0 are more readable, i prefer them.

src/intersections/segments.jl
dorn-gerhard: p is a vector of parameters, so i would call it ``parameters``

ah, you are right, no maximum / minimum necessary here.

src/intersections/segments.jl
dorn-gerhard: i will use the helper function ``approx``

test/intersections.jl
dorn-gerhard: grafik

they are collinear with negative slope, that caused the old version of intersectcollinear to be erroneous.

dorn-gerhard:
I will need a bit of time to work on the test cases, intersectiontypes for 2d are missing, i can provide the 5 test cases and the 
8 subcases for collinear.

Actually we are lucky that some test cases which compare segments work, as the order of points is still important (we could use the ``ringequal`` function later or map ``isequal`` to ``ringequal`` at least for Segments (to be discussed in issue #310 )
``@test s1 ∩ s2 == s2 ∩ s1 == Segment(P2(0,0), P2(1,0))``

src/intersections/segments.jl
```diff
@@ -1,161 +1,64 @@
...
-    if !wxzero && !wyzero
-      # configuration (1)
-      return @IT CrossingSegments intersectpoint(s1, s2) f
+function intersection(f, s1::Segment{N,T}, s2::Segment{N,T}) where {N,T}
+  a, b = s1(0), s1(1)
+  c, d = s2(0), s2(1)
+
+  λ₁, λ₂, r, rₐ = intersectparameters(a, b, c, d)
+  # helper function to round values close to x (one or zero)
+  approx(λ, x) = isapprox(λ, x, atol=atol(T)) ? x : λ
```
```suggestion

  # helper function to round values close to x (one or zero)
  mayberound(λ, x) = isapprox(λ, x, atol=atol(T)) ? x : λ
```

I think this name is more descriptive of the current behavior.

src/intersections/segments.jl
```diff
@@ -1,161 +1,64 @@
...
+
+  if r ≠ rₐ # not in same plane or parallel
+    return @IT NoIntersection nothing f #CASE 5
+  elseif r == rₐ == 1 # collinear
+    # find parameters λc and λd for points c and d in s1
+    # use dimension with largest vector component to avoid division by zero
+    v₁ = b-a
```
```suggestion
    v₁ = b - a
```

src/intersections/segments.jl
```diff
@@ -1,161 +1,64 @@
...
+    i = argmax(abs.(v₁))
+    λc, λd = ((c-a)[i], (d-a)[i]) ./ v₁[i]
```
juliohm: please add space before and after `-` to improve readability

src/intersections/segments.jl
juliohm: Is there a reason why we add the suffix 1 to this variable name? Can't we simply go with `v`?

juliohm:
Looking forward to this one 💯 it is nice to see a PR where we delete more lines than we add, and increase the generality of the algorithm at the same time.

Feel free to request a review when the tests are in place 👍🏽

juliohm:
@dorn-gerhard did you have time to take a look into this?

dorn-gerhard:
I added a bunch of test cases for 2D with special focus on colinear case, including tests on IntersectionType

grafik

I didn't touch the old 2D test cases, of which some may have become obsolete.

src/intersections/segments.jl
juliohm: @dorn-gerhard can you adjust this code style issue?

juliohm: there are multiple places where spaces can help readability

src/intersections/segments.jl
juliohm: here we don't need the suffix 1?

src/intersections/segments.jl
juliohm: what do you think of the name suggestion?

dorn-gerhard: yes maybe `approx`  is to similar to the latex command. i adopted the code

src/intersections/segments.jl
dorn-gerhard: i removed the subscript

src/intersections/segments.jl
dorn-gerhard: i added more spaces

test/intersections.jl
```diff
@@ -103,51 +108,184 @@
...
-    s3 = Segment(P3(0.5, 0.0, 0.0), P3(1.5, 0.0, 0.0))
-    s4 = Segment(P3(0.0, 1.0, 0.0), P3(0.0, -2.0, 0.0))
-    s5 = Segment(P3(-1.0, 1.0, 0.0), P3(2.0, -2.0, 0.0))
-    s6 = Segment(P3(0.0, 0.0, 0.0), P3(0.0, 1.0, 0.0))
-    s7 = Segment(P3(-1.0, 1.0, 0.0), P3(-1.0, -1.0, 0.0))
-    s8 = Segment(P3(-1.0, 1.0, 1.0), P3(-1.0, -1.0, 1.0))
-    s9 = Segment(P3(0.5, 1.0, 1.0), P3(0.5, -1.0, 1.0))
-    s10 = Segment(P3(0.0, 1.0, 0.0), P3(1.0, 1.0, 0.0))
-    s11 = Segment(P3(1.5, 0.0, 0.0), P3(2.5, 0.0, 0.0))
-    s12 = Segment(P3(1.0, 0.0, 0.0), P3(2.0, 0.0, 0.0))
```
juliohm: Can you please undo all these name changes that aren't related to the changes in the source? The less code we change in this PR the better.

test/intersections.jl
```diff
@@ -103,51 +108,184 @@
...
+    @test isnothing(s₁ ∩ s₁₁)
+    @test intersection(s₁, s₁₂) |> type == CornerTouchingSegments
+    @test s₁ ∩ s₁₂ ≈ P3(1.0, 0.0, 0.0)
 
     # type stability tests
-    s1 = Segment(P2(0,0), P2(1,0))
-    s2 = Segment(P2(0.5,0.0), P2(2,0))
-    @inferred someornone(s1, s2)
+    s₁ = Segment(P2(0,0), P2(1,0))
+    s₂ = Segment(P2(0.5,0.0), P2(2,0))
```
juliohm: same here, please undo these modifications to tests that aren't strictly necessary.

dorn-gerhard: I thought converging to the same convention namely using subscripts would be beneficial for the coding style.
but here you are: subscripts undone

juliohm: It is usually a good idea, but in this case the code style of the surrounding code is already defined. More importantly, the code changes that are proposed here are very delicate and we would like to keep our history of commits clean so that people understand what happened in the past. Changes to these tests would just pollute the history for someone that is just trying to debug a problem.

src/intersections/segments.jl
```diff
@@ -1,161 +1,64 @@
...
+      return @IT NoIntersection nothing f # CASE 5
+    elseif (λc == 0 && λd < 0) ||  (λd == 0 && λc < 0)
+      return @IT CornerTouchingSegments a f # CASE 3
+    elseif (λc == 1 && λd > 1) || (λd == 1 && λc > 1)
+      return @IT CornerTouchingSegments b f # CASE 3
     else
-      # configuration (5)
-      return @IT NoIntersection nothing f
+      parameters = sort([0,1,λc,λd])
+      return @IT OverlappingSegments Segment(s1(parameters[2]), s1(parameters[3])) f # CASE 4
```
```suggestion
      params = sort([0,1,λc,λd])
      p1 = s1(params[2])
      p2 = s1(params[3])
      return @IT OverlappingSegments Segment(p1, p2) f # CASE 4
```

test/intersections.jl
```diff
@@ -103,6 +108,139 @@
...
     @test s1 ∩ s2 ≈ s2 ∩ s1 ≈ P2(0.94798386, 0.5344541)
 
+    s₁ = Segment(P2(0,0), P2(3,4))
+    s₂ = Segment(P2(1,2), P2(3,-2))
+    s₃ = Segment(P2(2,0), P2(-2,0))
+    s₄ = Segment(P2(0,0), P2(1,2))
+    s₅ = Segment(P2(1,2), P2(3,4))
+    s₆ = Segment(P2(-1,-4/3), P2(0,0))
+    s₇ = Segment(P2(1,2), P2(0,4))
+    s₈ = Segment(P2(4,16/3), P2(3,4))
```
juliohm, line 118: Notice that the code style you adopted in this section is different than the surrounding code style. We can ignore it this time because these are just tests, but it is important to try to reuse the surrounding code style.

juliohm (approved):

juliohm:
Amazing work @dorn-gerhard , one more step towards a robust intersection library 💯

PR #318, opened 2022-11-01
Add intersection of Lines and Rays

dorn-gerhard:
This PR adds the intersection of Lines and Rays. 

Tests are added in 2D and 3D (special case of skew passing) according to the following image:

grafik


src/intersections.jl
```diff
@@ -135,6 +140,7 @@ include("intersections/boxes.jl")
 include("intersections/segments.jl")
 include("intersections/rays.jl")
 include("intersections/raysegment.jl")
+include("intersections/lineray.jl")
```
juliohm: Let's follow the naming pattern and use rayline.jl instead

src/intersections.jl
```diff
@@ -75,6 +75,11 @@ Type `IntersectionType` in a Julia session to see the full list.
   CornerTouchingRaySegment
   OverlappingRaySegment
 
+  # line-ray intersection
+  CrossingLineRay
+  TouchingLineRay
+  OverlappingLineRay
```
juliohm: Similarly, let's follow the naming pattern and start with Ray instead of Line, so that you have CrossingRayLine.

juliohm (changes requested):
Review attached.

src/intersections.jl
dorn-gerhard: ah i had the naming convention to put the more general objects first, so regarding lines > rays > segments my plan was to use
`lineray`, `linesegment` and `raysegment`.

this stems from my intuition, that a line intersects a ray and a segment.
What do you think?

juliohm: It would have been a good convention but we already have RayBox, SegmentTriangle and other names that start with a object that is "less general" for some sense of generality to an object that is more general. I think the easiest approach now is to rename to make it compatible with the surrounding code.

src/intersections/lineray.jl
```diff
@@ -0,0 +1,33 @@
...
+
+#=
+The intersection type can be one of four types:
+
+1. intersect at one inner point (CrossingLineRay -> Point)
+2. intersect at origin of ray (TouchingLineRay -> Point)
+3. overlap of line and ray (OverlappingLineRay -> Ray)
+4. do not overlap nor intersect (NoIntersection)
+=#
+function intersection(f, l1::Line{N,T}, r1::Ray{N,T}) where {N,T}
```
juliohm: We also need to rename the order of arguments here to match the new naming convention.

dorn-gerhard: done ;)

src/intersections.jl
dorn-gerhard: ok, I committed the requested changes

src/intersections/rayline.jl
```diff
@@ -0,0 +1,33 @@
...
+
+#=
+The intersection type can be one of four types:
+
+1. intersect at one inner point (CrossingRayLine -> Point)
+2. intersect at origin of ray (TouchingRayLine -> Point)
+3. overlap of line and ray (OverlappingRayLine -> Ray)
+4. do not overlap nor intersect (NoIntersection)
+=#
+function intersection(f, r1::Ray{N,T}, l1::Line{N,T}) where {N,T}
```
juliohm, line 13: can we get rid of the "1" suffix in the arguments? We only have one ray and one line in this case.

dorn-gerhard: well I would keep it, since `r` is already used for the rank.
Letter + number is a good convention to label primitives or polytopes.
I have done the same in `raysegments.jl`

juliohm: The clash with the rank variable is a good reason to keep the 1 suffix here. It is still strange, but no big deal.

juliohm (approved):

juliohm:
Thanks for these great improvements 💯

PR #319, opened 2022-11-04
adding the util function mayberound

dorn-gerhard:
I thought it might be useful to add the util function `mayberound`. 
We already use it in `segments.jl` (function is only locally defined), and it could be used as well in 
``rays.jl``, ``rayline.jl``, ``raysegment.jl``.


src/utils.jl
```diff
@@ -118,4 +118,25 @@ function householderbasis(n)
       u, v = v, u
   end
   u, v
-end
\ No newline at end of file
+end
+
```
juliohm, line 122:
```suggestion
```

src/utils.jl
```diff
@@ -118,4 +118,25 @@ function householderbasis(n)
...
+
+"""
+    mayberound(λ, x, tol)
+
+Round `λ` to `x` if it is within the tolerance `tol`
```
```suggestion
Round `λ` to `x` if it is within the tolerance `tol`.
```

src/utils.jl
```diff
@@ -118,4 +118,25 @@ function householderbasis(n)
...
+      error("Tolerances too large for ")
+    else
+      return x[findfirst(L)]
+    end
+  else
+    return λ
+  end
+end
+
+mayberound(λ::Vector, x, tol) = [mayberound(i, x, tol) for i in λ]
```
juliohm: We don't need this method as we can easily use dot syntax.

test/utils.jl
```diff
@@ -37,4 +37,13 @@
...
     @test u × v ≈ n ./ norm(n)
   end
+
+  @test mayberound(1.1, 1, 0.2) ≈ 1
+  @test mayberound(1.1, 1, 0.10000000000000001) ≈ 1.1
+  @test mayberound(1.1, 1, 0.05) ≈ 1.1
+
+  @test mayberound(1.1, 1:5, 0.2) ≈ 1
+
+  @test all(mayberound([1.1, 3.3, 3.1, 4.3], 1:5, 0.2) .≈ [1, 3.3, 3, 4.3])
```
```suggestion
  @test all(mayberound.([1.1, 3.3, 3.1, 4.3], 1:5, 0.2) .≈ [1, 3.3, 3, 4.3])
```

src/utils.jl
```diff
@@ -118,4 +118,25 @@ function householderbasis(n)
...
\ No newline at end of file
+end
+
+
+"""
+    mayberound(λ, x, tol)
+
+Round `λ` to `x` if it is within the tolerance `tol`
+"""
+function mayberound(λ::Number, x, tol)
```
```suggestion
function mayberound(λ, x, tol) 
```

juliohm (changes requested):
Thanks for extracting the function for reuse in other places.

test/utils.jl
dorn-gerhard: that yields an error without the vector function
 `@test all(mayberound.([1.1, 3.3, 3.1, 4.3], Ref(1:5), 0.2) .≈ [1, 3.3, 3, 4.3])`
would work

src/utils.jl
dorn-gerhard: done in the commit

src/utils.jl
dorn-gerhard: I removed it

src/utils.jl
```diff
@@ -118,4 +118,22 @@ function householderbasis(n)
...
\ No newline at end of file
+end
+
+"""
+    mayberound(λ, x, tol)
+
+Round `λ` to `x` if it is within the tolerance `tol`.
+"""
+function mayberound(λ, x, tol) 
+  L = isapprox.(λ, x, atol = tol)
```
```suggestion
  L = isapprox(λ, x, atol = tol)
```

src/utils.jl
```diff
@@ -118,4 +118,22 @@ function householderbasis(n)
...
+  if any(L)
+    if sum(L) > 1
+      error("Tolerances too large for ")
+    else
+      return x[findfirst(L)]
+    end
+  else
+    return λ
+  end
+end
```
juliohm: This logic here is overly complicated and slow. We just need to treat the scalar case and use broadcast at the call site inside the intersection algorithms if necessary.

juliohm (changes requested):
The helper function makes sense with the scalar case if it is being used in many places, otherwise I don't think we even need this function.

src/utils.jl
dorn-gerhard: I wanted to have the option to use
`mayberound(λ, [0,1], atol(T))`
instead of 
`mayberound(mayberound(λ, 0, atol(T)), 1, atol(T))`

but I guess a vectorized function is too complicated (and slow)

I now implemented the simple version: 
`mayberound(λ, x, tol) = isapprox(λ, x, atol = tol) ? x : λ`

Broadcasting does not work, but one can call the function iteratively.

src/utils.jl
```diff
@@ -118,4 +118,11 @@ function householderbasis(n)
...
-end
\ No newline at end of file
+end
+
+"""
+    mayberound(λ, x, tol)
+
+Round `λ` to `x` if it is within the tolerance `tol`.
+"""
+mayberound(λ, x, tol) = isapprox(λ, x, atol = tol) ? x : λ
```
juliohm: I think the argument `tol` should be renamed to `atol` to reflect the exact kind of tolerance that is being considered here. Also, it could default to `atol(T)` if you annotate lambda::T in the function signature.

dorn-gerhard: good idea, committed ;)

src/utils.jl
juliohm: Yes, in terms of design it is better to keep the signature simple and call the function twice. One thing that you can consider if this is often used with 0 and 1 is to define a local function `maybe0or1` that calls `mayberound` twice. But I am still not sure how often we are planning to use `mayberound` to justify adding it here as a utility function.

Are you expecting to call this function more than 2 times in our intersection algorithms?

dorn-gerhard: in `segments.jl` `maybe0or1` is needed 4 times, but i can define `maybe0or1` locally there.

in `raysegment.jl` `mayberound` is needed 5 times or 3 times and 1 `maybe0or1`, but here it can be called recursivly.

juliohm (approved):

Issue #320, opened 2022-11-09
Precision of segment intersection


henry2004y:
The following code shows difference between version 0.24 and 0.25:

```julia
seg1 = Segment(Point(2.0, 2.0), Point(3.0, 1.0))
seg2 = Segment(Point(2.125005,1.875003), Point(50000.0,30000.0))
seg3 = Segment(Point(2.125005,1.875003), Point(50002.125005,30001.875003))
# Meshes v0.24
seg1 ∩ seg2 #Nothing
seg1 ∩ seg3 #Nothing
# Meshes v0.25
seg1 ∩ seg2 #Nothing
seg1 ∩ seg3 #Point(2.125005, 1.875003)
```

Analytically, `Point(2.125, 1.875)` lies exactly along `seg1`, so Meshes v0.24 gives the correct result `Nothing`. This seems to be related to precisions of floating numbers?

juliohm:
@dorn-gerhard as you can see that PR we merged is quite delicate. Can you please take a look at this? It would be great if we could provide a hotfix.

juliohm:
Thank you @henry2004y for opening the issue. Please let us know if you have a fix, feel free to submit a PR as well.

dorn-gerhard:
the reason is that we currently use the relative parameter `λ` of a segment which is between zero and one and so indicates intersections relative to the length of a segment.
the intersection of seg1 and seg3 seen from seg3 is `λ = -9.999999999732445e-11` which is rounded according to the tolerance of Float64 (`atol(Float64) -> 10^-10`)

(Use `Meshes.intersectparameters(seg1(0), seg1(1), seg3(0), seg3(1)) to see the resulting parameters)

We can fix that by using the length of Segments (the defining points of Lines and Rays).
and use for instance
``` julia 
a, b = s1(0), s1(1/length(s1))
c, d = s2(0), s2(1/length(s2))
```
then the resulting parameters will have an actual length scale.

Please give feedback. 
Based on your thoughts, I can prepare a PR (changes are needed in: `rays.jl, lines.jl, segments.jl, raylines.jl, raysegment.jl`)

juliohm:
I remember having to use a squared tolerance somewhere in the old code to avoid these rounding issues because the old algorithm was computing areas. If tol is the tolerance for lengths, then tol^2 is the tolerance for areas.

@dorn-gerhard, if you know a good fix for the new algorithm already, feel free to start a PR. I am happy to review it.

[...]

PR #321, opened 2022-11-12
 Precision of segment intersection #320 

dorn-gerhard:
solves the issue #320 which was caused by using segment parameters `λ` that scale with the length of a segment. 
So the parameter λ₁ describing the segment `s1` did not represent actual length scale. 
In this PR segment parameters are given by their actual length.

Example (Issue 320): If a segment `s1` is very large (e.g. `length(s1) == 10^5`)
and a possible intersection point `P1` is quite close to the first point `s1(0)` of the segment (e.g. `length(P1 - s1(0)) == 10^-5`)
then this point has been regarded in terms of the segment parameter `λ₁` to be at distance 10^-10 since ``λ₁`` was scaled to the length of the segment. That caused the point ``P1`` to be interpreted to be identical with the first point of the segment.

Now (with this PR) the absolute distance between point `P1` and first segment point (`s1(0)`) is considered.

I added the tests from the issue in the test cases (different behaviour for different types (Float32 / Float64)

I will adopt the missing files
- rays
- lines
- raysegment
- rayline

in the next commits if the strategy how to solve the problem is approved


src/intersections/segments.jl
```diff
@@ -13,54 +13,57 @@ function intersection(f, s1::Segment{N,T}, s2::Segment{N,T}) where {N,T}
   a, b = s1(0), s1(1)
   c, d = s2(0), s2(1)
 
-  λ₁, λ₂, r, rₐ = intersectparameters(a, b, c, d)
-  # helper function to round values close to x (one or zero)
-  mayberound(λ, x) = isapprox(λ, x, atol=atol(T)) ? x : λ
+  len1, len2 = length(s1), length(s2)
+  b₀ = a + 1/len1 * (b - a) # corresponds to s1(1/length)
+  d₀ = c + 1/len2 * (d - c)
```
juliohm: Why not just write s1(1/length(s1)) ?

src/intersections/segments.jl
```diff
@@ -13,54 +13,57 @@ function intersection(f, s1::Segment{N,T}, s2::Segment{N,T}) where {N,T}
...
     # use dimension with largest vector component to avoid division by zero
-    v = b - a
+    v = b₀ - a
     i = argmax(abs.(v))
     λc, λd = ((c - a)[i], (d - a)[i]) ./ v[i]
-    λc = mayberound(mayberound(λc, 0), 1)
-    λd = mayberound(mayberound(λd, 0), 1)
-    if (λc > 1 && λd > 1) || (λc < 0 && λd < 0)
+    λc = mayberound(mayberound(λc, 0, atol(T)), len1, atol(T))
+    λd = mayberound(mayberound(λd, 0, atol(T)), len1, atol(T))
```
juliohm: We don't need atol(T) as it is default

src/intersections/segments.jl
```diff
@@ -13,54 +13,57 @@ function intersection(f, s1::Segment{N,T}, s2::Segment{N,T}) where {N,T}
...
       p1 = s1(params[2])
       p2 = s1(params[3])
       return @IT OverlappingSegments Segment(p1, p2) f # CASE 4
     end
   else # in same plane, not parallel
-    λ₁ = mayberound(mayberound(λ₁, 0), 1)
-    λ₂ = mayberound(mayberound(λ₂, 0), 1)
-    if λ₁ < 0 || λ₂ < 0 || λ₁ > 1 || λ₂ > 1
+    λ₁ = mayberound(mayberound(λ₁, 0, atol(T)), len1, atol(T))
+    λ₂ = mayberound(mayberound(λ₂, 0, atol(T)), len2, atol(T))
```
juliohm: same here, no need for atol(T)

juliohm (changes requested):
@dorn-gerhard appreciate if you can take a look into the review comments. Also, can you please explain why this modification addresses the issue for new comers? If you can add a comment somewhere saying why it is important to use the lengths of segments this way, it can help future contributors.

src/intersections/segments.jl
dorn-gerhard: `segment.jl` (in Polytopes) throws an error since 1 / length(s1) can be larger than 1:

``` julia
function (s::Segment)(t)
  if t < 0 || t > 1
    throw(DomainError(t, "s(t) is not defined for t outside [0, 1]."))
  end
  a, b = s.vertices
  a + t * (b - a)
end
```

src/intersections/segments.jl
dorn-gerhard: only works if I change it to
``` julia
λc = mayberound(mayberound(λc, zero(T)), len1)
λd = mayberound(mayberound(λd, zero(T)), len1)
```

I can do that

src/intersections/segments.jl
dorn-gerhard: I switch to zero(T) then also here (cause typeof(0) -> Int64)

dorn-gerhard:
I added a more detailed description in the beginning of the PR to explain why changes were necessary and what changes have been made.

src/intersections/rayline.jl
```diff
@@ -14,16 +14,20 @@ function intersection(f, r1::Ray{N,T}, l1::Line{N,T}) where {N,T}
   a, b = r1(0), r1(1)
```
juliohm: @dorn-gerhard can we get the arguments of these functions renamed to something more descriptive first? Let's rename r1 to ray and l1 to line so that we don't get lost with all the r variables in the code meaning ranks.

dorn-gerhard: I totally agree, better to use `ray`, `seg` and `line`.

src/intersections/rayline.jl
```diff
@@ -10,22 +10,26 @@ The intersection type can be one of four types:
...
-function intersection(f, r1::Ray{N,T}, l1::Line{N,T}) where {N,T}
-  a, b = r1(0), r1(1)
-  c, d = l1(0), l1(1)
+function intersection(f, ray::Ray{N,T}, line::Line{N,T}) where {N,T}
+  a, b = ray(0), ray(1)
+  c, d = line(0), line(1)
 
-  λ₁, λ₂, r, rₐ = intersectparameters(a, b, c, d)
+  # use a normalized point for ray parameter λ₁
+  len1 = norm(b - a)
```
juliohm: Ok, now that we have better names for the arguments of this function, we need to remove the suffix "1" from variable names because it is not helpful. We can rename `len1` to simply `l` for example.

src/intersections/rayline.jl
```diff
@@ -10,22 +10,26 @@ The intersection type can be one of four types:
...
+  b₀ = a + 1/len1 * (b - a)
+
+  λ₁, λ₂, r, rₐ = intersectparameters(a, b₀, c, d)
```
juliohm: I am really trying to understand what is happening here. I can say that this asymmetry where a new variable b0 is defined and d0 is not defined doesn't feel natural or mathematically pleasing. Can you help the reader of the code (not just me) understand what is the rationale behind this new b0 variable? You can add a comment before this line discussing this length issues, etc.

src/intersections/rays.jl
```diff
@@ -11,49 +11,54 @@ The intersection type can be one of six types:
...
+  a, b = ray1(0), ray1(1)
+  c, d = ray2(0), ray2(1)
 
-  λ₁, λ₂, r, rₐ = intersectparameters(a, b, c, d)
+  # normalize points to gain parameters λ₁, λ₂ corresponding to arc lengths
+  len1, len2 = norm(b - a), norm(d - c)
+  b₀ = a + 1/len1 * (b - a)
+  d₀ = c + 1/len2 * (d - c)
+
+  λ₁, λ₂, r, rₐ = intersectparameters(a, b₀, c, d₀)
```
juliohm: Now in this case here you add b0 and d0...

src/intersections/rayline.jl
juliohm: Or if you really want to mean the length of the first argument, then l₁ would be more aligned with the code style.

dorn-gerhard: in the context of all intersection functions I would prevent to name it `l` or `l₁`, simply because in other intersections files and tests we use `l₁` for lines and changing all tests to line is not the preferred way.

What do you think about `len₁`?

src/intersections/rayline.jl
dorn-gerhard: we only need a rescaling if an intersection object has a fixed point to check (like the origin of ray or the end points of a segment). 

Since lines do not have such critical points it's not necessary to normalize its parameter. Note, that we do not use the second parameter at all in `rayline.jl` (we could use it in line 30 and write 
`return @IT CrossingRayLine line(λ₂) f # CASE 1` instead of 
`return @IT CrossingRayLine ray(λ₁/len1) f # CASE 1`

src/intersections/rays.jl
dorn-gerhard: both intersecting objects have "critical" points where the length scale matters (`mayberound` is used)

src/intersections/rayline.jl
juliohm: I think we need to be consistent in the src directory and make sure that all intersection algorithms follow the same naming convention for function arguments with explicit line, ray, seg. Regarding the test directory, I think we can call l1, l2, ... without issues.

src/intersections/lines.jl
```diff
@@ -54,7 +54,7 @@ function intersectparameters(a::Point{Dim,T}, b::Point{Dim,T},
   # the zero entries of the diagonal of R
   _, R = qr([A y])
   
-  # for Dim == 2 one has to check the L1 norm of rows as 
+  # for Dim == 2 one has to check the line₁ norm of rows as
```
```suggestion
  # for Dim == 2 one has to check the L1 norm of rows as 
```

src/intersections/rayline.jl
```diff
@@ -10,22 +10,26 @@ The intersection type can be one of four types:
...
-  a, b = r1(0), r1(1)
-  c, d = l1(0), l1(1)
+function intersection(f, ray::Ray{N,T}, line::Line{N,T}) where {N,T}
+  a, b = ray(0), ray(1)
+  c, d = line(0), line(1)
 
-  λ₁, λ₂, r, rₐ = intersectparameters(a, b, c, d)
+  # rescaling of point b necessary to gain a parameter λ₁ representing the arc length
+  len₁ = norm(b - a)
+  b₀ = a + 1/len₁ * (b - a)
```
juliohm:
```suggestion
  l₁ = norm(b - a)
  b₀ = a + 1/l₁ * (b - a)
```

src/intersections/rayline.jl
```diff
@@ -10,22 +10,26 @@ The intersection type can be one of four types:
...
     return @IT NoIntersection nothing f # CASE 4
   elseif r == rₐ == 1 # collinear
-    return @IT OverlappingRayLine r1 f # CASE 3
+    return @IT OverlappingRayLine ray f # CASE 3
   else # in same plane, not parallel
-    λ₁ = isapprox(λ₁, 0, atol=atol(T)) ? 0 : λ₁
+    λ₁ = mayberound(λ₁, zero(T))
     if λ₁ > 0
-      return @IT CrossingRayLine r1(λ₁) f # CASE 1
+      return @IT CrossingRayLine ray(λ₁/len₁) f # CASE 1
```
juliohm:
```suggestion
      return @IT CrossingRayLine ray(λ₁/l₁) f # CASE 1
```

src/intersections/rays.jl
```diff
@@ -11,49 +11,54 @@ The intersection type can be one of six types:
...
-  c, d = r2(0), r2(1)
+function intersection(f, ray₁::Ray{N,T}, ray₂::Ray{N,T}) where {N,T}
+  a, b = ray₁(0), ray₁(1)
+  c, d = ray₂(0), ray₂(1)
 
-  λ₁, λ₂, r, rₐ = intersectparameters(a, b, c, d)
+  # normalize points to gain parameters λ₁, λ₂ corresponding to arc lengths
+  len₁, len₂ = norm(b - a), norm(d - c)
+  b₀ = a + 1/len₁ * (b - a)
+  d₀ = c + 1/len₂ * (d - c)
```
juliohm:
```suggestion
  l₁, l₂ = norm(b - a), norm(d - c)
  b₀ = a + 1/l₁ * (b - a)
  d₀ = c + 1/l₂ * (d - c)
```

src/intersections/raysegment.jl
```diff
@@ -11,36 +11,41 @@ The intersection type can be one of five types:
...
-  c, d = s1(0), s1(1)
+function intersection(f, ray::Ray{N,T}, seg::Segment{N,T}) where {N,T}
+  a, b = ray(0), ray(1)
+  c, d = seg(0), seg(1)
 
-  λ₁, λ₂, r, rₐ = intersectparameters(a, b, c, d)
+  # normalize points to gain parameters λ₁, λ₂ corresponding to arc lengths
+  len₁, len₂ = norm(b - a), length(seg)
+  b₀ = a + 1/len₁ * (b - a)
+  d₀ = c + 1/len₂ * (d - c)
```
juliohm:
```suggestion
  l₁, l₂ = norm(b - a), length(seg)
  b₀ = a + 1/l₁ * (b - a)
  d₀ = c + 1/l₂ * (d - c)
```

src/intersections/raysegment.jl
```diff
@@ -49,22 +54,21 @@ function intersection(f, r1::Ray{N,T}, s1::Segment{N,T}) where {N,T}
...
-    λ₂ = isapprox(λ₂, zero(T), atol=atol(T)) ? zero(T) : λ₂
-    λ₂ = isapprox(λ₂, one(T), atol=atol(T)) ? one(T) : λ₂
-    if λ₁ < 0 || (λ₂ < 0 || λ₂ > 1)
+    λ₁ = mayberound(λ₁, zero(T))
+    λ₂ = mayberound(mayberound(λ₂, zero(T)), len₂)
+    if λ₁ < 0 || (λ₂ < 0 || λ₂ > len₂)
       return @IT NoIntersection nothing f
     elseif λ₁ == 0
-      if λ₂ == 0 || λ₂ == 1
+      if λ₂ == 0 || λ₂ == len₂
```
juliohm:
```suggestion
      if λ₂ == 0 || λ₂ == l₂
```

src/intersections/raysegment.jl
```diff
@@ -49,22 +54,21 @@ function intersection(f, r1::Ray{N,T}, s1::Segment{N,T}) where {N,T}
     end
   # in same plane, not parallel
   else
-    λ₁ = isapprox(λ₁, zero(T), atol=atol(T)) ? zero(T) : λ₁
-    λ₂ = isapprox(λ₂, zero(T), atol=atol(T)) ? zero(T) : λ₂
-    λ₂ = isapprox(λ₂, one(T), atol=atol(T)) ? one(T) : λ₂
-    if λ₁ < 0 || (λ₂ < 0 || λ₂ > 1)
+    λ₁ = mayberound(λ₁, zero(T))
+    λ₂ = mayberound(mayberound(λ₂, zero(T)), len₂)
+    if λ₁ < 0 || (λ₂ < 0 || λ₂ > len₂)
```
juliohm:
```suggestion
    if λ₁ < 0 || (λ₂ < 0 || λ₂ > l₂)
```

src/intersections/raysegment.jl
```diff
@@ -49,22 +54,21 @@ function intersection(f, r1::Ray{N,T}, s1::Segment{N,T}) where {N,T}
...
+      if λ₂ == 0 || λ₂ == len₂
         return @IT CornerTouchingRaySegment a f # CASE 3
       else
         return @IT MidTouchingRaySegment a f # CASE 2
       end
     else
-      if λ₂ == 0 || λ₂ == 1
-        return @IT MidTouchingRaySegment (λ₂ < 0.5 ? c : d)  f # CASE 2
+      if λ₂ == 0 || λ₂ == len₂
+        return @IT MidTouchingRaySegment (λ₂ < len₂/2 ? c : d)  f # CASE 2
```
juliohm:
```suggestion
      if λ₂ == 0 || λ₂ == l₂
        return @IT MidTouchingRaySegment (λ₂ < l₂/2 ? c : d)  f # CASE 2
```

src/intersections/raysegment.jl
```diff
@@ -49,22 +54,21 @@ function intersection(f, r1::Ray{N,T}, s1::Segment{N,T}) where {N,T}
...
       else
-        return @IT CrossingRaySegment r1(λ₁) f # CASE 1, equal to s1(λ₂)
+        return @IT CrossingRaySegment ray(λ₁/len₁) f # CASE 1, equal to seg(λ₂/len₂)
```
juliohm:
```suggestion
        return @IT CrossingRaySegment ray(λ₁/l₁) f # CASE 1, equal to seg(λ₂/l₂)
```

src/intersections/segments.jl
```diff
@@ -9,58 +9,61 @@ The intersection type can be one of six types:
...
+function intersection(f, seg₁::Segment{N,T}, seg₂::Segment{N,T}) where {N,T}
+  a, b = seg₁(0), seg₁(1)
+  c, d = seg₂(0), seg₂(1)
 
-  λ₁, λ₂, r, rₐ = intersectparameters(a, b, c, d)
-  # helper function to round values close to x (one or zero)
-  mayberound(λ, x) = isapprox(λ, x, atol=atol(T)) ? x : λ
+  len₁, len₂ = length(seg₁), length(seg₂)
+  b₀ = a + 1/len₁ * (b - a) # corresponds to seg₁(1/length)
+  d₀ = c + 1/len₂ * (d - c)
```
```suggestion
  l₁, l₂ = length(seg₁), length(seg₂)
  b₀ = a + 1/l₁ * (b - a)
  d₀ = c + 1/l₂ * (d - c)
```

src/intersections/segments.jl
```diff
@@ -9,58 +9,61 @@ The intersection type can be one of six types:
...
-    v = b - a
+    v = b₀ - a
     i = argmax(abs.(v))
     λc, λd = ((c - a)[i], (d - a)[i]) ./ v[i]
-    λc = mayberound(mayberound(λc, 0), 1)
-    λd = mayberound(mayberound(λd, 0), 1)
-    if (λc > 1 && λd > 1) || (λc < 0 && λd < 0)
+    λc = mayberound(mayberound(λc, zero(T)), len₁)
+    λd = mayberound(mayberound(λd, zero(T)), len₁)
+    if (λc > len₁ && λd > len₁) || (λc < 0 && λd < 0)
```
```suggestion
    λc = mayberound(mayberound(λc, zero(T)), l₁)
    λd = mayberound(mayberound(λd, zero(T)), l₁)
    if (λc > len₁ && λd > len₁) || (λc < 0 && λd < 0)
```

src/intersections/segments.jl
```diff
@@ -9,58 +9,61 @@ The intersection type can be one of six types:
...
       return @IT NoIntersection nothing f # CASE 5
     elseif (λc == 0 && λd < 0) ||  (λd == 0 && λc < 0)
       return @IT CornerTouchingSegments a f # CASE 3
-    elseif (λc == 1 && λd > 1) || (λd == 1 && λc > 1)
+    elseif (λc == len₁ && λd > len₁) || (λd == len₁ && λc > len₁)
```
```suggestion
    elseif (λc == l₁ && λd > l₁) || (λd == l₁ && λc > l₁)
```

src/intersections/segments.jl
```diff
@@ -9,58 +9,61 @@ The intersection type can be one of six types:
...
       return @IT CornerTouchingSegments b f # CASE 3
     else
-      params = sort([0,1,λc,λd])
-      p1 = s1(params[2])
-      p2 = s1(params[3])
-      return @IT OverlappingSegments Segment(p1, p2) f # CASE 4
+      params = sort([0, 1, λc/len₁, λd/len₁])
```
```suggestion
      params = sort([0, 1, λc/l₁, λd/l₁])
```

src/intersections/segments.jl
```diff
@@ -9,58 +9,61 @@ The intersection type can be one of six types:
...
+      p₂ = seg₁(params[3])
+      return @IT OverlappingSegments Segment(p₁, p₂) f # CASE 4
     end
   else # in same plane, not parallel
-    λ₁ = mayberound(mayberound(λ₁, 0), 1)
-    λ₂ = mayberound(mayberound(λ₂, 0), 1)
-    if λ₁ < 0 || λ₂ < 0 || λ₁ > 1 || λ₂ > 1
+    λ₁ = mayberound(mayberound(λ₁, zero(T)), len₁)
+    λ₂ = mayberound(mayberound(λ₂, zero(T)), len₂)
+    if λ₁ < 0 || λ₂ < 0 || λ₁ > len₁ || λ₂ > len₂
```
```suggestion
    λ₁ = mayberound(mayberound(λ₁, zero(T)), l₁)
    λ₂ = mayberound(mayberound(λ₂, zero(T)), l₂)
    if λ₁ < 0 || λ₂ < 0 || λ₁ > l₁ || λ₂ > l₂
```

src/intersections/segments.jl
```diff
@@ -9,58 +9,61 @@ The intersection type can be one of six types:
...
       return @IT NoIntersection nothing f # CASE 5
     # 8 cases remain
     elseif λ₁ == 0
-      if λ₂ == 0 || λ₂ == 1
+      if λ₂ == 0 || λ₂ == len₂
```
```suggestion
      if λ₂ == 0 || λ₂ == l₂
```

src/intersections/segments.jl
```diff
@@ -9,58 +9,61 @@ The intersection type can be one of six types:
...
         return @IT CornerTouchingSegments a f # CASE 3
       else
         return @IT MidTouchingSegments a f # CASE 2
       end
-    elseif λ₁ == 1
-      if λ₂ == 0 || λ₂ == 1
+    elseif λ₁ == len₁
+      if λ₂ == 0 || λ₂ == len₂
```
```suggestion
    elseif λ₁ == l₁
      if λ₂ == 0 || λ₂ == l₂
```

src/intersections/segments.jl
```diff
@@ -9,58 +9,61 @@ The intersection type can be one of six types:
...
         return @IT CornerTouchingSegments b f # CASE 3
       else
         return @IT MidTouchingSegments b f # CASE 2
       end
-    elseif λ₂ == 0 || λ₂ == 1
+    elseif λ₂ == 0 || λ₂ == len₂
```
```suggestion
    elseif λ₂ == 0 || λ₂ == l₂
```

src/intersections/segments.jl
```diff
@@ -9,58 +9,61 @@ The intersection type can be one of six types:
...
       return @IT MidTouchingSegments (λ₂ == 0 ? c : d) f # CASE 2
     else
-      return @IT CrossingSegments s1(λ₁) f # CASE 1: equal to s2(λ₂)
+      return @IT CrossingSegments seg₁(λ₁/len₁) f # CASE 1: equal to seg₂(λ₂/len₂)
```
```suggestion
      return @IT CrossingSegments seg₁(λ₁/l₁) f # CASE 1: equal to seg₂(λ₂/l₂)
```

juliohm (changes requested):
Now we just need to replace all len1,len2 by l1, l2 and I think we are good to go. Thanks for cleaning this up and for the hotfix.

juliohm (approved):
Much cleaner 🎉

PR #322, opened 2022-11-14
Add Laplacian Smoothing Filter

felipeek:
Hello devs,

This is an implementation suggestion for the Laplacian Smoothing Filter. In practice this filter is just a special case of the Taubin filter with `\mu=0`. However, having `\lambda > \mu` is not allowed in the Taubin filter, as written in Taubin's original work, and correctly enforced by this package. Therefore, it is not possible to use the Taubin filter to 'fallback' to the laplacian filter with the current implementation of this package.

The proposed filter is simply a mimicked version of the existing Taubin filter, but simplified for `\mu=0`. Another implementation option would be to try to re-use Taubin's implementation, but given that the code is so small, I thought this would be clearer and more extensible.

I am currently using this code for a research project.

Thanks!

### DIFF
--- a/docs/src/transforms.md
+++ b/docs/src/transforms.md
@@ -74,3 +74,40 @@ viz(fig[1,1], mesh)
 viz(fig[2,1], smesh)
 fig
 ```
+
+## LaplacianSmoothing
+
+```@docs
+LaplacianSmoothing
+```
+
+```@example transforms
+using PlyIO
+
+# helper function to read *.ply files
+function readply(fname)
+  ply = load_ply(fname)
+  x = ply["vertex"]["x"]
+  y = ply["vertex"]["y"]
+  z = ply["vertex"]["z"]
+  points = Point3.(x, y, z)
+  connec = [connect(Tuple(c.+1)) for c in ply["face"]["vertex_indices"]]
+  SimpleMesh(points, connec)
+end
+
+# download mesh from the web
+file = download(
+  "https://raw.githubusercontent.com/juliohm/JuliaCon2021/master/data/beethoven.ply"
+)
+
+# read mesh from disk
+mesh = readply(file)
+
+# smooth mesh with 30 iterations
+smesh = mesh |> TaubinSmoothing(30)
+
+fig = Mke.Figure(resolution = (800, 1200))
+viz(fig[1,1], mesh)
+viz(fig[2,1], smesh)
+fig
+```
--- a/src/transforms/laplacian.jl
+++ b/src/transforms/laplacian.jl
@@ -1,0 +1,61 @@
+# ------------------------------------------------------------------
+# Licensed under the MIT License. See LICENSE in the project root.
+# ------------------------------------------------------------------
+
+"""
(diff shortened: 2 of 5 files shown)


docs/src/transforms.md
```diff
@@ -74,3 +74,40 @@ viz(fig[1,1], mesh)
...
+# download mesh from the web
+file = download(
+  "https://raw.githubusercontent.com/juliohm/JuliaCon2021/master/data/beethoven.ply"
+)
+
+# read mesh from disk
+mesh = readply(file)
+
+# smooth mesh with 30 iterations
+smesh = mesh |> TaubinSmoothing(30)
```
juliohm, line 107: This example is using TaubinSmoothing

docs/src/transforms.md
```diff
@@ -74,3 +74,40 @@ viz(fig[1,1], mesh)
 viz(fig[2,1], smesh)
 fig
 ```
+
+## LaplacianSmoothing
```
juliohm, line 78: Let's rename it to LaplaceSmoothing

src/transforms/laplacian.jl
```diff
@@ -0,0 +1,61 @@
...
+* Sorkine, O. 2005. [Laplacian Mesh Processing]
+  (http://dx.doi.org/10.2312/egst.20051044)
+"""
+struct LaplacianSmoothing{T} <: StatelessGeometricTransform
+  n::Int
+  λ::T
+
+  function LaplacianSmoothing{T}(n, λ) where {T}
+    new(n, λ)
+  end
```
juliohm, line 21: The inner constructor is only useful when we have additional lines of code to check the arguments. In this case you are only calling the built-in `new` function, so we can erase the inner constructor and it will exist by default.

src/transforms/laplacian.jl
```diff
@@ -0,0 +1,61 @@
...
+isrevertible(::Type{<:LaplacianSmoothing}) = true
+
+preprocess(::LaplacianSmoothing, mesh) =
+  laplacematrix(mesh, weights=:uniform)
+
+function applypoint(transform::LaplacianSmoothing, points, prep)
+  n = transform.n
+  λ = transform.λ
+  L = prep
+  _laplacian(points, L, n, λ), L
```
juliohm, line 37: let's rename to _laplace

juliohm (changes requested):
Thank you @felipeek, I think the best way forward is to refactor our current TaubinSmoothing to remove the restriction on lambda and mu. Basically, we need to

1. Rename TaubinSmoothing to LambdaMuSmoothing
2. Remove the restriction of lambda and mu from the inner constructor
3. Define a function TaubinSmoothing that checks the lambda and mu and then creates a LambdaMuSmoothing
4. Define a function LaplaceSmoothing that checks the lambda and mu and then creates a LambdaMuSmoothing

Can you help with these changes?

juliohm:
Fixed in https://github.com/JuliaGeometry/Meshes.jl/commit/1a0435ce886c6079b65e9f0122498ebb143eb92a

PR #324, opened 2022-11-15
add intersection of line and segment: linesegment.jl

dorn-gerhard:
This PR adds the last missing peace of intersections between lines, rays and segments: `linesegment.jl`

The included tests are illustrated in the following plot: 
grafik


juliohm (approved):

juliohm:
After this PR is merged we should probably decide on a good convention for the order of geometries involved in the intersection. Maybe a convention that favors geometries with lower "measure" as the first argument. In this case, we should use segmentline.jl instead of linesegment.jl. We can do this cleanup in a future separate PR 👍🏽

dorn-gerhard:
regarding name conventions of intersections:
I sticked to the following table (using the upper right triangle, first name comes from the row, second name from the column)

grafik

I would have preferred to favor more general geometries first, so
1. line
2. ray
3. segment
4. plane
5. box
6. triangle

But you decided to have `rayline` instead of `lineray`, 
so at the moment we have the order:
1. ray
2. line
3. segment
- plane
- box
- triangle

Please define the order 1 - 6, so this is defined once and for all time. If you want to change something on this naming convention we should do it soon before IntersectionTypes are implemented by some users.

dorn-gerhard:
and sorry for the format of the test files, I will fix that later / tomorrow.

juliohm:
> I would have preferred to favor more general geometries first, so
> 
> 1. line
> 2. ray
> 3. segment
> 4. plane
> 5. box
> 6. triangle

Maybe we should open an issue to converge on a convention. I understand the order proposed for the first 3 geometries in the list, but the last 3 are a bit unclear. If you can write down a pseudo-algorithm to decide the priority of geometries in this list we could document it somewhere.

juliohm:
> and sorry for the format of the test files, I will fix that later / tomorrow.

No problem, we are getting better at picking these formatting issues. They make a huge difference whenever we need to refactor big portions of the code base.

PR #339, opened 2022-12-16
Implement Ray-Plane intersection type and algorithm

jackbeagley:
Added appropriate types for Ray-Plane intersections, mostly taken from Segment-Plane intersections. Same tests added and used too, main difference is that rays extend past λ=1.

### DIFF
--- a/src/intersections/rayplane.jl
+++ b/src/intersections/rayplane.jl
@@ -12,9 +12,9 @@ function Meshes.intersection(f, r::Ray{3,T}, p::Plane{T}) where {T}
   # calculate components
   ln = direction(r) ⋅ n
   
-  # if ln is zero, the segment is parallel to the plane
+  # if ln is zero, the ray is parallel to the plane
   if isapprox(ln, zero(T), atol=atol(T))
-    # if the numerator is zero, the segment is coincident
+    # if the numerator is zero, the ray is coincident
     if isapprox(coordinates(pₒ - origin(r)) ⋅ n, zero(T), atol=atol(T))
       return @IT OverlappingRayPlane r f
     else
--- a/src/intersections/rayplane.jl
+++ b/src/intersections/rayplane.jl
@@ -5,7 +5,7 @@
 #=
 (https://en.wikipedia.org/wiki/Line-plane_intersection)
 =#
-function Meshes.intersection(f, r::Ray{3,T}, p::Plane{T}) where {T}
+function intersection(f, r::Ray{3,T}, p::Plane{T}) where {T}
   pₒ = coordinates(origin(p))
   n  = normal(p)
   
--- a/src/intersections/rayplane.jl
+++ b/src/intersections/rayplane.jl
@@ -1,0 +1,39 @@
+# ------------------------------------------------------------------
+# Licensed under the MIT License. See LICENSE in the project root.
+# ------------------------------------------------------------------
+
+#=
+(https://en.wikipedia.org/wiki/Line-plane_intersection)
+=#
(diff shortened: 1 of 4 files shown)


src/intersections/rayplane.jl
```diff
@@ -0,0 +1,39 @@
...
+
+#=
+(https://en.wikipedia.org/wiki/Line-plane_intersection)
+=#
+function Meshes.intersection(f, r::Ray{3,T}, p::Plane{T}) where {T}
+    pₒ = coordinates(origin(p))
+    n  = normal(p)
+    
+    # calculate components
+    ln = r.v ⋅ n
```
juliohm: Please use the public API to get the vector of the ray

src/intersections/rayplane.jl
```diff
@@ -0,0 +1,39 @@
...
+  
+    # if ln is zero, the segment is parallel to the plane
+    if isapprox(ln, zero(T), atol=atol(T))
+        # if the numerator is zero, the segment is coincident
+        if isapprox(coordinates(pₒ - r.p) ⋅ n, zero(T), atol=atol(T))
```
juliohm: public api for r.p

src/intersections/rayplane.jl
```diff
@@ -0,0 +1,39 @@
...
+            return @IT OverlappingRayPlane r f
+        else
+            return @IT NoIntersection nothing f
+        end
+    else
+        # calculate the ray parameter
+        λ = -(n ⋅ coordinates(r.p - pₒ)) / ln
```
juliohm: same here

src/intersections/rayplane.jl
```diff
@@ -0,0 +1,39 @@
...
+        if isapprox(λ, zero(T), atol=atol(T))
+        return @IT TouchingRayPlane r(zero(T)) f
+        end
+
+        # if λ is out of bounds for the ray, then there is no intersection
+        if (λ < zero(T))
+            return @IT NoIntersection nothing f
+        else
+            return @IT CrossingRayPlane r(λ) f
+        end
```
juliohm: incorrect indentation, please use the 2-space indentation style of the project

juliohm (changes requested):
Review attached.

jackbeagley:
> Review attached.

Thanks for those points @juliohm , I have just fixed them!

src/intersections/rayplane.jl
```diff
@@ -0,0 +1,39 @@
+# ------------------------------------------------------------------
+# Licensed under the MIT License. See LICENSE in the project root.
+# ------------------------------------------------------------------
+
+#=
+(https://en.wikipedia.org/wiki/Line-plane_intersection)
+=#
+function Meshes.intersection(f, r::Ray{3,T}, p::Plane{T}) where {T}
```
juliohm:
```suggestion
function intersection(f, r::Ray{3,T}, p::Plane{T}) where {T}
```

juliohm:
@jackbeagley you mean that this PR is a copy of the existing segment-plane code plus the additional flexibility to allow t>1? If that is the case we need to refactor the existing code to avoid duplication instead. We don't want to maintain to functions that are exactly the same with the exception of one line or so.

jackbeagley:
> @jackbeagley you mean that this PR is a copy of the existing segment-plane code plus the additional flexibility to allow t>1? If that is the case we need to refactor the existing code to avoid duplication instead. We don't want to maintain to functions that are exactly the same with the exception of one line or so.

Yes... that is a good point. Let me have a go at refactoring it and using this method we should get Lines for free too!

juliohm:
@jackbeagley can we close this in favor of a fresh PR with the refactored version?

PR #340, opened 2022-12-16
Rotation transformation

stla:
New transformation `Rotation`.

Note to @juliohm : I checked and there's no generic function to get the rotation matrix; one has to check the type with `if ... elseif` as I do.


juliohm:
> Note to @juliohm : I checked and there's no generic function to get the rotation matrix; one has to check the type with `if ... elseif` as I do.

There is because I used it in the past. You need to `convert(DCM, rotobj)` to get the DCM object.

src/transforms/rotations.jl
```diff
@@ -0,0 +1,82 @@
+# ------------------------------------------------------------------
+# Licensed under the MIT License. See LICENSE in the project root.
+# ------------------------------------------------------------------
+
+"""
+    Rotation(rotation)
+
+Perform the rotation represented by `rotation`, a
+`ReferenceFrameRotation` object from the `ReferenceFrameRotations` 
+library.
```
juliohm:
```suggestion
    Rotation(rot)

Rotate geometry or mesh with rotation `rot`
from ReferenceFrameRotations.jl
```

src/transforms/rotations.jl
```diff
@@ -0,0 +1,82 @@
...
+
+## Examples
+
+```julia
+using ReferenceFrameRotations
+Rotation(ReferenceFrameRotations.EulerAngleAxis(pi/4, [1, 0, 0]))
```
juliohm: We don't need to import the package in the examples. Also we should probably reexport ReferenceFrameRotations.jl if it is not exported already.

src/transforms/rotations.jl
```diff
@@ -0,0 +1,82 @@
...
+```
+
+"""
+struct Rotation{R} <: Meshes.GeometricTransform
```
juliohm:
```suggestion
struct Rotation{R} <: GeometricTransform
```

src/transforms/rotations.jl
```diff
@@ -0,0 +1,82 @@
...
+  rotation::R
```
juliohm:
```suggestion
  rot::R
```

src/transforms/rotations.jl
```diff
@@ -0,0 +1,82 @@
...
+"""
+    Rotation(axis, α; degrees=true)
+
+Perform the rotation of angle `α` around the axis directed by the 
+vector `axis`. The angle must be given in degrees if `degrees=true`, 
+otherwise in radians.
+
+"""
+function Rotation(axis::Vector{Real}, α::Real; degrees::Bool = true)
+  axisNorm = LinearAlgebra.norm2(axis)
```
juliohm: Please remove all package names, we don't need to add them as prefix.

src/transforms/rotations.jl
```diff
@@ -0,0 +1,82 @@
...
+  if axisNorm == 0
+    error("The `axis` vector is null.")
+  end
+  v = axis / axisNorm
+  if degrees
+    α = α * pi / 180
+  end
```
juliohm: All these checks should be performed already by ReferenceFrameRotations.jl, let them check it for us.

src/transforms/rotations.jl
```diff
@@ -0,0 +1,82 @@
...
+  R = ReferenceFrameRotations.EulerAngleAxis
+  rotation = R(α, v)
+  Rotation{R}(rotation)
```
juliohm:
```suggestion
  Rotation(EulerAngleAxis(α, v))
```

src/transforms/rotations.jl
```diff
@@ -0,0 +1,82 @@
...
+    M = ReferenceFrameRotations.angleaxis_to_dcm(rotation)
+    invM = ReferenceFrameRotations.inv(rotation)
+  elseif rtype <: ReferenceFrameRotations.EulerAngles
+    M = ReferenceFrameRotations.angle_to_dcm(rotation)
+    invM = ReferenceFrameRotations.inv(rotation)
+  elseif rtype <: ReferenceFrameRotations.Quaternion
+    M = ReferenceFrameRotations.quat_to_dcm(rotation)
+    invM = ReferenceFrameRotations.inv(rotation)
+  end
+  M, invM
```
juliohm:
```suggestion
  M = convert(DCM, transform.rot)
  M⁻¹ = inv(M)
  M, M⁻¹
```
Please double check, it should be something like that.

juliohm (changes requested):
Attached a couple of suggestions. Also, please rename the file to rotation.jl

stla:
> > Note to @juliohm : I checked and there's no generic function to get the rotation matrix; one has to check the type with `if ... elseif` as I do.
> 
> There is because I used it in the past. You need to `convert(DCM, rotobj)` to get the DCM object.

Not tested yet but I believe you. That said, we need the inverse matrix. I think it's better to use the inversion method for a specific type of rotation. The inverse matrix obtained from a `DCM` object is computed by inversion. The inverse matrix obtained from a `EulerAngleAxis` object is simply obtained as the matrix corresponding to the angle-axis rotation with opposite angle. This is more precise than numerical inversion. So if you agree, we need to check the type, but actually only one `if ... else` is needed, because there are only two functions for the inverse rotation: one for `DCM`, and the other one for all other types. Not sure I'm clear...

src/transforms/rotations.jl
stla: I've just checked. There's no check, you can pass a null vector or a non-normalized vector.

stla:
> > > Note to @juliohm : I checked and there's no generic function to get the rotation matrix; one has to check the type with `if ... elseif` as I do.
> > 
> > 
> > There is because I used it in the past. You need to `convert(DCM, rotobj)` to get the DCM object.
> 
> Not tested yet but I believe you. That said, we need the inverse matrix. I think it's better to use the inversion method for a specific type of rotation. The inverse matrix obtained from a `DCM` object is computed by inversion. The inverse matrix obtained from a `EulerAngleAxis` object is simply obtained as the matrix corresponding to the angle-axis rotation with opposite angle. This is more precise than numerical inversion. So if you agree, we need to check the type, but actually only one `if ... else` is needed, because there are only two functions for the inverse rotation: one for `DCM`, and the other one for all other types. Not sure I'm clear...

Here is what I mean:

```julia
function preprocess(transform::Rotation, object)
  rot = transform.rot
  local M, M⁻¹
  if typeof(rot) <: DCM
    M = rot
    M⁻¹ = inv_rotation(M)
  else
    M = convert(DCM, rot)
    M⁻¹ = convert(DCM, inv(rot))
  end
  M, M⁻¹
end
```

src/transforms/rotation.jl
```diff
@@ -0,0 +1,70 @@
...
+
+"""
+    Rotation(rot)
+
+Rotate geometry or mesh with rotation `rot` from ReferenceFrameRotations.jl. 
+
+## Examples
+
+```julia
+Rotation(ReferenceFrameRotations.EulerAngleAxis(pi/4, [1, 0, 0]))
```
juliohm:
```suggestion
Rotation(EulerAngleAxis(pi/4, [1, 0, 0]))
```

src/transforms/rotation.jl
```diff
@@ -0,0 +1,70 @@
...
+```
+
+"""
+struct Rotation{R} <: GeometricTransform
+  rot::R
+end
+
+"""
+    Rotation(axis, α; degrees=true)
```
juliohm: You need to choose a style and stick to it. Here you chose "axis" for one argument and `α` for another. You either choose descriptive names like `axis` and `angle` or symbolic names like `v` and `α`.

src/transforms/rotation.jl
```diff
@@ -0,0 +1,70 @@
...
+
+Perform the rotation of angle `α` around the axis directed by the 
+vector `axis`. The angle must be given in degrees if `degrees=true`, 
+otherwise in radians.
+
+"""
+function Rotation(axis::Vector{Real}, α::Real; degrees::Bool = true)
```
juliohm: Thinking more carefully, I think we don't need to create this additional outer constructor that does almost what the ReferenceFrameRotations.jl does with a few additional tweaks.

src/transforms/rotation.jl
```diff
@@ -0,0 +1,70 @@
...
+    α = α * pi / 180
+  end
+  Rotation(EulerAngleAxis(α, v))
+end
+
+isrevertible(::Type{<:Rotation}) = true
+
+function preprocess(transform::Rotation, object)
+  rot = transform.rot
+  local M, M⁻¹
```
juliohm: No need to use the keyword `local`, it is rarely needed in Julia code.

src/transforms/rotation.jl
```diff
@@ -0,0 +1,70 @@
...
+  if typeof(rot) <: DCM
+    M = rot
+    M⁻¹ = inv_rotation(M)
+  else
+    M = convert(DCM, rot)
+    M⁻¹ = convert(DCM, inv(rot))
+  end
```
juliohm: Notice that you tend to split your reasoning into different types. That is rarely needed in generic code. You can simply perform a single operation assuming that all types of rotation follow a consistent api:

```suggestion
  R, R⁻¹ = rot, inv_rotation(rot)
  M, M⁻¹ = convert.(DCM, (R, R⁻¹))
```

test/transforms.jl
```diff
@@ -41,4 +41,16 @@
...
+
+  @testset "Rotation" begin
+    # a rotation doesn't change the topology
+    trans = Rotation([1, 1, 1], 45)
+    @test TB.isrevertible(trans)
+    mesh  = readply(T, joinpath(datadir,"beethoven.ply"))
+    smesh = trans(mesh)
+    @test nvertices(smesh) == nvertices(mesh)
+    @test nelements(smesh) == nelements(mesh)
+    @test topology(smesh) == topology(mesh)
```
juliohm: We need other basic tests with trivial geometries to check that the coordinates have been correctly rotated. For example, you can create a simple `Triangle((0,0), (1,0), (0,1))` and apply the rotation to it. Then check that the vertices satisfy some condition.

juliohm (changes requested):
Review attached.

src/transforms/rotation.jl
stla: No! `inv_rotation` works only for `DCM`. For the other types, one has to use `inv`. Sorry I was not clear.

src/transforms/rotation.jl
stla: Even if `M` is introduced in a `if..else`? Strange, I was sure this was needed.

juliohm: A more idiomatic Julia code would define variables like:

```julia
var = if condition
  val1
else
  val2
end
```

juliohm: or even:

```julia
var = condition ? val1 : val2
```

src/transforms/rotation.jl
juliohm: Well, then we can replace inv_rotation by inv and everything should work right? You first call the correct inv method for the rotation type and then convert to DCM (matrix) representation for matrix multiplication. I think ReferenceFrameRotations.jl doesn't even need to convert the object to matrix to perform rotation. As far as I remember you can multiply rotation objects directly by vectors.

stla:
I've done the changes. I ran the unit tests and they are running for one hour now. Is it normal? That's the file `discretization.jl` which "blocks".

stla:
Ok I temporarily commented the other unit tests. There's a problem. I rotate a triangle and I check that the roated vertices that I manually enter belong to the rotated triangle, with the `in` function. But because of the numerical errors, there's a fail. Because a coordinate which is 0  theoretically is 6e-17.

stla:
Ok, instead of testing for the theoretical new vertices, I test for points interior to the new triangle.

src/transforms/rotation.jl
juliohm: @stla can you please double check this issue? I think you don't need to convert to DCM in order to apply the rotation. The ReferenceFrameRotations.jl package allows the application of the rotation with any rotation type.

test/transforms.jl
```diff
@@ -41,4 +41,24 @@
...
+    @test nelements(rmesh) == nelements(mesh)
+    @test topology(rmesh) == topology(mesh)
+
+    # check rotation on a triangle
+    triangle = Triangle(Point(0,0,0), Point(1,0,0), Point(0,1,0))
+    trans = Rotation(EulerAngleAxis(-pi/2, [0, 0, 1]))
+    rtriangle = trans(triangle)
+    @test in(Point(0,0,0), rtriangle)
+    @test in(Point(-0.01,0.9,0), rtriangle)
+    @test in(Point(-0.9,0.01,0), rtriangle)
```
juliohm: I think the right test here is to check that the vertices(rtriangle) are equal to specific points.

juliohm: You can do `@test p1 == p2` and the test will use approximate equality already implemented for `Point`

juliohm:
> I've done the changes. I ran the unit tests and they are running for one hour now. Is it normal?

It is not. On my machine the tests run in approximately 1m50s. Other people did run the tests on their own machines and the times are similar. See for example the timings in this other contribution: https://github.com/JuliaGeometry/Meshes.jl/pull/343

src/transforms/rotation.jl
stla: Was wrong. The `inv_rotation` function is generic, it works for all types of rotations. 
I already tried to multiply a rotation object by a vector and that didn't work. And I just checked the docs, there's no multiplication method for that.

src/transforms/rotation.jl
```diff
@@ -0,0 +1,42 @@
...
+"""
+struct Rotation{R} <: GeometricTransform
+  rot::R
+end
+
+isrevertible(::Type{<:Rotation}) = true
+
+function preprocess(transform::Rotation, object)
+  rot = transform.rot
+  R, R⁻¹ = rot, inv_rotation(rot)
```
juliohm: Their documentation is very explicit about the `inv_rotation` limitations:

https://juliaspace.github.io/ReferenceFrameRotations.jl/stable/man/inv_rotations

Can we stick to the suggested change where `inv(rot)` is used? I understand that `inv` works with any rotation type.

src/Meshes.jl
```diff
@@ -15,7 +15,7 @@ using Bessels: gamma
 using IterTools: ivec
 using StatsBase: AbstractWeights, Weights, quantile
 using Distances: PreMetric, Euclidean, Mahalanobis, evaluate
-using ReferenceFrameRotations: EulerAngles, DCM
+using ReferenceFrameRotations: EulerAngles, DCM, inv_rotation
```
```suggestion
using ReferenceFrameRotations: EulerAngles, DCM
```
This shouldn't be necessary if we stick to `inv`

test/transforms.jl
```diff
@@ -41,4 +41,25 @@
     @test nelements(smesh) == nelements(mesh)
     @test topology(smesh) == topology(mesh)
   end
+
+  @testset "Rotation" begin
+    # a rotation doesn't change the topology
+    trans = Rotation(EulerAngleAxis(pi/4, [1, 0, 0]))
```
juliohm: Please use the variable T in the tests to perform the operation with the correct floating point precision.

src/transforms/rotation.jl
stla: This doc is wrong. Here is the correct doc:

https://juliaspace.github.io/ReferenceFrameRotations.jl/stable/lib/library/#ReferenceFrameRotations.inv_rotation-Tuple{DCM}

test/transforms.jl
```diff
@@ -41,4 +41,25 @@
...
+    @test topology(rmesh) == topology(mesh)
+
+    # check rotation on a triangle
+    triangle = Triangle(Point(0, 0, 0), Point(1, 0, 0), Point(0, 1, 0))
+    trans = Rotation(EulerAngleAxis(-pi/2, [0, 0, 1]))
+    rtriangle = trans(triangle)
+    rpoints = vertices(rtriangle)
+    @test rpoints[1] == Point(0, 0, 0)
+    @test rpoints[2] == Point(0, 1, 0)
+    @test rpoints[3] == Point(-1, 0, 0)
```
juliohm: Same here, please use P3 to instantiate the points with the correct precision.

src/transforms/rotation.jl
juliohm: Please stick to `inv` as this is the universal function in Julia for inverses:

```julia
julia> using ReferenceFrameRotations

julia> r1 = EulerAngleAxis(1, [1, 1, 1])
EulerAngleAxis{Int64}:
  Euler angle : 1 rad  (57.2958°)
  Euler axis  : [1, 1, 1]

julia> r2 = EulerAngles(1, 1, 1)
EulerAngles{Int64}:
  R(Z) :  1 rad  ( 57.2958°)
  R(Y) :  1 rad  ( 57.2958°)
  R(X) :  1 rad  ( 57.2958°)

julia> r3 = Quaternion(1.0, 0.0, 0.0, 0.0)
Quaternion{Float64}:
  + 1.0 + 0.0⋅i + 0.0⋅j + 0.0⋅k

julia> r4 = DCM(I)
DCM{Bool}:
 1  0  0
 0  1  0
 0  0  1

julia> inv(r1)
EulerAngleAxis{Float64}:
  Euler angle : 1.0 rad  (57.2958°)
  Euler axis  : [-1.0, -1.0, -1.0]

julia> inv(r2)
EulerAngles{Int64}:
  R(X) : -1 rad  (-57.2958°)
  R(Y) : -1 rad  (-57.2958°)
  R(Z) : -1 rad  (-57.2958°)

julia> inv(r3)
Quaternion{Float64}:
  + 1.0 - 0.0⋅i - 0.0⋅j - 0.0⋅k

julia> inv(r4)
DCM{Float64}:
 1.0  0.0  0.0
 0.0  1.0  0.0
 0.0  0.0  1.0
```

test/transforms.jl
```diff
@@ -41,4 +41,25 @@
     @test nelements(smesh) == nelements(mesh)
     @test topology(smesh) == topology(mesh)
   end
+
+  @testset "Rotation" begin
+    # a rotation doesn't change the topology
+    trans = Rotation(EulerAngleAxis(T(pi/4), [1, 0, 0]))
```
```suggestion
    trans = Rotation(EulerAngleAxis(T(pi/4), T[1, 0, 0]))
```

test/transforms.jl
```diff
@@ -41,4 +41,25 @@
...
+    @test TB.isrevertible(trans)
+    mesh  = readply(T, joinpath(datadir,"beethoven.ply"))
+    rmesh = trans(mesh)
+    @test nvertices(rmesh) == nvertices(mesh)
+    @test nelements(rmesh) == nelements(mesh)
+    @test topology(rmesh) == topology(mesh)
+
+    # check rotation on a triangle
+    triangle = Triangle(P3(0, 0, 0), P3(1, 0, 0), P3(0, 1, 0))
+    trans = Rotation(EulerAngleAxis(T(-pi/2), [0, 0, 1]))
```
```suggestion
    trans = Rotation(EulerAngleAxis(T(-pi/2), T[0, 0, 1]))
```

juliohm:
@stla the ReferenceFrameRotations.jl package rotates the frame, so we always need to flip the sign of the angles to rotate the object in the opposite direction. I think we should swap the matrices M and inv(M) to achieve the desired result? If the user specifies a rotation with pi/2 then we need to rotate the frame by -pi/2, right?

stla:
> @stla the ReferenceFrameRotations.jl package rotates the frame, so we always need to flip the sign of the angles to rotate the object in the opposite direction. I think we should swap the matrices M and inv(M) to achieve the desired result? If the user specifies a rotation with pi/2 then we need to rotate the frame by -pi/2, right?

Ah ok. I was wondering why $x$ is mapped to $-y$ instead of $y$ by the rotation of angle $\pi/2$. I think you're right.

src/Meshes.jl
```diff
@@ -382,6 +382,7 @@ export
   LambdaMuSmoothing,
   LaplaceSmoothing,
   TaubinSmoothing,
+  Rotation,
```
juliohm: Let's rename it to RotateCoords for consistency with ScaleCoords

stla: Are you sure? I don't like it. `ScaleCoords` is nice: this is really a transformation of the coordinates, it depends on the Cartesian coordinates system. But a rotation is a transformation of the points, it does not depend on the Cartesian coordinates system.

juliohm: Can you elaborate on how rotation differs from scaling in this sense? It is not clear to me the distinction you are making.

stla: I mean a rotation is not defined as an operation on coordinates. It has a coordinates-free definition. We rotate a point or a vector, not a list of numbers.

juliohm: I see, you are probably referring to the geometric algebra definition? Here we are performing an operation on coordinates though, so our definition literally changes the coordinates, I think it is fine if we use RotateCoords for now. In the future, when we incorporate more geometric algebra into the package, we can come back and rename some of these transforms.

test/transforms.jl
```diff
@@ -41,4 +41,25 @@
...
+    # check rotation on a triangle
+    triangle = Triangle(P3(0, 0, 0), P3(1, 0, 0), P3(0, 1, 0))
+    trans = Rotation(EulerAngleAxis(T(pi/2), T[0, 0, 1]))
+    rtriangle = trans(triangle)
+    rpoints = vertices(rtriangle)
+    @test rpoints[1] == P3(0, 0, 0)
+    @test rpoints[2] == P3(0, 1, 0)
+    @test rpoints[3] == P3(-1, 0, 0)
+  end
+
```
```suggestion
```

src/Meshes.jl
juliohm: We can use the name RotateCoords temporarily. It is a good name to disambiguate with other transforms. It also has the benefit that end-users will simply ready ScaleCoords, RotateCoords and TranslateCoords and will immediately associate these transforms into the same group. Sometimes we need to be a bit less formal with these names to facilitate the lives of people who are not geometry experts.

juliohm:
Related to #348

src/Meshes.jl
stla: As you want. But this is a rotated number:
rfive

juliohm: hahah, good joke. In any case we rotated the coordinates of the associated mesh  and geometries to get that cool rotated 5

juliohm:
Triggering the tests again 👍🏽

src/transforms/rotation.jl
```diff
@@ -0,0 +1,42 @@
...
+end
+
+function revertpoint(::RotateCoords, newpoints, cache)
+  _, M⁻¹ = cache
+  [Point(M⁻¹ * coordinates(p)) for p in newpoints]
+end
+
+function reapplypoint(::RotateCoords, points, cache)
+  M, _ = cache
+  [Point(M * coordinates(p)) for p in points]
```
juliohm:
```suggestion
  R, _ = prep
  newpoints = [Point(R * coordinates(p)) for p in points]
  newpoints, prep
end

function revertpoint(::RotateCoords, newpoints, cache)
  _, R⁻¹ = cache
  [Point(R⁻¹ * coordinates(p)) for p in newpoints]
end

function reapplypoint(::RotateCoords, points, cache)
  R, _ = cache
  [Point(R * coordinates(p)) for p in points]
```

src/transforms/rotation.jl
```diff
@@ -0,0 +1,42 @@
...
+struct RotateCoords{R} <: GeometricTransform
+  rot::R
+end
+
+isrevertible(::Type{<:RotateCoords}) = true
+
+function preprocess(transform::RotateCoords, object)
+  rot = transform.rot
+  R, R⁻¹ = inv(rot), rot
+  convert.(DCM, (R, R⁻¹))
```
juliohm:
```suggestion
  convert.(DCM, (inv(rot), rot))
```

juliohm (approved):

juliohm:
Merged! 🎉 We can review these affine transforms in the future to see if we can find better names and a more smooth experience with active rotations instead of passive rotations as discussed in #344

PR #341, opened 2022-12-16
Torus

stla:
Torus primitive shape.

Not sure I did the things correctly...


stla:
Is it possible to have two constructors inside a `struct`? In order to handle the torus passing through three points, I think we should store the center of the torus and the rotation matrix which transform the "basic" torus to the torus passing through three points. Something like that:

```julia
struct Torus
    minorRadius::Real
    majorRadius::Real
    center::Vector{Real}
    rotmatrix::Matrix{Real}
    function Torus(minorRadius::Real, majorRadius::Real)
        new(minorRadius, majorRadius, [0,0,0], I)
    end
    function Torus(minorRadius::Real, p1::Point3, p2::Point3, p3::Point3)
        ## calculate majorRadius, center, rotmatrix...
        new(minorRadius, majorRadius, center, rotmatrix)
    end
end
```

I want to have the constructors inside the `struct` to prevent the user to give the center and the matrix. 

In this way we could have a function `in` for a general torus.

Does it make sense? Not sure I'm on the right way.

src/primitives/torus.jl
```diff
@@ -0,0 +1,54 @@
+# ------------------------------------------------------------------
+# Licensed under the MIT License. See LICENSE in the project root.
+# ------------------------------------------------------------------
+
+"""
+    Torus(majorRadius, minorRadius)
```
juliohm, line 6: Please follow the code style of the project. We don't use camel case for the variable names.

src/primitives/torus.jl
```diff
@@ -0,0 +1,54 @@
...
+
+A torus with `majorRadius` and `minorRadius` with the z axis as its axis of 
+rotation and the xy plane as its plane of reflection. 
+
```
juliohm, line 10:
```suggestion
```

src/primitives/torus.jl
```diff
@@ -0,0 +1,54 @@
+# ------------------------------------------------------------------
+# Licensed under the MIT License. See LICENSE in the project root.
+# ------------------------------------------------------------------
+
+"""
+    Torus(majorRadius, minorRadius)
+
+A torus with `majorRadius` and `minorRadius` with the z axis as its axis of 
+rotation and the xy plane as its plane of reflection.
```
juliohm, line 9: Worth adding a link to the Wikipedia definition like we do with other geometries.

src/primitives/torus.jl
```diff
@@ -0,0 +1,54 @@
...
+rotation and the xy plane as its plane of reflection. 
+
+"""
+struct Torus{T} <: Primitive{3,T}
+  majorRadius::T
+  minorRadius::T
+end
+
+function Torus(majorRadius::T1, minorRadius::T2) where {T1,T2}
+  T = promote(T1, T2)
```
juliohm, line 18: That is not how you use promote, the correct usage is `r1, r2 = promote(major, minor)`. Are you running these codes before submitting a PR for review?

src/primitives/torus.jl
```diff
@@ -0,0 +1,54 @@
...
+
+majorRadius(t::Torus) = t.majorRadius
+
+# https://en.wikipedia.org/wiki/Torus
+function measure(t::Torus)
+  R, r = t.majorRadius, t.minorRadius
+  2π^2 * R * r^2
+end
+
+Base.length(t::Torus) = measure(t)
```
juliohm, line 38: a torus doesn't have a length
```suggestion
```

src/primitives/torus.jl
```diff
@@ -0,0 +1,54 @@
...
+
+volume(t::Torus) = measure(t)
```
juliohm, line 40: a torus doesn't have a volume
```suggestion
```

src/primitives/torus.jl
```diff
@@ -0,0 +1,54 @@
...
+
+#= to have such a function for a general torus (I have in mind the torus passing 
+  through three points), we need more than the two radii
+function Base.in(p::Point, s::Sphere)
+  x = coordinates(p)
+  c = coordinates(s.center)
+  r = s.radius
+  sum(abs2, x - c) ≈ r^2
+end
+ =#
```
juliohm, line 54:
```suggestion
```

src/sampling/regular.jl
```diff
@@ -45,6 +45,25 @@ function sample(::AbstractRNG, bezier::BezierCurve{Dim,T},
   (bezier(t) for t in trange)
 end
 
+function sample(::AbstractRNG, torus::Torus{T},
+                method::RegularSampling) where {T}
+  sz = fitdims(method.sizes, paramdim(torus))
+  R, r = majorRadius(torus), minorRadius(torus)
+  kxy = R^2 - r^2
+  kz = sqrt(kxy) * r
+  s = sqrt(kxy) / r
```
juliohm, line 54:
```suggestion
  kz = √kxy * r
  s = √kxy / r 
```

src/sampling/regular.jl
```diff
@@ -45,6 +45,25 @@ function sample(::AbstractRNG, bezier::BezierCurve{Dim,T},
...
+  kz = sqrt(kxy) * r
+  s = sqrt(kxy) / r 
+
+  V = T <: AbstractFloat ? T : Float64
+  umin, umax = V(-s * π), V(s * π)
+  vmin, vmax = V(-π), V(π)
+  urange = range(umin, stop=umax, length=sz[1])
+  vrange = range(vmin, stop=vmax, length=sz[2])
+
+  r⃗(u, v) = Vec{3,T}(kxy * sin(u/s), kxy * cos(u/s), kz * sin(v)) / (R - r*cos(v))
```
juliohm, line 62:
```suggestion
  r⃗(u, v) = Point{3,T}(kxy * sin(u/s), kxy * cos(u/s), kz * sin(v)) / (R - r*cos(v))
```

src/transforms.jl
```diff
@@ -94,4 +94,5 @@ reapply(transform::StatelessGeometricTransform, object, cache) =
 # ----------------
 
 include("transforms/stdcoords.jl")
-include("transforms/smoothing.jl")
\ No newline at end of file
+include("transforms/smoothing.jl")
+include("rotations.jl")
```
juliohm, line 98: These changes are not related to the Torus addition. Please try to start a new Git branch before submitting a PR.

juliohm (changes requested):
This PR is mixed with the other changes about rotation. I will close it so that you can start a clean one. Keep in mind that GitHub has a powerful feature that allows you to edit GitHub repos directly by simply pressing "." in the repo page. That can save you some trouble with Git and make it easier to start PRs without mixing branches.

src/primitives/torus.jl
stla: Not when I'm sure of me :-D I should not be so self-confident sometimes.

src/primitives/torus.jl
stla: What is your rationale here? A torus can mean the surface or the volume bounded by this surface.

juliohm: The mathematical definition of a Torus is that of a surface: https://en.wikipedia.org/wiki/Torus

We only provide the alias `volume` for geometries with parametric dimension 3. The torus has parametric dimension 2 as a surface, hence only `area` should be introduced as an alias to `measure`.

stla:
> This PR is mixed with the other changes about rotation. I will close it so that you can start a clean one. Keep in mind that GitHub has a powerful feature that allows you to edit GitHub repos directly by simply pressing "." in the repo page. That can save you some trouble with Git and make it easier to start PRs without mixing branches.

I understand now. That's because I did `git checkout -b torus` when I was in the `rotations` branch.

PR #342, opened 2022-12-17
Circumsphere

stla:
3D sphere passing through four given points


juliohm:
This PR is mixed up with changes from other PRs.

stla:
Really ? This is a different branch. I thought this was ok like this. But no problem, I will wait for the previous pr.

juliohm:
Click on the "Files changed" tab here and you will find the effective changes introduced by the PR. They contain Torus, CircumSphere, Rotation, ...

stla:
I remade this branch and when running the tests, the `LinearAlgebra` function `norm_sqr` is not found. Why that?

juliohm:
The LinearAlgebra stdlib does not have a norm_sqr function, at least not here on Julia v1.8

stla:
Why I have this function? It is here:

https://github.com/JuliaLang/julia/blob/master/stdlib/LinearAlgebra/src/generic.jl#L458

Anyway this is not a good function. I thought it avoided the square root but no, it is defined as `norm(x)^2`.

juliohm:
You are probably using internal functions that are not public with LinearAlgebra.internal_func. That is yet another reason to not use modules as prefixes. A module exports a set of functions as the public api.

PR #345, opened 2022-12-20
Scaling transformation

stla:
New transformation: `Scaling(sx, sy, sz)`. Allows for example to get the mesh of an ellipsoid from the mesh of a sphere.


src/Meshes.jl
```diff
@@ -382,6 +382,7 @@ export
   LambdaMuSmoothing,
   LaplaceSmoothing,
   TaubinSmoothing,
+  Scaling,
```
juliohm: Please rename to ScaleCoords to avoid confusion with TableTransforms.Scale

src/transforms/scaling.jl
```diff
@@ -0,0 +1,58 @@
+# ------------------------------------------------------------------
+# Licensed under the MIT License. See LICENSE in the project root.
+# ------------------------------------------------------------------
+
+"""
+    Scaling(sx, sy, sz)
```
juliohm: Please use a vector of scaling factors instead to avoid a 3D-specific implementation.

src/transforms/scaling.jl
```diff
@@ -0,0 +1,58 @@
...
+  sx::T
+  sy::T
+  sz::T
+end
+
+function Scaling(sx::T1, sy::T2, sz::T3) where {T1<:Real,T2<:Real,T3<:Real}
+  if sx <= 0 || sy <= 0 || sz <= 0
+    error("The scaling factors must be nonnegative.")
+  end
+  T = promote(sx, sy, sz)
```
juliohm: Probably does not compile? Can you please make sure that the code in your PRs is compiling before submitting it for review?

src/transforms/scaling.jl
```diff
@@ -0,0 +1,58 @@
...
+  Scaling(T(sx), T(sy), T(sz))
+end
+
+isrevertible(::Type{<:Scaling}) = true
+
+function preprocess(transform::Scaling, object)
+  [transform.sx, transform.sy, transform.sz]
```
juliohm: Yes, this vector is the dimension-agnostic list of scaling factors that should be a field of the ScaleCoords struct.

test/transforms.jl
```diff
@@ -41,4 +41,25 @@
...
+    # check scaling on a triangle
+    triangle = Triangle(P3(0, 0, 0), P3(1, 0, 0), P3(0, 1, 1))
+    trans = Scaling(T(1), T(2), T(3))
+    striangle = trans(triangle)
+    spoints = vertices(striangle)
+    @test spoints[1] == P3(0, 0, 0)
+    @test spoints[2] == P3(1, 0, 0)
+    @test spoints[3] == P3(0, 2, 3)
+  end
+
```
```suggestion
```

juliohm (changes requested):
Review attached.

src/transforms/scaling.jl
stla: It compiles, I ran the unit tests.

stla:
I'm wondering whether we should add a center for this transformation (here the center is (0,0,0)).

Moreover, there are two kinds of scaling transformations:

- axes-scaling, this one
- scaling given by a center, a direction vector, and a scale factor

juliohm:
> I'm wondering whether we should add a center for this transformation (here the center is (0,0,0)).
> 
> Moreover, there are two kinds of scaling transformations:
> 
> * axes-scaling, this one
> * scaling given by a center, a direction vector, and a scale factor

These are good questions. We should seek the general case, and provide helper outer constructors for the simpler specific cases. I believe that a combination of ScaleCoords, RotateCoords and TranslateCoords can achieve the general case. We already support pipelines so you can write something like:

```julia
# you can get the arrow → by typing \to and pressing TAB
pipe = ScaleCoords(...) → RotateCoords(...) → TranslateCoords(...)

mesh |> pipe
```

stla:
> > I'm wondering whether we should add a center for this transformation (here the center is (0,0,0)).
> > Moreover, there are two kinds of scaling transformations:
> > 
> > * axes-scaling, this one
> > * scaling given by a center, a direction vector, and a scale factor
> 
> These are good questions. We should seek the general case, and provide helper outer constructors for the simpler specific cases. I believe that a combination of ScaleCoords, RotateCoords and TranslateCoords can achieve the general case. We already support pipelines so you can write something like:
> 
> ```julia
> # you can get the arrow → by typing \to and pressing TAB
> pipe = ScaleCoords(...) → RotateCoords(...) → TranslateCoords(...)
> 
> mesh |> pipe
> ```

I don't think so. In 2D, the general affine transformation can be decomposed in a translation, a scaling (coords), a shearing, and a rotation. In 3D I don't know.

src/transforms/scaling.jl
stla: Do you mean a vector for the definition of the `struct`, and we keep the constructor with the three scaling factors? I prefer the three scaling factors because the vector made of the scaling factors has no geometrical interpretation, this is just a list of numbers, not a geometric object.

juliohm: Yes, the struct itself stores a vector of scaling parameters in a way that is dimension-agnostic. Then we can have as many outer constructors as needed for convenience that take list of arguments and convert to the internal vector.

src/transforms/scaling.jl
```diff
@@ -0,0 +1,58 @@
...
+```
+"""
+struct Scaling{T<:Real} <: GeometricTransform
+  sx::T
+  sy::T
+  sz::T
+end
+
+function Scaling(sx::T1, sy::T2, sz::T3) where {T1<:Real,T2<:Real,T3<:Real}
+  if sx <= 0 || sy <= 0 || sz <= 0
```
juliohm: Please use the LaTeX \le instead of <=

src/transforms/scaling.jl
juliohm: You are using promote incorrectly again. I don't know how you are getting the correct results because the promote function returns a tuple of numbers and you are using T as a type. Please, please, please... run these codes locally before submitting the PR. The time here is really scarce to review code that has not been even run by the person who submitted the PR. Keep in mind that this project is used by many people and we don't want serious bugs crashing engineering systems out there.

stla: Weird, you're right. I always run the unit tests before submitting. I don't understand why they didn't crash here!

juliohm: The unit tests will only crash if you add new tests covering the new functionality. Otherwise you will just be testing old code that is already known. Every PR usually comes with a new set of tests. This one here is missing new tests.

stla: Are you joking? ;-) There's a `@testset "Scaling"` in this PR, and I ran it.

juliohm: Ah that is true, then I don't understand how tests could have passed. The code is clearly incorrect treating T as a type when it is a tuple of numbers.

juliohm: I triggered the tests on GitHub again to see what happens.

stla: Is it possible that `Scaling(T(1), T(2), T(3))` has applied the `struct` and not my function? I don't see any other explanation. Not sure I'm clear.

src/transforms/scaling.jl
```diff
@@ -0,0 +1,55 @@
...
+struct ScaleCoords{T<:Real} <: GeometricTransform
+  factors::AbstractVector{T}
+end
+
+function ScaleCoords(sx::T1, sy::T2, sz::T3) where {T1<:Real,T2<:Real,T3<:Real}
+  if sx ≤ 0 || sy ≤ 0 || sz ≤ 0
+    error("The scaling factors must be nonnegative.")
+  end
+  ScaleCoords([sx, sy, sz])
+end
```
juliohm: Try to use Vararg here. You can write:

```julia
ScaleCoords(factors...) = ScaleCoords{eltype(factors)}(factors)
```

Also, try to move the assertion to the inner constructor, that way all constructors will have the assertion.

src/transforms/scaling.jl
```diff
@@ -0,0 +1,55 @@
...
+
+isrevertible(::Type{<:ScaleCoords}) = true
+
+function preprocess(transform::ScaleCoords, object)
+  transform.factors
+end
```
juliohm:
```suggestion
preprocess(transform::ScaleCoords, object) = transform.factors
```

src/transforms/scaling.jl
```diff
@@ -0,0 +1,55 @@
...
+
+function applypoint(::ScaleCoords, points, prep)
+  scale_vector = prep
+  newpoints = [Point(scale_vector .* coordinates(p)) for p in points]
+  newpoints, prep
```
juliohm:
```suggestion
  factors = prep
  newpoints = [Point(factors .* coordinates(p)) for p in points]
  newpoints, prep
```

src/transforms/scaling.jl
```diff
@@ -0,0 +1,55 @@
...
+end
+
+function revertpoint(::ScaleCoords, newpoints, cache)
+  scale_vector = cache
+  iscale_vector = 1 ./ scale_vector
+  [Point(iscale_vector .* coordinates(p)) for p in newpoints]
```
juliohm:
```suggestion
  factors = cache
  [Point((1 ./ factors) .* coordinates(p)) for p in newpoints]
```

src/transforms/scaling.jl
```diff
@@ -0,0 +1,55 @@
...
+end
+
+function reapplypoint(::ScaleCoords, points, cache)
+  scale_vector = cache
+  [Point(scale_vector .* coordinates(p)) for p in points]
```
juliohm:
```suggestion
  factors = cache
  [Point(factors .* coordinates(p)) for p in points]
```

juliohm (changes requested):
Review attached.

src/transforms/scaling.jl
stla: Do you mean `ScaleCoords{eltype(factors)}([factors...])`?

juliohm:
```julia
julia> f(x::Vector) = sum(x)
f (generic function with 1 method)

julia> f(x...) = f(collect(x))
f (generic function with 2 methods)

julia> f(1,2,3)
6
```
You can use collect instead

src/transforms/scaling.jl
```diff
@@ -0,0 +1,54 @@
...
+
+- `sx`: scaling factor for the first coordinate
+- `sy`: scaling factor for the second coordinate
+- `sz`: scaling factor for the third coordinate
+
+## Examples
+
+```julia
+ScaleCoords(1, 2, 3)
+```
```
juliohm: Update docstring to reflect the constructor with Vector besides the constructor with Vararg

src/transforms/scaling.jl
```diff
@@ -0,0 +1,54 @@
...
+"""
+struct ScaleCoords{T<:Real} <: GeometricTransform
+  factors::AbstractVector{T}
```
juliohm:
```suggestion
struct ScaleCoords{T} <: GeometricTransform
  factors::Vector{T}
```
AbstractVector as a field is not good for performance.

src/transforms/scaling.jl
```diff
@@ -0,0 +1,54 @@
...
+  function ScaleCoords{T}(factors::AbstractVector{T}) where {T<:Real}
```
juliohm:
```suggestion
  function ScaleCoords{T}(factors) where {T}
```

src/transforms/scaling.jl
```diff
@@ -0,0 +1,54 @@
...
+    if any(factors .≤ 0)
```
juliohm:
```suggestion
    if any(≤(0), factors)
```

src/transforms/scaling.jl
```diff
@@ -0,0 +1,54 @@
...
+      error("The scaling factors must be nonnegative.")
```
juliohm:
```suggestion
      throw(ArgumentError("Scaling factors must be positive."))
```

src/transforms/scaling.jl
```diff
@@ -0,0 +1,54 @@
...
+    end
+    new(factors)
+  end
+end
+
+ScaleCoords(factors...) = ScaleCoords{eltype(factors)}([factors...])
```
juliohm:
```suggestion
ScaleCoords(factors...) = ScaleCoords{eltype(factors)}(collect(factors))
```

juliohm:
@stla just so that you know, you can press "." here on pull request page to be redirected to a vscode editor on the browser where you can continue editing the PR. I will commit the changes I suggested above and you can benefit from these changes right away by pressing ".". If you don't like the idea of using the web editor, please make sure to pull the changes to your local repository before committing more changes.

test/transforms.jl
```diff
@@ -41,4 +41,24 @@
     @test nelements(smesh) == nelements(mesh)
     @test topology(smesh) == topology(mesh)
   end
+
+  @testset "ScaleCoords" begin
+    # a scaling doesn't change the topology
```
juliohm: Please add a test at the beginning testing that both constructors work, i.e. ScaleCoords([1,2,3]) == ScaleCoords(1,2,3)

juliohm:
This PR is almost ready! 🎉 Looking forward to the next round of reviews.

juliohm:
Related to #348

src/transforms/scaling.jl
```diff
@@ -0,0 +1,53 @@
...
+
+function revertpoint(::ScaleCoords, newpoints, cache)
+  factors = cache
+  [Point((1 ./ factors) .* coordinates(p)) for p in newpoints]
+end
+
+function reapplypoint(::ScaleCoords, points, cache)
+  factors = cache
+  [Point(factors .* coordinates(p)) for p in points]
+end
```
juliohm:
```suggestion
function applypoint(::ScaleCoords, points, prep)
  s = prep
  newpoints = [Point(s .* coordinates(p)) for p in points]
  newpoints, prep
end

function revertpoint(::ScaleCoords, newpoints, cache)
  s = cache
  [Point((1 ./ s) .* coordinates(p)) for p in newpoints]
end

function reapplypoint(::ScaleCoords, points, cache)
  s = cache
  [Point(s .* coordinates(p)) for p in points]
end
```

stla:
> @stla just so that you know, you can press "." here on pull request page to be redirected to a vscode editor on the browser where you can continue editing the PR. I will commit the changes I suggested above and you can benefit from these changes right away by pressing ".". If you don't like the idea of using the web editor, please make sure to pull the changes to your local repository before committing more changes.

How should I pull these changes? I've just synchronized my fork and I pulled the master branch. Should I pull the `scaling` branch?

juliohm:
> How should I pull these changes? I've just synchronized my fork and I pulled the master branch. Should I pull the `scaling` branch?

I think that you should have pulled just the commits in this branch,  but I am not sure. You could rebase with the master branch if there are conflicts, or address the conflicts here directly on the web editor. I started using the web editor more to avoid Gitology.

src/transforms/scaling.jl
```diff
@@ -0,0 +1,53 @@
...
+
+- `sx`: scaling factor for the first coordinate
+- `sy`: scaling factor for the second coordinate
+- `sz`: scaling factor for the third coordinate
+
+## Examples
+
+```julia
+ScaleCoords(1, 2, 3)
+```
```
juliohm:
```suggestion
"""
    ScaleCoords([s1, s2, ...])
    ScaleCoords(s1, s2, ...)

Transform geometry or mesh by scaling coordinates `(x1, x2, ...) ↦ (s1*x1, s2*x2, ...)`. 
```

src/transforms/scaling.jl
```diff
@@ -0,0 +1,41 @@
+# ------------------------------------------------------------------
+# Licensed under the MIT License. See LICENSE in the project root.
+# ------------------------------------------------------------------
+
+"""
+    ScaleCoords([s1, s2, ...])
+    ScaleCoords(s1, s2, ...)
+
+Transform geometry or mesh by scaling coordinates `(x1, x2, ...) ↦ (s1*x1, s2*x2, ...)`.
```
juliohm:
```suggestion
Transform geometry or mesh by scaling coordinates
`(x1, x2, ...) ↦ (s1*x1, s2*x2, ...)`. 
```

test/transforms.jl
```diff
@@ -41,4 +41,24 @@
...
+    trans = ScaleCoords(T(1), T(2), T(3))
```
```suggestion
  @testset "ScaleCoords" begin
    # test constructors
    trans1 = ScaleCoords(T[1,2,3])
    trans2 = ScaleCoords(T(1), T(2), T(3))
    @test trans1 == trans2
    
    # a scaling doesn't change the topology
    trans = ScaleCoords(T(1), T(2), T(3))
```

src/transforms/scaling.jl
```diff
@@ -0,0 +1,42 @@
...
+  factors::Vector{T}
+  function ScaleCoords{T}(factors) where {T}
+    if any(≤(0), factors)
+      throw(ArgumentError("Scaling factors must be positive."))
+    end
+    new(factors)
+  end
+end
+
+ScaleCoords(factors...) = ScaleCoords{eltype(factors)}(collect(factors))
```
juliohm:
```suggestion
ScaleCoords(factors::AbstractVector) =
  ScaleCoords{eltype(factors)}(factors)

ScaleCoords(factors...) =
  ScaleCoords(collect(factors))
```

src/transforms/scaling.jl
```diff
@@ -0,0 +1,46 @@
...
+"""
+    ScaleCoords([s1, s2, ...])
+    ScaleCoords(s1, s2, ...)
+
+Transform geometry or mesh by scaling coordinates
+`(x1, x2, ...) ↦ (s1*x1, s2*x2, ...)`. 
+"""
+struct ScaleCoords{T} <: GeometricTransform
+  factors::Vector{T}
+  function ScaleCoords{T}(factors) where {T}
```
juliohm:
```suggestion
  factors::Vector{T}
  
  function ScaleCoords{T}(factors) where {T}
```

src/transforms/scaling.jl
```diff
@@ -0,0 +1,47 @@
+# ------------------------------------------------------------------
+# Licensed under the MIT License. See LICENSE in the project root.
+# ------------------------------------------------------------------
+
+"""
+    ScaleCoords([s1, s2, ...])
```
juliohm:
```suggestion
```

test/transforms.jl
```diff
@@ -21,28 +21,33 @@
...
-    # smoothing doesn't change the topology
-    trans = LaplaceSmoothing(30)
+  
+  @testset "ScaleCoords" begin
+    # test constructors
+    trans1 = ScaleCoords(T[1,2,3])
+    trans2 = ScaleCoords(T(1), T(2), T(3))
+    @test trans1 == trans2
+    
+    # a scaling doesn't change the topology
```
```suggestion
  
  @testset "ScaleCoords" begin
    # scaling doesn't change the topology
```

juliohm (approved):

PR #346, opened 2022-12-20
3D sphere constructor with 4 points

stla:
Sphere passing through four given points.

### DIFF
--- a/src/primitives/sphere.jl
+++ b/src/primitives/sphere.jl
@@ -52,8 +52,8 @@ function Sphere(p1::Point{3}, p2::Point{3}, p3::Point{3}, p4::Point{3})
   v2 = p2 - p4
   v3 = p3 - p4
   r⃗ = ((v3 ⋅ v3) * (v1 × v2) +
-           (v2 ⋅ v2) * (v3 × v1) +
-           (v1 ⋅ v1) * (v2 × v3)) / 12vol
+        (v2 ⋅ v2) * (v3 × v1) +
+        (v1 ⋅ v1) * (v2 × v3)) / 12vol
   radius = norm(r⃗)
   center = p4 + r⃗
   Sphere(center, radius)
--- a/src/primitives/sphere.jl
+++ b/src/primitives/sphere.jl
@@ -52,7 +52,7 @@ function Sphere(p1::Point{3}, p2::Point{3}, p3::Point{3}, p4::Point{3})
   v2 = p2 - p4
   v3 = p3 - p4
   u = (v3 ⋅ v3) * (v1 × v2) + (v2 ⋅ v2) * (v3 × v1) + (v1 ⋅ v1) * (v2 × v3)
-  r⃗ = u / 12 / vol
+  r⃗ = u / 12vol
   radius = norm(r⃗)
   center = p4 + r⃗
   Sphere(center, radius)
--- a/src/primitives/sphere.jl
+++ b/src/primitives/sphere.jl
@@ -37,6 +37,36 @@ end
 
 Sphere(p1::Tuple, p2::Tuple, p3::Tuple) = Sphere(Point(p1), Point(p2), Point(p3))
 
+"""
+    Sphere(p1, p2, p3, p4)
+
+A 3D sphere passing through points `p1`, `p2`, `p3` and `p4`.
+"""
+function Sphere(p1::Point{3}, p2::Point{3}, p3::Point{3}, p4::Point{3})
+  v1 = coordinates(p1)
+  v2 = coordinates(p2)
+  v3 = coordinates(p3)
+  v4 = coordinates(p4)
+  M = hcat(v1, v2, v3, v4)
+  a = 2 * det(vcat(M, ones((1, 4))))
+  if a == 0
+    error("The four points are coplanar.")
+  end
+  q = [dot(v1, v1), dot(v2, v2), dot(v3, v3), dot(v4, v4)] / a
+  x = M[1, :]
+  y = M[2, :]
+  z = M[3, :]
+  dx = det(hcat(q, y, z, ones(4)))
+  dy = -det(hcat(q, x, z, ones(4)))
+  dz = det(hcat(q, x, y, ones(4)))
+  c = det(hcat(q, x, y, z))
+  radius = sqrt(dx^2 + dy^2 + dz^2 - 2 * c) 
+  center = Point(dx, dy, dz)
+  Sphere(center, radius)
+end
+
+Sphere(p1::Tuple, p2::Tuple, p3::Tuple, p4::Tuple) = Sphere(Point(p1), Point(p2), Point(p3), Point(p4))
+
 paramdim(::Type{<:Sphere{Dim}}) where {Dim} = Dim - 1
 
 isconvex(::Type{<:Sphere}) = false
--- a/src/primitives/sphere.jl
+++ b/src/primitives/sphere.jl
@@ -43,16 +43,16 @@ Sphere(p1::Tuple, p2::Tuple, p3::Tuple) = Sphere(Point(p1), Point(p2), Point(p3)
 A 3D sphere passing through points `p1`, `p2`, `p3` and `p4`.
 """
 function Sphere(p1::Point{3}, p2::Point{3}, p3::Point{3}, p4::Point{3})
+  vol = volume(Tetrahedron(p1, p2, p3, p4))
+  T = typeof(vol)
+  if isapprox(vol, zero(T); atol = atol(T))
+    throw(ArgumentError("The four points are coplanar."))
+  end
   v1 = p1 - p4
   v2 = p2 - p4
   v3 = p3 - p4
-  d = det(hcat(v1, v2, v3))
-  T = typeof(d)
-  if isapprox(d, zero(T); atol = atol(T))
-    error("The four points are coplanar.")
-  end
-  x = (v3 ⋅ v3) * cross(v1, v2) + (v2 ⋅ v2) * cross(v3, v1) + (v1 ⋅ v1) * cross(v2, v3)
-  r⃗ = x / 2 / d
+  u = (v3 ⋅ v3) * (v1 × v2) + (v2 ⋅ v2) * (v3 × v1) + (v1 ⋅ v1) * (v2 × v3)
+  r⃗ = u / 12 / vol
   radius = norm(r⃗)
   center = p4 + r⃗
   Sphere(center, radius)
--- a/src/primitives/sphere.jl
+++ b/src/primitives/sphere.jl
@@ -43,26 +43,18 @@ Sphere(p1::Tuple, p2::Tuple, p3::Tuple) = Sphere(Point(p1), Point(p2), Point(p3)
 A 3D sphere passing through points `p1`, `p2`, `p3` and `p4`.
 """
 function Sphere(p1::Point{3}, p2::Point{3}, p3::Point{3}, p4::Point{3})
-  x1 = coordinates(p1)
-  x2 = coordinates(p2)
-  x3 = coordinates(p3)
-  x4 = coordinates(p4)
-  X = hcat(x1, x2, x3, x4)
-  a = 2 * det(vcat(X, ones((1, 4))))
-  T = typeof(a)
-  if isapprox(a, zero(T); atol = atol(T))
+  v1 = p1 - p4
+  v2 = p2 - p4
+  v3 = p3 - p4
+  d = det(hcat(v1, v2, v3))
+  T = typeof(d)
+  if isapprox(d, zero(T); atol = atol(T))
     error("The four points are coplanar.")
   end
-  q = [x1 ⋅ x1, x2 ⋅ x2, x3 ⋅ x3, x4 ⋅ x4] / a
-  y1 = X[1, :]
-  y2 = X[2, :]
-  y3 = X[3, :]
-  d1 = det(hcat(q, y2, y3, ones(4)))
-  d2 = -det(hcat(q, y1, y3, ones(4)))
-  d3 = det(hcat(q, y1, y2, ones(4)))
-  c = det(hcat(q, y1, y2, y3))
-  radius = sqrt(d1^2 + d2^2 + d3^2 - 2 * c) 
-  center = Point(d1, d2, d3)
+  x = (v3 ⋅ v3) * cross(v1, v2) + (v2 ⋅ v2) * cross(v3, v1) + (v1 ⋅ v1) * cross(v2, v3)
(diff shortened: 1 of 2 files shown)


src/primitives/sphere.jl
```diff
@@ -37,6 +37,36 @@ end
...
+"""
+    Sphere(p1, p2, p3, p4)
+
+A 3D sphere passing through points `p1`, `p2`, `p3` and `p4`.
+"""
+function Sphere(p1::Point{3}, p2::Point{3}, p3::Point{3}, p4::Point{3})
+  v1 = coordinates(p1)
+  v2 = coordinates(p2)
+  v3 = coordinates(p3)
+  v4 = coordinates(p4)
```
juliohm: Let's rename this to x1, x2, x3, x4. We use v1, v2, v3, v4 notation to mean vectors throughout the codebase.

src/primitives/sphere.jl
```diff
@@ -37,6 +37,36 @@ end
...
+  M = hcat(v1, v2, v3, v4)
```
juliohm: The matrix of x1,x2,x3, ... can then be denoted by X

src/primitives/sphere.jl
```diff
@@ -37,6 +37,36 @@ end
...
+  a = 2 * det(vcat(M, ones((1, 4))))
+  if a == 0
```
juliohm: Equality with floating point is not ideal. Maybe you meant isapprox(a, zero(T), atol=atol(T))

src/primitives/sphere.jl
```diff
@@ -37,6 +37,36 @@ end
...
+    error("The four points are coplanar.")
+  end
+  q = [dot(v1, v1), dot(v2, v2), dot(v3, v3), dot(v4, v4)] / a
```
juliohm: The dot product can be written with LaTex by writting \dot and pressing tab

src/primitives/sphere.jl
```diff
@@ -37,6 +37,36 @@ end
...
+  x = M[1, :]
+  y = M[2, :]
+  z = M[3, :]
+  dx = det(hcat(q, y, z, ones(4)))
+  dy = -det(hcat(q, x, z, ones(4)))
+  dz = det(hcat(q, x, y, ones(4)))
```
juliohm: Please adjust the variable names after fixing the other issues above.

src/primitives/sphere.jl
```diff
@@ -37,6 +37,36 @@ end
...
+  c = det(hcat(q, x, y, z))
+  radius = sqrt(dx^2 + dy^2 + dz^2 - 2 * c) 
+  center = Point(dx, dy, dz)
+  Sphere(center, radius)
+end
+
+Sphere(p1::Tuple, p2::Tuple, p3::Tuple, p4::Tuple) = Sphere(Point(p1), Point(p2), Point(p3), Point(p4))
```
juliohm:
```suggestion
Sphere(p1::Tuple, p2::Tuple, p3::Tuple, p4::Tuple) =
  Sphere(Point(p1), Point(p2), Point(p3), Point(p4))
```

juliohm (changes requested):
Awesome PR @stla ❤️ Let's work on it a bit more as it is almost ready to get merged.

src/primitives/sphere.jl
stla: What should I take for `T` here?
And are you sure `isapprox(x, 0)` is reliable? I vaguely remember there is a problem when comparing with zero, but I don't remember, and I'm not sure.

juliohm: T is the coordtype(p1) for example. It is usually Float64 or Float32 in our tests.

isapprox is not ideal either, but better than == at least. We will consider exact coordinates at some point in the future as well, but for now we are using floating point.

src/primitives/sphere.jl
stla: I don't understand. This makes a dot above a symbol, like an accented letter with a dot for the accent.

src/primitives/sphere.jl
stla: Ah yes, `atol` is absolute tolerance, this should be nice.

src/primitives/sphere.jl
juliohm: Try x \dot y (with whitespaces)

stla: Still get this strange dot.

src/primitives/sphere.jl
stla: What if `coordtype(p2)` is different from `coordtype(p1)`? Isn't it better to take `T = typeof(a)`?

src/primitives/sphere.jl
juliohm: Sorry, my mistake, I meant \cdot for the linear algebra dot product in LaTeX

stla: It's `\cdot`, no?

src/primitives/sphere.jl
juliohm: That sounds good too

src/primitives/sphere.jl
```diff
@@ -37,6 +37,38 @@ end
...
+A 3D sphere passing through points `p1`, `p2`, `p3` and `p4`.
+"""
+function Sphere(p1::Point{3}, p2::Point{3}, p3::Point{3}, p4::Point{3})
+  x1 = coordinates(p1)
+  x2 = coordinates(p2)
+  x3 = coordinates(p3)
+  x4 = coordinates(p4)
+  X = hcat(x1, x2, x3, x4)
+  a = 2 * det(vcat(X, ones((1, 4))))
+  T = typeof(a)
```
juliohm: You are trying to compute the volume of the prism right? We prefer to be explicit in our code, so try to rewrite this as volume(Tetrahedron(p1, p2, p3, p4))

src/primitives/sphere.jl
```diff
@@ -37,6 +37,38 @@ end
...
+    error("The four points are coplanar.")
+  end
+  q = [x1 ⋅ x1, x2 ⋅ x2, x3 ⋅ x3, x4 ⋅ x4] / a
+  y1 = X[1, :]
+  y2 = X[2, :]
+  y3 = X[3, :]
+  d1 = det(hcat(q, y2, y3, ones(4)))
+  d2 = -det(hcat(q, y1, y3, ones(4)))
+  d3 = det(hcat(q, y1, y2, ones(4)))
+  c = det(hcat(q, y1, y2, y3))
```
juliohm: All these area/volume calculations can be expressed with explicit geometries like area(Triangle(...)). Please try to refactor the code to simplify it and to facilitate future maintenance.

src/primitives/sphere.jl
stla: Clearly this determinant is zero if and only if the the volume of the tetrahedron is zero, but I don't know whether these are the same things.

src/primitives/sphere.jl
stla: Hmm.. I don't know. I think I borrowed this code a long time ago from Asymptote. I don't know what these determinants represent.

src/primitives/sphere.jl
juliohm: From a quick look, the volume you are computing here is an integer multiple of the volume of the tetrahedron. What I am saying is that it is useful to write algorithms in terms of high-level functions to increase coverage and to avoid fixing a bug in an isolated place.

src/primitives/sphere.jl
stla: Ah no i found this code on the [Wolfram site](https://mathworld.wolfram.com/Circumsphere.html).

juliohm: @stla be aware that you cannot copy/paste code from other projects without considering their license terms. If this code comes from Asymptote, it is GPL code, and we can't merge as is.

stla: There's a [formula here](http://rodolphe-vaillant.fr/entry/127/find-a-tetrahedron-circumcenter) which looks easier.

src/primitives/sphere.jl
```diff
@@ -37,6 +37,30 @@ end
...
+"""
+    Sphere(p1, p2, p3, p4)
+
+A 3D sphere passing through points `p1`, `p2`, `p3` and `p4`.
+"""
+function Sphere(p1::Point{3}, p2::Point{3}, p3::Point{3}, p4::Point{3})
+  v1 = p1 - p4
+  v2 = p2 - p4
+  v3 = p3 - p4
+  d = det(hcat(v1, v2, v3))
```
juliohm: This line represents the volume of a geometry? Can you write it in terms of volume so that more code gets reused? Also, probably a good idea to rename the result to vol

src/primitives/sphere.jl
```diff
@@ -37,6 +37,30 @@ end
...
+  T = typeof(d)
+  if isapprox(d, zero(T); atol = atol(T))
+    error("The four points are coplanar.")
```
juliohm:
```suggestion
    throw(ArgumentError("four points are coplanar"))
```
Please follow previous review suggestions. We try to throw specific errors usually.

src/primitives/sphere.jl
```diff
@@ -37,6 +37,30 @@ end
...
+  end
+  x = (v3 ⋅ v3) * cross(v1, v2) + (v2 ⋅ v2) * cross(v3, v1) + (v1 ⋅ v1) * cross(v2, v3)
```
juliohm: You can use \times for the cross product. Also the resulting variable can have a different name. We use x to represent coordinates in our codebase

juliohm (changes requested):
Review attached

juliohm:
This one is also almost ready! A small push and you can get two PRs merged soon :)

src/primitives/sphere.jl
stla: I looked at Wikipedia for the volume of a tetrahedron and I didn't see that this determinant is a volume.

stla: Ah no you're right. This determinant is `1/6` of the volume.

juliohm: https://mathinsight.org/relationship_determinants_area_volume

juliohm: Yes, you are basically checking if the four points lie on the same plane. This is equivalent to checking if the corresponding tetrahedron has zero volume. So the entire code can be reduced to something like:

```julia
if volume(Tetrahedron(p1, p2, p3, p4)) > 0
  # do something
else
  # throw error
end
```

src/primitives/sphere.jl
```diff
@@ -37,6 +37,30 @@ end
...
+function Sphere(p1::Point{3}, p2::Point{3}, p3::Point{3}, p4::Point{3})
+  vol = volume(Tetrahedron(p1, p2, p3, p4))
+  T = typeof(vol)
+  if isapprox(vol, zero(T); atol = atol(T))
+    throw(ArgumentError("The four points are coplanar."))
+  end
+  v1 = p1 - p4
+  v2 = p2 - p4
+  v3 = p3 - p4
+  u = (v3 ⋅ v3) * (v1 × v2) + (v2 ⋅ v2) * (v3 × v1) + (v1 ⋅ v1) * (v2 × v3)
```
juliohm: We can probably express this big expression in terms of volumes? What is the intuition behind this set of cross products? Perhaps there is a determinant hidden here too?

juliohm: Try to think in terms of geometric concepts. This linear algebra stuff becomes an implementation detail. If we can write code in terms of geometries it will be much easier to generalize it in the future.

stla: No idea.

juliohm: It is taking the planes v1xv2, v3xv1 and v2xv3 and scaling by the other vectors. If there was a minus sign somewhere we could try to map to determinant, but at a first glance it is not something straightforward.

src/primitives/sphere.jl
```diff
@@ -37,6 +37,30 @@ end
...
+  r⃗ = u / 12 / vol
```
juliohm:
```suggestion
  r⃗ = u / 12vol
```

src/primitives/sphere.jl
```diff
@@ -37,6 +37,30 @@ end
...
+  vol = volume(Tetrahedron(p1, p2, p3, p4))
+  T = typeof(vol)
+  if isapprox(vol, zero(T); atol = atol(T))
+    throw(ArgumentError("The four points are coplanar."))
+  end
+  v1 = p1 - p4
+  v2 = p2 - p4
+  v3 = p3 - p4
+  u = (v3 ⋅ v3) * (v1 × v2) + (v2 ⋅ v2) * (v3 × v1) + (v1 ⋅ v1) * (v2 × v3)
+  r⃗ = u / 12vol
```
juliohm:
```suggestion
  r⃗ = ((v3 ⋅ v3) * (v1 × v2) +
           (v2 ⋅ v2) * (v3 × v1) +
           (v1 ⋅ v1) * (v2 × v3)) / 12vol
```

src/primitives/sphere.jl
```diff
@@ -37,6 +37,31 @@ end
...
+  T = typeof(vol)
+  if isapprox(vol, zero(T); atol = atol(T))
+    throw(ArgumentError("The four points are coplanar."))
+  end
+  v1 = p1 - p4
+  v2 = p2 - p4
+  v3 = p3 - p4
+  r⃗ = ((v3 ⋅ v3) * (v1 × v2) +
+           (v2 ⋅ v2) * (v3 × v1) +
+           (v1 ⋅ v1) * (v2 × v3)) / 12vol
```
juliohm:
```suggestion
  r⃗ = ((v3 ⋅ v3) * (v1 × v2) +
        (v2 ⋅ v2) * (v3 × v1) +
        (v1 ⋅ v1) * (v2 × v3)) / 12vol
```

src/primitives/sphere.jl
```diff
@@ -37,6 +37,31 @@ end
...
+    throw(ArgumentError("The four points are coplanar."))
+  end
+  v1 = p1 - p4
+  v2 = p2 - p4
+  v3 = p3 - p4
+  r⃗ = ((v3 ⋅ v3) * (v1 × v2) +
+        (v2 ⋅ v2) * (v3 × v1) +
+        (v1 ⋅ v1) * (v2 × v3)) / 12vol
+  radius = norm(r⃗)
+  center = p4 + r⃗
```
juliohm:
```suggestion
  center = p4 + r⃗
  radius = norm(r⃗)
```

juliohm (approved):

juliohm:
Another nice addition 👏🏽

PR #350, opened 2022-12-23
Refactor Line-Plane to work for Segment, Rays too

jackbeagley:
Some of this is new territory for me, the code is quite awkward and I think some of these helper functions would be better written as macros, however I can't get my head around how to do that.. any suggestions would be appreciated! @juliohm 
Happy with overall layout and logic though.


src/intersections/lineraysegmentplane.jl
```diff
@@ -0,0 +1,111 @@
...
+  return x(0)
+end
+
+function lineraysegmentorigin(x::Ray)
+  return origin(x)
+end
+
+function lineraysegmentorigin(x::Segment)
+  return vertices(x)[1]
+end
```
juliohm: Can these three methods be replaced by the generic x(0) that works for all three types?

src/intersections/lineraysegmentplane.jl
```diff
@@ -0,0 +1,111 @@
...
+  return x(1) - x(0)
+end
+
+function lineraysegmentdirection(x::Ray)
+  return direction(x)
+end
+
+function lineraysegmentdirection(x::Segment)
+  return -reduce(-, coordinates.(vertices(x)))
+end
```
juliohm: Same here, can all these three methods be replaced by the single generic x(1) - x(0)?

src/intersections/lineraysegmentplane.jl
```diff
@@ -0,0 +1,111 @@
...
+  return OverlappingSegmentPlane
+end
+
+function lineraysegmenttouching(x::Ray)
+  return TouchingRayPlane
+end
+
+function lineraysegmenttouching(x::Segment)
+  return TouchingSegmentPlane
+end
```
juliohm: Yes, these type factory can certainly be replaced by a macro.

src/intersections/lineraysegmentplane.jl
jackbeagley: Totally can be replaced with an x(1) - x(0), but I think it introduces unnecessary operations? Ray already contains the direction information and for segment subtracting requires less operations than calculating `a + t * (b - a)` for t = 0 and t = 1. In my application (and I'm sure others too) the runtime is already quite long and it seems to me making these lower level functions simpler will help :)

juliohm: Isn't this a case of premature optimization? If we just make sure that the code is general and works for the three types first, we can come later to benchmark and spot the bottlenecks. My guess is that this tiny detail won't be the bottleneck of most applications. Let's make the code work first before we start with micro-optimizations 👍🏽

jackbeagley:
@juliohm I have made changes for direction and origin as you suggested and refactored the body of the code too. I tried using macros and had some success, however as the type of `g` is unknown at parse time I don't think they are the right tool. Instead I took out as much logic as possible and put it into type specific functions that process the parameter lambda. Let me know what you think! Also happy new years!

src/intersections/lineraysegmentplane.jl
```diff
@@ -0,0 +1,114 @@
...
+function Meshes.intersection(f, g::G, p::Plane{T}) where {T, G<:Union{Line{3,T}, Ray{3,T}, Segment{3,T}}}
+  p₀ = coordinates(origin(p))
+  n  = normal(p)
+
+  # Get the origin and direction of geometry g
+  g₀ = g(0)
+  gdir = g(1) - g(0)
+  
+  # calculate components
+  ln = gdir ⋅ n
```
juliohm:
```suggestion
  p = coordinates(origin(p))
  n = normal(p)
  o = g(0)
  d = g(1) - g(0)
  a = d ⋅ n
```

src/intersections/lineraysegmentplane.jl
```diff
@@ -0,0 +1,114 @@
...
+  return OverlappingLinePlane
+end
+
+function planeintersectionoverlapping(::Ray)
+  return OverlappingRayPlane
+end
+
+function planeintersectionoverlapping(::Segment)
+  return OverlappingSegmentPlane
+end
```
juliohm:
```suggestion
_overlapping(::Line) = OverlappingLinePlane
_overlapping(::Ray) = OverlappingRayPlane
_overlapping(::Segment) = OverlappingSegmentPlane
```

src/intersections/lineraysegmentplane.jl
```diff
@@ -0,0 +1,114 @@
...
+
+"""
+    checkparameterplaneintersection(f, l::Line, λ)
+
+Return the intersection for a `Plane` and non-parallel `Line`, `l` given a line parameter `λ`. As a line is infinite, a `CrossingLinePlane` will always be returned.
+"""
+function checkparameterplaneintersection(f, l::Line, λ)
```
juliohm: Let's rename this internal function to something short like `_intersection` and replace the docstrings by simple comments for maintainers. Docstrings aren't very useful here given that this is an internal function.

juliohm (changes requested):
@jackbeagley it is looking much nicer now! 💯 

Can you refactor the code a bit more and convert the PR to a normal PR to run the test suite? We are getting closer to a merge state.

src/intersections/lineraysegmentplane.jl
```diff
@@ -0,0 +1,96 @@
...
+function intersection(f, g::G, p::Plane{T}) where {T, G<:Union{Line{3,T}, Ray{3,T}, Segment{3,T}}}
+  p₀ = coordinates(origin(p))
+  n  = normal(p)
+
+  # get the origin and direction of geometry g
+  g₀ = g(0)
+  d = g(1) - g(0)
+  
+  # evaluate denominator
+  a = d ⋅ n
```
jackbeagley: @juliohm I have done a few of the variable name changes as you suggested, but left in some comments and kept some of the variable names as is - it at least helps me read and understand it and I feel it is terse enough.

src/intersections/lineraysegmentplane.jl
jackbeagley: This is much tidier, I should have thought of this much quicker, thanks!

src/intersections/lineraysegmentplane.jl
juliohm: That is lovely @jackbeagley , thanks for the improvement.

src/intersections/lineraysegmentplane.jl
```diff
@@ -0,0 +1,96 @@
...
+  else
+    return @IT CrossingSegmentPlane s(λ) f
+  end
+end
+
+#=
+(https://en.wikipedia.org/wiki/Line-plane_intersection)
+=#
+function intersection(f, g::G, p::Plane{T}) where {T, G<:Union{Line{3,T}, Ray{3,T}, Segment{3,T}}}
+  p₀ = coordinates(origin(p))
```
juliohm: It seems that the algorithm only uses p0 - g0, so we don't need to take the coordinates early here.
```suggestion
  p₀ = origin(p)
```

src/intersections/lineraysegmentplane.jl
```diff
@@ -0,0 +1,96 @@
...
+  if isapprox(a, zero(T), atol=atol(T))
+    # if the numerator is zero, then g is overlapping
+    if isapprox(coordinates(p₀ - g₀) ⋅ n, zero(T), atol=atol(T))
+      return @IT _overlapping(g) g f
+    else
+      return @IT NoIntersection nothing f
+    end
+  else
+    # calculate the length parameter
+    λ = -(n ⋅ coordinates(g₀ - p₀)) / a
```
juliohm:
```suggestion
    if isapprox((p₀ - g₀) ⋅ n, zero(T), atol=atol(T))
      return @IT _overlapping(g) g f
    else
      return @IT NoIntersection nothing f
    end
  else
    # calculate the length parameter
    λ = -(n ⋅ (g₀ - p₀)) / a
```
Now that both p0 and g0 are points, we know that subtraction produces a vector. No need to use coordinates.

src/intersections.jl
```diff
@@ -148,7 +157,8 @@ include("intersections/raysegment.jl")
 include("intersections/rayline.jl")
 include("intersections/linesegment.jl")
 include("intersections/raybox.jl")
-include("intersections/segmentplane.jl")
+# include("intersections/segmentplane.jl")
```
```suggestion
```

src/intersections/lineraysegmentplane.jl
```diff
@@ -0,0 +1,96 @@
+# ------------------------------------------------------------------
+# Licensed under the MIT License. See LICENSE in the project root.
+# ------------------------------------------------------------------
+
+#=
+Return appropriate type for a geometry overlapping with a `Plane`.
+Ideally this would be a macro, but the geometry type isn't known at parse time, so it is implemented via multiple-dispatch instead
```
juliohm:
```suggestion
Ideally this would be a macro, but the geometry type isn't known at parse time,
so it is implemented via multiple-dispatch instead
```

src/intersections/lineraysegmentplane.jl
```diff
@@ -0,0 +1,96 @@
...
+    return @IT NoIntersection nothing f
+  else
+    return @IT CrossingSegmentPlane s(λ) f
+  end
+end
+
+#=
+(https://en.wikipedia.org/wiki/Line-plane_intersection)
+=#
+function intersection(f, g::G, p::Plane{T}) where {T, G<:Union{Line{3,T}, Ray{3,T}, Segment{3,T}}}
```
juliohm:
```suggestion
const LineLike{T} = Union{Line{3,T}, Ray{3,T}, Segment{3,T}}
function intersection(f, g::LineLike{T}, p::Plane{T}) where {T}
```

test/intersections.jl
```diff
@@ -820,6 +865,46 @@
...
+    @test intersection(r, p) |> type == NoIntersection
+    @test isnothing(r ∩ p)
+
+    # plane as first argument
+    p = Plane(P3(0, 0, 1), V3(1, 0, 0), V3(0, 1, 0))
+    r = Ray(P3(0, 0, 0), V3(0, 2, 2))
+    @test intersection(p, r) |> type == CrossingRayPlane
+    @test r ∩ p == p ∩ r == P3(0, 1, 1)
+  end
+
```
juliohm: Can we move these RayPlanes section to after LinePlanes for consistency? It will follow the same Line-Ray-Segment order in the source code.

jackbeagley: Done!

src/intersections/lineraysegmentplane.jl
jackbeagley: I have committed your suggestions as is, it's much tidier now, cheers! 🍻

src/intersections/lineraysegmentplane.jl
```diff
@@ -0,0 +1,98 @@
...
+  if isapprox(a, zero(T), atol=atol(T))
+    # if the numerator is zero, then g is overlapping
+    if isapprox((p₀ - g₀) ⋅ n, zero(T), atol=atol(T))
+      return @IT _overlapping(g) g f
+    else
+      return @IT NoIntersection nothing f
+    end
+  else
+    # calculate the length parameter
+    λ = -(n ⋅ (g₀ - p₀)) / a
```
juliohm:
```suggestion
    λ = ((p₀ - g₀) ⋅ n) / a
```
Can we flip the order of arguments here to match the other expression above in line 87? Can these two calculations be extracted out before the first if-else block?

src/intersections/lineraysegmentplane.jl
```diff
@@ -0,0 +1,98 @@
...
+(https://en.wikipedia.org/wiki/Line-plane_intersection)
+=#
+const LineLike{T} = Union{Line{3,T}, Ray{3,T}, Segment{3,T}}
+function intersection(f, g::LineLike{T}, p::Plane{T}) where {T}
+  p₀ = origin(p)
+  n  = normal(p)
+
+  # get the origin and direction of geometry g
+  g₀ = g(0)
+  d = g(1) - g(0)
```
juliohm:
```suggestion

  # origin and direction of geometry
  g₀ = g(0)
  d  = g(1) - g(0)
```

src/intersections/lineraysegmentplane.jl
```diff
@@ -0,0 +1,98 @@
...
+  end
+end
+
+#=
+(https://en.wikipedia.org/wiki/Line-plane_intersection)
+=#
+const LineLike{T} = Union{Line{3,T}, Ray{3,T}, Segment{3,T}}
+function intersection(f, g::LineLike{T}, p::Plane{T}) where {T}
+  p₀ = origin(p)
+  n  = normal(p)
```
juliohm, line 81:
```suggestion
 # origin and normal of plane
  p₀ = origin(p)
  n  = normal(p)
```

src/intersections/lineraysegmentplane.jl
```diff
@@ -0,0 +1,99 @@
...
+=#
+const LineLike{T} = Union{Line{3,T}, Ray{3,T}, Segment{3,T}}
+function intersection(f, g::LineLike{T}, p::Plane{T}) where {T}
+ # origin and normal of plane
+  p₀ = origin(p)
+  n  = normal(p)
+
+  # origin and direction of geometry
+  g₀ = g(0)
+  d  = g(1) - g(0)
```
juliohm:
```suggestion
  # origin and direction of geometry
  g₀ = g(0)
  d  = g(1) - g(0)
  
 # origin and normal of plane
  p₀ = origin(p)
  n  = normal(p)
```

src/intersections/lineraysegmentplane.jl
```diff
@@ -0,0 +1,99 @@
...
+end
+
+#=
+(https://en.wikipedia.org/wiki/Line-plane_intersection)
+=#
+const LineLike{T} = Union{Line{3,T}, Ray{3,T}, Segment{3,T}}
+function intersection(f, g::LineLike{T}, p::Plane{T}) where {T}
+  # origin and direction of geometry
+  g₀ = g(0)
+  d  = g(1) - g(0)
```
juliohm:
```suggestion
function intersection(f, l::LineLike{T}, p::Plane{T}) where {T}
  # origin and direction of line
  l₀ = l(0)
  d  = l(1) - l(0)
```

src/intersections/lineraysegmentplane.jl
juliohm: I believe that we can introduce two constants before the first if-else block:

```julia
a = (p₀ - g₀) ⋅ n
b = d ⋅ n
```

and then use these constants in the following lines for clarity.

jackbeagley: This is a nice suggestion. Done!

src/intersections/lineraysegmentplane.jl
```diff
@@ -0,0 +1,100 @@
...
+function intersection(f, l::LineLike{T}, p::Plane{T}) where {T}
+  # origin and direction of line
+  l₀ = l(0)
+  d  = l(1) - l(0)
+  
+  # origin and normal of plane
+  p₀ = origin(p)
+  n  = normal(p)
+  
+  # evaluate numerator and denominator seperately
```
juliohm:
```suggestion
  # evaluate numerator and denominator separately
```

src/intersections/lineraysegmentplane.jl
```diff
@@ -0,0 +1,100 @@
...
+  a = (p₀ - l₀) ⋅ n
+  b = d ⋅ n
+  
+  # if a is zero, then g is parallel to p
```
juliohm:
```suggestion
  # if denominator is zero, then g is parallel to p
```

src/intersections/lineraysegmentplane.jl
```diff
@@ -0,0 +1,100 @@
...
+    if isapprox(a, zero(T), atol=atol(T))
+      return @IT _overlapping(l) l f
+    else
+      return @IT NoIntersection nothing f
+    end
+  else
+    # calculate the length parameter
+    λ = a / b
+
+    return _intersection(f, l, λ)
```
juliohm:
```suggestion
    # calculate the length parameter
    return _intersection(f, l, a / b)
```

src/intersections/lineraysegmentplane.jl
```diff
@@ -0,0 +1,98 @@
...
+  
+  # origin and normal of plane
+  p₀ = origin(p)
+  n  = normal(p)
+  
+  # evaluate numerator and denominator separately
+  a = (p₀ - l₀) ⋅ n
+  b = d ⋅ n
+  
+  # if denominator is zero, then g is parallel to p
```
juliohm:
```suggestion
```

src/intersections/lineraysegmentplane.jl
```diff
@@ -0,0 +1,98 @@
...
+  if isapprox(b, zero(T), atol=atol(T))
+    # if the numerator is zero, then g is overlapping
```
juliohm:
```suggestion
```

src/intersections/lineraysegmentplane.jl
```diff
@@ -0,0 +1,98 @@
...
+    if isapprox(a, zero(T), atol=atol(T))
+      return @IT _overlapping(l) l f
+    else
+      return @IT NoIntersection nothing f
+    end
+  else
+    # calculate the length parameter
```
juliohm:
```suggestion
```

src/intersections/lineraysegmentplane.jl
```diff
@@ -0,0 +1,95 @@
...
+function intersection(f, l::LineLike{T}, p::Plane{T}) where {T}
+  # origin and direction of line
+  l₀ = l(0)
+  d  = l(1) - l(0)
+  
+  # origin and normal of plane
+  p₀ = origin(p)
+  n  = normal(p)
+  
+  # evaluate numerator and denominator separately
```
juliohm:
```suggestion
  # auxiliary parameters
```

src/intersections/lineraysegmentplane.jl
```diff
@@ -0,0 +1,95 @@
...
+    return @IT NoIntersection nothing f
+  else
+    return @IT CrossingSegmentPlane s(λ) f
+  end
+end
+
+#=
+(https://en.wikipedia.org/wiki/Line-plane_intersection)
+=#
+const LineLike{T} = Union{Line{3,T}, Ray{3,T}, Segment{3,T}}
```
juliohm:
```suggestion
```

src/intersections/lineraysegmentplane.jl
```diff
@@ -0,0 +1,95 @@
+# ------------------------------------------------------------------
+# Licensed under the MIT License. See LICENSE in the project root.
+# ------------------------------------------------------------------
+
```
juliohm, line 4:
```suggestion

const LineLike{T} = Union{Line{3,T}, Ray{3,T}, Segment{3,T}}

```

juliohm (approved):

juliohm:
Thanks @jackbeagley for the awesome addition! 💯

PR #352, opened 2022-12-25
Triangle subdivision

stla:
Is it normal that we see the hidden edges with CairoMakie?

ptqrefine-Float64


juliohm:
CairoMakie can only fake 3D. I personally never understood the rationale
behind this backend. It kind of goes against the whole Makie design as an
interactive 3D visualization system.

Em sáb., 24 de dez. de 2022 22:59, stla ***@***.***> escreveu:

src/Meshes.jl
```diff
@@ -365,6 +365,7 @@ export
   TriRefinement,
   QuadRefinement,
   CatmullClark,
+  PTQ,
```
juliohm: We prefer more descriptive and explicit names in Julia. Is there a standard name for this method? Ideally we retain the name of the original paper that introduced it.

src/refinement/ptq.jl
```diff
@@ -0,0 +1,61 @@
+# ------------------------------------------------------------------
+# Licensed under the MIT License. See LICENSE in the project root.
+# ------------------------------------------------------------------
+
+"""
+    PTQ()
+
+Refinement of triangle meshes.
+A triangle is subdivided into four triangles.
```
juliohm: Please try to follow the docstring convention of other methods. In particular, it is very helpful when we provide references in the literature. Is there a published paper discussing/using this method?

src/refinement/ptq.jl
```diff
@@ -0,0 +1,61 @@
...
+"""
+struct PTQ <: RefinementMethod end
+
+function refine(mesh, ::PTQ)
+  primal_triangle_quadrisection(mesh, (p1, p2) -> Point((coordinates(p1) + coordinates(p2))/2))
```
juliohm: Are there plans to support other types of "middle" function?  I think this function primal_triangle_quadrisection could be the refine function. No need to introduce a separate function.

juliohm (changes requested):
Review attached.

src/refinement/ptq.jl
```diff
@@ -0,0 +1,61 @@
...
+end
+
+function ordered_pair(i ,j)
+  i < j ? (i, j) : (j, i)
+end
+
+function primal_triangle_quadrisection(mesh, middle)
+  if !(eltype(mesh) <: Triangle)
+    throw(ArgumentError("The mesh is not triangle."))
```
juliohm: Can we generalize this refinement method to support arbitrary polygonal meshes?  I think we can call it EarRefinement, an algorihtm that clips ears of the polygon as triangles and then triangulates the interior polygon with FanTriangulation.

src/Meshes.jl
stla: It is called PTQ in several papers, and in CGAL.

src/refinement/ptq.jl
stla: Yes, the hyperbolic midpoint to construct hyperbolic triangles. Also if you take three points on the unit sphere and you iterate the PTQ with the normalized midpoint, you get a spherical triangle.

src/refinement/ptq.jl
stla: By calling `triangulate()` at the beginning?

src/Meshes.jl
stla: And this a subdivision algorithm of the so-called family of **loop** subdivision algorithms.

src/refinement/ptq.jl
stla: The CGAL documentation refers to a master thesis:

Charles Loop. *Smooth subdivision surfaces based on triangles. Master's thesis, University of Utah, 1987.*

Lol, I thought that the world "loop" was refering to a for-loop...

src/Meshes.jl
juliohm: > It is called PTQ in several papers, and in CGAL.

The R language calls many things, ABC, DET, PTYS, ..., but this is not necessarily a good practice. It makes things unsearchable and the Julia community is working really hard to avoid unnecessary acronyms everywhere. We should rename it to a more descriptive searchable name. Charles Loop's thesis describes a couple of algorithms in chapter 3, which one is the one in this PR? https://charlesloop.com/thesis.pdf

src/refinement/ptq.jl
juliohm: Awesome, we should definitely add this his thesis to the docstring.

src/refinement/ptq.jl
juliohm: We can refactor our refinement methods later to add this flexibility and avoid code duplication. As far as I remember our CatmullClark method is the same as TriRefinement with a custom "middle" function as well.

src/refinement/ptq.jl
juliohm: That would be an interesting alternative as well.

src/Meshes.jl
stla: I never talked about R here. I didn't see any name in the thesis. I first found the name *Primal Triangle Quadrisection* in the CGAL documentation, and googling this name yields some results.

stla: Maybe `TriQuad`? For "triangle quadrisection"?

juliohm: > I never talked about R here.

I know :) Just pointing out that acronyms are terrible choices and we should avoid whenever possible. We use them sometimes, but it is worth checking if there is a more descriptive name to (1) honor the author of the algorithm and (2) make the code more searchable.

> I didn't see any name in the thesis.

image

His chapter 3 describes a couple of algorithms. You can see that he mentions Catmull-Clark there, which we already have implemented here. Maybe the algorithm you implemented is one of the other ones?

juliohm: > Maybe `TriQuad`? For "triangle quadrisection"?

I think it is a bit confusing. I read it as a refinement procedure that produces triangles and quadrangles, but that is not the case here.

stla: `TriQuadrisectionRefinement`?

src/refinement/triquadrisection.jl
```diff
@@ -0,0 +1,66 @@
...
+
+* Charles Loop. 1987. [Smooth subdivision surfaces based on 
+  triangles](https://charlesloop.com/thesis.pdf). 
+  Master's thesis, University of Utah.
+"""
+struct TriQuadrisection <: RefinementMethod end
+
+function refine(mesh, ::TriQuadrisection)
+  if !(eltype(mesh) <: Triangle)
+    mesh = simplexify(mesh)
```
juliohm: We shouldn't overwrite user input. I think we need to save the result in a new variable, maybe tmesh

src/refinement/triquadrisection.jl
```diff
@@ -0,0 +1,66 @@
...
+  end
+
+  # retrieve geometry and topology
+  points = vertices(mesh)
+  npoints = length(points)
+  connec = topology(mesh)
+
+  # convert to half-edge structure
+  t = convert(HalfEdgeTopology, connec)
```
juliohm:
```suggestion

  # convert to half-edge structure
  t = convert(HalfEdgeTopology, topology(mesh))
```

src/refinement/triquadrisection.jl
```diff
@@ -0,0 +1,66 @@
...
+  ntriangles = nelements(t)
+
+  # add middle points of edges
+  middles = Dict{Tuple{Int,Int},Int}()
+  ∂₁₀ = Boundary{1,0}(t)
+  for edge_index in 1:nfacets(t)
```
juliohm: rename edge_index to eind, or edge, or ind, ... some name that doesn't have underscores

src/refinement/triquadrisection.jl
```diff
@@ -0,0 +1,66 @@
...
+    triangles[k + 3] = (v3, m31, m23)
+    triangles[k + 4] = (m12, m23, m31)
+  end
+
+  SimpleMesh(points, connect.(triangles, Triangle))
+end
+
+function _ordered_pair(i ,j)
+  i < j ? (i, j) : (j, i)
+end
```
juliohm:
```suggestion
_ordered(i ,j) = i < j ? (i, j) : (j, i)
```

src/refinement/triquadrisection.jl
```diff
@@ -0,0 +1,66 @@
...
+
+Refinement of a mesh by preliminarly triangulating it if needed and
+then subdividing each triangle into four triangles.
+
+## References
+
+* Charles Loop. 1987. [Smooth subdivision surfaces based on 
+  triangles](https://charlesloop.com/thesis.pdf). 
+  Master's thesis, University of Utah.
+"""
```
juliohm: ❤️

test/refinement.jl
```diff
@@ -76,4 +76,21 @@
...
+    ref1   = refine(mesh, TriQuadrisection())
+    ref2   = refine(ref1, TriQuadrisection())
+    ref3   = refine(ref2, TriQuadrisection())
+
+    if visualtests
+      fig = Mke.Figure(resolution=(900, 300))
+      viz(fig[1,1], ref1, showfacets=true)
+      viz(fig[1,2], ref2, showfacets=true)
+      viz(fig[1,3], ref3, showfacets=true)
+      @test_reference "data/ptqrefine-$T.png" fig
```
juliohm:
```suggestion
      @test_reference "data/triquadrefine-$T.png" fig
```

src/refinement/triquadrisection.jl
```diff
@@ -0,0 +1,65 @@
...
+struct TriQuadrisection <: RefinementMethod end
+
+function refine(mesh, ::TriQuadrisection)
+  if !(eltype(mesh) <: Triangle)
+    tmesh = simplexify(mesh)
+  else
+    tmesh = mesh
+  end
```
juliohm:
```suggestion
  tmesh = eltype(mesh) <: Triangle ? mesh : simplexify(mesh)
```

src/refinement/triquadrisection.jl
```diff
@@ -0,0 +1,65 @@
...
+  t = convert(HalfEdgeTopology, topology(tmesh))
+  ntriangles = nelements(t)
+
+  # add middle points of edges
+  middles = Dict{Tuple{Int,Int},Int}()
+  ∂₁₀ = Boundary{1,0}(t)
+  for eind in 1:nfacets(t)
+    e1, e2 = ∂₁₀(eind)
+    p1, p2 = points[e1], points[e2]
+    midpoint = Point((coordinates(p1) + coordinates(p2))/2)
```
juliohm:
```suggestion
    midpoint = center(Segment(points[i], points[j]))
```

src/refinement/triquadrisection.jl
```diff
@@ -0,0 +1,65 @@
...
+  # retrieve vertices
+  points = vertices(tmesh)
+  npoints = length(points)
+
+  # convert topology to half-edge structure
+  t = convert(HalfEdgeTopology, topology(tmesh))
+  ntriangles = nelements(t)
+
+  # add middle points of edges
+  middles = Dict{Tuple{Int,Int},Int}()
```
juliohm:
```suggestion
  midpoints = Dict{Tuple{Int,Int},Int}()
```

src/refinement/triquadrisection.jl
```diff
@@ -0,0 +1,65 @@
...
+  ∂₁₀ = Boundary{1,0}(t)
+  for eind in 1:nfacets(t)
+    e1, e2 = ∂₁₀(eind)
+    p1, p2 = points[e1], points[e2]
```
juliohm:
```suggestion
    i, j = sort(∂₁₀(eind))
```

src/refinement/triquadrisection.jl
```diff
@@ -0,0 +1,65 @@
...
+    midpoint = Point((coordinates(p1) + coordinates(p2))/2)
+    push!(points, midpoint)
+    npoints = npoints + 1
+    middles[_ordered_pair(e1, e2)] = npoints
```
juliohm:
```suggestion
    midpoints[(i,j)] = npoints
```

src/refinement/triquadrisection.jl
```diff
@@ -0,0 +1,65 @@
...
+  end
+
+  # construct subtriangles of faces
+  ∂₂₀ = Boundary{2,0}(t)
+  triangles = Vector{Tuple{Int,Int,Int}}(undef, 4 * ntriangles)
+  for tind in 1:ntriangles
+    v1, v2, v3 = ∂₂₀(tind)
+    m12 = middles[_ordered_pair(v1, v2)]
+    m23 = middles[_ordered_pair(v2, v3)]
+    m31 = middles[_ordered_pair(v3, v1)]
```
juliohm:
```suggestion
    m12 = midpoints[_ordered_pair(v1, v2)]
    m23 = midpoints[_ordered_pair(v2, v3)]
    m31 = midpoints[_ordered_pair(v3, v1)]
```

juliohm (changes requested):
We also need to delete old image files pqtrefine-*.png that were generate in a previous version of the code.

src/refinement/triquadrisection.jl
stla: `center` is not implemented for `Segment`. Can I implement it in this PR? Or I open a new one? Hmm, yes, I think a new one is better.

src/refinement/triquadrisection.jl
```diff
@@ -0,0 +1,60 @@
...
+  # add midpoints of edges
+  midpoints = Dict{Tuple{Int,Int},Int}()
+  ∂₁₀ = Boundary{1,0}(t)
+  for eind in 1:nfacets(t)
+    i, j = sort(∂₁₀(eind))
+    midpoint = Segment(points[i], points[j])(0.5)
+    push!(points, midpoint)
+    npoints = npoints + 1
+    midpoints[(i, j)] = npoints
+  end
```
juliohm: I think we can avoid the construction of this dictionary here. We can simply rely on the insertion order. Instead of looping over facets, try to loop over elements (i.e. triangles) and for each of these elements, you insert 4 new points to the end of the vector of points. In a second loop you can then lookup the points in the vector with the correct offsets, and that should be even faster and more memory efficient than then dictionary solution.

stla: I'm not convinced. Some faces share a common edge. If I follow the algorithm you outline, if I correctly understand, I will compute multiple times the midpoints of the edges. No?

juliohm:
I've read Loop's thesis and I think we should honor his work by calling this method `LoopSubdivision`

src/refinement/triquadrisection.jl
juliohm: You're right. I had the other refinement method in mind where new points lie inside the elements.

src/refinement/loopsubdivision.jl
```diff
@@ -0,0 +1,62 @@
...
+  # convert topology to half-edge structure
+  t = convert(HalfEdgeTopology, topology(tmesh))
+
+  # add midpoints of edges
+  midpoints = Dict{Tuple{Int,Int},Int}()
+  ∂₁₀ = Boundary{1,0}(t)
+  for eind in 1:nfacets(t)
+    i, j = sort(∂₁₀(eind))
+    edge = Segment(points[i], points[j])
+    push!(points, center(edge))
```
juliohm: A final question: in the Loop's thesis, the subdivision procedure is similar to Catmull-Clark in the sense that the initial mesh is considered a de Boor net and the iteration process smooths it out with midpoints that are not necessarily on the edges. Perhaps we should do the same here and "smooth" the points.  Alternatively, we could find a different name for the method that is not LoopSubdivision. Which method do you prefer?

At some point in the future we need to refactor these refinement methods to give more flexibility in the way the new vertices are computed. We have a lot of code repetition currently. For example, QuadRefinement and CatmullClark are basically the same algorithm, except for the new vertex expressions.

stla:
I took a look at the Loop subdivision, e.g. [Loop.cpp](https://github.com/icemiliang/loop_subdivision/blob/master/src/LOOP.cpp). Does it look doable in Meshes.jl?

juliohm:
> I took a look at the Loop subdivision, e.g. [Loop.cpp](https://github.com/icemiliang/loop_subdivision/blob/master/src/LOOP.cpp). Does it look doable in Meshes.jl?

I think we have more generic interfaces for some of the implementation details. Your PR here is almost ready, I think we just need to update the coordinates of the vertices like Loop did. The subdivision itself is already implemented as far as I remember.

stla:
The subdivision is far to be enough I think. I simply took the middle of the edges. But the construction of the locations of the vertices of the Loop subdivision is much more complicated. For each edge it depends on the two faces adjacent to this edge (if this is not a border edge).

juliohm:
> The subdivision is far to be enough I think. I simply took the middle of the edges. But the construction of the locations of the vertices of the Loop subdivision is much more complicated. For each edge it depends on the two faces adjacent to this edge (if this is not a border edge).

Yes, we have a similar vertex calculation in our CatmullClark subdivision algorithm. Maybe you could get inspiration from there? The main challenge is understand how Loop computes the vertex, the implementation should be easy with our topological relations.

juliohm:
Alternatively, we could rename this algorithm here to something else that is not Loop's subdivision. Our current list of refinement methods is:

1. TriRefinement: converts n-gons into triangles
2. QuadRefinement: converts n-gons into quadrangles
3. CatmullClark: same as above + smoothing

Given that the current implementation in this PR always splits n-gons into triangles we need a name that distinguishes it from the existing TriRefinement.

src/refinement/loopsubdivision.jl
```diff
@@ -0,0 +1,62 @@
...
+
+## References
+
+* Charles Loop. 1987. [Smooth subdivision surfaces based on 
+  triangles](https://charlesloop.com/thesis.pdf). 
+  Master's thesis, University of Utah.
+"""
+struct LoopSubdivision <: RefinementMethod end
+
+function refine(mesh, ::LoopSubdivision)
```
juliohm:
```suggestion
function refine(mesh, ::LoopSubdivision)
  @assert paramdim(mesh) == 2 "LoopSubdivision only defined for surface meshes"
  
```

src/refinement/loopsubdivision.jl
```diff
@@ -0,0 +1,64 @@
+# ------------------------------------------------------------------
+# Licensed under the MIT License. See LICENSE in the project root.
+# ------------------------------------------------------------------
+
+"""
+  LoopSubdivision()
```
juliohm:
```suggestion
  TriSubdivision()
```

src/refinement/loopsubdivision.jl
```diff
@@ -0,0 +1,64 @@
...
+## References
+
+* Charles Loop. 1987. [Smooth subdivision surfaces based on 
+  triangles](https://charlesloop.com/thesis.pdf). 
+  Master's thesis, University of Utah.
+"""
+struct LoopSubdivision <: RefinementMethod end
+
+function refine(mesh, ::LoopSubdivision)
+  @assert paramdim(mesh) == 2 "LoopSubdivision only defined for surface meshes"
```
juliohm:
```suggestion
struct TriSubdivision <: RefinementMethod end

function refine(mesh, ::TriSubdivision)
  @assert paramdim(mesh) == 2 "TriSubdivision only defined for surface meshes"
```

src/Meshes.jl
```diff
@@ -365,6 +365,7 @@ export
   TriRefinement,
   QuadRefinement,
   CatmullClark,
+  LoopSubdivision,
```
```suggestion
  TriSubdivision,
```

src/refinement.jl
```diff
@@ -23,3 +23,4 @@ function refine end
 include("refinement/tri.jl")
 include("refinement/quad.jl")
 include("refinement/catmullclark.jl")
+include("refinement/loopsubdivision.jl")
```
juliohm:
```suggestion
include("refinement/trisubdivision.jl")
```

test/refinement.jl
```diff
@@ -76,4 +76,21 @@
       @test_reference "data/catmullclark-3-$T.png" fig
     end
   end
+
+  @testset "LoopSubdivision" begin
```
juliohm:
```suggestion
  @testset "TriSubdivision" begin
```

test/refinement.jl
```diff
@@ -76,4 +76,21 @@
...
+    points = P3[(-1,-1,-1),(1,1,-1),(1,-1,1),(-1,1,1)]
+    connec = connect.([(1,2,3),(3,2,4),(4,2,1),(1,3,4)])
+    mesh   = SimpleMesh(points, connec)
+    ref1   = refine(mesh, LoopSubdivision())
+    ref2   = refine(ref1, LoopSubdivision())
+    ref3   = refine(ref2, LoopSubdivision())
```
juliohm:
```suggestion
    ref1   = refine(mesh, TriSubdivision())
    ref2   = refine(ref1, TriSubdivision())
    ref3   = refine(ref2, TriSubdivision())
```

test/refinement.jl
```diff
@@ -76,4 +76,21 @@
...
+
+    if visualtests
+      fig = Mke.Figure(resolution=(900, 300))
+      viz(fig[1,1], ref1, showfacets=true)
+      viz(fig[1,2], ref2, showfacets=true)
+      viz(fig[1,3], ref3, showfacets=true)
+      @test_reference "data/loopsubdivision-$T.png" fig
```
juliohm:
```suggestion
      @test_reference "data/trisubdivision-$T.png" fig
```

juliohm (changes requested):
@stla let's just rename LoopSubdivision to TriSubdivision and rename the corresponding files. After that we can merge right away.

This method is already useful to have. We can come back to LoopSubdivision in the future.

juliohm (approved):

stla:
The TriRefinement is also useful for the clipping, to add some vertices when there are not enough.

PR #353, opened 2022-12-25
Mesh cleaning utilities

stla:
I propose two functions for mesh cleaning:

- `clean(mesh)` to remove the unused vertices - provided in this PR
- `gather(mesh)` to merge the duplicated vertices - not in this PR

The `clean(mesh)` function is used for the clipping.

### DIFF
--- a/src/cleaning/clean.jl
+++ b/src/cleaning/clean.jl
@@ -1,0 +1,43 @@
+# ------------------------------------------------------------------
+# Licensed under the MIT License. See LICENSE in the project root.
+# ------------------------------------------------------------------
+
+"""
+    clean(mesh)
+
+Remove the unused vertices of `mesh`.
+"""
+function clean(mesh)
+  # get the faces of the mesh
+  topo = convert(HalfEdgeTopology, topology(mesh))
+  ∂₂₀ = Boundary{2,0}(topo)
+  nfaces = nelements(topo)
+
+  # vector to store the indices of the used vertices
+  used = Vector{Int}(undef, 0)
+
+  # dictionary to map the old indices to the new indices
+  newindices = Dict{Int,Int}()
+
+  # vector to store the connectivities of the cleaned mesh
+  connectivities = Vector{Connectivity}(undef, 0)
+
+  # l=length(used) will be incremented
+  l = 0
+
+  # iterate over faces
+  for f in 1:nfaces
+    face = ∂₂₀(f)
+    for index in face
+      if !in(index, used)
+        push!(used, index)
+        l = l + 1
+        newindices[index] = l
+      end
+    end
+    connectivity = connect(tuple([newindices[i] for i in face]...))
+    push!(connectivities, connectivity)
+  end
+
+  SimpleMesh(vertices(mesh)[used], connectivities)
+end
--- a/src/cleaning/gather.jl
+++ b/src/cleaning/gather.jl
@@ -1,0 +1,60 @@
+# ------------------------------------------------------------------
+# Licensed under the MIT License. See LICENSE in the project root.
+# ------------------------------------------------------------------
+
+"""
+    gather(mesh)
+
+Merge the duplicated vertices of `mesh`.
+"""
+function gather(mesh)
+  # get mesh vertices
+  points = vertices(mesh)
+  npoints = length(points)
+
+  # vector indicating the duplicated vertices
+  duplicated = fill(false, npoints)
+
+  # dictionary to map the old indices to the new indices
+  newindices = Dict{Int,Int}()
+
+  # initialize new indices
+  newindex = 1
+
+  # iterate over vertices
+  for index in 1:(npoints-1)
+    if !duplicated[index]
+      tail = points[(index+1):npoints]
+      duplicates = index .+ findall(==(points[index]), tail)
+      duplicated[duplicates] .= true
+      push!(duplicates, index)
+      for i in duplicates
+        newindices[i] = newindex
+      end
+      newindex = newindex + 1
+    end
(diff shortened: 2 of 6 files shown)


src/cleaning/clean.jl
```diff
@@ -0,0 +1,43 @@
+# ------------------------------------------------------------------
+# Licensed under the MIT License. See LICENSE in the project root.
+# ------------------------------------------------------------------
+
+"""
+    clean(mesh)
+
+Remove the unused vertices of `mesh`.
+"""
+function clean(mesh)
```
juliohm, line 10: Please use a more descriptive name for the function. If it removes unused vertices, then we should probably call it `rmvertices`. More generally, I think we should discuss a "clean" api that implements our current Transform api. It would be nice to be able to create cleaning pipelines for meshes.

src/cleaning/clean.jl
```diff
@@ -0,0 +1,43 @@
...
+  used = Vector{Int}(undef, 0)
+
+  # dictionary to map the old indices to the new indices
+  newindices = Dict{Int,Int}()
+
+  # vector to store the connectivities of the cleaned mesh
+  connectivities = Vector{Connectivity}(undef, 0)
+
+  # l=length(used) will be incremented
+  l = 0
```
juliohm, line 26: Variable name can be more descriptive, for example, nused

src/cleaning/clean.jl
```diff
@@ -0,0 +1,43 @@
...
+  # get the faces of the mesh
+  topo = convert(HalfEdgeTopology, topology(mesh))
+  ∂₂₀ = Boundary{2,0}(topo)
+  nfaces = nelements(topo)
+
+  # vector to store the indices of the used vertices
+  used = Vector{Int}(undef, 0)
+
+  # dictionary to map the old indices to the new indices
+  newindices = Dict{Int,Int}()
```
juliohm, line 20: rename to newinds

src/cleaning/clean.jl
```diff
@@ -0,0 +1,43 @@
...
+Remove the unused vertices of `mesh`.
+"""
+function clean(mesh)
+  # get the faces of the mesh
+  topo = convert(HalfEdgeTopology, topology(mesh))
+  ∂₂₀ = Boundary{2,0}(topo)
+  nfaces = nelements(topo)
+
+  # vector to store the indices of the used vertices
+  used = Vector{Int}(undef, 0)
```
juliohm, line 17: rename to usedinds

src/cleaning/clean.jl
```diff
@@ -0,0 +1,43 @@
...
+
+  # dictionary to map the old indices to the new indices
+  newindices = Dict{Int,Int}()
+
+  # vector to store the connectivities of the cleaned mesh
+  connectivities = Vector{Connectivity}(undef, 0)
```
juliohm, line 23: we use connec for list of connectivities

src/cleaning/clean.jl
```diff
@@ -0,0 +1,43 @@
...
+"""
+    clean(mesh)
+
+Remove the unused vertices of `mesh`.
+"""
+function clean(mesh)
+  # get the faces of the mesh
+  topo = convert(HalfEdgeTopology, topology(mesh))
+  ∂₂₀ = Boundary{2,0}(topo)
+  nfaces = nelements(topo)
```
juliohm, line 14: rename to nelems or simply ne

src/cleaning/clean.jl
```diff
@@ -0,0 +1,43 @@
...
+  newindices = Dict{Int,Int}()
+
+  # vector to store the connectivities of the cleaned mesh
+  connectivities = Vector{Connectivity}(undef, 0)
+
+  # l=length(used) will be incremented
+  l = 0
+
+  # iterate over faces
+  for f in 1:nfaces
```
juliohm, line 29: for e in 1:nelems

src/cleaning/clean.jl
```diff
@@ -0,0 +1,43 @@
...
+    face = ∂₂₀(f)
```
juliohm, line 30: this is not a face, please use a more descriptive name like inds to help the reader of the code that this variable stores a list of indices

src/cleaning/clean.jl
```diff
@@ -0,0 +1,43 @@
...
+    for index in face
+      if !in(index, used)
```
juliohm, line 32: use the latex index \in used

src/cleaning/gather.jl
```diff
@@ -0,0 +1,60 @@
+# ------------------------------------------------------------------
+# Licensed under the MIT License. See LICENSE in the project root.
+# ------------------------------------------------------------------
+
+"""
+    gather(mesh)
+
+Merge the duplicated vertices of `mesh`.
```
juliohm, line 8: I don't understand what the function does. Also, we can't merge PRs with multiple additions. We need a PR for a single function with well-defined scope and good tests.

juliohm (changes requested):

juliohm:
@stla I am closing this one in favor of a more composable transform interface. We need to introduce transforms like RemoveUnusedVertices, MergeDuplicatedFaces, ...

src/cleaning/clean.jl
stla: I will add the unused faces removal to this function, following **VCG**. What about `collect_garbage(mesh)` for the name? Or what?

src/cleaning/gather.jl
stla: See the example in the unit tests, you will understand.

PR #356, opened 2022-12-26
rotation between two planes

stla:
Needed for the general torus.

This is the so-called "Reorient transform" on my blog, discussed on zulip.


juliohm:
That is awesome. Can you please start a fresh PR based on our latest master branch? There are conflicts with recent changes we made.

PR #360, opened 2022-12-26
Cone primitive

stla:
Just started the `Cone`. I've implemented the cone mesh but I don't understand where should I put it in the package. 

1. Should I create a discretization method specific to the cone?
2. Should I also define `ConeSurface`?
3. What is `discretizewithin`?
4. I included the number of sides of the base of the cone in the `struct`. Should I rather include it as a parameter of the discretization?


src/Meshes.jl
```diff
@@ -141,11 +141,11 @@ export
 
   # primitives
   Primitive,
-  Line, Ray, Plane, BezierCurve,
+  Line, Ray, Plane, BezierCurve, Cone, 
   Box, Ball, Sphere, Cylinder, CylinderSurface,
   ncontrols, degree, Horner, DeCasteljau,
   radius, bottom, top, axis, isright, sides,
-  measure, diagonal, origin, direction,
+  measure, diagonal, origin, direction,
```
```suggestion
  measure, diagonal, origin, direction,
```

src/Meshes.jl
```diff
@@ -141,11 +141,11 @@ export
 
   # primitives
   Primitive,
-  Line, Ray, Plane, BezierCurve,
+  Line, Ray, Plane, BezierCurve, Cone,
```
```suggestion
  Line, Ray, Plane, BezierCurve, Cone,
```

src/primitives/cone.jl
```diff
@@ -0,0 +1,30 @@
...
+  Cone(vertex, center, radius, nsides, closed)
+
+Cone with `vertex` and base with `center` and `radius` made of `nsides` 
+segments. If `closed`, the base is included.
+"""
+struct Cone{T,I<:Integer} <: Primitive{3,T}
+  vertex::Point{3,T}
+  center::Point{3,T}
+  radius::T
+  nsides::I
```
juliohm:
```suggestion
struct Cone{T} <: Primitive{3,T}
  vertex::Point{3,T}
  center::Point{3,T}
  radius::T
  nsides::Int
```

src/primitives/cone.jl
```diff
@@ -0,0 +1,30 @@
...
+  closed::Bool
+end
+
+function Cone(vertex::Tuple, center::Tuple, radius, nsides, closed)
```
juliohm, line 19: Take a look at Sphere for an example of outer constructors. We first need to make sure that the outer constructor with geometry types has the correct result and then we can wrap it in a second outer constructor with tuples.

juliohm (changes requested):
Another great addition 👍🏽 

A few review comments to follow the conventions in other primitives.

juliohm:
> 1. Should I create a discretization method specific to the cone?

Take a look at how Sphere and Cylinder are discretized. Maybe all you need to do is tell how points are sampled inside the cone solid (or in the case of a ConeSurface, over the surface).

Are you aware of the VSCode debugger? It is pretty useful to learn the code flow. You can type `@enter discretize(sphere)` in the VSCode REPL and follow the steps of specific methods being called. It is very educational when you are not 100% familiar with the codebase.

> 2\. Should I also define `ConeSurface`?

If the cone is defined as a solid, then we also need its boundary as a ConeSurface.

> 3\. What is `discretizewithin`?

This is a method that discretizes a 2D region based on its boundary, given as a Multi{Chain}.

src/primitives/cone.jl
```diff
@@ -0,0 +1,30 @@
...
+
+Cone with `vertex` and base with `center` and `radius` made of `nsides` 
+segments. If `closed`, the base is included.
+"""
+struct Cone{T} <: Primitive{3,T}
+  vertex::Point{3,T}
+  center::Point{3,T}
+  radius::T
+  nsides::Int
+  closed::Bool
```
juliohm, line 16: The definition of a Cone relies on an apex and a plane. We can assume that the base of the cone is a perfect circle, but it could be any simple geometry like a polygon: https://en.wikipedia.org/wiki/Cone

A cone with polygonal base is called a pyramid. We have a Pyramid type already but we assume a quadrangle base currently. We should probably generalize it to accept any polygon.

juliohm:
Already available in master.

PR #361, opened 2022-12-26
Connected components

stla:


juliohm:
I think this one requires more discussion to improve the usability. Can you share an example of input and output with the current implementation? Some visualization that could clarify the ultimate goal?

stla:
> I think this one requires more discussion to improve the usability. Can you share an example of input and output with the current implementation? Some visualization that could clarify the ultimate goal?

```julia
using MarchingCubes
using Meshes
using MeshViz
using GLMakie
using ColorSchemes
using Makie

# isosurface function (slice of a seven-dimensional toratope)
function f(x, y, z, a)
   (sqrt(
     (sqrt((sqrt((x*sin(a))^2 + (z*cos(a))^2) - 5)^2 + (y*sin(a))^2) - 2.5)^2 + 
       (x*cos(a))^2) - 1.25
   )^2 + (sqrt((sqrt((z*sin(a))^2 + (y*cos(a))^2) - 2.5)^2) - 1.25)^2
end

# make the isosurface
n = 200
voxel = zeros(Float32, n, n, n)
rg = LinRange(-10, 10, n)
for k in 1:n, j in 1:n, i in 1:n
  voxel[i, j, k] = f(rg[i], rg[j], rg[k], pi/2)
end
mc = MC(voxel, Int32)
march(mc, 0.25)
mesh = MarchingCubes.makemesh(Meshes, mc)

# compute connected components
meshes = connected_components(mesh)
ncomponents = length(meshes)

# plot the connected components with different colors
colors = cgrad(:viridis, ncomponents, categorical = true)
function draw()
  fig, _, _ = viz(meshes[1]; color = colors[1])
  for i in 2:ncomponents
    viz!(meshes[i]; color = colors[i])
    Makie.save("connected_components.png", fig)
  end
end
```

connected_components

stla:
It was a bit slow.

In MeshLab, there is the "Remove small connected components" operation. It can be useful. To implement such a feature we would need a function returning the diameter of a mesh.

stla:
I think I found an improvement. Once we create the mesh corresponding to the first connected component, it is not necessary to search among all faces when we want to create the mesh corresponding to the second connected component: we can drop the faces of the previously created mesh. And so on.

src/connected_components.jl
```diff
@@ -0,0 +1,68 @@
...
+
+  # make the graph
+  graph = SimpleGraph(nvertices(mesh))
+  for eind in 1:nfacets(topo)
+    i, j = ∂₁₀(eind)
+    add_edge!(graph, i, j)
+  end
+
+  graph
+end
```
juliohm, line 68: I think we can close this PR and start a new one that introduces functions to return the various graphs associated with a mesh. Given a mesh, there are various kinds of graphs that we can take out of it. In this specific algorithm here it seems that you are taking the primal graph of edges connecting vertices.

In this new PR we can assess the Graphs.jl dependency, and define a clean interface to query the different graph types.

PR #363, opened 2022-12-26
Torus primitive

stla:
Still have to implement some tests. 
I visually checked on some plots and this seems ok.


stla:
In fact, it will be more convenient to implement some tests for the torus passing through three points, which is not implemented in this PR yet.

stla:
I close since I have updated this PR after pulling the new `master` branch. Will make this new PR right now.

PR #364, opened 2022-12-26
Torus primitive

stla:
Note: I'm not sure of the wording "equator of a torus".


stla:
I made this picture using the torus passing through three points. It demonstrates the code is correct.
Capture du 2022-12-27 00-07-05

src/discretization.jl
```diff
@@ -108,6 +108,9 @@ discretize(ball::Ball{2}) =
 discretize(sphere::Sphere{3}) =
   discretize(sphere, RegularDiscretization(50))
 
+discretize(torus::Torus) =
+  discretize(torus, RegularDiscretization(60, 30))
```
juliohm:
```suggestion
  discretize(torus, RegularDiscretization(50))
```
We default to 50 samples per parametric dimension. Is it too bad for visualization?

src/primitives/torus.jl
```diff
@@ -0,0 +1,75 @@
...
+  center::Point{3,T}
+  normal::Vec{3,T}
+  major::T
+  minor::T
+end
+
+function Torus(center::Tuple, normal::Tuple, major::T1, minor::T2) where {T1,T2}
+  T = promote_type(eltype(center), eltype(normal), T1, T2, Float32)  
+  Torus(Point(T.(center)), Vec(T.(normal)), T(major), T(minor))
+end
```
juliohm: We need an outer constructor with original geometry types. See Sphere and Ball for example. After this constructor is written, we can simply write a second outer constructor with tuples.

src/primitives/torus.jl
```diff
@@ -0,0 +1,75 @@
...
+  v13 = p3 - p1
+  p12 = coordinates(p1 + v12/2)
+  p13 = coordinates(p1 + v13/2)
+  offset = coordinates(p1) ⋅ n⃗
+  A = transpose(hcat(n⃗, v12, v13))
+  b = [offset, p12 ⋅ v12, p13 ⋅ v13]
+  c = Point(inv(A) * b)
+  major = norm(p1 - c)
+  #
+  T = typeof(major)
```
juliohm, line 45: This pattern is not ideal. I see you are relying on it in many PRs. We would like to define constructors where T is informed with a `where {T}` clause in general. The case where arguments of the constructor have mixed coordinate type is not supported, at least for now.

src/primitives/torus.jl
```diff
@@ -0,0 +1,75 @@
...
+area(t::Torus) = measure(t)
+
+function Base.in(p::Point, t::Torus)
+  R, r = radii(t)
+  c = center(t)
+  n⃗ = t.normal
+  rotation = Rotate(n⃗, Vec(0, 0, 1))
+  M = convert(DCM, rotation.rot)
+  x, y, z = M * (p - c)
+  (R - √(x^2 + y^2))^2 + z^2 ≤ r^2
```
juliohm:
```suggestion
  c, n = t.center, t.normal
  R, r = t.major, t.minor
  rotation = Rotate(n, Vec(0, 0, 1))
  M = convert(DCM, rotation.rot)
  x, y, z = M * (p - c)
  (R - √(x^2 + y^2))^2 + z^2 ≤ r^2
```

src/sampling/regular.jl
```diff
@@ -172,4 +172,28 @@ end
...
+  s = √kxy / r 
+
+  V = T <: AbstractFloat ? T : Float64
+  umin, umax = V(-s * π), V(s * π)
+  vmin, vmax = V(-π), V(π)
+  urange = range(umin, stop=umax, length=sz[1]+1)[1:sz[1]]
+  vrange = range(vmin, stop=vmax, length=sz[2]+1)[1:sz[2]]
+
+  c = center(torus)
+  normal = torus.normal
```
juliohm: We need a public function to access the normal of a torus. We can't access it here as a field.

src/sampling/regular.jl
```diff
@@ -172,4 +172,28 @@ end
...
+  R, r = radii(torus)
+  kxy = R^2 - r^2
+  kz = √kxy * r
+  s = √kxy / r 
+
+  V = T <: AbstractFloat ? T : Float64
+  umin, umax = V(-s * π), V(s * π)
+  vmin, vmax = V(-π), V(π)
+  urange = range(umin, stop=umax, length=sz[1]+1)[1:sz[1]]
+  vrange = range(vmin, stop=vmax, length=sz[2]+1)[1:sz[2]]
```
juliohm: Please try to follow the convention in other regular samplers. This indexing with [1:sz[1]] looks strange

src/transforms/rotate.jl
```diff
@@ -35,7 +35,7 @@ function Rotate(u::Vec, v::Vec)
   u⃗ = normalize(u)
   v⃗ = normalize(v)
   re = √((1 + u⃗ ⋅ v⃗)/2)
-  im = (u⃗ × v⃗) / 2re
+  im = (v⃗ × u⃗) / 2re
```
juliohm: Is this a bug in our Rotate transform? If yes, we need to fix it in a separate PR.

test/primitives.jl
```diff
@@ -353,4 +353,22 @@
     c = CylinderSurface(1)
     @test coordtype(c) == Float64
   end
+
+  @testset "Torus" begin
+    # check the `in` function
```
juliohm:
```suggestion
    # basic tests
```

test/primitives.jl
```diff
@@ -353,4 +353,22 @@
...
+    t = Torus(T.((1,1,1)), T.((1,0,0)), 2, 1)
+    @test P3(1,1,-1) ∈ t
+    @test P3(1,1,1) ∉ t
+    # torus passing through three points
```
juliohm:
```suggestion
    @test P3(1,1,1) ∉ t
    
    # torus passing through three points
```

juliohm (changes requested):
Review attached.

src/discretization.jl
stla: Didn't try but the second parameter is for the small circle (minor radius) and the first parameter for the big circle (major radius) so it seems sensible to take more subdivisions for the first.

src/transforms/rotate.jl
stla: No, that's still due to the fact that `ReferenceFramesRotation` rotates the frame.

stla: I'm tired I didn't sleep... In fact I think there's a problem: in our `Rotate` stuff we inverse the rotation. But here I don't use our `Rotate` stuff... So I think this rotation will be inverted if we apply it to a mesh. But me I don't apply it to a mesh, you see?

src/sampling/regular.jl
stla: But when this is periodic on a range e.g. `0 - 2pi` one has to drop one of the numbers at the edge of the range otherwise there's a duplicated vertex in the mesh.

src/transforms/rotate.jl
juliohm: Yes, I think we need a new utility function in utils.jl that takes two vectors and returns the appropriate DCM for the rotation of the object. We can then use this utility function in the constructor of Rotate.

src/sampling/regular.jl
juliohm: Yes, but we do that in the range itself. See Cylinder for example where the correct range is constructed for one of the dimensions.

src/transforms/rotate.jl
juliohm: I've added a new utility function called `uvrotation` with the code from Rotate. You can now rebase this PR on top of the latest master branch to make use of it.

src/sampling/regular.jl
stla: I see. I will do the change but I prefer my way ^^

src/discretization.jl
juliohm: I don't know if we should add an exception. It seems a bit arbitrary anyways. Let's leave as the others and we can always fine tune later.

src/primitives/torus.jl
```diff
@@ -0,0 +1,76 @@
...
+  # circumcenter of p1, p2, p3
+  n⃗ = normal(Plane(p1, p2, p3))
+  v12 = p2 - p1
+  v13 = p3 - p1
+  p12 = coordinates(p1 + v12/2)
+  p13 = coordinates(p1 + v13/2)
+  offset = coordinates(p1) ⋅ n⃗
+  A = transpose(hcat(n⃗, v12, v13))
+  b = [offset, p12 ⋅ v12, p13 ⋅ v13]
+  c = Point(inv(A) * b)
```
juliohm, line 41: from the code comment shouldn't this block of code be just `center(Sphere(p1, p2, p3))`?

juliohm (changes requested):
Review attached.

test/primitives.jl
```diff
@@ -353,4 +353,26 @@
     c = CylinderSurface(1)
     @test coordtype(c) == Float64
   end
+
+  @testset "Torus" begin
+    # basic tests
+    t = Torus(T.((1,1,1)), T.((1,0,0)), 2, 1)
```
juliohm:
```suggestion
    t = Torus(P3(1,1,1), V3(1,0,0), T(2), T(1))
```

test/primitives.jl
```diff
@@ -353,4 +353,26 @@
...
+    @test P3(1,1,-1) ∈ t
+    @test P3(1,1,1) ∉ t
+    
+    # torus passing through three points
+    p₁ = P3(0,0,0)
+    p₂ = P3(1,2,3)
+    p₃ = P3(3,2,1)
+    t = Torus(p₁, p₂, p₃, 1)
```
juliohm:
```suggestion
    t = Torus(p₁, p₂, p₃, T(1))
```

src/sampling/regular.jl
```diff
@@ -145,6 +145,32 @@ function sample(::AbstractRNG, cylsurf::CylinderSurface{T},
...
+  urange = range(umin, stop=umax-δu, length=sz[1])
+  vrange = range(vmin, stop=vmax-δv, length=sz[2])
+
+  # rotation to align n with direction (-1, 0, 1)
+  Q = uvrotation(n⃗, Vec{3,V}(-1, 0, 1))
+
+  r⃗(u, v) = Vec{3,V}(kxy * cos(u/s), kxy * sin(u/s), kz * sin(v)) / (R - r*cos(v))
+
+  ivec(c + Q * r⃗(u, v) for u in urange, v in vrange)
+end
```
juliohm, line 172: Missing tests for this new sampling method.

src/primitives/torus.jl
stla: There's no unique sphere passing through three points in 3D :-)

stla: Should we implement `circumcenter` for a `Triangle`? For the tetrahedron we already use it, in the sphere passing through four points.

juliohm: That is a great idea. Let's add a new constructor for Triangle and then come back here to finish this PR 👍🏽 

In the meantime, this PR is missing tests for the regular sampling method too.

juliohm:
@stla did you have some time to take a look into this? Maybe we should split this PR into two, one introducing the new primitive with basic tests and another introducing the RegularSampling method? Looking forward to your thoughts.

stla:
I'm going to work on it.

Could you please show me how to deal with the constructor? I'm lost. I did:

```julia
function Torus(center::Tuple, normal::Tuple, major::T1, minor::T2) where {T1,T2}
  T = promote_type(eltype(center), eltype(normal), T1, T2, Float32)  
  Torus(Point(T.(center)), Vec(T.(normal)), T(major), T(minor))
end
```
What should I do instead?

stla:
> Maybe we should split this PR into two, one introducing the new primitive with basic tests and another introducing the RegularSampling method? Looking forward to your thoughts.

Ok, I will create another branch since I have to rebase anyway.

src/sampling/regular.jl
stla: I did something stupid with git and I'm trying to repair that... 
What is `R^1` in this file? Is it me who wrote that?? And why aligning the normal with `(-1,0,1)` instead of `(0,0,1)`? I'm lost...

stla: However I remember I did an image with tori passing through three point and this worked fine.

juliohm: I don't know the answer to these questions, it is been a while since the last time I reviewed the PR. But the code is probably intact from your original implementation.

juliohm:
I suggest closing this PR and starting two new ones. One that introduces the primitive for review and another one later that introduces the sampling method.

src/sampling/regular.jl
stla: No, I never wrote `R^1` and `-1,0,1`, that's very strange. Anyway I re-did this PR and it works, I'm going to close this one.

PR #369, opened 2023-01-26
circumcenter for 2D triangles

stla:
I'm preparing the 3D circumcenter now. It is needed for the torus passing through three points.

PS: I know I forgot to uncomment in `runtests.jl`.


stla:
Should we add `circumradius`?

Project.toml
```diff
@@ -9,6 +9,7 @@ CircularArrays = "7a955b69-7140-5f4e-a0ed-f168c5e2e749"
 Distances = "b4f34e82-e78d-54a5-968a-f98e89d6e8f7"
 IterTools = "c8e1da08-722c-5040-9ed9-7db0dc04731e"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
+MeshViz = "9ecf9c4f-6e5a-4b5e-83ae-06f2c7a661d8"
```
juliohm, line 12: We can't add MeshViz as a dependency :)

src/polytopes/ngon.jl
```diff
@@ -80,6 +80,33 @@ function measure(t::Triangle{3})
   norm((B - A) × (C - A)) / 2
 end
 
+function circumcenter(t::Triangle{2})
+  A, B, C = t.vertices
+  sin2A = sin(2 * ∠(C, A, B))
+  sin2B = sin(2 * ∠(A, B, C))
+  sin2C = sin(2 * ∠(B, C, A))
+  sines = [sin2A, sin2B, sin2C]
```
juliohm, line 88:
```suggestion
  sines = sin(2 .* innerangles(boundary(t)))
```
Maybe this is more straightforward? Its worth checking if the performance is similar.

stla:
Or rather `circumcircle`?

src/polytopes/ngon.jl
```diff
@@ -80,6 +80,33 @@ function measure(t::Triangle{3})
...
+  x₁, y₁ = coordinates(A)
+  x₂, y₂ = coordinates(B)
+  x₃, y₃ = coordinates(C)
+  x = [x₁, x₂, x₃]
+  y = [y₁, y₂, y₃]
+  Point((x ⋅ sines) / sum(sines), (y ⋅ sines) / sum(sines))
```
juliohm, line 94:
```suggestion
  xs = coordinates.(t.vertices)
  Point((xs ⋅ sines) / sum(sines))
```
Maybe this works?

src/polytopes/ngon.jl
```diff
@@ -80,6 +80,33 @@ function measure(t::Triangle{3})
...
+end
+
+function circumcenter(t::Triangle{3})
+  A, B, C = t.vertices
+  vAB = B - A
+  vAC = C - A
+  mAB = coordinates(A + vAB/2)
+  mAC = coordinates(A + vAC/2)
+  n⃗ = normal(t)
+  offset = coordinates(A) ⋅ n⃗
```
juliohm, line 104: I would try to use single letter for all variables given that the function is mostly made of single letter variables.

juliohm (changes requested):
Awesome addition 💯 left a few small suggestions of improvement.

Project.toml
stla: ?? Never changed this file! Not intentional.

juliohm:
circumcircle seems more natural as a unit

Em qui., 26 de jan. de 2023 11:22, stla ***@***.***> escreveu:

src/polytopes/ngon.jl
stla: I don't know. Personally I find it is hard to understand this code.

juliohm:
I wonder if we could simply add a construtor for Circle instead with 3
vertices?

Em qui., 26 de jan. de 2023 11:27, stla ***@***.***> escreveu:

src/polytopes/ngon.jl
stla: Hmm.. I like `vAB`: vector from A to B. And `mAB` is for middle.

juliohm: I referred to the variable `offset`, it is kind of an outlier in the code when all variables are short.

juliohm:
Should we avoid introducing a new circumcenter function and instead just add a new constructor to the Circle geometry with 3 vertices? I think it is more general and clean.

stla:
> Should we avoid introducing a new circumcenter function and instead just add a new constructor to the Circle geometry with 3 vertices? I think it is more general and clean.

Yes, agreed. But we should also have a 2D circle for the circumcircle of a 2D triangle.

juliohm:
That is true. We probably need a constructor for Circle and another one for Sphere.

Will close this PR so that we can start fresh in another one.

juliohm:
I would focus on the 3D case though with Circle given that it is blocking the Torus PR.

stla:
Just seen your `Circle` code. I find it not very comfortable, given by a plane and a radius, and the center defined as the origin of the plane. What is the origin of the plane passing through three points for example? Why didn't you define a circle by the center and the radius?

stla:
Ah yes, you're not wrong... center+radius does not define a circle in 3D!

stla:
But I really don't like the plane origin. A plane has no origin geometrically speaking, it makes sense when a Cartesian coordinates system is chosen. The origin is just a choice of the user. I need the plane passing through three points and it's not clear what is its origin. 

I would define a 3D circle by *center+radius+plane*.

juliohm:
Exactly! The Circle in 3D requires a plane and a radius. We need to take 3
points and find the plane with origin at the circumcenter.

Em qui., 26 de jan. de 2023 12:18, stla ***@***.***> escreveu:

juliohm:
> I would define a 3D circle by _center+radius+plane_.

That would be a good improvement I guess. Feel free to propose the change in the same PR as you add more constructors to the Circle.

juliohm:
The problem is that the Plane already contains the origin in our representation of planes. So it is redundant to ask for center + plane. Maybe we can stick to plane + radius and just add another constructor with 3 vertices?

stla:
The origin of the plane is not intrinsic to the plane: if you change the origin of the plane, you don't change the plane.

The origin of a plane in the package is nothing but a point on the plane, no? Does it have the role of an "origin"?

I need the plane passing through three points. What is its origin? (I didn't check). We could set it to the circumcenter, but again this is not intrinsic to the plane, and this would be not clear for another developer.

juliohm:
The word origin in this case is used to refer to any point in the plane, but some downstream uses of it assume it is an origin for the base of vectors that spans the plane. We need a point in the plane anyways to place it in space. The origin can be any point derived from the 3 vertices, I would pick the centroid of the vertices as a candidate.

If you have another suggestion to remove the reference point from the plane representation, please feel free to share it.

juliohm:
I'd take the circumcenter of the 3 vertices as the origin of the plane
understanding that it is just a point on the plane where the parametric
representation plane(u, v) starts. That is, origin(plane) == plane(0, 0)

Em qui., 26 de jan. de 2023 12:34, stla ***@***.***> escreveu:

juliohm:
Maybe we should remove the accessor function origin(plane) and use plane(0,
0) in downstream code. Will see if that is possible in a PR today.

Em sex., 27 de jan. de 2023 06:55, Júlio Hoffimann <
***@***.***> escreveu:

juliohm:
Done in https://github.com/JuliaGeometry/Meshes.jl/commit/7dcabe68d2b5b7f8fdd1ca73863290c0aad97b83

The new Circle constructor could take a plane, a center and a radius as you suggested 👍🏽

stla:
Nice :+1:  Sorry for the inactivity, not very motivated these days.

stla:
> Done in [7dcabe6](https://github.com/JuliaGeometry/Meshes.jl/commit/7dcabe68d2b5b7f8fdd1ca73863290c0aad97b83)
> 
> The new Circle constructor could take a plane, a center and a radius as you suggested 👍🏽

You didn't merge this commit? I've just fetched and I don't see this new Circle constructor.

stla:
Ah, perhaps I misunderstood. You wrote *"could"*. You want me to implement this new constructor, right?

stla:
> I'd take the circumcenter of the 3 vertices as the origin of the plane understanding that it is just a point on the plane where the parametric representation plane(u, v) starts. That is, origin(plane) == plane(0, 0) 

This is the centroid which is located at `(0,0)`, not the circumcenter. No?

juliohm:
> Ah, perhaps I misunderstood. You wrote _"could"_. You want me to implement this new constructor, right?

That would be awesome 🥇 

> This is the centroid which is located at `(0,0)`, not the circumcenter. No?

I meant that we could find the circumcenter from the 3 vertices and then define the origin of the plane accordingly. Feel free to propose alternative origins. We can review in the PR :)

stla:
> I meant that we could find the circumcenter from the 3 vertices and then define the origin of the plane accordingly. Feel free to propose alternative origins. We can review in the PR :)

Hmm but I find the centroid is more sensible, because as you say this is the `(0,0)` point.

juliohm:
Makes sense, I think the centroid is more sensible too 👍🏽

PR #370, opened 2023-02-11
Add "Indexing with Regex" support for the `Data` trait

eliascarv:


juliohm (approved):
Clean and to the point ❤️

PR #371, opened 2023-02-14
3D circumcircle

stla:


src/primitives/circle.jl
```diff
@@ -5,24 +5,54 @@
...
     Circle(plane, radius)
 
-A circle on a `plane` with given `radius`.
+A circle on a `plane` with given `center` and `radius`.
 
 See also [`Disk`](@ref).
 """
 struct Circle{T} <: Primitive{3,T}
   plane::Plane{T}
+  center::Point{3,T}
```
juliohm, line 14: Should this center be expressed in terms of parametric coordinates (u,v)? I find it a bit strange that we need to provide a 3D center after choosing a Plane of reference. We only have 2 degrees of freedom to choose the center after the Plane is given.

juliohm:
Closing this PR as it is proposing non-ideal changes to the primitives. Appreciate if you can create a new PR with just the circumcenter functionality, i.e. a new constructor for the Circle that accepts 3 points.

stla:
You mean we abandon the Circle constructor with a plane?

juliohm:
I don't know if I follow... Can't we take 3 points and then construct the plane to call the other constructor?

stla:
Ah you mean we keep the constructor taking the origin as the center? Ok. I will see how to deal to match the center.

PR #374, opened 2023-02-27
Torus primitive

stla:
It works:

Capture du 2023-02-27 23-15-11


stla:
I know I use `promote_type` and you don't like it, but I don't know how to do.

src/Meshes.jl
```diff
@@ -144,11 +144,11 @@ export
   Ray, Line,
   BezierCurve,
   Plane, Box,
-  Ball, Sphere, Disk, Circle,
+  Ball, Sphere, Disk, Circle, Torus,
   Cylinder, CylinderSurface,
   Cone, ConeSurface,
```
juliohm, line 149:
```suggestion
  Ball, Sphere, Disk, Circle,
  Cylinder, CylinderSurface,
  Cone, ConeSurface,
  Torus,
```

src/primitives/torus.jl
```diff
@@ -0,0 +1,67 @@
...
+  center::Point{3,T}
+  normal::Vec{3,T}
+  major::T
+  minor::T
+end
+
+function Torus(center::Tuple, normal::Tuple, major::T1, minor::T2) where {T1,T2}
+  T = promote_type(eltype(center), eltype(normal), T1, T2, Float32)  
+  Torus(Point(T.(center)), Vec(T.(normal)), T(major), T(minor))
+end
```
juliohm, line 22: Please follow the pattern in Ball, i.e., first a constructor with Point and Vec and type parameter T, then a second constructor with Tuple.

src/primitives/torus.jl
```diff
@@ -0,0 +1,67 @@
...
+minor radius `minor`.
+"""
+function Torus(p1::Point{3}, p2::Point{3}, p3::Point{3}, minor)
+  c = Circle(p1, p2, p3)
+  O = center(c)
+  major = radius(c)
+  n⃗ = normal(Plane(p1, p2, p3))
+  T = typeof(major)
+  Torus(O, Vec{3,T}(n⃗), major, T(minor))
+end
```
juliohm, line 37: I don't think we need this constructor, at least in this PR.

src/sampling/regular.jl
```diff
@@ -172,4 +172,28 @@ end
 function sample(rng::AbstractRNG, grid::CartesianGrid,
                 method::RegularSampling)
   sample(rng, boundingbox(grid), method)
+end
+
+function sample(::AbstractRNG, torus::Torus{T},
+                method::RegularSampling) where {T}
```
juliohm, line 178: let's add this sampling method in a separate PR as suggested with proper tests and more careful review. This PR should only introduce the Torus type with the appropriate constructors and acessor functions.

stla:
Observe that in the constructor function of the torus given three points, the normal to the plane is computed two times: first time when computing the circle passing through the three points, second time explicitely. Therefore I'm wondering whether we should add a `normal` function for 3D circles? Hmm... not sure that would avoid the double computation in fact.

juliohm:
@stla let's stick to the single constructor that is low-level for now. Please close this PR and follow our contribution guidelines in the next one:

https://github.com/JuliaGeometry/Meshes.jl#contributing

In particular, PRs should do one single thing and include tests for full coverage of the thing that was added.

As I suggested above, one PR could add the Torus primitive and all relevant accessor functions to be tested with non-trivial examples. Another PR could add the sampling method with another set of tests.

stla:
The code for the sphere is:

```julia
Sphere(center::Point{Dim,T}, radius) where {Dim,T} = Sphere(center, T(radius))

Sphere(center::Tuple, radius) = Sphere(Point(center), radius)
```

But here there are two radii. So I don't see how to do the analogue of the first line.

stla:
And there is `normal`.

stla:
Like this ?

```julia
Torus(center::Point{3,T}, normal::Vec{3,T}, major, minor) where {T} = 
  Torus(center, normal, T(major), T(minor))

function Torus(center::Tuple, normal::Tuple, major, minor)
  Torus(Point(center), Vec(normal), major, minor)
end
```

Issue #308, opened 2022-10-10
Suggestion for new intersection types


dorn-gerhard:
I am planning to work on the intersection of Ray - Triangle to provide more
intersection types. Currently it just works in 3D which is basically fine though I don't see issues with a 2D implementation.

But most important if a Ray is parallel to a triangle in the sense that they lie in the same plane, then the 
intersection should be a segment.

So basically there are two main cases, namely whether Ray and triangle **lie in the same plane** (2D case) **or not**

Further I would indicate whether the boundary of the triangle is involved in the intersection.

Here are my suggested intersection types:

A) ray is in the plane of the trianlge (also 2D case, overlapping and touching)
  - CASE 1 do not touch or overlap (**NoIntersection**)
  - CASE 2 touch in one corner (**CornerTouchingRayTriangle** -> Point)
  - CASE 3 touch on an edge point (**EdgeTouchingRayTriangle** -> Point)
  - CASE 4 overlap with the triangle (**OverlappingRayTriangle** -> Segment)
  - CASE 5 overlap with an edge of the triangle (**OverlappingRayTriangleEdge** -> Segment)
  - CASE 6 overlap with the triangle includes a corner (**OverlappingRayTriangleCorner** -> Segment)

B) ray is not in the plane of the triangle (crossing and touching)
  - CASE 1 do not cross or touch (**NoIntersection**)
  - CASE 2 touch in one corner (**CornerTouchingRayTriangle** -> Point
  - CASE 3 touch on an edge point (**EdgeTouchingRayTriangle** -> Point)
  - CASE 7 cross in one inner point of the triangle (**CrossingRayTriangle** -> Point)

Cases A would be new and are not treated so far (I think). Cases B would change IntersectingRayTriangle to the three suggested cases.
What do you think?

juliohm:
You are the master of intersections, I am trusting the type hierarchy you designed carefully. Thank you for putting the time into it.

kylebeggs:
Just wanted to say thanks for working on this. With these planned updates, #293 will finally be able to be finished!

dorn-gerhard:
After finishing intersections between ``rays``, ``lines`` and ``segments``, I will work on intersections of these three geometries with ``plane`` and ``triangles``.

Therefore I will think about a new (more general) `intersectparameters` function

juliohm:
@kylebeggs has solved this issue already. We have more combinations of ray-triangle intersection types already. 🎉

PR #377, opened 2023-03-02
updated ray-triangle intersect types

kylebeggs:
This PR attempts to complete #308 partially (just some of the ray-triangle types) as well as adds two more types which handle if the origin of the ray intersects an edge or a vertex/corner of the triangle.
1. `CornerOriginTouchingRayTriangle`
2. `EdgeOriginTouchingRayTriangle`


src/intersections/raytriangle.jl
```diff
@@ -49,7 +56,15 @@ function intersection(f, r::Ray{3,T}, t::Triangle{3,T}) where {T}
     return @IT NoIntersection nothing f
   end
 
+  any(o .≈ vs) && (return @IT CornerOriginTouchingRayTriangle r(λ) f)
+
+  λ ≈ zero(T) && (return @IT EdgeOriginTouchingRayTriangle r(λ) f)
+
+  coords = (u, v, det - u - v)
+  count(x -> x ≈ zero(T), coords) == 1 && (return @IT EdgeTouchingRayTriangle r(λ) f)
+  count(x -> x ≈ det, coords) == 1 && (return @IT CornerTouchingRayTriangle r(λ) f)
```
juliohm: I think we can make coords a Vector and then use isapprox as it is done in other places in the codebase.

src/intersections/raytriangle.jl
```diff
@@ -5,6 +5,13 @@
 #=
 Möller, T. & Trumbore, B., 1997.
 (https://www.tandfonline.com/doi/abs/10.1080/10867651.1997.10487468)
+
+Cases
+1. CrossingRayTriangle - middle of ray intersects middle of triangle
+2. CornerOriginTouchingRayTriangle - origin of ray intersects corner of triangle
+3. EdgeOriginTouchingRayTriangle - origin of ray intersects edge of triangle
+4. EdgeTouchingRayTriangle - middle of ray intersects edge of triangle
+5. CornerTouchingRayTriangle - middle of ray intersects corner of triangle
```
juliohm: Can we come up with better names for cases 2 to 5? I know it is hard to come up with these names, but we should try to remove the words Edge and Corner if possible. The prefix Origin is fine because the Ray has an origin, but the triangle only has vertices.

src/intersections/raytriangle.jl
```diff
@@ -49,7 +56,15 @@ function intersection(f, r::Ray{3,T}, t::Triangle{3,T}) where {T}
     return @IT NoIntersection nothing f
   end
 
+  any(o .≈ vs) && (return @IT CornerOriginTouchingRayTriangle r(λ) f)
```
juliohm: all these cases where we use isapprox we also need to use atol(T) for the correct precision.

juliohm (changes requested):
Thanks for the improvement @kylebeggs ❤️ left a few quick suggestions. Let's see if we can find better names, it is a tricky task, I know...

src/intersections/raytriangle.jl
kylebeggs: I'm not sure that is necessary. `isapprox` finds the least precise type and sets the tol based on that.

src/intersections/raytriangle.jl
kylebeggs: Yes, I agree, I was trying to stay as much in convention with what was suggested in previous issues (4 and 5 were suggested in [#308](https://github.com/JuliaGeometry/Meshes.jl/issues/308)). How about:

2. VertexOriginTouchingRayTriangle - origin of ray intersects corner of triangle
3. SegmentOriginTouchingRayTriangle - origin of ray intersects edge of triangle
4. SegmentTouchingRayTriangle - middle of ray intersects edge of triangle
5. VertexTouchingRayTriangle - middle of ray intersects corner of triangle
   
Honestly, I'm not sure what else to call the cases where the edge/segment of the triangle are involved. I don't really care so if you have a suggestion we can just go with it.

juliohm: I think the following names are an improvement as they focus on what is being intersected and how. For example, CornerTouching means that the corner of the triangle is being touched and EdgeCrossing means that the edge of the triangle is being crossed:

1. CrossingRayTriangle
2. CornerTouchingRayTriangle
3. EdgeTouchingRayTriangle
4. EdgeCrossingRayTriangle
5. CornerCrossingRayTriangle

src/intersections/raytriangle.jl
juliohm: It is because we use `isapprox(a, b, atol=atol(T))` with the function `atol(T)` defined in the package. We have tolerances for 64 and 32 bits that are not necessarily the same as default tolerances in numerical code.

src/intersections/raytriangle.jl
kylebeggs: Makes sense to me! I'll update with this.

src/intersections/raytriangle.jl
kylebeggs: Oh, ok. Can you link where `atol` is defined in the package just so I know what we are doing here?

kylebeggs: Oh, I see. What is the purpose of this? There is a good reason I;m sure, but this should be stated in the documentation (unless it is and I missed it) because its not obvious to a contributor/user.

src/intersections/raytriangle.jl
kylebeggs: Now that I think about it more, can we replace 'Touching" with 'Origin' here? I feel like 'Origin' is more explicit and 'Touching' is more vague.

src/intersections/raytriangle.jl
juliohm: I think we just fixed a precision in the source and used it consistently. If at some point we need to change this precision we can by changing it in a single place.

src/intersections/raytriangle.jl
juliohm: We already using Touching for other types and it follows nicely with the idea of what is happening besides the geometry that is being touched. In the above convention we have a geometry followed by a verb always.

src/Meshes.jl
```diff
@@ -338,7 +338,11 @@ export
   CrossingRayBox,
   TouchingRayBox,
   IntersectingSegmentTriangle,
-  IntersectingRayTriangle,
+  CrossingRayTriangle,
+  CornerOriginRayTriangle,
```
juliohm: Please rename origin to touching again, CornerOriginRayTriangle doesn't say much about the geometric configuration.

juliohm (changes requested):
Let's just rename Origin to Touching to follow the convention of other intersection types and to be more explicit about what is happening, which is a corner being touched.

src/Meshes.jl
```suggestion
  CornerTouchingRayTriangle,
```

src/Meshes.jl
```diff
@@ -338,7 +338,11 @@ export
...
+  CornerCrossingRayTriangle,
+  EdgeCrossingRayTriangle,
+  EdgeOriginRayTriangle,
```
```suggestion
  EdgeTouchingRayTriangle,
```

src/intersections.jl
```diff
@@ -93,7 +93,11 @@ Type `IntersectionType` in a Julia session to see the full list.
   IntersectingSegmentTriangle
 
   # ray-triangle intersection
-  IntersectingRayTriangle
+  CrossingRayTriangle
+  CornerOriginRayTriangle
```
```suggestion
  CornerTouchingRayTriangle
```

src/intersections.jl
```diff
@@ -93,7 +93,11 @@ Type `IntersectionType` in a Julia session to see the full list.
...
+  CornerCrossingRayTriangle
+  EdgeCrossingRayTriangle
+  EdgeOriginRayTriangle
```
```suggestion
  EdgeTouchingRayTriangle
```

src/intersections/raytriangle.jl
```diff
@@ -5,6 +5,13 @@
 #=
 Möller, T. & Trumbore, B., 1997.
 (https://www.tandfonline.com/doi/abs/10.1080/10867651.1997.10487468)
+
+Cases
+1. CrossingRayTriangle - middle of ray intersects middle of triangle
+2. CornerOriginRayTriangle - origin of ray intersects corner of triangle
```
juliohm:
```suggestion
2. CornerTouchingRayTriangle - origin of ray intersects corner of triangle
```

src/intersections/raytriangle.jl
```diff
@@ -5,6 +5,13 @@
...
+3. EdgeOriginRayTriangle - origin of ray intersects edge of triangle
```
juliohm:
```suggestion
3. EdgeTouchingRayTriangle - origin of ray intersects edge of triangle
```

src/intersections/raytriangle.jl
```diff
@@ -49,7 +56,20 @@ function intersection(f, r::Ray{3,T}, t::Triangle{3,T}) where {T}
     return @IT NoIntersection nothing f
   end
 
+  if any(isapprox.(o, vs, atol=atol(T)))
+    return @IT CornerOriginRayTriangle r(λ) f
```
juliohm:
```suggestion
    return @IT CornerTouchingRayTriangle r(λ) f
```

src/intersections/raytriangle.jl
```diff
@@ -49,7 +56,20 @@ function intersection(f, r::Ray{3,T}, t::Triangle{3,T}) where {T}
...
+  elseif isapprox(λ, zero(T), atol=atol(T))
+    return @IT EdgeOriginRayTriangle r(λ) f
```
juliohm:
```suggestion
    return @IT EdgeTouchingRayTriangle r(λ) f
```

test/intersections.jl
```diff
@@ -753,21 +753,36 @@
...
     @test r ∩ t ≈ P3(0.5, 0.5, 0.5)
 
     # parallel, offset from t, no intersection
     r = Ray(P3(0.0, 0.5, 1.0), V3(1.0, 0.0, 0.0))
     @test intersection(r, t) |> type == NoIntersection
     @test isnothing(r ∩ t)
+
+    # origin of ray intersects with vertex of triangle
+    r = Ray(P3(0.0, 0.0, 0.0), V3(0.0, 0.0, -1.0))
+    @test intersection(r, t) |> type == CornerOriginRayTriangle
```
```suggestion
    @test intersection(r, t) |> type == CornerTouchingRayTriangle
```

test/intersections.jl
```diff
@@ -753,21 +753,36 @@
...
+    @test r ∩ t ≈ P3(0.0, 0.0, 0.0)
+
+    # origin of ray intersects with edge of triangle
+    r = Ray(P3(0.5, 0.0, 0.0), V3(0.0, 0.0, -1.0))
+    @test intersection(r, t) |> type == EdgeOriginRayTriangle
```
```suggestion
    @test intersection(r, t) |> type == EdgeTouchingRayTriangle
```

juliohm (changes requested):

src/intersections.jl
```suggestion
  CornerTouchingRayTriangle
```

src/intersections/raytriangle.jl
```diff
@@ -49,7 +56,20 @@ function intersection(f, r::Ray{3,T}, t::Triangle{3,T}) where {T}
...
+  elseif isapprox(λ, zero(T), atol=atol(T))
+    return @IT EdgeTouchingRayTriangle r(λ) f
+  end
+
+  coords = Vec(u, v, det - u - v)
+  if count(x -> isapprox(x, zero(T), atol=atol(T)), coords) == 1
+    return @IT EdgeCrossingRayTriangle r(λ) f
+  elseif count(x -> isapprox(x, det, atol=atol(T)), coords) == 1
+    return @IT CornerCrossingRayTriangle r(λ) f
+  end
```
juliohm:
```suggestion
  w = Vec(u, v, det - u - v)
  if count(x -> isapprox(x, zero(T), atol=atol(T)), w) == 1
    return @IT EdgeCrossingRayTriangle r(λ) f
  elseif count(x -> isapprox(x, det, atol=atol(T)), w) == 1
    return @IT CornerCrossingRayTriangle r(λ) f
  end
```

juliohm (approved):

kylebeggs:
🎉 where is the merge button?!

juliohm:
Merged it! 🎉 Thanks for the improvements! I think we can finish #293 now 👌🏽

kylebeggs:
Oh so you have to do it? Thanks! yep, will turn to #293 now.

PR #378, opened 2023-03-04
formatter!

kylebeggs:
Here is an initial example. I think the best route to go is simply using another style guide such as the 3 main ones supported ([second line in this link](https://domluna.github.io/JuliaFormatter.jl/stable/)). I personally like the Blue style and like indent of 4, but seeing as the one thing that seemed consistent in this package was indent=2 spaces so I kept that. Take a browse and see what you don't like and we can further modify the toml!


juliohm:
It is tricky to modify so many files in one go, but we can certainly review our current style and can see if it is possible to customize JuliaFormatter to match it. For example, we don't use the `return` keyword everywhere, we would like to export more than a single symbol per line, we don't need to use ; to separate kwargs, ...

kylebeggs:
Yeah, it is certainly tricky to change so many files at once - we can approach this on a PR basis meaning just moving forward only the files that were changed in a PR gets updated to the formatting rules we come up with here. Eventually all the files would be updated. We can just format all the files to see an example of how it works and once we settle on all the format rules, we can just only commit the toml file.

juliohm:
That sounds like a good idea. Thanks for working on it.

Em sáb., 4 de mar. de 2023 12:29, Kyle Beggs ***@***.***>
escreveu:

PR #379, opened 2023-03-04
Add .JuliaFormatter.toml

### LINKED ISSUE #376: Add a formatting style
Some of the source code has formatting errors / the style is not consistent. We should add a `.JuliaFormatter.toml` to the repo with the settings we want and require contributors run that before merging a PR.

kylebeggs:
Starting with a clean PR from #378, without comitting the format changes in the files. Whoever is interested can run this on their local machine without comitting to see the changes it would make.

Fix #376

### DIFF
--- a/.JuliaFormatter.toml
+++ b/.JuliaFormatter.toml
@@ -1,0 +1,10 @@
+style = "sciml"
+indent = 2
+margin = 92
+always_use_return = false
+short_to_long_function_def = false
+annotate_untyped_fields_with_any = false
+whitespace_typedefs = false
+whitespace_ops_in_indices = false
+whitespace_in_kwargs = false
+separate_kwargs_with_semicolon = false
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@
 .DS_Store
 Manifest.toml
 .vscode
+*.code-workspace
 docs/build
 /.benchmarkci
 /benchmark/*.json


.JuliaFormatter.toml
```diff
@@ -0,0 +1,10 @@
+style = "sciml"
```
juliohm, line 1: should we rename this to meshes?

juliohm: @kylebeggs are these styles fixed in the template or the name can vary freely?

kylebeggs: This grabs the pre-built style named "sciml". You can create your own style with your own name, but I am not familiar with how to do it. I just read in the docs that it is possible.

juliohm: We should try to create our own style before merging then 👍🏽

kylebeggs: Ok, we should label this to solicit for help because I'm not sure how to do that and I don't have the time right now.

hyrodium: There are three available styles in JuliaFormatter.jl: [YAS](https://domluna.github.io/JuliaFormatter.jl/dev/yas_style/), [Blue](https://domluna.github.io/JuliaFormatter.jl/dev/blue_style/) and [SciML](https://domluna.github.io/JuliaFormatter.jl/dev/sciml_style/).

If we want our own named style, we need to create a repository that describes our style (e.g. [Blue style](https://github.com/invenia/BlueStyle)) and a PR to JuliaFormatter.jl to support our style. I think this is too hard and not what we would like to have in this PR.

We can just use the default or some style, and modify it by adding configs like `indent = 2`.

juliohm: Thank you @hyrodium, we will take a look at the existing styles and see if we can simply adopt one and adjust a few options.

juliohm:
Overseeded by #416

PR #381, opened 2023-03-07
add TouchingRayTriangle

kylebeggs:
Adds a case where the origin intersects the middle of a triangle to form a complete set of ray-triangle intersections as there are 3 possible things to intersects on the the triangle (vertex, edge, or middle) and 2 possible things to intersects on the ray (origin, ray itself / crossing) for a total of 6 combinations which we now have.

### DIFF
--- a/src/intersections/raytriangle.jl
+++ b/src/intersections/raytriangle.jl
@@ -8,10 +8,11 @@ Möller, T. & Trumbore, B., 1997.
 
 Cases
 1. CrossingRayTriangle - middle of ray intersects middle of triangle
-2. CornerTouchingRayTriangle - origin of ray intersects corner of triangle
-3. EdgeTouchingRayTriangle - origin of ray intersects edge of triangle
-4. EdgeCrossingRayTriangle - middle of ray intersects edge of triangle
-5. CornerCrossingRayTriangle - middle of ray intersects corner of triangle
+2. TouchingRayTriangle - origin of ray intersects middle of triangle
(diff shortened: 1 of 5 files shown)


src/intersections/raytriangle.jl
```diff
@@ -56,13 +57,19 @@ function intersection(f, r::Ray{3,T}, t::Triangle{3,T}) where {T}
...
   end
 
+  # assemble barycentric weights
+  w = Vec(u, v, det - u - v)
+
   if any(isapprox.(o, vs, atol=atol(T)))
     return @IT CornerTouchingRayTriangle r(λ) f
   elseif isapprox(λ, zero(T), atol=atol(T))
-    return @IT EdgeTouchingRayTriangle r(λ) f
+    if all(w .> zero(T))
```
juliohm:
```suggestion
    if all(>(0), w)
```

src/intersections/raytriangle.jl
```diff
@@ -56,13 +57,19 @@ function intersection(f, r::Ray{3,T}, t::Triangle{3,T}) where {T}
...
   end
 
+  # assemble barycentric weights
+  w = Vec(u, v, det - u - v)
+
   if any(isapprox.(o, vs, atol=atol(T)))
     return @IT CornerTouchingRayTriangle r(λ) f
   elseif isapprox(λ, zero(T), atol=atol(T))
-    return @IT EdgeTouchingRayTriangle r(λ) f
+    if all(>(0), w)
```
juliohm:
```suggestion
    if all(>(zero(T)), w)
```

PR #382, opened 2023-03-08
add sideof method for point-mesh

kylebeggs:


src/utils.jl
```diff
@@ -77,6 +77,44 @@ function sideof(p::Point{2,T}, c::Chain{2,T}) where {T}
   ifelse(isapprox(w, zero(T), atol=atol(T)), :OUTSIDE, :INSIDE)
 end
 
+"""
+    sideof(point, mesh)
+Determines whether a `point` is inside, outside the surface of a `mesh`. returns :INSIDE, :OUTSIDE, or :ON. Uses a ray-casting algorithm.
```
```suggestion

Determines whether a `point` is inside, outside or on the surface of a `mesh`.
Possible results are `:INSIDE`, `:OUTSIDE`, or `:ON`.

## Notes

Uses a ray-casting algorithm.
```

src/utils.jl
```diff
@@ -77,6 +77,44 @@ function sideof(p::Point{2,T}, c::Chain{2,T}) where {T}
...
+"""
+function sideof(point::Point{3,T}, mesh::Mesh{3,T}) where {T}
+  if !(eltype(mesh) <: Triangle)
+    throw(ArgumentError(
+          "This function only works for surface meshes with triangles as elements."))
+  end
+  z = last.(coordinates.(extrema(mesh)))
+  r = Ray(point, Vec(zero(T), zero(T), 2 * (z[2] - z[1])))
+  vs = vertices(mesh)
+  touching_types = (EdgeTouchingRayTriangle, CornerTouchingRayTriangle, TouchingRayTriangle)
```
```suggestion
```

src/utils.jl
```diff
@@ -77,6 +77,44 @@ function sideof(p::Point{2,T}, c::Chain{2,T}) where {T}
...
 
+"""
+    sideof(point, mesh)
+Determines whether a `point` is inside, outside the surface of a `mesh`. returns :INSIDE, :OUTSIDE, or :ON. Uses a ray-casting algorithm.
+"""
+function sideof(point::Point{3,T}, mesh::Mesh{3,T}) where {T}
+  if !(eltype(mesh) <: Triangle)
+    throw(ArgumentError(
+          "This function only works for surface meshes with triangles as elements."))
+  end
```
```suggestion
  end
  
```

src/utils.jl
```diff
@@ -77,6 +77,44 @@ function sideof(p::Point{2,T}, c::Chain{2,T}) where {T}
...
+  touching_types = (EdgeTouchingRayTriangle, CornerTouchingRayTriangle, TouchingRayTriangle)
+
+  intersects = false
+  edge_crosses = 0
+  corners_crossing = Point{3,T}[]
+  for elem in mesh
+    I = intersection(r, elem)
+    if type(I) == CrossingRayTriangle
+      intersects = !intersects
+    elseif type(I) ∈ touching_types
```
```suggestion
    elseif type(I) ∈ [EdgeTouchingRayTriangle, CornerTouchingRayTriangle, TouchingRayTriangle]
```

src/utils.jl
```diff
@@ -77,6 +77,44 @@ function sideof(p::Point{2,T}, c::Chain{2,T}) where {T}
...
+          "This function only works for surface meshes with triangles as elements."))
+  end
+  z = last.(coordinates.(extrema(mesh)))
+  r = Ray(point, Vec(zero(T), zero(T), 2 * (z[2] - z[1])))
+  vs = vertices(mesh)
+  touching_types = (EdgeTouchingRayTriangle, CornerTouchingRayTriangle, TouchingRayTriangle)
+
+  intersects = false
+  edge_crosses = 0
+  corners_crossing = Point{3,T}[]
```
juliohm: please avoid underscores in variables names

src/utils.jl
```diff
@@ -77,6 +77,44 @@ function sideof(p::Point{2,T}, c::Chain{2,T}) where {T}
...
+    elseif type(I) == EdgeCrossingRayTriangle
+      edge_crosses += 1
+    elseif type(I) == CornerCrossingRayTriangle
+      id = findfirst(p -> p ≈ I.geom, vs)
+      if vs[id] ∉ corners_crossing
+          push!(corners_crossing, vs[id])
+          intersects = !intersects
+      end
+    end
+  end
```
juliohm, line 117:
```suggestion
  end
  
```

test/utils.jl
```diff
@@ -48,4 +48,45 @@
...
+    @test sideof(P3(0.1, 0.1, 0.0), mesh) == :ON
+
+    points = P3[(0, 0, 0), (1, 0, 0), (0, 1, 0), (0, 0, 1)]
+    mesh = SimpleMesh(points, connec)
+    # ray collinear with edge
+    @test sideof(P3(0.0, 0.0, 0.1), mesh) == :INSIDE
+    @test sideof(P3(0.0, 0.0, -0.1), mesh) == :OUTSIDE
+
+    # construct test mesh with tetrahedron for which this function should error as it only
+    #   tests for triangle meshes
```
```suggestion
    # sideof only defined for triangle meshes
```

test/utils.jl
```diff
@@ -48,4 +48,45 @@
   @test uvrotation(V2(1,0), V2(0,1)) ≈ T[0 -1; 1 0]
   @test uvrotation(V2(0,1), V2(1,0)) ≈ T[0 1; -1 0]
   @test uvrotation(V3(1,0,0), V3(0,1,0)) ≈ T[0 1 0; -1 0 0; 0 0 1]
+
+  @testset "Point in mesh 3D" begin
```
juliohm: no need to create a testset at this moment. we can move all the sideof implementations to a separate file later.

juliohm (changes requested):
Thanks @kylebeggs , review attached.

src/utils.jl
kylebeggs: I think we should we use a tuple - () instead of array - [] here so the compiler knows it is a fixed length. Perhaps it doesn't make a difference in this specific case, but I think its better to use an immutable container when you can.

src/utils.jl
kylebeggs: What is the style for variable names? This discussion motivates the need for #379 !

test/utils.jl
kylebeggs: so just remove testset but keep in this file? sure.

src/utils.jl
```diff
@@ -77,6 +77,51 @@ function sideof(p::Point{2,T}, c::Chain{2,T}) where {T}
   ifelse(isapprox(w, zero(T), atol=atol(T)), :OUTSIDE, :INSIDE)
 end
 
+"""
+    sideof(point, mesh)
+
+Determines whether a `point` is inside, outside or on the surface of a `mesh`.
+Possible results are `:INSIDE`, `:OUTSIDE`, or `:ON`.
+
+## Notes
```
```suggestion
### Notes
```

src/utils.jl
```diff
@@ -77,6 +77,51 @@ function sideof(p::Point{2,T}, c::Chain{2,T}) where {T}
...
+  for elem in mesh
+    I = intersection(r, elem)
+    if type(I) == CrossingRayTriangle
+      intersects = !intersects
+    elseif type(I) ∈ (EdgeTouchingRayTriangle, CornerTouchingRayTriangle, TouchingRayTriangle)
+      return :ON
+    elseif type(I) == EdgeCrossingRayTriangle
+      edgecrosses += 1
+    elseif type(I) == CornerCrossingRayTriangle
+      id = findfirst(p -> p ≈ I.geom, vs)
```
juliohm: Please use the public API instead of accessing fields directly like I.geom

src/utils.jl
```diff
@@ -77,6 +77,51 @@ function sideof(p::Point{2,T}, c::Chain{2,T}) where {T}
...
+  vs = vertices(mesh)
+
+  intersects = false
+  edgecrosses = 0
+  cornerscrossing = Point{3,T}[]
+  for elem in mesh
+    I = intersection(r, elem)
+    if type(I) == CrossingRayTriangle
+      intersects = !intersects
+    elseif type(I) ∈ (EdgeTouchingRayTriangle, CornerTouchingRayTriangle, TouchingRayTriangle)
```
juliohm: Please break this tuple into multiple lines, align on the first letter.

src/utils.jl
```diff
@@ -77,6 +77,51 @@ function sideof(p::Point{2,T}, c::Chain{2,T}) where {T}
...
+
+  z = last.(coordinates.(extrema(mesh)))
+  r = Ray(point, Vec(zero(T), zero(T), 2 * (z[2] - z[1])))
+  vs = vertices(mesh)
+
+  intersects = false
+  edgecrosses = 0
+  cornerscrossing = Point{3,T}[]
+  for elem in mesh
+    I = intersection(r, elem)
```
juliohm: Should we use the 3-arg version of intersection to throw away other branches at compile time? Check the docs of intersection for examples.

src/utils.jl
```diff
@@ -77,6 +77,51 @@ function sideof(p::Point{2,T}, c::Chain{2,T}) where {T}
...
+      if vs[id] ∉ cornerscrossing
+          push!(cornerscrossing, vs[id])
+          intersects = !intersects
+      end
+    end
+  end
+
+  # check how many edges we crossed
+  isodd(round(Int, edgecrosses / 2)) && (intersects = !intersects)
+  intersects == true ? (return :INSIDE) : (return :OUTSIDE)
```
```suggestion
  intersects ? :INSIDE : :OUTSIDE
```

src/utils.jl
kylebeggs: I don't think it would be able to since the compiler wouldn't know the type of the intersection until we actually run the intersection so I think both methods would be equivalent performance wise. I will update it for style reasons, however.

kylebeggs: I think the 3-arg version is only beneficial in performance when you have a mesh with different types of elements (like triangles and quads, etc) but since we limit this to triangles only (for now) then it shouldn't make a difference.

juliohm: I don't think this is related to multiple element types. The 3-arg version simply sends the relevant branches that are explicitly written to a new function barrier. So it can generate typed code instead of Any returns.

src/utils.jl
```diff
@@ -77,6 +77,55 @@ function sideof(p::Point{2,T}, c::Chain{2,T}) where {T}
...
+
+  z = last.(coordinates.(extrema(mesh)))
+  r = Ray(point, Vec(zero(T), zero(T), 2 * (z[2] - z[1])))
+  vs = vertices(mesh)
+
+  intersects = false
+  edgecrosses = 0
+  cornerscrossing = Point{3,T}[]
+  for elem in mesh
+    result = intersection(r, elem) do I
```
```suggestion
  for e in mesh
    result = intersection(r, e) do I
```

src/utils.jl
```diff
@@ -77,6 +77,55 @@ function sideof(p::Point{2,T}, c::Chain{2,T}) where {T}
...
+
+  intersects = false
+  edgecrosses = 0
+  cornerscrossing = Point{3,T}[]
+  for e in mesh
+    result = intersection(r, e) do I
+      if type(I) == CrossingRayTriangle
+        intersects = !intersects
+      elseif type(I) ∈ (EdgeTouchingRayTriangle, CornerTouchingRayTriangle,
+                        TouchingRayTriangle)
```
```suggestion
      elseif type(I) ∈ (EdgeTouchingRayTriangle,
                        CornerTouchingRayTriangle,
                        TouchingRayTriangle)
```

src/utils.jl
```diff
@@ -77,6 +77,56 @@ function sideof(p::Point{2,T}, c::Chain{2,T}) where {T}
...
+      elseif type(I) == CornerCrossingRayTriangle
+        id = findfirst(p -> p ≈ get(I), vs)
+        if vs[id] ∉ cornerscrossing
+            push!(cornerscrossing, vs[id])
+            intersects = !intersects
+        end
+      end
+      return nothing
+    end
+    result !== nothing && (return :ON)
```
```suggestion
    !isnothing(result && return :ON
```

src/utils.jl
```diff
@@ -77,6 +77,56 @@ function sideof(p::Point{2,T}, c::Chain{2,T}) where {T}
...
+      elseif type(I) == CornerCrossingRayTriangle
+        id = findfirst(p -> p ≈ get(I), vs)
+        if vs[id] ∉ cornerscrossing
+            push!(cornerscrossing, vs[id])
+            intersects = !intersects
+        end
+      end
+      return nothing
+    end
+    !isnothing(result && return :ON
```
```suggestion
    !isnothing(result) && return :ON
```

src/utils.jl
```diff
@@ -77,6 +77,56 @@ function sideof(p::Point{2,T}, c::Chain{2,T}) where {T}
...
+        id = findfirst(p -> p ≈ get(I), vs)
+        if vs[id] ∉ cornerscrossing
+            push!(cornerscrossing, vs[id])
+            intersects = !intersects
+        end
+      end
+      return nothing
+    end
+    !isnothing(result) && return :ON
+  end
```
```suggestion
  v = vertices(mesh)

  intersects = false
  edgecrosses = 0
  cornerscrossing = Point{3,T}[]
  for e in mesh
    result = intersection(r, e) do I
      if type(I) == CrossingRayTriangle
        intersects = !intersects
      elseif type(I) ∈ (EdgeTouchingRayTriangle,
                        CornerTouchingRayTriangle,
                        TouchingRayTriangle)
        return :ON
      elseif type(I) == EdgeCrossingRayTriangle
        edgecrosses += 1
      elseif type(I) == CornerCrossingRayTriangle
        id = findfirst(p -> p ≈ get(I), v)
        if v[id] ∉ cornerscrossing
            push!(cornerscrossing, v[id])
            intersects = !intersects
        end
      end
      return nothing
    end
    !isnothing(result) && return :ON
  end
```

src/utils.jl
```diff
@@ -77,6 +77,56 @@ function sideof(p::Point{2,T}, c::Chain{2,T}) where {T}
...
+
+  # check how many edges we crossed
+  isodd(round(Int, edgecrosses / 2)) && (intersects = !intersects)
+  intersects ? (return :INSIDE) : (return :OUTSIDE)
```
juliohm: can you write the English-equivalent of this algorithm? I feel that it could be simplified, it is currently not straightforward to digest. The variable names mixing single-letter variables with long-words variables also makes it difficult to follow. Maybe cornerscrossing should be a Set? You are constantly checking if the point belongs to the list and then adding it. Maybe the lines with `intersects = !intersects` should be more explicit with `true` and `false` instead?

kylebeggs: Sure, probably should have explained in English in the description of this PR anyways. The algorithm really is easy in the basic idea, it is the 'tricky' cases that make it confusing/difficult to handle. We just need to check how many intersections there are of a ray with the surface - if odd, the point is inside, if even, outside. This is why we can't set intersects to true or false because we don't keep track of how many intersections and are just flipping back and forth like even and odd. In our case, we also want to check if the point is on the surface, but that part is easy to see (lines 107-110).

The tricky parts are when the ray hits an edge or a corner of a triangle of the surface mesh.
1. If it hits an edge - we have to consider that an edge is shared by 2 triangles so when we reach the other triangle in the loop over all elements that intersection will get counted again. To handle this, I am just counting the total number of edge intersects (lines 111-112) and then divide by 2 to get the actual amount of intersections of edges (line 126). Now if this number is even, it "cancels" out but if odd, we count it as another intersection in the global tracking and "flip" the switch of `intersects` as if the ray simply intersected the middle of a triangle (line 126).
2. if the ray hit a corner of a triangle - this is the part where there is probably a better way to do this. Since the number of triangles that share a corner is not always the same like an edge, you can't just divide by that number at the end. There are 2 main ways I thought of to go about this:

- what I did here: when I do come across a ray hitting a corner, we check if that point has already been "hit" (if we are currently considering an element that shares the corner we already came across) (line 115) and if not, then we add it to the list of corners we have intersected (line 116) and flip the global tracker of intersects (line 117).
- each time we hit a corner, we check to see how many elements it shares the corner with via coboundary (lets call that N) and we can add N-1 to a sum (called A), and also add 1 to a sum counting how many corners (including duplicates from shared elemnts) we have hit (call it C) so that we can find how many unique corners we hit via C-A. add again if that number is odd then we flip the global `intersects`.

Example:

Number of shared elements for each corner:
c1 = 3
c2 = 4
c3 = 6

Example cases
hit all: 13-10=3 is odd -> flip global intersects switch
hit c1and c2: 7-5=2 is even -> do nothing (it "cancels")
hit c1 only: 3-2=1 is odd -> flip global intersects switch

I don't know the complexity of finding the number of shared elements via coboundary, so it could be better I don;t know. But my latest commit gets rid of the useless `findfirst` and instead compares approximately to the point the ray-triangle returns which should be the same as the vertex in the mesh up to numerical precision for that type. We can benchmark and go through all that, but I think it is pointless, because this algorithm is meant to be **robust**, not **performant**. if you really care about performance, there are much faster algorithms, but they are not as robust (they are approximate). For what its worth, I have implemented this approximate type of approach for my personal use case and I find it works well enough for me (but I am using medium-ish size surface meshes ~30k elements) and this algorithms accuracy scales with number of elements (not good for a box or something like that).

kylebeggs: See "Fast winding numbers for soups and clouds" - Gavin Barill, 2018. I implemented a naive (not as performant or accurate) approach of this myself. I am looking to implement this, but it is not completely foolproof like the algorithm we have here. ideally we have both and the user can choose what is more important to them.

src/utils.jl
```diff
@@ -77,6 +77,55 @@ function sideof(p::Point{2,T}, c::Chain{2,T}) where {T}
...
+        intersects = !intersects
+      elseif type(I) ∈ (EdgeTouchingRayTriangle,
+                        CornerTouchingRayTriangle,
+                        TouchingRayTriangle)
+        return :ON
+      elseif type(I) == EdgeCrossingRayTriangle
+        edgecrosses += 1
+      elseif type(I) == CornerCrossingRayTriangle
+        p = get(I)
+        if !any(isapprox.(p, cornerscrossing, atol=atol(T)))
```
juliohm: We can use `!any(==(p), cornerscrossing)` here because our Point == is implemented in terms of isapprox already. I would also rename `cornerscrossing` to `points` or simply `ps` to be more succinct with variable names.

src/utils.jl
```diff
@@ -77,6 +77,55 @@ function sideof(p::Point{2,T}, c::Chain{2,T}) where {T}
...
+  end
+
+  z = last.(coordinates.(extrema(mesh)))
+  r = Ray(point, Vec(zero(T), zero(T), 2 * (z[2] - z[1])))
+
+  intersects = false
+  edgecrosses = 0
+  cornerscrossing = Point{3,T}[]
+  for e in mesh
+    result = intersection(r, e) do I
```
juliohm: I don't think this do block returns something meaningful? Should we just erase the `result` variable? The logic of this block is still a bit confusing. It does a bunch of checks and then interrupts the loop with explicit `return` statements, but why do we need to save the result if nothing is actually stored as an intersection?

src/utils.jl
kylebeggs: I am usually a fan of longer, more dewscriptive names as it helps a future developer understand what that variable is for, but I don't think its a huge deal here so I'll change it.

src/utils.jl
kylebeggs: Perhaps you can help me here - its the only way I knew to make it work. The check if the point is `:ON` the surface (lines 106-109) should just return the final results immediately, so I don't know how to break out of the do block other than returning the results in the do block and then checking if the result from the do block is `:ON` or `nothing` and if `!isnothing` then return `:ON` (line 121).

src/utils.jl
```diff
@@ -77,6 +77,55 @@ function sideof(p::Point{2,T}, c::Chain{2,T}) where {T}
...
+      elseif type(I) ∈ (EdgeTouchingRayTriangle,
+                        CornerTouchingRayTriangle,
+                        TouchingRayTriangle)
+        return :ON
+      elseif type(I) == EdgeCrossingRayTriangle
+        edgecrosses += 1
+      elseif type(I) == CornerCrossingRayTriangle
+        p = get(I)
+        if !any(==(p), ps)
+            push!(ps, p)
```
juliohm: @kylebeggs should ps be a Set? This pattern of checking if the point is in the list and then pushing it to the list should be natural with the Set data structure, right? You can always push to the Set and it will contain unique members all the time.

src/utils.jl
```diff
@@ -77,6 +77,55 @@ function sideof(p::Point{2,T}, c::Chain{2,T}) where {T}
...
+  end
+
+  z = last.(coordinates.(extrema(mesh)))
+  r = Ray(point, Vec(zero(T), zero(T), 2 * (z[2] - z[1])))
+
+  intersects = false
+  edgecrosses = 0
+  ps = Point{3,T}[]
+  for e in mesh
+    result = intersection(r, e) do I
```
```suggestion
  for t in mesh
    result = intersection(r, t) do I
```

src/utils.jl
```diff
@@ -77,6 +77,55 @@ function sideof(p::Point{2,T}, c::Chain{2,T}) where {T}
...
+  edgecrosses = 0
+  ps = Point{3,T}[]
+  for t in mesh
+    result = intersection(r, t) do I
+      if type(I) == CrossingRayTriangle
+        intersects = !intersects
+      elseif type(I) ∈ (EdgeTouchingRayTriangle,
+                        CornerTouchingRayTriangle,
+                        TouchingRayTriangle)
+        return :ON
```
juliohm: what is the effect of the return statement here, is it breaking the do block or the entire for loop?

src/utils.jl
```diff
@@ -77,6 +77,55 @@ function sideof(p::Point{2,T}, c::Chain{2,T}) where {T}
...
+            intersects = !intersects
+        end
+      end
+      return nothing
+    end
+    !isnothing(result) && return :ON
+  end
+
+  # check how many edges we crossed
+  isodd(round(Int, edgecrosses / 2)) && (intersects = !intersects)
```
juliohm: Do you want `edgecrosses \div 2` here by any chance? Or you really need to round for correctness?

src/utils.jl
kylebeggs: Yes that makes sense.

src/utils.jl
kylebeggs: Ideally, I would want to break the entire loop, but I don't know how to do that. It onbly breaks out of the do block here. thats why we need to return the do block to a value and check on line 121.

src/utils.jl
kylebeggs: Yes I forgot about \div.

src/utils.jl
kylebeggs: For some reason `Set` gives the wrong answer here
```
Point(0.25, 0.25, 0.9999999999999999)
Set{Point3} with 1 element:
  Point(0.25, 0.25, 1.0)
p ∉ ps = true
```
 but with a vector we get the correct result:
```
Point(0.25, 0.25, 0.9999999999999999)
1-element Vector{Point3}:
 Point(0.25, 0.25, 1.0)
p ∉ ps = false
```

Any ideas? I thought both of a `Set` and a `Vector` should compare with the same tolerance? This example probably demonstrates we should just use `isapprox` anyways.

juliohm: Maybe this is due to the difference between == and isequal. I think Set uses isequal, and we probably need to implement it for Point as well.

src/utils.jl
juliohm: Do we actuallly gain type stability here using the 3-arg version of the intersection function as originally proposed? If we don't get type stability, I think we can just revert back to the 2-arg version to make the code more readable?

kylebeggs: The 2 arg is type stable. The reason for using the 3 arg is compiler optimizations as you mentioned before. Given my comments that this algorithm's advantage is robustness and we should implement an approximate algorithm for the users that need speed, I think the 2 arg is fine to use here.

src/utils.jl
kylebeggs: == and isequal just treat NaN, missing, and 0.0 differently. I don't think it should matter in the example above? i know its not ideal, but we can use the vector for now?

juliohm: Sounds good, we just need to replace == by \isapprox. See https://github.com/JuliaGeometry/Meshes.jl/pull/384

We will refactor some basic concepts in the near future.

src/utils.jl
juliohm: Yes, I think it is wise to stick to the 2-arg version here, at least in this first implementation. We can then come back to it later after it is merged and propose small improvements.

juliohm:
This is a very useful feature @kylebeggs , thanks again for carefully taking care of the review comments. We will use this feature in a project soon, looking forward to merging it 💯

src/utils.jl
kylebeggs: I typically prefer that development style anyways (avoiding premature optimizations).

src/utils.jl
```diff
@@ -77,6 +77,52 @@ function sideof(p::Point{2,T}, c::Chain{2,T}) where {T}
...
+
+### Notes
+
+Uses a ray-casting algorithm.
+"""
+function sideof(point::Point{3,T}, mesh::Mesh{3,T}) where {T}
+  if !(eltype(mesh) <: Triangle)
+    throw(ArgumentError(
+          "This function only works for surface meshes with triangles as elements."))
+  end
```
juliohm: Last question. In the case we have a polygon mesh with quadrangles or other types of polygons we can simply call `simplexify(mesh)` to obtain a triangle mesh before running the algorithm. So maybe we should check that the parametric dimension of the mesh is 2 first, and if it is only made of triangles we proceed, otherwise we simplexify and continue. What do you think?

```julia
@assert paramdim(mesh) == 2 "sideof only defined for surface meshes"

(eltype(mesh) <: Triangle) || return sideof(point, simplexify(mesh))
```

kylebeggs: Yeah that looks good!

test/utils.jl
```diff
@@ -48,4 +48,48 @@
...
+  @test sideof(P3(0.0, 0.0, 0.1), mesh) == :INSIDE
+  @test sideof(P3(0.0, 0.0, -0.1), mesh) == :OUTSIDE
+
+  # sideof for meshes that have elements > 3-gons.
+  points = P3[(0, 0, 0), (1, 0, 0), (0, 1, 0), (0.25, 0.25, 1), (1, 1, 0)]
+  connec = connect.([(1, 2, 4), (1, 4, 3), (2, 3, 4), (1, 2, 5, 3)])
+  mesh = SimpleMesh(points, connec)
+  @test sideof(P3(0.25, 0.25, 0.1), mesh) == :INSIDE
+
+  # sideof only defined for triangle meshes
```
```suggestion
  # sideof only defined for surface meshes
```

juliohm (approved):

juliohm:
Awesome addition @kylebeggs 💯 🎉

PR #392, opened 2023-03-31
Breaking Change: Don't use `isapprox` in `==`

eliascarv:


src/utils.jl
```diff
@@ -92,7 +92,7 @@ function sideof(point::Point{3,T}, mesh::Mesh{3,T}) where {T}
   (eltype(mesh) <: Triangle) || return sideof(point, simplexify(mesh))
 
   z = last.(coordinates.(extrema(mesh)))
-  r = Ray(point, Vec(zero(T), zero(T), 2 * (z[2] - z[1])))
+  r = Ray(point, Vec{3,T}(zero(T), zero(T), 2 * (z[2] - z[1])))
```
juliohm, line 95: @eliascarv this one here is necessary? I thought that Julia would be able to infer the types from the other expressions.

juliohm (approved):

src/utils.jl
eliascarv: True, it was an unnecessary change.
Could you remove this change on the master branch?

Issue #135, opened 2021-05-25
Fix angle between 3D vectors


juliohm:
We need to fix the current implementation to return the correct oriented angle.

Feel free to tackle the issue if you have time available.

lmejn:
I'd be happy to have a go at this, but I can't seem to reproduce an incorrect result. Are you referring to the `∠` function? https://github.com/JuliaGeometry/Meshes.jl/blob/f78377cc83cb3dc3976aee425faaf8031de3a6ee/src/angles.jl#L25-L28
Would you able to provide an example of when the angle calculation is incorrect?

juliohm:
Hi @lmejn , thanks for checking the issue. If I remember correctly we would like to have this angle function behaving like it behaves in 2D. That is, we wanted it to return the angle in a specific orientation, but I guess it depends on arbitrary choices of reference. In 2D when you type `∠(A, B, C)` that is different than `∠(C, B, A)`. Maybe we should forget about this behavior in 3D and always return the smallest angle as we currently do?

lmejn:
Ah I see. Yes, this would depend on the arbitrary choice of reference: the angle from `∠(A, B, C)` could be measured about either normal vectors, `BA × BC` or `BC × BA` (which of course equals `-(BA × BC)`).

[One way to get a signed angle](https://stackoverflow.com/a/33920320) would be to give a reference unit vector (`vref`):
```julia
function ∠(A::P, B::P, C::P, vref::Vec) where {P<:Point{3}}
  BA, BC = A - B, C - B
  atan((BA × BC) ⋅ vref, BA ⋅ BC)
end
```
To get the angle between vectors, `vref` should be the vector normal to `BA`, `BC`, but you can decide which normal.

I can't think of a case where this would be useful though. I would be inclined to forget about this behaviour, unless there is an actual use case for this kind of computation.

juliohm:
You are correct @lmejn , I think this behavior is not motivated by any algorithm we've implemented so far. I will close the issue and we can re-open if it becomes relevant.

PR #396, opened 2023-04-13
Angle

JeffreySarnoff:


src/angles.jl
```diff
@@ -26,14 +26,44 @@ See https://en.wikipedia.org/wiki/Atan2.
...
+@inline unitize(u::V) where {V<:Vec} = u.coords ./ norm2(u)
+@inline unitize(u::NTuple) = u ./ norm2(u)
+
+@inline isnegangle(u::V, v::V) where {V<:Vec3} = false
+@inline isnegangle(u::V, v::V) where {V<:Vec2} =
+    u.coords[1] * v.coords[2] < u.coords[2] * v.coords[1]
+
+@inline isnegangle(u::NTuple{3,T}, v::NTuple{3,T}) where {T} = false
+@inline isnegangle(u::NTuple{2,T}, v::NTuple{2,T}) where {T} =
+    u.coords[1] * v.coords[2] < u.coords[2] * v.coords[1]
```
juliohm: Can we get rid of this method with NTuple?

src/angles.jl
```diff
@@ -26,14 +26,44 @@ See https://en.wikipedia.org/wiki/Atan2.
...
+
+@inline norm2(u::V) where {V<:Vec} = sqrt(foldl(+, abs2.(u.coords)))
+@inline norm2(u::NTuple) = sqrt(foldl(+, abs2.(u)))
+
+@inline unitize(u::V) where {V<:Vec} = u.coords ./ norm2(u)
+@inline unitize(u::NTuple) = u ./ norm2(u)
+
+@inline isnegangle(u::V, v::V) where {V<:Vec3} = false
+@inline isnegangle(u::V, v::V) where {V<:Vec2} =
+    u.coords[1] * v.coords[2] < u.coords[2] * v.coords[1]
```
juliohm: We try to never access internal fields of a struct, instead we stick to the public api of the Vec type. In this case our Vec is a custom StaticVector, it behaves like a normal Julia vector.

src/angles.jl
```diff
@@ -26,14 +26,44 @@ See https://en.wikipedia.org/wiki/Atan2.
...
+Vec2 angles return a value in range [-π, π].
+Vec3 angles return a value in range [0, π].
 
 ## Examples
 
 ```julia
 ∠(Vec(1,0), Vec(0,1)) == π/2
 ```
+
+Thank you Jeffrey Sarnoff for contributing this work.
```
juliohm: Perhaps you have a reference that we can cite in the docstring? We don't usually put the names of the author in the docstring, but a citation to a paper or online material. See for example: https://github.com/JuliaGeometry/Meshes.jl/blob/a0487c6824d6ee9d7389edc25ae937f1e4cf26fd/src/topologies.jl#L10-L14

juliohm (changes requested):
Thank you @JeffreySarnoff for the improvement 💯 Attached a few suggestions.

JeffreySarnoff:
Using the tests on master with the master branch of Meshes.jl:
The 2D vector tests fail and the 3D vector tests fail.
I do not know what to use for testing.

JeffreySarnoff:
Your definitions of angle for 2D and of angle for 3D are unclear to me.
- angle(u::Vec2D, v::Vec2D) is not measuring counterclockwise from u toward v as far a I can tell (it measures clockwise from v toward u for angles that are reported >= 0, and angles that are reported negative seem to rely on the quadrants for each vec)

For 2D, can you provide some simple diagrams that show how you are determining the angle measures.
same for 3D or  a complete if then else note.

Really I cannot approach the other stuff without having resolved this.

src/angles.jl
```diff
@@ -8,32 +8,53 @@
...
+
+    ifelse(isnegangle(uunit, vunit), -a, a)
+end
+
+@inline norm2(u::Vec) = sqrt(foldl(+, abs2.(u)))
+
+@inline unitize(u::Vec) = u ./ norm2(u)
+
+@inline isnegangle(u::Vec{3,T}, v::Vec{3,T}) where {T} = false
+@inline isnegangle(u::Vec{2,T}, v::Vec{2,T}) where {T} = u[1] * v[2] < u[2] * v[1]
```
juliohm:
```suggestion
@inline isnegangle(u::Vec{2}, v::Vec{2}) = u[1] * v[2] < u[2] * v[1]
@inline isnegangle(u::Vec{3}, v::Vec{3}) = false
```

src/angles.jl
```diff
@@ -8,32 +8,53 @@
...
+    vunit = unitize(v)
+
+    y = norm2(uunit .- vunit)
+    x = norm2(uunit .+ vunit)
+
+    a = 2 * atan(y, x)
+
+    a = !(signbit(a) || signbit(pi - a)) ? a : (signbit(a) ? zero(T) : (T)(pi))
+
+    ifelse(isnegangle(uunit, vunit), -a, a)
```
juliohm:
```suggestion
  T = eltype(u)

  uunit = unitize(u)
  vunit = unitize(v)

  y = norm2(uunit .- vunit)
  x = norm2(uunit .+ vunit)

  a = 2 * atan(y, x)

  a = !(signbit(a) || signbit(pi - a)) ? a : (signbit(a) ? zero(T) : (T)(pi))

  ifelse(isnegangle(uunit, vunit), -a, a)
```

juliohm:
If I remember correctly, our 2D angle is the smallest angle from u to v. Maybe that is the difference in the definition you have in mind? These angle definitions are very tricky to communicate. Maybe we need to have more explicit names for these angle functions after we draw them in a diagram.

Do you have suggestions on how we should name the various angles formed by two vectors?

src/angles.jl
```diff
@@ -8,32 +8,53 @@
...
+  a = 2 * atan(y, x)
+
+  a = !(signbit(a) || signbit(pi - a)) ? a : (signbit(a) ? zero(T) : (T)(pi))
+
+  ifelse(isnegangle(uunit, vunit), -a, a)
+end
+
+@inline norm2(u::Vec) = sqrt(foldl(+, abs2.(u)))
+
+@inline unitize(u::Vec) = u ./ norm2(u)
```
juliohm, line 57: Given that our `Vec` is a `StaticVector` from StaticArrays.jl I think we can use their implementation of LinearAlgebra.normalize and LinearAlgebra.norm here.

JeffreySarnoff:
> If I remember correctly, our 2D angle is the smallest angle from u to v.
It is not. The angle is from v to u (which may be a mistake).

juliohm:
We probably need to review the angle definitions we have and provide a new set of angle functions to cover the different angles between u and v. You have more experience with angle calculations, feel free to suggest a terminology for all the angles between the vectors and propose a set of functions. We can help reviewing the names and implementations.

juliohm:
@JeffreySarnoff do you have a suggestion for the different angles between u and v? Do you have an example where our current 2D implementation is not producing the result you expected?

JeffreySarnoff:
here is some info on angles in 2D
I am travelling -- will continue in a few days

named angles
Kinds of 2D angles
[Full Circle Angular Measure In 2D.pdf](https://github.com/JuliaGeometry/Meshes.jl/files/11269302/Full.Circle.Angular.Measure.In.2D.pdf)
[Signed Angular Measure In 2D.pdf](https://github.com/JuliaGeometry/Meshes.jl/files/11269304/Signed.Angular.Measure.In.2D.pdf)
[Angular Measure In 2D with atan.pdf](https://github.com/JuliaGeometry/Meshes.jl/files/11269306/Angular.Measure.In.2D.with.atan.pdf)

JeffreySarnoff:
Currently
```
xpos = V2(1,0); ypos = V2(0,1); xneg = V2(-1,0); yneg = V2(0,-1);


julia> ∠(xpos,xneg)
3.1415927                # ok
julia> ∠(xneg,xpos)
-3.1415927              # ok
julia> ∠(ypos,yneg)
-3.1415927              # incorrect, should be >0, should ==  ∠(xpos,xneg)
julia> ∠(yneg,ypos)
3.1415927              # incorrect, should be <0, should ==  ∠(xneg,xpos)
```

juliohm:
Thank you for pointing out this issue with the ypos/yneg case. It should indeed give similar results to the xpos/xneg case.

juliohm:
@JeffreySarnoff , here is a sketch of some angle definitions:

null

The scenario (II) is currently implemented in Meshes.jl for 2D angles. We simply define the angle as the smallest angle in magnitude between the vectors and then place the sign depending on the orientation (CCW vs CW). The scenario (I) on the other hand always measures angles in CCW orientation. The scenario you have in mind is any of the two?

JeffreySarnoff:
UVAB

a is angle(u,v) the counterclockwise angle from u to v, a > 0
b is angle(v,u) the clockwise angle from v to u, b < 0, b = -a

let aa = 2pi - a, the counterclockwise angle from v to u, aa > 0
let bb = -(aa) == a - 2pi == -b -2pi , the clockwise angle from u to v, bb < 0

juliohm:
Agree with all definitions. What do we need to do?

Em sáb., 22 de abr. de 2023 17:14, Jeffrey Sarnoff ***@***.***>
escreveu:

JeffreySarnoff:
Where 2D angles are signed, it is much more consistent to have 3D angles be signed rather than unsigned.  In 2D the sign is determined by orientation (counterclockwise is +, clockwise is -).  If you lift 2D (x,y) coordinates to 3D (x,y,z) as (x,y,0), then continuing with the 2D u,v example [above](https://github.com/JuliaGeometry/Meshes.jl/pull/396#issuecomment-1518765556):
the cross product of u->v (angle a) is cross([ux, uy, 0], [vx, vy, 0]) which for our diagram where v is counterclockwise from u is ideally [0, 0, 1].  In the same way cross([vx, vy, 0], [ux, uy, 0]) is ideally [0, 0, -1].  The sign of the z coordinate of the oriented cross product is the sign of the 3D angle.  We can use this for 3D vectors with nonzero z coordinates, and it works by always respecting the right hand rule.

I don't know if there is internal 3D logic that would be arithmetically confused by 3D vectors with sign -- there should not be, anywhere that occur is using flawed or incomplete logic anyway and should be fixed.

juliohm:
If you can double check the 2D implementation in this PR and check what is causing tests to fail, we can merge it first, and then consider a separate PR with the 3D case. These changes are subtle but important to fix, I agree.

mfsch:
I came across the issue of how angles are defined in 3D while investigating #405. I don’t think it makes sense to define a signed angle in 3D the way @JeffreySarnoff is suggesting, as the relative orientation of an object with respect to the z-axis is a pretty arbitrary and would change with a rotation of the coordinate axes, even if the object itself is unchanged. The sign as defined in 2D is a much more fundamental property.

Instead, I think it may be better to revisit the proposal in #135, which adds another argument to the angle function with a reference vector. The sign is then given by the alignment with that reference vector, and choosing (0,0,1) as the reference would recover the behavior that @JeffreySarnoff proposed. This approach also allows for computing the inner angles of a polygon by using the normal vector as reference, as attempted [here](https://github.com/JuliaGeometry/Meshes.jl/commit/5aa10fcbf7db8dbafafd30674ffba34f9c31ab96) and I think it should generalize to higher dimensions. Maybe it can also be generalized “down” to 2D (and perhaps 1D, with an angle of ±π)?

JeffreySarnoff:
> the relative orientation of an object with respect to the z-axis is a pretty arbitrary and would change with a rotation of the coordinate axes

that is true. it is not what I intended.  given two vectors in 3D space that share a common origination point (for simplicity, less simply  that extend to intersect, or more generally have a place of nearest approach), determine a third vector that is orthogonal to both (use the cross product on the unit length versions of the vectors) and shares that common origination point.  For numerical stability when the two original vectors are narrowly separated you may need to spread the two original vectors along their common plane before taking the cross product.

The (positive) z-axis comes into play as the target of the 3D rotation (there are more than one such rotation, a convention is adopted -- e.g. least total angular displacement).  There is a composite rotation or an appropriate unit quaternion which takes the the cross-product derived vector to coincide with the positive z axis.  That rotation carries the original vectors along with it.  In this resituation, the orientation of those vectors (which is the initial vector and which is the vector obtained by carrying the initial vector through their separating angle) gives that separating angle its sign.  This keeps a deep continuity with 2D and 3D, and eventually allows interior and exterior solid angles to play nicely.

juliohm:
@mfsch the idea of adding an extra argument that defaults to (0,0,1) seems very elegant and intuitive. Let's go with that.

@JeffreySarnoff what do you think we should do regarding this PR? Originally we intended to just replace our current implementation with yours for greater numerical stability, but then entered into a design issue about which angles make sense. We can either come back to the original intent and finish the PR preserving the old definitions of angles, or close this PR and start a new one where we change the design first with the additional reference vector, and then reintroduce your implementation that is more robust.

If you guys can coordinate the efforts, I am happy to review the improvements.

JeffreySarnoff:
I do not have the requisite availability.

mfsch:
I just looked through the discussion about the 2D case and from what I understand the issue is only how diametrically opposed vectors are handled, i.e. that the angles are currently in the range [-π,π] but should be in the range (-π,π] or [-π,π) (probably the former). Can you confirm this is the case @JeffreySarnoff? If not, can you provide an example that is not handled correctly for which |α|<π?

For the 3D case, I think we’d have to decide if the reference vector should only specify the sign or whether the angle should be the signed angle in the plane perpendicular to the normal vector, i.e. `atan((u × v) · nref, u ⋅ v)`, as suggested in #135 and as I’ve been using to compute inner angles of polygons. In any case, I think the default should be the unsigned angle in [0,π], at least in 3D. Specifying (0,0,1) as the reference even just for the sign produces pretty weird behavior when u×v lies in the x–y-plane, where slight variations (e.g. floating-point errors) would flip the sign of angles of any size, not just around π.

I’m not an expert in computational geometry though and this seems like a pretty fundamental question that must have been discussed extensively in papers and textbooks, so it would probably be best to find a solid reference before proceeding? My hunch is that angles should by default always be the unsigned smallest angle between two vectors, and that a signed angle between two N-dimensional vectors needs an (N−2)-dimensional reference (e.g. N−2 vectors) to define the orientation of the 2-dimensional plane in which they live and assign a sign to the angle.

JeffreySarnoff:
I will forward a few well-written explanations. Whatever the academic
formulation, it is of great import to use computations that avoid
inaccuracies when vectors are very close (where the smaller angle between
them is quite small) or are nearly opposed.  Never rely on acos alone.
Using the two argument form of atan is best, and becomes even better when
all the vectors are normalized, and becomes better again when all inputs to
atan have themselves been normalized (especially the cross product of two
unit vectors).
The boundary cases of -pi +pi, 0 +2pi need to be checked and clamped as
appropriate to the conventions used, otherwise there will be results of
both e.g. 0 and +2pi which can mess up other computations that rely on
getting values in [0.0 .. 2pi) or in [-pi .. +pi).  Where the range is [0.0
.. 2pi), the signed range should be [-pi .. pi); otherwise you cannot map
one range to the other without a wrinkle and that is a problem to be
avoided.  Basically, as one would not see (0.0 .. 2pi] as a natural choice
(it leaves a hole at 0), the closed-open intervals are preferred and more
robust.

juliohm:
Thank you all for putting energy into this fundamental concept. Looking forward to an improved version that combines your ideas into a general implementation that is also numerically robust 🥇

juliohm:
Closing in favor of #409

Issue #354, opened 2022-12-25
Cleaning and repairing meshes


juliohm:
In #353 some cleaning functions were proposed to eliminate vertices of a mesh that are orphan or to "merge" them somehow. Instead of providing this functionality as isolated functions, we should aim for a more composable transform interface.

I propose that we start listing the possible cleaning operations that we would like to implement in this issue, and then we can open separate PRs for each of them. A good list of cleaning operations is available in the MeshLab project as discussed in our Zulip channel.

- [x] 0 = only duplicated vertices and faces are removed
- [x] 1 = remove unreferenced vertices
- [ ] 2 = remove non-manifold faces
- [ ] 3 = remove degenerate faces
- [ ] 4 = remove non-manifold vertices
- [ ] 5 = split non-manifold vertices by threshold
- [ ] 6 = merge close vertices (given a radius)
- [x] 7 = coherently orient faces

stla:
I believe MeshLab has the same cleaning operations than the R package Rvcg.

I already implemented **0** (I called it `gather(mesh)`) and **1**. I don't know what is a manifold vertex/face. **6** can be obtained by a minor modification of **0**. For **7** I don't know how to do.

juliohm:
> I don't know what is a manifold vertex/face

https://sinestesia.co/blog/tutorials/non-manifold-meshes-and-how-to-fix-them/

Basically, a mesh is said to be manifold when it represents a "simple" surface, i.e. when you look at it locally you can imagine a Cartesian plane with two dimensions. When a vertex shares more than 3 edges or when an edge shares more than 2 faces, it means that we have some sort of feature that doesn't belong to the "surface", e.g. a face that is hanging in a T shape.

stla:
Thanks.
Regarding the names for these functions, in **Rvcg** there's a unique function named `vcgClean`. The wanted operations are given by one or more numbers between 0 and 7. Personally I don't really like.

juliohm:
Yes, I think we can aim for a more composable set of transforms. The naming will be difficult but we can certainly do better.

stla:
Hello,
I have a function which removes the unused vertices of a mesh. How should I name it and where should I put it if I want to do a PR? Currently it is named `removeUnusedVertices`. This function is necessary for the clipping (another function I have).

juliohm:
[...]

PR #399, opened 2023-04-28
Repair{1} transform: remove unused vertices

stla:
As discussed in #354


src/Meshes.jl
```diff
@@ -405,6 +405,7 @@ export
   LambdaMuSmoothing,
   LaplaceSmoothing,
   TaubinSmoothing,
+  Repair,
```
juliohm: Move to after StdCoords

src/transforms.jl
```diff
@@ -98,3 +98,4 @@ include("transforms/translate.jl")
 include("transforms/stretch.jl")
 include("transforms/stdcoords.jl")
 include("transforms/smoothing.jl")
+include("transforms/repair.jl")
```
juliohm, line 100: move to after stdcoords.jl

src/transforms/repair.jl
```diff
@@ -0,0 +1,58 @@
...
+- K = 7: faces are coherently oriented
+"""
+struct Repair{K} <: StatelessGeometricTransform end
+
+# implement operation K = 1
+function apply(transform::Repair{1}, mesh)
+  # get the faces
+  topo = convert(HalfEdgeTopology, topology(mesh))
+  ∂₂₀ = Boundary{2,0}(topo)
+  nfaces = nelements(topo)
```
juliohm:
```suggestion
  nelem = nelements(topo)
```

src/transforms/repair.jl
```diff
@@ -0,0 +1,58 @@
...
+
+  # indices of used vertices will be stored in this vector
+  used = Vector{Int}(undef, 0)
+
+  # initialize dictionary to map old indices to new indices
+  newindices = Dict{Int,Int}()
+
+  # to store the connectivities of the cleaned mesh
+  connectivities = Vector{Connectivity}(undef, 0)
```
juliohm:
```suggestion
  connec = Vector{Connectivity}(undef, 0)
```

src/transforms/repair.jl
```diff
@@ -0,0 +1,58 @@
...
+
+  # l = length(used) will be incremented
+  l = 0
```
juliohm:
```suggestion
  nused = 0
```

src/transforms/repair.jl
```diff
@@ -0,0 +1,58 @@
...
+
+  # iterate over faces
+  for f in 1:nfaces
```
juliohm:
```suggestion
  for e in 1:nelem
```

src/transforms/repair.jl
```diff
@@ -0,0 +1,58 @@
...
+    face = ∂₂₀(f)
+    for index in face
+      if !in(index, used)
```
juliohm: please use the latex \notin insted of !in

src/transforms/repair.jl
```diff
@@ -0,0 +1,58 @@
...
+        l = l + 1
+        newindices[index] = l
+      end
+    end
+    connec = connect(tuple([newindices[i] for i in face]...))
+    push!(connectivities, connec)
+  end
+
+  # output mesh
+  SimpleMesh(vertices(mesh)[used], connectivities)
```
juliohm: create a new variable points = vertices(mesh)[used] before passing it to the SimpleMesh

test/transforms.jl
```diff
@@ -81,4 +81,14 @@
...
+
+  @testset "Repair{1} (unused vertices)" begin
+    # a tetrahedron with an unused vertex
+    verts = P3.([(0, 0, 0), (0, 0, 1), (5, 5, 5), (0, 1, 0), (1, 0, 0)])
+    mesh = SimpleMesh(verts, connect.([(1, 2, 4), (1, 2, 5), (1, 4, 5), (2, 4, 5)], Triangle))
+    rmesh = mesh |> Repair{1}
+    @test nvertices(rmesh) == nvertices(mesh) - 1
+    @test nelements(rmesh) == nelements(mesh)
+    @test topology(rmesh) == topology(mesh)
+  end
```
juliohm: move testset to before Smoothing testset

juliohm (changes requested):
Nice addition. Review attached.

src/transforms/repair.jl
juliohm: also, the apply function returns the result and the cache to revert the operation when that is possible. if that is not possible, we return `nothing`.

test/transforms.jl
```diff
@@ -81,4 +81,14 @@
     @test nelements(smesh) == nelements(mesh)
     @test topology(smesh) == topology(mesh)
   end
+
+  @testset "Repair{1} (unused vertices)" begin
+    # a tetrahedron with an unused vertex
+    verts = P3.([(0, 0, 0), (0, 0, 1), (5, 5, 5), (0, 1, 0), (1, 0, 0)])
+    mesh = SimpleMesh(verts, connect.([(1, 2, 4), (1, 2, 5), (1, 4, 5), (2, 4, 5)], Triangle))
+    rmesh = mesh |> Repair{1}
```
```suggestion
    rmesh = mesh |> Repair{1}()
```

test/transforms.jl
```diff
@@ -61,7 +61,17 @@
     vnew2 = TB.reapply(trans, vset, cache)
     @test vnew == vnew2
   end
-  
+
+  @testset "Repair{1}" begin
+    # a tetrahedron with an unused vertex
+    verts = P3.([(0, 0, 0), (0, 0, 1), (5, 5, 5), (0, 1, 0), (1, 0, 0)])
+    mesh = SimpleMesh(verts, connect.([(1, 2, 4), (1, 2, 5), (1, 4, 5), (2, 4, 5)], Triangle))
```
```suggestion
    mesh  = SimpleMesh(verts, connect.([(1, 2, 4), (1, 2, 5), (1, 4, 5), (2, 4, 5)], Triangle))
```

stla:
Thanks. Looks like there's an error: `nvertices(rmesh)` is equal to 3. That's weird, I tested this function several times and I didn't notice that.

stla:
That makes no sense :-( When I test the example outside the package there are 4 vertices, not 3.

src/transforms/repair.jl
```diff
@@ -0,0 +1,61 @@
...
+  # indices of used vertices will be stored in this vector
+  used = Vector{Int}(undef, 0)
+
+  # initialize dictionary to map old indices to new indices
+  newindices = Dict{Int,Int}()
+
+  # to store the connectivities of the cleaned mesh
+  connectivities = Vector{Connectivity}(undef, 0)
+
+  # iterate over faces
```
juliohm:
```suggestion
  # iterate over elements
```

stla:
Did you change something on my branch? My push has been rejected.

juliohm:
Yes, I've been updating the PR here as you can see in the history above. Please try to pull the changes first before committing again or press . on GitHub so that it opens the editor in the PR version online.

stla:
Ok. In fact `rmesh` is  not a mesh. `topology(rmesh)` fails as well as `nelements(rmesh)`.

What is `rmesh, nothing`?

juliohm:
The apply function returns the result of the transform and a cache to
revert the transform. In this case the cache is nothing because the revert
is not implemented.

Em sex., 28 de abr. de 2023 11:49, stla ***@***.***> escreveu:

stla:
Ok. It seems to work now. But `topology(rmesh) == topology(mesh)` is not true because the vertex indices have been adjusted. I'm going to push.

src/transforms/repair.jl
```diff
@@ -0,0 +1,61 @@
...
+
+# implement operation K = 1
+function apply(transform::Repair{1}, mesh)
+  # get the faces
+  topo = convert(HalfEdgeTopology, topology(mesh))
+  ∂₂₀ = Boundary{2,0}(topo)
+  nelem = nelements(topo)
+
+  # indices of used vertices will be stored in this vector
+  used = Vector{Int}(undef, 0)
```
juliohm: replace this by a Set{Int} instead, which as fast checks for `in`

src/transforms/repair.jl
```diff
@@ -0,0 +1,61 @@
...
+  for e in 1:nelem
+    elem = ∂₂₀(e)
+    for i in elem
+      if i ∉ used
+        push!(used, i)
+        nused += 1
+        newindices[i] = nused
+      end
+    end
+    c = connect(tuple([newindices[i] for i in elem]...))
```
juliohm: what happens when the dictionary doesn't have the index i?

juliohm: also, use ntuple(i->newindices[i], elem) instead to create the tuple statically

src/transforms/repair.jl
```diff
@@ -0,0 +1,61 @@
...
+  # get the faces
+  topo = convert(HalfEdgeTopology, topology(mesh))
+  ∂₂₀ = Boundary{2,0}(topo)
+  nelem = nelements(topo)
+
+  # indices of used vertices will be stored in this vector
+  used = Vector{Int}(undef, 0)
+
+  # initialize dictionary to map old indices to new indices
+  newindices = Dict{Int,Int}()
```
juliohm: rename this variable to simply inds

test/transforms.jl
```diff
@@ -61,7 +61,16 @@
     vnew2 = TB.reapply(trans, vset, cache)
     @test vnew == vnew2
   end
-  
+
+  @testset "Repair{1}" begin
+    # a tetrahedron with an unused vertex
+    verts = P3.([(0, 0, 0), (0, 0, 1), (5, 5, 5), (0, 1, 0), (1, 0, 0)])
```
```suggestion
    verts = P3[(0, 0, 0), (0, 0, 1), (5, 5, 5), (0, 1, 0), (1, 0, 0)]
```

test/transforms.jl
```diff
@@ -61,7 +61,16 @@
...
+    mesh  = SimpleMesh(verts, connect.([(1, 2, 4), (1, 2, 5), (1, 4, 5), (2, 4, 5)], Triangle))
```
juliohm: move this connect argument to a variable connec before creating the mesh

juliohm (changes requested):
A few more changes needed before we can merge.

juliohm:
Also, please add the new Repair transform to the list of transforms in the docs after the changes above are made.

src/transforms/repair.jl
stla: I believe it's not possible that the dictionary doesn't have the index i (it's late and I can't seriously think about that right now).

Your code `ntuple(i->newindices[i], elem)` does not work.

src/transforms/repair.jl
stla: We cannot do `vertices(mesh)[used]` if `used` is a set.

juliohm (approved):

PR #400, opened 2023-04-28
Remove Vec constructor for AbstractVector

### LINKED ISSUE #398: Should we avoid Vec constructor for AbstractVector?
This is related to JuliaEarth/GeoTables.jl#13, but in general it seems that is not a good idea to have non typestable contructor like the one defined for `Vec` over `AbstracVector`:
https://github.com/JuliaGeometry/Meshes.jl/blob/cb556d34900c64b6f40ac827360039c0fb3549f3/src/vectors.jl#L66.

 For example, `StaticArrays.jl` avoids it (https://github.com/JuliaArrays/StaticArrays.jl/blob/ed92217f94e2d70847bcf2b7716f47a410d52d25/docs/src/pages/api.md?plain=1#L237-L239). 

> We have avoided adding `SVector(v::AbstractVector)` as a valid constructor to
help users avoid the type instability (and potential performance disaster, if
used without care) of this innocuous looking expression.

@juliohm Should we avoid using the constructor for `AbstractVector` or limit it for few dimensions?

ErickChacon:
Fixes #398 and it is related to JuliaEarth/GeoTables.jl#13. We decided to remove `Vec` constructor over `AbstractVector` because it is not type-stable. The same approach is taken by `StaticArrays.jl`. 

It is just a initial PR to show that we can get rid of the constructor for `AbstractVector`, some improvements need to be done.

### DIFF
--- a/src/intersections/lines.jl
+++ b/src/intersections/lines.jl
@@ -63,7 +63,8 @@ function intersectparameters(a::Point{Dim,T}, b::Point{Dim,T},
 
   # calculate parameters of intersection or closest point
   if r ≥ 2
-    λ = A \ y
+    # λ = A \ y
+    λ = A \ SVector(y.coords)
   else # parallel or collinear
     λ = SVector(zero(T), zero(T))
   end
--- a/src/intersections/lines.jl
+++ b/src/intersections/lines.jl
@@ -48,7 +48,7 @@ calculated in order to identify the intersection type:
 function intersectparameters(a::Point{Dim,T}, b::Point{Dim,T}, 
                              c::Point{Dim,T}, d::Point{Dim,T}) where {Dim,T}
   A = [(b - a) (c - d)]
-  y = SVector(c - a)
+  y = c - a
 
   # calculate the rank of the augmented matrix by checking
   # the zero entries of the diagonal of R
--- a/src/intersections/lines.jl
+++ b/src/intersections/lines.jl
@@ -48,7 +48,7 @@ calculated in order to identify the intersection type:
 function intersectparameters(a::Point{Dim,T}, b::Point{Dim,T}, 
                              c::Point{Dim,T}, d::Point{Dim,T}) where {Dim,T}
   A = [(b - a) (c - d)]
-  y = c - a
+  y = SVector(c - a)
 
   # calculate the rank of the augmented matrix by checking
   # the zero entries of the diagonal of R
@@ -63,7 +63,7 @@ function intersectparameters(a::Point{Dim,T}, b::Point{Dim,T},
 
   # calculate parameters of intersection or closest point
   if r ≥ 2
-    λ = A \ SVector(y.coords)
+    λ = A \ y
   else # parallel or collinear
     λ = SVector(zero(T), zero(T))
   end
--- a/src/intersections/lines.jl
+++ b/src/intersections/lines.jl
@@ -63,7 +63,6 @@ function intersectparameters(a::Point{Dim,T}, b::Point{Dim,T},
 
   # calculate parameters of intersection or closest point
   if r ≥ 2
-    # λ = A \ y
     λ = A \ SVector(y.coords)
   else # parallel or collinear
     λ = SVector(zero(T), zero(T))
--- a/test/polytopes.jl
+++ b/test/polytopes.jl
@@ -424,7 +424,8 @@
       0.0  0.2  0.2  0.4  0.4  0.6  0.6  0.8  0.8  0.6  0.4  0.2  0.0  1.0  1.0  0.0
       0.0  0.2  0.4  0.4  0.2  0.2  0.4  0.4  0.2  0.2  0.2  0.2  0.0  0.0  1.0  1.0
     ]
-    @test vertices(chain) == Point.(eachcol(target))
+    # @test vertices(chain) == Point.(eachcol(target))
+    @test vertices(chain) == Point.(NTuple{2, T}.(eachcol(target)))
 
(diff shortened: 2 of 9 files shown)


src/intersections/lines.jl
```diff
@@ -63,7 +63,7 @@ function intersectparameters(a::Point{Dim,T}, b::Point{Dim,T},
 
   # calculate parameters of intersection or closest point
   if r ≥ 2
-    λ = A \ y
+    λ = A \ SVector(y.coords)
```
juliohm: This one we thought it would be a good idea to convert both A and y in the beginning of the function right?

test/polytopes.jl
```diff
@@ -424,7 +424,7 @@
       0.0  0.2  0.2  0.4  0.4  0.6  0.6  0.8  0.8  0.6  0.4  0.2  0.0  1.0  1.0  0.0
       0.0  0.2  0.4  0.4  0.2  0.2  0.4  0.4  0.2  0.2  0.2  0.2  0.0  0.0  1.0  1.0
     ]
-    @test vertices(chain) == Point.(eachcol(target))
+    @test vertices(chain) == Point.(NTuple{2, T}.(eachcol(target)))
```
```suggestion
    @test vertices(chain) == Point.(Tuple.(eachcol(target)))
```

src/intersections/lines.jl
ErickChacon: `A` is already a `SMatrix`, and I am now tranforming `y` from the beginning. Check the updated version please.

src/intersections/lines.jl
```diff
@@ -48,7 +48,7 @@ calculated in order to identify the intersection type:
 function intersectparameters(a::Point{Dim,T}, b::Point{Dim,T}, 
                              c::Point{Dim,T}, d::Point{Dim,T}) where {Dim,T}
   A = [(b - a) (c - d)]
-  y = c - a
+  y = SVector(c - a)
```
juliohm: @ErickChacon can you comment again why this is necessary? StaticArrays.jl forces a conversion before A \ y? Would that be something to be fixed in StaticArrays.jl in the future?

juliohm (approved):

src/intersections/lines.jl
ErickChacon: @juliohm, it is related to this line (https://github.com/JuliaArrays/StaticArrays.jl/blob/b4e49ce11fc88fc826b6353312f97e7101473509/src/solve.jl#L73).

```julia
similar_type(b, T, $outsize)(Matrix(a) \ b)
```

1. `a` gets converted to `Matrix`
2. solve operation is done using LinearAlgebra.jl
3.  finally, it try to convert the output (`Vector`) to the type of `b` (Vec).

So, step 3 does not work because we do not provide the method to convert an `AbstractVector` to `Vec`.

juliohm: And it works with vanilla `SVector`? I understood that `SVector` doesn't provide the method to convert `AbstractVector` to `SVector` either?

juliohm:
We also need to update the docstring to remove the constructor.

src/intersections/lines.jl
ErickChacon: They do not have non type-stable convertor, but they have a convertor that for a given `Type` and `size`. The function `similar_type` makes that conversion, but note that you need to provide `type` and `size` which they can know in advance because of matrix operations. We also have the constructor for a given Type and size, but not the non-type stable convertor.

For this reason, we still have our constructor (
https://github.com/ErickChacon/Meshes.jl/blob/db16ac4f9959a8802a59d8aa0d1e9066d3cf39dd/src/vectors.jl#L56-L59) that is used in 
https://github.com/ErickChacon/Meshes.jl/blob/db16ac4f9959a8802a59d8aa0d1e9066d3cf39dd/src/intersections.jl#L267.

juliohm: I don't know if I followed the difference between our custom `Vec` and their `SVector`. Are we missing some method implementation regarding this `similar_type` that would make the code work without the explicit conversion to `SVector`?

ErickChacon: Maybe we can use something simirlar to `similar_type`, I am not sure. I will have to check how that function work in `StaticArrays.jl`.

ErickChacon: There is some information here: https://github.com/JuliaArrays/StaticArrays.jl/blob/b4e49ce11fc88fc826b6353312f97e7101473509/docs/src/pages/api.md?plain=1#L99-L121

juliohm: @eliascarv since you were the one who originally looked into it, do you have an idea of what we would need to implement to support the A \ y without converting y to SVector explicitly?

ErickChacon: @juliohm I think the issue is fixed. It is actually not necessarily at all to modify this function. Let me explain:

1. Initially, I used to get the error for this function when I removed all `AbstractVector`  constructors meaning that there was no constructor like `Vec{Dim, T}(coords:AbstractVector)`, so I solved the problem using `y = SVector(c-a).
2. In other tests, I notice that it was necessary to use `Vec{Dim, T}(coords:AbstractVector)`, so I decided to put it back.
3. I did not notice that defining `Vec{Dim, T}(coords:AbstractVector)` helped `StaticArrays.jl` to work normally without the need of `SVector`. It is used in `similar_type(b, T, $outsize)(Matrix(a) \ b)`.

Hence we do not need to change this function at all.

juliohm:
Thank you @ErickChacon , amazing contribution 🥇

PR #401, opened 2023-04-28
another convex hull / GrahamScan function added

ig-or:
faster convex hull function added. I tested it an it works fast and reliable for my case in 10 millions  hulls.    
what is not perfect:

- not using existing Meshes.jl data structures like `Point ` etc. 
- Probably not work correctly with corner cases (like making a hull from just two points)

### DIFF
--- a/src/hulls/graham.jl
+++ b/src/hulls/graham.jl
@@ -62,3 +62,84 @@ function hull(pset::PointSet{2,T}, ::GrahamScan) where {T}
 
   PolyArea(c)
 end
+
+
+"""
+  hull_simple(p, h, s, v)
+
+calculate the convex hull for the matrix of points.  This is actually an implementation of the pseudocode from  [https://en.wikipedia.org/wiki/Graham_scan]  (https://en.wikipedia.org/wiki/Graham_scan).
+
+This is a fast and convinient method when point coordinates are in a matrix, and you have to create 10000000 hulls in a second.  
+
+parameters:   
+ * p is the matrix [y x w]  of the points; where 'y' is the point dimension and w is the number of points.  
+ * h is the preallocated vector of the indices of the points in p [w]
+ * s is the preallocated vector of the indices of the points in p, which makes a hull   [w]
+ * v preallocated vector of Folat64 [w]
+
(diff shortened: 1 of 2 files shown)


juliohm:
@ig-or appreciate the attempt to add a faster implementation here. Can you please consider improving our existing Graham scan implementation instead?

The low level function you implemented could be refactored and then used behind a more high-level interface with Vector{Point} etc.

DanielVandH:
This implementation seems to fail for this (contrived) example:

```julia
n = 4
p1 = [27.643564356435643, -21.881188118811881]
p2 = [83.366336633663366, 15.544554455445542]
p3 = [4.0, 4.0]
p4 = [73.415841584158414, 8.8613861386138595]
p = hcat(p1, p2, p3, p4)

h = zeros(Int32, n)
v = zeros(Float64, n)
s = zero(h)
u = hull_simple(p, h, s, v)
s = s[1:u]
push!(s, s[begin])

fig = Figure()
ax = Axis(fig[1, 1])
scatter!(ax, p[1, :], p[2, :], color=:red)
lines!(ax, p[1, s], p[2, s])

fig
```
image

Meshes.jl's existing implementation works, though.

```julia
_p = PointSet(p)
viz(hull(_p, GrahamScan()))
```

image

juliohm:
Thank you for constructing the example @DanielVandH .

@ig-or I will close the PR because it is not fitting our interface, and because it has the bug mentioned above. I am happy to review future PRs with improvements to our hull implementation.

ig-or:
Thanks for looking at my code. I fixed this issue with several points on one line on my side. But as I said make this fitting your interface might be more challenging (but possible). The main idea of the increase of speed is to remove all the memory allocations from the code. 

I realized my use case is not common. I'm creating millions of convex hulls from just a few points. Usually people have millions of points and need make one convex hull out of it.

DanielVandH:
FYI @ig-or, you might be interested to test on some of the cases discussed in this paper https://doi.org/10.1016/j.comgeo.2007.06.003 (it's where my example from above comes from) for convex hulls. They are all for convex hulls of very small point sets.

ig-or:
Thanks Daniel!

PR #404, opened 2023-05-01
Add Prism geometry

mfsch:
As mentioned on Zulip, I have implemented some algorithms to compute the signed distance field of a mesh on a regular grid. This code currently lives in a [separate repository](https://github.com/mfsch/MeshRasterization.jl), but some (all?) of the functionality might make more sense directly in Meshes.jl.

Based on the response on Zulip, I now submitting a first part of this code as a PR. This code contains a `Prism` type that defines a geometry with a polygonal base and a vector that defines the offset to the other end of the prism. I tried to make it a bit more general than in my rasterization package and filled in some gaps in how N-gons in 3D space are handled.

 A few points perhaps worth discussing:

- The type only stores the vertices of one side and uses a displacement vector to represent the rest of the geometry. I thought it’s better to only represent a canonical form of the geometry internally, but perhaps this can be an issue for some applications?
- The `vertices` function returns a `Vec` with the full set of points. I would like to keep that function allocation-free, but maybe it would make sense to switch to a regular array if N gets too large?
- The `normal`/`innerangles` computation would probably make more sense to implement for `Chain` instead of `Ngon`, but as things are currently set up, creating the `Chain` with `boundary` does a number of allocations. I would much prefer it if computing `normal` and `isconvex` could be done without allocation. To have `boundary(ngon)` without allocation would probably require changing `Chain` such that a closed chain can be represented with a `CircularVector` internally – perhaps this is worth considering?
- I think I currently don’t really need the `edge_connectivity` and `facet_connectivity` functions, so the former could be removed and the latter wrapped into the `boundary(prism)` function. However, I think it may be useful to allow building connectivities directly from polytopes rather than forcing the detour through a `SimpleMesh`. The naming is probably not ideal and there may be some conceptual muddling between the geometry and its boundary, but it would be useful to have a way to iterate over parts of a geometry (e.g. the sides of a polytope) without building up data structures that require allocations.
- It generally wasn’t that clear to me how generators vs. arrays are handled in Meshes.jl. When implementing algorithms that loop over the elements of a mesh, I would like to avoid allocations that happen for each element, as they really drag down performance in my experience. However, generators can be a bit more annoying to use and I guess it would make sense for functions like `innerangles` to consistently return either an array or a generator.

Let me know any changes you want me to make. I can also move the normals/angles stuff to a separate PR, and I also just saw there’s another PR open discussing angles between vectors, but from what I saw it doesn’t address the issue with `innerangles` in 3D.


src/Meshes.jl
```diff
@@ -159,6 +159,7 @@ export
   Octagon, Nonagon, Decagon,
   Chain, PolyArea,
   Tetrahedron, Pyramid, Hexahedron,
+  Prism,
```
```suggestion
  Tetrahedron, Hexahedron,
  Pyramid, Prism,
```

juliohm (changes requested):
Thank you @mfsch for this awesome first PR. I suggest that we break it down into smaller PRs changing different parts of the codebase. For example, if we need to improve the orientation and innerangles functions, then we need a separate PR just for that, and then a new PR for the Prism struct with methods.

Our contribution guidelines request that each PR is self-contained and does only one thing. This is to ensure that we can revert things in the future if necessary.

juliohm:
> * The type only stores the vertices of one side and uses a displacement vector to represent the rest of the geometry. I thought it’s better to only represent a canonical form of the geometry internally, but perhaps this can be an issue for some applications?

At first glance, this is a good design decision. I thought the same that we should represent the base with a `<:Polygon` and the displacement with a `Vec`.

> * The `vertices` function returns a `Vec` with the full set of points. I would like to keep that function allocation-free, but maybe it would make sense to switch to a regular array if N gets too large?

This function is widespread across the codebase and always returns a dynamically-allocated vector of `Point`. Our `Vec` type represents a geometric vector in R^n (think of an arrow), it is not a Julia container.

If you need allocation-free implementations somewhere, it is easy to do so inside the function. We don't want to change the return of `vertices` globally to be statically-allocated.

> * The `normal`/`innerangles` computation would probably make more sense to implement for `Chain` instead of `Ngon`, but as things are currently set up, creating the `Chain` with `boundary` does a number of allocations. I would much prefer it if computing `normal` and `isconvex` could be done without allocation. To have `boundary(ngon)` without allocation would probably require changing `Chain` such that a closed chain can be represented with a `CircularVector` internally – perhaps this is worth considering?

We can handle these allocation-free requirements inside the function of interest (e.g., fancy algorithm consuming a `Prism`). Our `Chain` currently makes use of `CircularVector` in the vertices, but there is probably place for improvement to avoid more allocations here and there. Maybe this bullet point deserves an issue and a proposal? We can address this issue first before adding the `Prism`. 

> * I think I currently don’t really need the `edge_connectivity` and `facet_connectivity` functions, so the former could be removed and the latter wrapped into the `boundary(prism)` function. However, I think it may be useful to allow building connectivities directly from polytopes rather than forcing the detour through a `SimpleMesh`. The naming is probably not ideal and there may be some conceptual muddling between the geometry and its boundary, but it would be useful to have a way to iterate over parts of a geometry (e.g. the sides of a polytope) without building up data structures that require allocations.

It is nice that you care a lot about allocations. We can certainly improve the codebase with your feedback. Regarding the separation between `Geometry` and `Mesh`, we do that on purpose to make the code more organized. For us, geometries like Sphere, Ball, Cylinder are memory-light representations of geometries that you can discretize at some point into a mesh if needed. In the case of a Box for example, we only need to store the two corners, but if someone needs to loop over the faces, we need to materialize the boundary and operate on the resulting Mesh object. This design cleans up a lot of dispatch rules, and has incredible flexibility so far.

> * It generally wasn’t that clear to me how generators vs. arrays are handled in Meshes.jl. When implementing algorithms that loop over the elements of a mesh, I would like to avoid allocations that happen for each element, as they really drag down performance in my experience. However, generators can be a bit more annoying to use and I guess it would make sense for functions like `innerangles` to consistently return either an array or a generator.

We try to stick to one of the two always for specific functions. For example, `vertices` is one function that I think always returns a concrete vector, but I am not 100% sure. We certainly need to review if some of these functions are returning varying output types, and fix that on separate PRs.

> Let me know any changes you want me to make. I can also move the normals/angles stuff to a separate PR, and I also just saw there’s another PR open discussing angles between vectors, but from what I saw it doesn’t address the issue with `innerangles` in 3D.

It would be awesome if you could start smaller PRs tackling very specific issues. We could then discuss in more depth the pros and cons of different designs together. Regarding angles, take a look at #396 as well where we are brainstorming what our current angle implementations should return conceptually.

mfsch:
Thanks for your feedback! I’ve converted this PR into a draft while we address some of the smaller items with separate PRs. How do you handle dependent PRs in general? Do you create a PR for the first feature and only create a PR for the second feature once the first PR is merged, or do you already submit a second PR that contains both features and keep rebasing that one?

For the next step, I can try to prepare a PR that would change Chain such that a chain with `V<:CircularVector` is a valid representation of a closed chain. This should then make it possible to define `boundary(ngon)` in a non-allocating way and `normal` & `innerangles` can be implemented for Chain without making them expensive when called for Ngon.

I’ve also added a separate proposal #406 that would avoid the need to define a non-allocating `vertices` function.

Issue #406, opened 2023-05-01
Define a function to get a single vertex of a polytope


mfsch:
At the moment, most Polytopes are represented internally with a list of vertices. However, in some cases it may make sense to choose a different internal representation to avoid storing redundant vertices and to make it harder to violate assumed invariants of a geometry. Concrete examples are the existing Box and the Prism geometry proposed in #404. For those, the `vertices` function is no longer a trivial computation that simply returns the internal representation, and generally has to allocate a vector.

I think it would make sense to introduce a function `vertex(geo::Geometry, ind::Int)` to get the i-th vertex of a geometry. This can be defined as `vertices(geo)[ind]` by default, but it allows geometries such as Box and Prism to define `vertex` to dynamically compute the i-th vertex and have `vertices(geo) = collect(vertex(geo, ind) for ind in 1:nvertices(geo))`. The naming `vertex` would be consistent with a similar `facet(m, ind)` function defined for meshes.

This also makes it possible to iterate over the vertices of a geometry without allocating a vector without having to change the return type of `vertices` to a generator or statically-allocated array for geometries like Box & Prism.

juliohm:
I like the idea @mfsch , it generalizes well what we currently have with faces 👍🏽

juliohm:
@mfsch I've closed the last PR because of our recent improvements with the Box type. Feel free to submit another PR with the `vertex` function if you feel it is still needed to avoid allocations, or feel free to close this issue if that is not the case anymore.

mfsch:
@juliohm, the main reason for this suggestion was the Prism type, not the Box. In my use of the Prism type, I need a way to access the vertices in an efficient manner, without building and allocating a new array of vertices each time. I think the `vertex` function is the cleanest way of enabling this without changing the return type of `vertices`.

juliohm:
Makes total sense @mfsch , let's start a clean PR with the new vertex function, this time without touching on Box as it doesn't have vertices anymore.

mfsch:
I’ve updated [the corresponding branch](https://github.com/mfsch/Meshes.jl/tree/vertex-function). Do you just want to reopen the previous PR? I think that should update it to the current version of the `mfsch:vertex-function` branch.

juliohm:
Apparently GitHub doesn't allow to reopen the PR in this case. Can you submit a new one?

mfsch:
Addressed in #414.

PR #408, opened 2023-05-02
Define function to get specific vertex of polytope

mfsch:
As discussed in #406, this defines a new function to get a specific vertex of a polytope. It also implements the `vertices` function of the Box in terms of this new function, generalizing it to N dimensions.

A few open questions:
- Maybe it is better to leave the implementation of `vertices(polytope)` as it was? The new implementation would allow new polytopes to automatically get the implementation of `vertices` if they define `vertex` and `nvertices` and should not have any runtime overhead since the conditional can be resolved at compile-time, but has circular dependencies if those two methods are not defined.
- For the Box, I think it would be cleaner to define the vertex order to consistently switch between the lower and upper boundary along the first dimension, i.e. remove the conditional at lines 61–63 (e.g. [(0,0),(1,0),(0,1),(1,1)] in 2D). This would require a small change in the `boundary` function. However, I’m not sure if there is any other code that relies on the order of the box vertices, so maybe it’s better to keep the order as it is?
- Is there a reason that the Box is classified as a primitive instead of a polytope?

### DIFF
--- a/src/polytopes.jl
+++ b/src/polytopes.jl
@@ -39,11 +39,24 @@ Return the parametric dimension or rank of the polytope.
 paramdim(::Type{<:Polytope{K}}) where {K} = K
 
 """
+    vertex(polytope, ind)
+
+Return the vertex of a `polytope` at index `ind`.
+"""
+vertex(p::Polytope, ind) = vertices(p)[ind]
+
+"""
     vertices(polytope)
 
 Return the vertices of a `polytope`.
 """
-vertices(p::Polytope) = p.vertices
+function vertices(p::Polytope)
+  if hasfield(typeof(p), :vertices)
+    p.vertices
+  else
+    collect(vertex(p, ind) for ind in 1:nvertices(p))
+  end
+end
 
 """
     nvertices(polytope)
--- a/src/primitives/box.jl
+++ b/src/primitives/box.jl
@@ -47,32 +47,23 @@ diagonal(b::Box) = norm(b.max - b.min)
 
 sides(b::Box) = Tuple(b.max - b.min)
 
-vertices(b::Box{1}) = [b.min, b.max]
+nvertices(b::Box{Dim}) where Dim = 2^Dim
 
-function vertices(b::Box{2})
-  A = coordinates(b.min)
-  B = coordinates(b.max)
-  Point.([
-    (A[1], A[2]),
-    (B[1], A[2]),
-    (B[1], B[2]),
-    (A[1], B[2]),
-  ])
-end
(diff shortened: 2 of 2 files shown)


src/polytopes.jl
```diff
@@ -38,12 +38,25 @@ Return the parametric dimension or rank of the polytope.
...
 Return the vertices of a `polytope`.
 """
-vertices(p::Polytope) = p.vertices
+function vertices(p::Polytope)
+  if hasfield(typeof(p), :vertices)
+    p.vertices
+  else
+    collect(vertex(p, ind) for ind in 1:nvertices(p))
+  end
+end
```
juliohm, line 59:
```suggestion
vertices(p::Polytope) = [vertex(p, ind) for ind in 1:nvertices(p)]
```

juliohm:
> * Is there a reason that the Box is classified as a primitive instead of a polytope?

I love it how you are spotting inconsistencies in our codebase ❤️ I realized that all our `Primitive` types do not define the `vertices` function, except for `Box`, which is an indication that maybe it should be a `Polytope` instead. Originally, the `Box` was considered a `Primitive` because of its "lazy" representation where we could save memory and build the vertices on the fly, but I am starting to think that it could simply be a `Polytope` for greater consistency.

We already have "boxes" that are more flexible like `Segment`, `Quadrangle` and `Hexahedron` for 1D, 2D, and 3D, respectively. They are more flexible in the sense that they do not need to be axis-aligned nor they need to have right angles.

Can we go one step back and investigate the possibility of `Box` as a `Polytope`? If you could draft a PR where this change is made, we could assess the impact in algorithms downstream. We could then come back to the issue of the new `vertex` function proposed in this present PR.

In summary:

1. Move `Box <: Primitive` to `Box <: Polytope` and see if all tests pass.
2. Define the new `vertex` function for `Polytope` as its done in this PR.
3. Investigate the order of vertices in `Box` in a third PR to see if tests pass.

Sorry for requesting so many different PRs, it is because you're touching on very fundamental design issues that we didn't realize, and so I think it is important to handle them separately before continuing with the changes in these opened PRs 🙏🏽

juliohm:
I am performing some investigation with the Box type to see if it is worth simply eliminating the `vertices` implementation for it. Will let you know when the investigation is concluded.

juliohm:
@mfsch I removed the `vertices` implementation of `Box`, it is not internal to the `boundary` definitions. Can you please rebase this PR on top of the latest master branch?

juliohm:
@mfsch I will close this PR given that it is changing Box and we refactored it in the master branch. Let's see if there is still the need for a `vertex` function or if it was just a design issue with Box that motivated its creation.

Issue #409, opened 2023-05-03
Angle definitions


juliohm:
Moving the discussion from Zulip, where we want to address a few issues regarding our current angle implementations:

1. Improving the numerical robustness of the implementation.
2. Defining the open/closed ends of the interval in which angles are returned
3. Defining how signed & unsigned angles are handled in dimensions higher than 2

cc: @mfsch, @JeffreySarnoff

mfsch:
Regarding 2: I think for angles that disregard orientation, the range [0, π] is the only real option. For oriented angles, the unsigned range [0, 2π) is the natural choice and I think it makes sense to have extra checks that exclude the upper boundary 2π in favor of 0, perhaps even within a certain tolerance (e.g. with `isapprox`). In #396, @JeffreySarnoff mentioned that this implies that signed angles should therefore be in [−π, π), i.e. also closed at the lower end, to avoid issues when mapping one range to the other. I’m not sure if that’s the case, since it seems to me that the only difference is whether you use `<` or `<=` in the conditional when converting one to the other, but perhaps I’m overlooking something? Otherwise, I think the range (−π, π] feels more natural to me and I think that’s the one I would expect without reading the documentation. In my view, the open questions therefore are (a) whether the signed interval should be closed at the lower or higher end, and (b) whether the open end of the interval should be handled with `==` or `isapprox`.

mfsch:
@juliohm, did you ever make a decision more generally about you want to approach floating-point precision more generally within Meshes.jl? I’ve noticed that `Point(0.1 + 0.2) ∈ Segment((0.0,), (0.3,))` gives `false`, so it seems that in that case at least you have zero tolerance for floating-point errors. The rasterization algorithm I implemented constructs a number of prism shapes and loops over points in each of those prisms, so it only works if no points can “fall in the gap” between two prisms due to floating-point errors. I therefore have a tolerance argument in the “point-in-prism” check I implemented.

juliohm:
@mfsch I think we can move this discussion about tolerance to Zulip to keep the issue focused on angles. But answering your question, we try to use isapprox always in the current codebase. If some algorithm is using isequal or == we need to fix. At some point in the future we may consider interval arithmetic, but we are moving one step at a time.

juliohm:
[...]

Issue #405, opened 2023-05-01
`isconvex` does not work for polygons in higher-dimensional spaces


mfsch:
For a 2D polygon, the `isconvex` function works as expected:

```julia
julia> isconvex(Ngon(Point2[(0,1),(1,1),(1,0),(2,2)]))
false
```

However, the same polygon in a 3D space is not considered convex:

```julia
julia> isconvex(Ngon(Point3[(0,1,0),(1,1,0),(1,0,0),(2,2,0)]))
true
```

This is because `isconvex` relies on `innerangles`, but the angle function `∠(u, v)` cannot produce a signed angle in three dimensions. Instead of relying on the orientation of the polygon (CW/CCW does not really have a meaning in 3D), we can just rely on the normal vector of the polygon instead. I’ve added an implementation of this approach in my PR #404 for adding the Prism geometry, since I wanted to have a working `isconvex` for the Prism. I’ve also changed the argument types of the existing `innerangle` and `orientation` functions to give an error if the space is not 2D. However, I think for the current approach to work in N dimensions we would have to add a more general implementation of `normal` and `×`.

I also just saw that the issue in 3D has come up before in #135 – perhaps this issue should be reopened and the proposed approach of adding a reference vector could be reconsidered? This is basically what I ended up doing to fix the 3D case, except that I didn’t modify the angle function and kept everything in `innerangles`.

juliohm:
This issue is related to #396 where we are currently discussing what should be the angle definitions adopted in 2D and 3D. Your input there is very welcome.

juliohm:
Depends on #409

juliohm:
Update: the function now works with all primitives and some 3D polytopes that are convex from the type itself.

juliohm:
Fixed already:

```julia
julia> isconvex(Ngon((0,1,0),(1,1,0),(1,0,0),(2,2,0)))
false
```

PR #410, opened 2023-05-03
Clarify current use of signed angles

mfsch:
Currently, signed angles are only returned in 2D. While #396 has sparked a discussion on various aspects of the `angle` function, this PR aims just to clarify the current situation: In 3D, unsigned angles in [0,π] are returned and the functions that rely on signed angles only work in 2D. This partially addresses #405, where `isconvex` in 3D should now return an error instead of wrong results. Further improvements will be discussed in #409.

### DIFF
--- a/src/angles.jl
+++ b/src/angles.jl
@@ -8,7 +8,8 @@
 Angle ∠ABC between rays BA and BC.
 See https://en.wikipedia.org/wiki/Angle.
 
-Uses the two-argument form of `atan` returning value in range [-π, π].
+Uses the two-argument form of `atan` returning value in range [-π, π]
+in 2D and [0, π] in 3D.
 See https://en.wikipedia.org/wiki/Atan2.
 
 ## Examples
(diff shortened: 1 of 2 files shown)


juliohm:
Thank you @mfsch , I am wondering if we have  plans to reintroduce the generic Dim in a future PR that fixes the 3D case. It seems that we are restricting the code to get an error, which is nice, but it may be worth it to fix the actual result of isconvex with fixes in other functions?

mfsch:
For 3D, the changes I had in #404 fix `innerangles` and therefore `isconvex` for Ngon, also adding `normal` for Ngons with more than 3 vertices. The fixed `innerangles` would best be added for Chain rather than Ngon though, but the current implementation of closed Chains makes boundary(ngon) allocate, which would then make `normal`, `isconvex`, and functions much more computationally expensive than they should be, unless they are implemented twice for Ngon and Chain. I have an almost-finished PR to consider changing Chain such that it internally uses a CircularVector to represent closed chains rather than requiring a new vector with a duplicated first node, which would then make it practically free to construct a boundary(ngon) and make Chain the natural place to implement these methods.

juliohm:
Let's try to concentrate on tiny changes in these PRs, otherwise it is very hard to foresee the impacts downstream. If you can fix one small function at a time, providing evidence that the change is for the better, we will feel more safe to merge and proceed to the next improvement. Right now the PRs are mixing different parts of the code and I am having difficulty to assess the risks involved with the changes.

juliohm:
For example, in this present PR, the goal is to improve the docstring + restrict the implementations of other unrelated functions. I am happy to merge the docstring improvement, but am still wondering if the restriction is the best way forward. If we plan to fix the other functions later so that they become N-dimensional again, we will need to undo the restriction Dim->2 and get back to 2->Dim

juliohm (approved):

PR #411, opened 2023-05-04
Use CircularVector internally for closed Chain

mfsch:
I’ve prepared a PR to propose some changes to the internal representation of the Chain. The idea is to represent closed chains with a CircularVector instead of repeating the vertex at which the chain is closed. This has the advantage that creating a closed Chain from an existing vector of vertices no longer needs to allocate, making the `boundary(ngon)` function very cheap. As a result, we can implement functions such as `normal(ngon)`, `innerangles(ngon)` and `isconvex(ngon)` in 3D for Chain without making them expensive for Ngon. I also think it is cleaner overall, but that’s somewhat subjective. This does however result in small changes in the behavior of Chain so it is not fully backwards compatible.

Some comments/questions:

- The main disadvantage is that the constructor is no longer type-stable, unless it is called with a CircularVector. On the other hand, the `vertices` function is now type-stable, which it wasn’t before. It might therefore be good to add a way of constructing an open Chain that remains open even if the first and last vertices are the same but is type-stable, which could be called in cases where we know we don’t want to have a closed Chain.
- The biggest observable change is that `open!` and `close!` no longer exist, since we cannot change the type of V once a Chain is constructed. However, the non-modifying `open` and `close` no longer need to copy the underlying data so they become virtually free since all they do is wrap/unwrap the data with a CircularVector.
- The other somewhat incompatible change is that before the constructor `Chain(::CircularVector)` would create a copy of the vertices instead of reusing them whereas now `Chain(::Vector)` creates a copy in the case that the first and last point match (this is why the `unique!` test case had to be modified). If the Chain is known to be closed a priori, this can be avoided by calling the constructor with a CircularVector.
- It might also make sense to have a more ergonomic way of creating closed Chains. My favorite solution for that would be to define Ring as a type alias for Chain with CircularVector, and use `Ring(points)` as a constructor for closed chains.
- `vertices` and `nvertices` still have the same behavior, but they can now rely on the shared implementation in `polytopes.jl`. Do we still want to keep a separate function for them here so we can keep the documentation that is specific to Chain?
- I think the year for the Held citation should be 2001 instead of 1998, but I haven’t changed it here since it’s not really related.
- I kept `npoints` even though that no longer corresponds to the length of the internal vector, but maybe it’s best to remove it (it is not exported) and change the remaining uses to `nvertices`? It might also be more consistent to change the `Base.show` to not duplicate the first/last vertex and print “Closed 4-Chain” instead of “5-Chain” or perhaps “4-Ring”. The only non-printing use of `npoints` is in `simplexify`, which should probably be changed to use `nvertices`.
- The `unique!` function no longer removes the repeated first/last point for closed chains, since that would change the type `V` and is therefore impossible. However, I think that’s the more natural behavior anyway, and the only use in the code was of the form `close!(unique!(open!(c)))`. Otherwise `unique!` should have remained the same, but I’m not sure it is meant to have the behavior it currently has, where it removes not only consecutive identical points but also points that are repeated later along the chain, and in that case I think it is not well-defined which position is kept and which one removed.

The other alternative to this approach with the CircularVector would be to not touch the type of the vertex vector and instead add a `closed::Bool` field to the struct. This keeps the advantages (cheap to create, cheap to open/close), but decouples the open/closed status from the type of the Chain. In turn, the `vertices` function would lose type-stability again, unless we decide to stop returning a CircularVector for closed chains. It would then make sense to also add a `vertex(chain, ind)` function (as proposed in #406) to have a performant way of accessing the vertices. This function could provide the same looping indexing as CircularVector for closed Chains. I think I like this option almost better in hindsight, and the changes would be very similar to this PR.


mfsch:
I’ve tried out the approach I mentioned at the end of the PR in a [separate commit](https://github.com/JuliaGeometry/Meshes.jl/compare/master...mfsch:Meshes.jl:boolean-chain). I feel like this is the better solution, since it does not make the open/closed status of the Chain part of its type and therefore avoids type-instabilities in the constructor. It also provides a very natural way of creating closed chains or chains that are kept open through an optional `close` keyword argument in the constructor. If that keyword argument is provided, the constructor will always reuse the provided vector and never allocate a new one. The `vertices` function remains type-unstable as it currently is, but the `vertex` function provides a way of accessing vertices (including looping indexing) without allocations and type-instabilities.

There’s also the option of making the the `closed` field mutable, either by wrapping it in a Ref or making the whole struct mutable, which would allow to preserve the modifying `open!` and `close!` functions. However, since opening/closing now no longer changes the vertices array, I think simply providing an `open`/`close` function that returns a new Chain that uses the same `vertices` array is probably cleaner.

I suggest that I update the code in this PR to the version with the Bool field, unless you think feel like the CircularVector approach is more promising.

juliohm:
What about we simply store the closedness information in the type of the vector field? For example, we can simply work with Chain(vertices) as the only constructor and save whatever vector type we pass, it can be Vector or CircularVector. Then we can use dispatch on the various functions, example:

```julia
isclosed(::Chain{Dim,T,<:CircularVector}) = true
isclosed(::Chain) = false
```

We could even introduce the type alias:

```julia
const Ring{Dim,T} = Chain{Dim,T,<:CircularVector}
```

to clean up the code.

At the end, we would only need to specialize the constructor with Vararg, which is a constructor intended for end-users. This constructor `Chain((0,0), (1,1), (2,2), (0,0))` would then do the dynamic check to see if start == end to wrap the vertices in a CircularVector if necessary.

src/polytopes/chain.jl
```diff
@@ -4,18 +4,34 @@
 
 """
     Chain(p1, p2, ..., pn)
+    Chain(points)
 
-A polygonal chain from a sequence of points `p1`, `p2`, ..., `pn`.
+A polygonal chain from a sequence of points `p1`, `p2`, ..., `pn` or a vector
+of `points`. If the first point is equal to the last point or if `points` is of
+type `CircularVector`, the chain is considered closed.
```
juliohm:
```suggestion
A polygonal chain from a sequence of points `p1`, `p2`, ..., `pn`.
If the first point is equal to the last point or if `points` is of
type `CircularVector`, the chain is considered closed.
```

src/polytopes/chain.jl
```diff
@@ -4,18 +4,34 @@
...
 struct Chain{Dim,T,V<:AbstractVector{Point{Dim,T}}} <: Polytope{1,Dim,T}
   vertices::V
+
+  function Chain(vertices::AbstractVector{Point{Dim,T}}) where {Dim,T}
+    if !isa(vertices, CircularVector) && length(vertices) >= 2 &&
+        first(vertices) == last(vertices)
+      vertices = CircularVector(vertices[begin:end-1])
+    end
+    new{Dim,T,typeof(vertices)}(vertices)
+  end
```
juliohm: We need to implement this with multiple-dispatch at compile-time instead of checking the types if `isa`. This is not idiomatic Julia.

src/polytopes/chain.jl
```diff
@@ -31,32 +47,7 @@ This function is provided for IO purposes. Most algorithms
...
-"""
-function vertices(c::Chain)
-  if isclosed(c)
-    vs = @view c.vertices[begin:end-1]
-    CircularVector(vs)
-  else
-    c.vertices
-  end
-end
+npoints(c::Chain) = length(c.vertices) + isclosed(c)
```
juliohm: Can you elaborate on why we are doing these changes and removing nvertices and vertices here?

src/polytopes/chain.jl
juliohm:
```suggestion
```

src/polytopes/chain.jl
```diff
@@ -4,18 +4,34 @@
...
+    new{Dim,T,typeof(vertices)}(vertices)
+  end
 end
 
 # specialize Polytope outer constructor to use standard vector
-Chain(vertices::Vararg{P}) where {P<:Point} = Chain(collect(vertices))
-
-Chain(vertices::CircularVector) = Chain([collect(vertices); first(vertices)])
+function Chain(vertices::Vararg{P}) where {P<:Point}
+  if length(vertices) >= 2 && first(vertices) == last(vertices)
```
juliohm:
```suggestion
  if first(vertices) == last(vertices)
```

src/polytopes/chain.jl
```diff
@@ -91,7 +82,7 @@ Tells whether or not the chain is closed.
 
 A closed chain is also known as a ring.
 """
-isclosed(c::Chain) = first(c.vertices) == last(c.vertices)
+isclosed(c::Chain{Dim,T,V}) where {Dim,T,V} = V <: CircularVector
```
juliohm:
```suggestion
isclosed(::Chain) = false
isclosed(::Ring) = true
```

src/polytopes/chain.jl
```diff
@@ -139,7 +130,8 @@ See https://en.wikipedia.org/wiki/Winding_number.
 """
 function windingnumber(p::Point{2,T}, c::Chain{2,T}) where {T}
```
juliohm: We should update this function so that it only accepts Ring. The winding number is not defined for open chains.

src/polytopes/chain.jl
```diff
@@ -173,9 +165,9 @@ Default method is `WindingOrientation()`.
 orientation(c::Chain) = orientation(c, WindingOrientation())
```
juliohm: same for orientation, it should only accept Ring

src/polytopes/chain.jl
```diff
@@ -190,11 +182,11 @@ end
 """
     unique!(chain)
 
-Remove duplicate vertices in the `chain`.
+Remove duplicate vertices in the `chain`. Closed chains remain closed.
```
juliohm, line 174:
```suggestion
Remove duplicate vertices in the `chain`.
```

src/polytopes/chain.jl
```diff
@@ -220,54 +212,37 @@ end
 """
     unique(chain)
 
-Return a new `chain` without duplicate vertices.
+Return a new `chain` without duplicate vertices. Closed chains remain closed.
```
juliohm, line 204:
```suggestion
Return a new `chain` without duplicate vertices.
```

test/polytopes.jl
```diff
@@ -448,7 +443,7 @@
     points = P2[(1,1),(2,2),(2,2),(3,3),(1,1)]
     poly   = PolyArea(points)
     unique!(poly)
-    @test first(chains(poly)) == Chain(points)
+    @test first(chains(poly)) == Chain(P2[(1,1),(2,2),(3,3),(1,1)])
```
juliohm: this was a buggy test?

juliohm (changes requested):
Attached first round of reviews.

test/polytopes.jl
mfsch: An auto-closing constructor has to either modify the input array with pop! (doesn’t work for inputs that are static or views), store a `@view points[begin:end-1]` internally, or create a copy of `points[begin:end-1]`. Storing the view makes a function like `unique!` impossible, so I went with the last option. The test relies on the fact that both the original `points` array and the internal array of the Chain are modified by `unique!`, which is no longer the case.

src/polytopes/chain.jl
mfsch: I figured since this is a breaking change in the behavior of `unique` it might be good to mention it?

src/polytopes/chain.jl
mfsch: This is explained in the PR description:

> `vertices` and `nvertices` still have the same behavior, but they can now rely on the shared implementation in `polytopes.jl`. Do we still want to keep a separate function for them here so we can keep the documentation that is specific to Chain?

src/polytopes/chain.jl
mfsch: I’m pretty sure `isa` is always resolved at compile-time (where possible) and it is used frequently in Base and throughout the Julia ecosystem. In this case I agree that it might be cleaner to rely on multiple dispatch but in general I think multiple dispatch makes sense where most/all of the functionality of a function is different and `isa` makes more sense when only a small part differs and that small part doesn’t make sense as a separate function.

src/polytopes/chain.jl
```diff
@@ -4,59 +4,31 @@
 
 """
     Chain(p1, p2, ..., pn)
+    Chain(points::AbstractVector)
```
juliohm:
```suggestion
```

We can review the signatures of all polytopes in a separate PR 👍🏽

src/polytopes/chain.jl
```diff
@@ -4,59 +4,31 @@
...
 
-A polygonal chain from a sequence of points `p1`, `p2`, ..., `pn`.
+A polygonal chain from a sequence of points `p1`, `p2`, ..., `pn`. If `points`
+is of type `CircularVector`, the chain is considered closed.
```
juliohm:
```suggestion
A polygonal chain from a sequence of points `p1`, `p2`, ..., `pn`.
If `points` is of type `CircularVector`, the chain is considered closed.
```

src/polytopes/chain.jl
```diff
@@ -4,59 +4,31 @@
...
-chain, no matter if it is closed or open.
-
-See also [`nvertices`](@ref).
 
 ### Notes
 
-This function is provided for IO purposes. Most algorithms
-should be written in terms of `nvertices` and `vertices`
-as they are consistent with each other.
+- `Ring` is a type alias for closed chains.
```
juliohm:
```suggestion
- A `Ring` is a type alias for closed `Chain`.
```

src/polytopes/chain.jl
```diff
@@ -4,59 +4,31 @@
...
+  vertices::V
+end
 
-Return the number of vertices of the `chain`. In the case
-that the chain is closed, the number of vertices is one
-less the total number of points used to represent the chain.
 """
-nvertices(c::Chain) = npoints(c) - isclosed(c)
+    Ring(p1, p2, ..., pn)
+    Ring(points::AbstractVector)
```
juliohm:
```suggestion
```

src/polytopes/chain.jl
```diff
@@ -4,59 +4,31 @@
...
 
+A closed polygonal chain from a sequence of points `p1`, `p2`, ..., `pn`. This
+is a type alias for a `Chain` with a `CircularVector`.
```
juliohm:
```suggestion
A closed polygonal chain from a sequence of points `p1`, `p2`, ..., `pn`.
```

src/polytopes/chain.jl
```diff
@@ -4,59 +4,31 @@
...
-function vertices(c::Chain)
-  if isclosed(c)
-    vs = @view c.vertices[begin:end-1]
-    CircularVector(vs)
-  else
-    c.vertices
-  end
-end
+(::Type{Ring})(vertices::AbstractVector) = Chain(CircularVector(vertices))
+(::Type{Ring})(vertices::CircularVector) = Chain(vertices)
```
juliohm: Isn't the syntax Ring(vertices::AbstractVector) supported? Just double checking if we can remove the ::Type wrapper.

src/polytopes/chain.jl
```diff
@@ -220,56 +191,35 @@ end
...
-    open!(chain)
-
-Open the `chain`, i.e. remove the last vertex.
-"""
-function open!(c::Chain)
-  pop!(c.vertices)
-  c
-end
+Base.close(c::Chain) = c
+Base.close(c::Ring) = Chain(CircularVector(c.vertices))
```
juliohm: I think these two should be swapped? If we get a Ring we just return it, otherwise we get a Chain that is open and wrap the vertices in a CircularVector.

src/polytopes/chain.jl
```diff
@@ -220,56 +191,35 @@ end
...
     open(chain)
 
-Same as [`open!`](@ref) but operates on a copy
-of the original `chain`.
+Open the `chain`, i.e. remove the segment going from the last to the first vertex.
 """
-Base.open(c::Chain) = open!(deepcopy(c))
+# call `open` again to avoid issues in case of nested CircularVector
+Base.open(c::Chain) = open(Chain(parent(c.vertices)))
+Base.open(c::Ring) = c
```
juliohm: Same here, I think the definitions are swapped.

src/polytopes/chain.jl
```diff
@@ -220,56 +191,35 @@ end
...
 
 Reverse the `chain` vertices in place.
 """
-function Base.reverse!(c::Chain)
-  reverse!(c.vertices)
-  c
-end
+Base.reverse!(c::Chain) = reverse!(c.vertices)
+# do not change which vertex comes first for closed chains
+Base.reverse!(c::Ring) = reverse!(@view c.vertices[2:end])
```
juliohm:
```suggestion
Base.reverse!(c::Ring) = reverse!(view(c.vertices, begin+1:end))
```

src/polytopes/polyarea.jl
```diff
@@ -21,14 +23,11 @@ in the real world, including issues with:
...
-struct PolyArea{Dim,T,C<:Chain{Dim,T}} <: Polygon{Dim,T}
+struct PolyArea{Dim,T,C<:Ring{Dim,T}} <: Polygon{Dim,T}
   outer::C
   inners::Vector{C}
 
-  function PolyArea{Dim,T,C}(outer, inners, fix) where {Dim,T,C}
-    @assert isclosed(outer) "invalid outer chain"
-    @assert all(isclosed, inners) "invalid inner chains"
-
+  function PolyArea{Dim,T,C}(outer::Ring, inners::Ring, fix) where {Dim,T,C}
```
juliohm:
```suggestion
  function PolyArea{Dim,T,C}(outer, inners, fix) where {Dim,T,C}
```

juliohm:
@mfsch it would be lovely to incorporate this breaking change in the next release. Any chance you have some time to get back to it next week?

mfsch:
Alright, I’ve rebased this PR and finished the changes that introduce the `Ring` type alias and no longer automatically close the `Chain` type, as discussed. As you can see, these changes are quite disruptive as they touch almost every usage of `Chain` and `PolyArea`. Most usages became simpler/cleaner, but I still think it’s a pretty aggressive way to break compatibility, since code based on the current semantics silently changes its behavior.

A few more comments:
- Some of the remaining uses of `Chain` (as opposed to `Ring`) should probably also be converted to `Ring`, for example in the type signatures of algorithms that implicitly assume that the `Chain` is closed. I haven’t changed those, since they were arguably wrong or at least underdocumented already and I got the impression that you’d rather not have such changes as part of this PR.
- I’ve kept the `npoints` function around, again to keep the changes more limited, but I think it should also be removed since it no longer reflects the internal structure and was already unexported.
- I’m not too crazy about the approach of using the “implementation detail” of the `CircularVector` to represent the completely different semantics of an open and closed chain. In retrospect, I would prefer having two completely separate types, `OpenChain` and `ClosedChain`/`Ring`, that are subtypes of an abstract `Chain`. The `Chain` constructor could then be type-unstable and check whether `first(vertices) == last(vertices)`, whereas the constructors of the concrete types would be type-stable. This would also avoid the disruptive change for `Chain` at least, though `PolyArea` is still silently changing its behavior.
- If you wanted, you could also change the PolyArea constructors that take point vectors instead of chains to `pop!` the last point if it matches the first (or use a view). That away the PolyArea would mostly retain its current behavior.

src/polytopes/chain.jl
mfsch: Done in newer version.

src/polytopes/polyarea.jl
mfsch: Done in newer version.

src/polytopes/chain.jl
mfsch: Done in newer version.

src/polytopes/chain.jl
mfsch: Done in newer version.

src/polytopes/chain.jl
mfsch: Done in newer version.

src/polytopes/chain.jl
mfsch: Done in newer version.

src/polytopes/chain.jl
mfsch: Done in newer version.

src/polytopes/chain.jl
mfsch: Done in newer version.

src/polytopes/chain.jl
mfsch: Done in newer version.

src/polytopes/chain.jl
mfsch: I’ve kept this because this is a change in the behavior, but feel free to remove it if you don’t want this behavior documented.

src/polytopes/chain.jl
mfsch: I’ve kept this because this is a change in the behavior, but feel free to remove it if you don’t want this behavior documented.

src/polytopes/chain.jl
mfsch: I’ve made the changes to docstrings more minimal in the newer version.

src/polytopes/chain.jl
mfsch: I’ve made the changes to docstrings more minimal in the newer version – feel free to change further.

src/polytopes/chain.jl
mfsch: I’ve made the changes to docstrings more minimal in the newer version – feel free to change further.

src/polytopes/chain.jl
mfsch: I’ve made the changes to docstrings more minimal in the newer version – feel free to change further.

test/runtests.jl
```diff
@@ -67,6 +69,7 @@ include("dummy.jl")
 
 # list of tests
 testfiles = [
+  "polytopes.jl",
```
juliohm:
```suggestion
```

juliohm (approved):
This is perhaps the best PR we have ever received in Meshes.jl 👏🏽 Thank you very much @mfsch for the great improvement.

PR #413, opened 2023-05-04
Use Bool field for representing closed Chain

mfsch:
This PR contains the alternative approach to #411 that doesn’t make the open/closed property of the Chain part of its type and uses a boolean field instead.


juliohm:
The PR with the Ring type seems very promising already, I will close this one as an inferior alternative. We can reopen if we realize #411 is inferior.

PR #414, opened 2023-05-04
Define function to get specific vertex of polytope

mfsch:
As discussed in #406, this introduces a function to access individual vertices of a Polytope without retrieving the whole vector of vertices. This makes it possible to add Polytopes that use a different internal representation than a vector of all their vertices, but still need a way to access individual vertices without building and allocating the whole list. This currently affects the Prism type I would like to add (see #404), and it would also affect Box if we wanted to align it with the other polytopes in the future. One version of the proposed changes to Chain (#413) would also benefit from this.

More generally, this approach could serve as a template for other functions that generally build a list of items to return but also need a way to access individual items efficiently without building the whole list. For example, a `segment(chain, ind)` function could return a specific segment of a chain without building a vector of segments. Functions like `issimple(chain)` could then be implemented without `collect`ing the segments and eventually without allocations if we make `segment(chain, ind)` non-allocating. But that’s something for the future to consider – I’m just mentioning it to show that the pattern might be more broadly useful.

One option to consider is to keep the `vertex` function unexported for now. That way, code within Meshes.jl has an efficient way of accessing vertices without making any promises to external users. An even more conservative option is to only introduce a `vertex` function for types that need it and perhaps introduce the function for all Polytopes later if we find a need to do so. This would mean postponing this PR for now and simply use `vertex(polytope, ind)` as a design pattern within the implementation of polytopes that benefit from it, similar to the current use of `npoints`. I think I would prefer this option, but I get the impression that the current design tries to avoid unexported functions for polytopes/geometries?


juliohm:
That is awesome @mfsch and the PR is already in pretty good shape. I think we can add the function to the Mesh abstract interface as well given that it also has vertices. There is no problem in exporting the new function name.

juliohm (approved):

juliohm:
Just waiting for the tests before merging it 💯

mfsch:
Thanks!

Issue #344, opened 2022-12-20
Improve rotation standards


juliohm:
We have a couple of issues with vanilla ReferenceFrameRotations.jl rotation objects:

1. They describe rotations of frames and we always need to flip signs of angles to flip geometries and meshes
2. They are defined for 3D embeddings only, and we had to introduce our own 2D rotations in [src/rotations.jl](https://github.com/JuliaGeometry/Meshes.jl/blob/1432107f191130f3fe3792b00816b4057b236a51/src/rotations.jl)

I am opening this issue to discuss a general improvement regarding our rotation interface to facilitate the lives of end-users who are mostly interested in rotating geometries and meshes, not frames.

As a starting proposal I suggest that we introduce a new set of rotations in `src/rotations.jl` to flip the signs of angles in existing rotation types from ReferenceFrameRotations.jl. For example:

```julia
IEulerAngleAxis(angle, axis) = EulerAngleAxis(-angle, axis)
```

We can export this list of standardized "inverse" rotations and drop support for vanilla rotation types from ReferenceFrameRotations.jl. This doesn't mean that we will stop using ReferenceFrameRotations.jl, it just means that end-users will only interface with the features of the package via a set of well-defined geometry-oriented rotation types as opposed to frame-oriented rotation types.

@ronisbr in case you have a suggestion for an alternative solution that involves changes to ReferenceFrameRotations.jl itself, please let us know. These angle signs are very error-prone and our users are not usually thinking in terms of frames. If you feel that these `IRotation` types should live in ReferenceFrameRotations.jl, we can work on a PR over there.

cc: @stla

ronisbr:
Hi @juliohm !

I think the problem is that you want to rotate vectors, whereas ReferenceFrameRotations.jl rotates reference frames. It is what we call active and passive rotations. Those are different exactly by that inverse signal.

https://en.wikipedia.org/wiki/Active_and_passive_transformation

Are you sure you want here to rotate the reference frame instead of rotating vectors in the same reference frame?

juliohm:
Hi Ronan! We want to rotate vectors mostly. That is why we need other
rotation types that differ from the ones in ReferenceFrameRotations.jl

Em qua., 21 de dez. de 2022 12:44, Ronan Arraes Jardim Chagas <
***@***.***> escreveu:

ronisbr:
[...]

PR #415, opened 2023-05-05
Use Rotations.jl for rotations

hyrodium:
This PR fixes https://github.com/JuliaGeometry/Meshes.jl/issues/344.
I'm keeping this PR as a draft because removing `ReferenceFrameRotations.DCM` is not finished yet.
I just would like to share the progress and get some feedback on the changes on this PR.


juliohm:
Looking forward to these improvements @hyrodium ❤️ 

It would be nice if you could summarize the new types from Rotations.jl that replace the old types from ReferenceFrameRotations.jl in a comment here when you are done 👌🏽

src/transforms/rotate.jl
```diff
@@ -6,15 +6,15 @@
...
 
 ## Examples
 
 ```julia
-Rotate(EulerAngleAxis(pi/4, [1, 0, 0]))
+Rotate(rand(QuatRotation{Float64}))
 ```
 """
-struct Rotate{R} <: StatelessGeometricTransform
+struct Rotate{R<:Rotation} <: StatelessGeometricTransform
```
hyrodium, line 19: Now the parameter of `Rotate` requires a subtype of `Rotations.Rotation`. This abstract type `Rotation{N}` represents a special orthonormal group $SO(N)$.

```julia
julia> using StaticArrays, Rotations

julia> Rotation{3} <: StaticMatrix{3,3} <: AbstractMatrix
true
```

src/neighborhoods/metricball.jl
```diff
@@ -36,40 +36,29 @@ struct MetricBall{R,M} <: Neighborhood
...
 
+function MetricBall(radii::SVector{Dim,T}, R::Rotation{Dim}=_default_rotation(Val{Dim}(), T)) where {Dim,T}
   # scaling matrix
   Λ = Diagonal(one(T) ./ radii .^ 2)
 
-  # rotation matrix
-  R = convert(DCM, rot)
-
-  # sanity check
-  @assert size(R) == (Dim, Dim) "invalid rotation for radii"
```
hyrodium: `Rotation{Dim}` enforces the size of the transformation matrix, so we don't need the sanity check.

test/rotations.jl
```diff
@@ -1,11 +1,11 @@
...
-  R  = convert(DCM, cw)
+  # ClockwiseAngle
+  R = Angle2d(T(-π/2))
   @test eltype(R) == T
   @test R ≈ T[0 1; -1 0]
 
-  ccw = CounterClockwiseAngle(T(π/2))
-  R  = convert(DCM, ccw)
+  # CounterClockwiseAngle
+  R = Angle2d(T(π/2))
```
hyrodium: I have updated these tests to follow the updated `src/*.jl`, but these tests are now independent of Meshes.jl. 
Should we remove these tests?

src/utils.jl
```diff
@@ -186,13 +186,9 @@ end
...
 end
 
 function uvrotation(u::Vec{3}, v::Vec{3})
-  u⃗ = normalize(u)
-  v⃗ = normalize(v)
-  re = √((1 + u⃗ ⋅ v⃗) / 2)
-  im = (u⃗ × v⃗) / 2re
-  convert(DCM, Quaternion(re, im...))
+  return rotation_between(v, u)
 end
```
hyrodium: There is a similar function `Rotations.rotation_between`, but the order of arguments is different.
I kept the current behavior, but `rotation_between(u, v)` seems better for its consistency.

```julia
julia> using Meshes, LinearAlgebra, Pkg

julia> Pkg.status("Meshes")
Status `~/tmp/Project.toml`
  [eacbb407] Meshes v0.28.1

julia> uvrotation(normalize(Vec2(1,2)), normalize(Vec2(2,1))) * Vec2(1,2)  # Correct
2-element Vec2 with indices SOneTo(2):
 2.0
 1.0

julia> uvrotation(normalize(Vec3(1,2,3)), normalize(Vec3(2,1,3))) * Vec3(1,2,3)  # Inconsistent
3-element Vec3 with indices SOneTo(3):
 -0.14285714285714293
  2.7142857142857144
  2.5714285714285716

julia> uvrotation(normalize(Vec3(2,1,3)), normalize(Vec3(1,2,3))) * Vec3(1,2,3)  # Desired behavior
3-element Vec3 with indices SOneTo(3):
 2.0
 1.0
 3.0
```

src/transforms/rotate.jl
```diff
@@ -6,15 +6,15 @@
...
 
 Rotate geometry or mesh with rotation `rot`
-from ReferenceFrameRotations.jl. 
+from Rotations.jl.
 
 ## Examples
 
 ```julia
-Rotate(EulerAngleAxis(pi/4, [1, 0, 0]))
+Rotate(rand(QuatRotation{Float64}))
```
hyrodium: I added an example with `QuatRotation <: Rotation{3}` which represents a rotation with a quaternion.

src/neighborhoods/metricball.jl
```diff
@@ -36,40 +36,29 @@ struct MetricBall{R,M} <: Neighborhood
...
-  else
-    rotation
-  end
+function _default_rotation(::Val{2}, T)
+  one(Angle2d{T})
+end
+
+function _default_rotation(::Val{3}, T)
+  one(QuatRotation{T})
+end
```
hyrodium: The default rotation can be obtained from `_default_rotation` function. The error not for 2 or 3-dimensional rotation is now `MethodError` instead of `ErrorException("not implemented")`.

src/rotations.jl
```diff
@@ -1,46 +0,0 @@
-# ------------------------------------------------------------------
-# Licensed under the MIT License. See LICENSE in the project root.
-# ------------------------------------------------------------------
-
-# -------------
-# 2D ROTATIONS
-# -------------
-
-# temporary fix for https://github.com/JuliaSpace/ReferenceFrameRotations.jl/issues/22
```
hyrodium, line 9: I'm understanding `ClockwiseAngle` and `CounterClockwiseAngle` are a temporary fix for 2-dimensional rotation, and can be replaced with `Rotations.Angle2d`.

src/rotations.jl
```diff
@@ -1,46 +0,0 @@
...
-Base.inv(ccw::CounterClockwiseAngle) = ClockwiseAngle(ccw.θ)
-
-Base.convert(::Type{<:DCM}, ccw::CounterClockwiseAngle) =
-  convert(DCM, ClockwiseAngle(-ccw.θ))
-
-# -------------
-# 3D ROTATIONS
-# -------------
-
-TaitBryanAngles(θs...) = EulerAngles((-1 .* θs)..., :ZXY) |> inv
```
hyrodium, line 45: Rotations.jl provides types for both Tait-Bryan angles and Euler angles. See https://juliageometry.github.io/Rotations.jl/stable/3d_euler/#Rotation-around-three-axes-(Euler-Angles) for more information.

hyrodium (commented):
This PR is ready to review! I added some comments for the changes.

src/utils.jl
```diff
@@ -186,13 +186,9 @@ end
 Return rotation matrix from vector `u` to vector `v`.
 """
 function uvrotation(u::Vec{2}, v::Vec{2})
-  convert(DCM, CounterClockwiseAngle(∠(u, v)))
+  Angle2d(∠(u,v))
```
hyrodium: This can be replaced with `rotation_between` after https://github.com/JuliaGeometry/Rotations.jl/pull/257 is finished.

Project.toml
```diff
@@ -24,7 +24,6 @@ CircularArrays = "1.3"
 Distances = "0.10"
 IterTools = "1.3"
 NearestNeighbors = "0.4"
-ReferenceFrameRotations = "3"
```
juliohm, line 27: We need a version in the compat section for Rotations.jl

src/Meshes.jl
```diff
@@ -15,7 +15,7 @@ using Bessels: gamma
 using IterTools: ivec
 using StatsBase: AbstractWeights, Weights, quantile
 using Distances: PreMetric, Euclidean, Mahalanobis, evaluate
-using ReferenceFrameRotations: EulerAngles, DCM, Quaternion
+using Rotations: rotation_between, Angle2d, QuatRotation, Rotation
```
```suggestion
using Rotations: Angle2d, QuatRotation, Rotation, rotation_between
```

src/neighborhoods/metricball.jl
juliohm: So QuatRotation is the equivalent of EulerAngles in the 3D case?

src/neighborhoods/metricball.jl
```diff
@@ -36,40 +36,29 @@ struct MetricBall{R,M} <: Neighborhood
...
-      EulerAngles(zeros(T, Dim)...)
-    else
-      throw(ErrorException("not implemented"))
-    end
-  else
-    rotation
-  end
+function _default_rotation(::Val{2}, T)
+  one(Angle2d{T})
+end
```
juliohm:
```suggestion
default_rotation(::Val{2}, T) = one(Angle2d{T})
```

src/neighborhoods/metricball.jl
juliohm:
```suggestion
default_rotation(::Val{3}, T) = one(QuatRotation{T})
```

juliohm: Can you explain why we need to wrap the rotation type in `one`?

src/neighborhoods/metricball.jl
```diff
@@ -36,40 +36,29 @@ struct MetricBall{R,M} <: Neighborhood
...
+
+function _default_rotation(::Val{3}, T)
+  one(QuatRotation{T})
+end
 
+function MetricBall(radii::SVector{Dim,T}, R::Rotation{Dim}=_default_rotation(Val{Dim}(), T)) where {Dim,T}
```
juliohm:
```suggestion
function MetricBall(radii::SVector{Dim,T},
                    R=default_rotation(Val{Dim}(), T)) where {Dim,T}
```

src/neighborhoods/metricball.jl
```diff
@@ -36,40 +36,29 @@ struct MetricBall{R,M} <: Neighborhood
...
-  @assert size(R) == (Dim, Dim) "invalid rotation for radii"
-
   # Mahalanobis metric
   metric = Mahalanobis(Symmetric(R'*Λ*R))
 
   MetricBall{typeof(radii),typeof(metric)}(radii, metric)
 end
 
-MetricBall(radii::NTuple{Dim,T}, rotation=nothing) where {Dim,T} =
+MetricBall(radii::NTuple{Dim,T}, rotation=_default_rotation(Val{Dim}(), T)) where {Dim,T} =
```
juliohm:
```suggestion
MetricBall(radii::NTuple{Dim,T}, rotation=default_rotation(Val{Dim}(), T)) where {Dim,T} =
```

src/neighborhoods/metricball.jl
```diff
@@ -36,40 +36,29 @@ struct MetricBall{R,M} <: Neighborhood
...
   MetricBall(SVector(radii), rotation)
 
 # avoid silent calls to inner constructor
-MetricBall(radii::AbstractVector{T}, rotation=nothing) where {T} =
+MetricBall(radii::AbstractVector{T}, rotation=_default_rotation(Val{Dim}(), T)) where {T} =
```
juliohm:
```suggestion
MetricBall(radii::AbstractVector{T}, rotation=default_rotation(Val{Dim}(), T)) where {T} =
```

src/transforms/rotate.jl
juliohm: Is there an example with a more familiar rotation type? Maybe `AngleAxis`? Quaternions are a bit exoteric for non-mathematicians.

src/utils.jl
```suggestion
  Angle2d(∠(u, v))
```

src/utils.jl
```diff
@@ -186,13 +186,9 @@ end
...
 end
 
 function uvrotation(u::Vec{3}, v::Vec{3})
-  u⃗ = normalize(u)
-  v⃗ = normalize(v)
-  re = √((1 + u⃗ ⋅ v⃗) / 2)
-  im = (u⃗ × v⃗) / 2re
-  convert(DCM, Quaternion(re, im...))
+  return rotation_between(v, u)
```
juliohm: Nice, it would be great if we could remove uvrotation in favor of rotation_between, also is there a reason for the order (v, u) instead of (u, v)?

src/utils.jl
juliohm: Do you think rotation_between could use the (u, v) order in a new release? I am finding it a bit strange that the order is reversed. What is the rationale?

test/rotations.jl
juliohm: Yes, I think we should remove these tests now that the types belong to Rotations.jl

juliohm (changes requested):
Lovely PR @hyrodium , it is almost ready.

Project.toml
hyrodium: Fixed in https://github.com/JuliaGeometry/Meshes.jl/pull/415/commits/95b4db281395a62501f3091d1c3354f7a2e83e71.

hyrodium: Adding tests with [Aaua.jl](https://github.com/JuliaTesting/Aqua.jl) would be helpful.

>Check that all external packages listed in deps have corresponding `compat` entry

src/primitives/torus.jl
```diff
@@ -62,7 +62,7 @@ area(t::Torus) = measure(t)
 function Base.in(p::Point, t::Torus)
   c, n⃗ = t.center, t.normal
   R, r = t.major, t.minor
-  M = uvrotation(Vec(0, 0, 1), n⃗)
```
juliohm, line 62: I think you misunderstood my request. I was wondering if rotation_between could have a different order of arguments, our uvrotation has the expected order of arguments.

hyrodium: Okay.., but is it okay to have inconsistent behavior? (https://github.com/JuliaGeometry/Meshes.jl/pull/415#discussion_r1186701645)
I was thinking that the current argument order of `uvrotation(::Vec2, ::Vec2)` is correct, but `uvrotation(::Vec3, ::Vec3)` is not.

juliohm: Sorry, maybe I misunderstood something. I thought that Rotations.rotation_between had the same purpose of Meshes.uvrotation and that we could remove our uvrotation in favor of rotation_between. I then started to wonder why rotation_between expects (v, u) instead of (u, v). Is rotation_between(v, u) the rotation necessary to move v to u?

hyrodium: >Is rotation_between(v, u) the rotation necessary to move v to u?

Yes, that's right. `Rotations.rotation_between(u, v)` is a rotation that moves `u` to `v`. Here's an examlpe:

```julia
julia> using Rotations, StaticArrays

julia> u = SVector(1,2,3);

julia> v = SVector(3,1,2);

julia> rotation_between(u, v) * u ≈ v
true
```

However, `Meshes,uvroatation(u::Vec3, v::Vec3)` is a rotation that moves `v` to `u`. (NOT `u` to `v`)

```julia
julia> using Meshes

julia> u = Vec3(1,2,3);

julia> v = Vec3(3,1,2);

julia> uvrotation(u, v) * u ≈ v
false

julia> uvrotation(u, v) * v ≈ u
true
```

Note that `Meshes,uvroatation(u::Vec2, v::Vec2)` is a rotation that moves `u` to `v`.

```julia
julia> u = Vec2(1,3);

julia> v = Vec2(3,1);

julia> uvrotation(u, v) * u ≈ v
true

julia> uvrotation(u, v) * v ≈ u
false
```

juliohm: Oh then this yet another reason to eliminate our uvrotation completely from the codebase. Can we remove it and use rotation_between everywhere?

src/Meshes.jl
```diff
@@ -15,7 +15,7 @@ using Bessels: gamma
 using IterTools: ivec
 using StatsBase: AbstractWeights, Weights, quantile
 using Distances: PreMetric, Euclidean, Mahalanobis, evaluate
-using ReferenceFrameRotations: EulerAngles, DCM, Quaternion
+using Rotations: Angle2d, QuatRotation, Rotation, rotation_between
```
```suggestion
using Rotations: Rotation, QuatRotation, Angle2d, rotation_between
```

src/sampling/regular.jl
```diff
@@ -136,7 +136,7 @@ function sample(::AbstractRNG, torus::Torus{T},
 
   c = center(torus)
   n⃗ = normal(torus)
-  M = uvrotation(n⃗, Vec{3,T}(0, 0, 1))
+  M = uvrotation(Vec{3,T}(0, 0, 1), n⃗)
```
juliohm, line 139:
```suggestion
  M = rotation_between(Vec{3,T}(0, 0, 1), n⃗)
```

src/utils.jl
```diff
@@ -186,13 +186,9 @@ end
 Return rotation matrix from vector `u` to vector `v`.
 """
 function uvrotation(u::Vec{2}, v::Vec{2})
-  convert(DCM, CounterClockwiseAngle(∠(u, v)))
+  Angle2d(∠(u, v))
 end
```
juliohm: Let's erase uvrotation completely in favor of rotation_between

juliohm:
@hyrodium I am available today in case you want to finish this PR for a final review. Thanks again for the great improvement.

src/primitives/torus.jl
hyrodium: Sorry for the late reply. I agree with removing `uvrotaiton`, and I'd like to merge https://github.com/JuliaGeometry/Rotations.jl/pull/257 first. (x-ref https://github.com/JuliaGeometry/Meshes.jl/pull/415#discussion_r1186704202)

test/rotations.jl
hyrodium: Removed in https://github.com/JuliaGeometry/Meshes.jl/pull/415/commits/5236644144dc7ffd1a7e1462b52062c1f75fab2b.

src/neighborhoods/metricball.jl
hyrodium: The usage of `_default_rotation` is totally internal. so I have named it with the prefix `_` to represent that.

[Prepended underscore in Base functions](https://discourse.julialang.org/t/prepended-underscore-in-base-functions/19645/3):

>In general, an prepended underscore is typically used to indicate some kind of internal implementation detail, especially functions that make special assumptions about the arguments and so aren’t safe for external use.

Are there any conventions to avoid these underscores in Meshes.jl?

juliohm: We avoid underscores as much as possible in Meshes.jl, that is why I suggested removing it. We recently merged a formal code style with JuliaFormatter, so these things will gradually converge.

juliohm: BTW, we need to rebase this PR on top of the latest master to fix these formatting conflicts.

juliohm:
@hyrodium can you please rebase this PR on top of the latest master branch?

juliohm:
@hyrodium any chance we could get this merged this week? We are planning a new breaking release and it would be nice to incorporate the changes here.

hyrodium:
Sorry for being late. I have merged the current `master` branch to resolve the conflicts.

hyrodium:
I will update this PR in a few days to reflect the review comments.

src/neighborhoods/metricball.jl
hyrodium: The current line-length limitation is [`margin = 120`](https://github.com/JuliaGeometry/Meshes.jl/blob/a3e9130219ebe8a3cd91a4d0338c3a0af13f5f6a/.JuliaFormatter.toml#LL2C9-L2C9), so we don't need to add the line break here.

src/transforms/rotate.jl
hyrodium: I have added more examples to the docstring.

src/neighborhoods/metricball.jl
hyrodium: >So QuatRotation is the equivalent of EulerAngles in the 3D case?

Not equivalent. Rotations with quaternions are more useful than Euler angles in many cases. See [gimbal lock](https://en.wikipedia.org/wiki/Gimbal_lock) for more information.

src/neighborhoods/metricball.jl
hyrodium: >We avoid underscores as much as possible in Meshes.jl

Okay, I can remove the underscores in this PR, but could you tell me the reason to avoid underscores? If we really need to avoid them, how can we modify internal functions without breaking releases?

juliohm: We can simply modify the internal functions, they are not exported and not intended for end-users.

src/transforms/rotate.jl
```diff
@@ -3,18 +3,21 @@
 # ------------------------------------------------------------------
 
 """
-    Rotate(rot)
+    Rotate <: StatelessGeometricTransform
+    Rotate(rot::Rotations.Rotation)
```
juliohm:
```suggestion
    Rotate(rot)
```

src/transforms/rotate.jl
```diff
@@ -31,13 +34,13 @@ vector `u` to the plane passing through the origin with normal vector `v`.
 Rotate(Vec(1, 0, 0), Vec(1, 1, 1))
 ```
 """
-Rotate(u::Vec, v::Vec) = Rotate(uvrotation(u, v))
+Rotate(u::Vec, v::Vec) = Rotate(rotation_between(v, u))
```
juliohm:
```suggestion
Rotate(u::Vec, v::Vec) = Rotate(rotation_between(u, v))
```
Here we want the rotation to rotate vector u to vector v, right?

src/transforms/rotate.jl
```diff
@@ -31,13 +34,13 @@ vector `u` to the plane passing through the origin with normal vector `v`.
...
 
 isrevertible(::Type{<:Rotate}) = true
 
 function preprocess(transform::Rotate, object)
   rot = transform.rot
-  convert.(DCM, (inv(rot), rot))
+  return inv(rot), rot
```
juliohm:
```suggestion
  inv(rot), rot
```

test/utils.jl
```diff
@@ -37,9 +37,9 @@
   @test mayberound(1.1, 1, 0.10000000000000001) ≈ 1.1
   @test mayberound(1.1, 1, 0.05) ≈ 1.1
 
-  @test uvrotation(V2(1, 0), V2(0, 1)) ≈ T[0 -1; 1 0]
-  @test uvrotation(V2(0, 1), V2(1, 0)) ≈ T[0 1; -1 0]
-  @test uvrotation(V3(1, 0, 0), V3(0, 1, 0)) ≈ T[0 1 0; -1 0 0; 0 0 1]
+  @test rotation_between(V2(1, 0), V2(0, 1)) ≈ T[0 -1; 1 0]
+  @test rotation_between(V2(0, 1), V2(1, 0)) ≈ T[0 1; -1 0]
+  @test rotation_between(V3(0, 1, 0), V3(1, 0, 0)) ≈ T[0 1 0; -1 0 0; 0 0 1]
```
```suggestion
```

juliohm (approved):

juliohm:
Thank you @hyrodium ! ❤️

PR #423, opened 2023-05-14
Remove extra comma

eliascarv:


juliohm (approved):

juliohm:
The extra comma was introduced by the JuliaFormatter? Just curious how much we can trust on this tool.

juliohm:
> The extra comma was introduced by the JuliaFormatter? Just curious how much we can trust on this tool.

I don't think so. I think we added the extra comma accidentally ourselves. Thanks for the fix!

eliascarv:
Yes, the comma was already there:
Captura de tela de 2023-05-14 11-12-29
Note: this print is from PR #422

PR #424, opened 2023-05-24
[WIP] Add `FarthestPath`

eliascarv:


src/paths/midpoint.jl
```diff
@@ -0,0 +1,20 @@
...
+
+sumofdists(point, points) = sum(evaluate(Euclidean(), point, p) for p in points)
+
+function traverse(object, ::MidPointPath)
+  path = [1]
+  nelms = nelements(object)
+  points = [centroid(object, i) for i in 1:nelms]
+  for _ in 2:nelms
+    pathpoints = points[path]
+    otherinds = setdiff(1:nelms, path)
```
juliohm: I would create a vector `other` before the loop with the initial list `2:nelms`. I would then pop! from this vector and push! to the `path` in the loop to avoid unnecessary allocations.

eliascarv:
Closed in favor of #427.

PR #425, opened 2023-05-25
Homogeneous sampling inside a Ball in 2D

ErickChacon:
Currently homogeneous sampling inside a `Ball` is done discretizing the `Ball`. This is not necessarily as we can easily sample homogeneously inside a ball using polar coordinates.

### DIFF
--- a/src/sampling/homogeneous.jl
+++ b/src/sampling/homogeneous.jl
@@ -43,7 +43,7 @@ end
 function sample(rng::AbstractRNG, ball::Ball{2,T}, method::HomogeneousSampling) where {T}
   function randpoint()
     u₁, u₂ = rand(rng, T, 2)
-    ball(sqrt(u₁), u₂)
+    ball(√u₁, u₂)
   end
   (randpoint() for _ in 1:(method.size))
 end
@@ -51,7 +51,7 @@ end
 function sample(rng::AbstractRNG, ball::Ball{3,T}, method::HomogeneousSampling) where {T}
   function randpoint()
     u₁, u₂, u₃ = rand(rng, T, 3)
-    ball(u₁^(1/3), acos(1 - 2u₂) / π, u₃)
+    ball(∛u₁, acos(1 - 2u₂) / T(π), u₃)
   end
   (randpoint() for _ in 1:(method.size))
 end
--- a/src/sampling/homogeneous.jl
+++ b/src/sampling/homogeneous.jl
@@ -40,22 +40,20 @@ function sample(rng::AbstractRNG, triangle::Triangle{Dim,T}, method::Homogeneous
   (randpoint() for _ in 1:(method.size))
 end
 
-function sample(rng::AbstractRNG, ball::Ball{Dim,T}, method::HomogeneousSampling) where {Dim,T}
-  if Dim == 2
-    function randpoint2()
-      u₁, u₂ = rand(rng, T, 2)
-      ball(sqrt(u₁), u₂)
-    end
-    (randpoint2() for _ in 1:(method.size))
-  elseif Dim == 3
-    function randpoint3()
-      u₁, u₂, u₃ = rand(rng, T, 3)
-      ball(u₁^(1/3), acos(1 - 2u₂) / π, u₃)
-    end
-    (randpoint3() for _ in 1:(method.size))
-  else
-    throw(ErrorException("not implemented"))
+function sample(rng::AbstractRNG, ball::Ball{2,T}, method::HomogeneousSampling) where {T}
+  function randpoint()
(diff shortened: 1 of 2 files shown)


ErickChacon:
Veryfiying that the algorithm is working, found also an issue plotting circles with `MeshViz.jl`.

```julia
D = Ball(Point(20,20), 10)
ps = sample(D, HomogeneousSampling(500)) |> collect
fig = viz(D)
viz!(ps, color = :red)
```

makie

ErickChacon:
> Veryfiying that the algorithm is working, found also an issue plotting circles with `MeshViz.jl`.

The problem is actually the `discretize` function as mentioned in JuliaGeometry/Meshes.jl#426, This pull request only solves sampling points homogeneously  inside a 2d Ball, but does not solve the issue with `discretize`.

src/sampling/homogeneous.jl
```diff
@@ -40,6 +40,20 @@ function sample(rng::AbstractRNG, triangle::Triangle{Dim,T}, method::Homogeneous
   (randpoint() for _ in 1:(method.size))
 end
 
+function sample(rng::AbstractRNG, ball::Ball{Dim,T}, method::HomogeneousSampling) where {Dim,T}
```
juliohm: Can you please reimplement this in terms of the new parametric representation of Ball? In the master branch you can ball(u, v) or ball(u, v, w) to get a point in the ball using polar or spherical coordinates.

ErickChacon: Sure, I will do that.

juliohm: Thank you! We need some tests after that 👌🏽

src/sampling/homogeneous.jl
```diff
@@ -40,6 +40,24 @@ function sample(rng::AbstractRNG, triangle::Triangle{Dim,T}, method::Homogeneous
...
+  if Dim == 2
+    function randpoint2()
+      u₁, u₂ = rand(rng, T, 2)
+      ball(sqrt(u₁), u₂)
```
juliohm: Hi @ErickChacon can you clarify the sqrt?

src/sampling/homogeneous.jl
```diff
@@ -40,6 +40,24 @@ function sample(rng::AbstractRNG, triangle::Triangle{Dim,T}, method::Homogeneous
...
+    end
+    (randpoint2() for _ in 1:(method.size))
+  elseif Dim == 3
+    function randpoint3()
+      u₁, u₂, u₃ = rand(rng, T, 3)
+      ball(u₁^(1/3), acos(1 - 2u₂) / π, u₃)
```
juliohm: also, these expressions in terms of u1, u2, u3. We need to stretch these dimensions so that points are distributed "homogeneously"?

src/sampling/homogeneous.jl
ErickChacon: Yes. In order to simulate homogeneusly, we need the marginal distributions for the radio and the angle. It can be demonstrated that the marginal distribution for the radio is $2r/R^2$ which can be simulated as $\sqrt{U}$ using the inverse of the cumulative distribution function.

src/sampling/homogeneous.jl
```diff
@@ -40,6 +40,24 @@ function sample(rng::AbstractRNG, triangle::Triangle{Dim,T}, method::Homogeneous
   (randpoint() for _ in 1:(method.size))
 end
 
+function sample(rng::AbstractRNG, ball::Ball{Dim,T}, method::HomogeneousSampling) where {Dim,T}
+  if Dim == 2
```
juliohm: Let's use multiple-dispatch with Dim for better clarity. That is, two method definitions one for Ball{2,T} and another for Ball{3,T}

src/sampling/homogeneous.jl
ErickChacon: The same as in 2D. For example, in the 3D case, it can be shown that the marginal distribution of the radio is $3r^2/R^3$ which has a CDF as $r^3/R^3$ so it can be simulated using $RU^{1/3}$.

juliohm (changes requested):
Thank you @ErickChacon , can you add some tests now? We will be able to merge afterwards.

ErickChacon:
> Thank you @ErickChacon , can you add some tests now? We will be able to merge afterwards.

Yes, I was doing that but there was a problem to keep the Float32 type for the 3D case. It is sorted now using `cbrt`.

ErickChacon:
@juliohm, I think this PR is completed now. I added tests, and previous tests are working adequately too.

juliohm (approved):

PR #427, opened 2023-05-26
Add `MultiGridPath`

eliascarv:


src/paths/multigrid.jl
```diff
@@ -0,0 +1,25 @@
...
+
+to-do
+"""
+struct MultiGridPath <: Path end
+
+function traverse(grid::Grid{Dim}, ::MultiGridPath) where {Dim}
+  path = Int[]
+  sizes = size(grid)
+  steps = sizes .- 1
+  nelems = nelements(grid)
```
juliohm:
```suggestion
  nelems = prod(sizes)
```

src/paths/multigrid.jl
```diff
@@ -0,0 +1,25 @@
...
+  linear = LinearIndices(sizes)
+  while length(path) < nelems
+    cinds = CartesianIndices(ntuple(d -> 1:steps[d]:sizes[d], Dim))
```
juliohm:
```suggestion
    ranges = ntuple(d -> 1:steps[d]:sizes[d], Dim)
    cinds = CartesianIndices(ranges)
```

src/paths/multigrid.jl
```diff
@@ -0,0 +1,25 @@
...
+    inds = [linear[cind] for cind in cinds]
+    append!(path, setdiff(inds, path))
```
juliohm:
```suggestion
    linds = [linear[cind] for cind in cinds]
    append!(path, setdiff(linds, path))
```

src/paths/multigrid.jl
```diff
@@ -0,0 +1,25 @@
...
+    steps = ceil.(Int, steps ./ 2)
```
juliohm, line 35: Worth comparing with steps \div 2

eliascarv: For non square grids, this will break the code:
```
julia> 1 ÷ 2
0

julia> ceil(Int, 1 / 2)
1
```

src/paths/multigrid.jl
```diff
@@ -0,0 +1,26 @@
...
+"""
+    MultiGridPath()
+
+to-do
+"""
+struct MultiGridPath <: Path end
+
+function traverse(grid::Grid{Dim}, ::MultiGridPath) where {Dim}
+  path = Int[]
+  sizes = size(grid)
```
eliascarv: I think `dims` is betrer than `sizes`. What do you think @juliohm?

juliohm: Yes, agree here.

src/paths/multigrid.jl
```diff
@@ -0,0 +1,26 @@
...
+  path = Int[]
+  dims = size(grid)
+  steps = dims .- 1
+  nelems = prod(dims)
+  linear = LinearIndices(dims)
+  while length(path) < nelems
+    ranges = ntuple(d -> 1:steps[d]:dims[d], Dim)
+    cinds = CartesianIndices(ranges)
+    linds = [linear[cind] for cind in cinds]
+    append!(path, setdiff(linds, path))
```
juliohm: Let's avoid this allocation here. Simply loop over cinds, convert to linear and check if it is already in the path. It it is not, then push!.

src/paths/multigrid.jl
```diff
@@ -0,0 +1,26 @@
...
+"""
+struct MultiGridPath <: Path end
+
+function traverse(grid::Grid{Dim}, ::MultiGridPath) where {Dim}
+  path = Int[]
+  dims = size(grid)
+  steps = dims .- 1
+  nelems = prod(dims)
+  linear = LinearIndices(dims)
+  while length(path) < nelems
```
juliohm:
```suggestion
  dims = size(grid)
  nelem = prod(dims)
  
  steps = dims .- 1
  linear = LinearIndices(dims)
  while length(path) < nelem
```

test/paths.jl
```diff
@@ -40,4 +40,19 @@
...
+  @testset "MultiGridPath" begin
+    path = MultiGridPath()
+    grid = CartesianGrid{T}(3, 3)
+    @test traverse(grid, path) == [1, 3, 7, 9, 2, 4, 5, 6, 8]
+
+    if visualtests
+      for d in (5, 7, 9)
+        grid = CartesianGrid{T}(d, d)
+        quads = [grid[i] for i in traverse(grid, path)]
+        fig = viz(quads, color=1:length(quads))
```
juliohm, line 53:
```suggestion
        vgrid = view(grid, traverse(grid, path))
        fig = viz(vgrid, color=1:nelements(vgrid))
```

PR #431, opened 2023-05-26
Regular sampling 

### LINKED ISSUE #426: Fix RegularSampling for Ball
`discretize` a 2D Ball is not working adequately. For example, when sampling points homogeneously inside a ball returns incorrect points.

```julia
D = Ball(Point(20,20), 10)
ps = sample(D, HomogeneousSampling(500)) |> collect
viz(D)
viz!(ps, color = :red)
```
incorrect

ErickChacon:
Fix #426.

However further discussion is required to tidy up our regular sampling implementation.

@juliohm, as you can see the issue can be easily fixed by defining an adequate scaling for points. However, we do not use the `Ball` interface and only rely on sampling over a `Sphere`. Is this adequate, or would it be better to have the implementation for `Ball` independent of `Sphere`?

### DIFF
--- a/src/primitives/ball.jl
+++ b/src/primitives/ball.jl
@@ -38,11 +38,11 @@ area(b::Ball{2}) = measure(b)
 
 volume(b::Ball{3}) = measure(b)
 
-function Base.in(p::Point, b::Ball)
-  x = coordinates(p)
-  c = coordinates(b.center)
+function Base.in(p::Point{Dim,T}, b::Ball{Dim,T}) where {Dim,T}
+  c = b.center
   r = b.radius
-  sum(abs2, x - c) ≤ r^2
+  s = norm(p - c)
+  s < r || isapprox(s, r, atol = atol(T))
 end
 
 boundary(b::Ball) = Sphere(b.center, b.radius)
--- a/src/sampling/regular.jl
+++ b/src/sampling/regular.jl
@@ -56,7 +56,7 @@ function sample(::AbstractRNG, ball::Ball{Dim,T}, method::RegularSampling) where
   # reuse samples on the boundary
   points = sample(Sphere(c, r), RegularSampling(sz[1:(Dim - 1)]))
 
-  scale(p, s) = Point(s * coordinates(p))
+  scale(p, s) = c + s * (p - c)
 
   ivec(scale(p, s) for p in points, s in srange)
 end
--- a/test/primitives.jl
+++ b/test/primitives.jl
@@ -258,9 +258,17 @@
     @test b(T(0), T(0)) ≈ P2(0, 0)
     @test b(T(1), T(0)) ≈ P2(2, 0)
 
+    b = Ball(P2(7, 7), T(1.5))
+    ps = b.(1, rand(T, 100))
+    all(∈(b), ps)
+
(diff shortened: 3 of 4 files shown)


ErickChacon:
This is the comparison with previous scaling:

scaling-wrong

and with the current scaling:

scaling-right

juliohm (approved):

juliohm:
That is great for now @ErickChacon , it fixes the issue. Can you also add a test so that we don't fall into the same trap again in the future after refactoring?

ErickChacon:
Also adding a comparison for 3D for other people as a reference. The Ball is defined with center at `(50,50,50)` but we can see that the result is not centered in this value.

scaling-wrong-3d

With the fix, we can see that the samples are now around the defined center `50,50,50)`

scaling-right-3d

juliohm:
@ErickChacon as soon as you add the tests, we can do a final review and merge 💯

ErickChacon:
> @ErickChacon as soon as you add the tests, we can do a final review and merge 100

Sure. I am checking something else with respect of the values of the radius for Ball in 3D. For example when I provide `n = 2`, it just uses one value for the radius, but when I use `n=3` it uses 3 values for the radius.

juliohm:
I don't recall exactly the convention, but I think we did choose these numbers in order to always be able to connect the dots in the RegularDiscretization routine. Basically, we sample points regularly along the parametric dimensions and connect them.

ErickChacon:
Ok, I just will add some tests then. 

Edit:

I added some tests but in the 3D case not all points are lying inside the the Ball. I will check what is the problem.

juliohm:
> I added some tests but in the 3D case not all points are lying inside the the Ball. I will check what is the problem.

Ok, let me know when the PR is ready for review again.

ErickChacon:
@juliohm, the reason why not all points lie inside the Ball in 3D is just due to numerical approximation. For example generating points using our ball interface does not ensure we obtain (numerically) points inside the Ball.

```julia
ball = Ball(P3(10, 10, 10), 1.5)
all(∈(ball), ball.(1,0:0.1:1,0:0.1:1))
# false
```

Any suggestion of how to proceed? For this particular PR, it might be enough if the points lie inside the Ball with an additional increment in the radius. However, in general, it would be better if we can ensure points created with the ball interface lie inside the ball.

juliohm:
Maybe we need to review the Base.in implementation for ball? I thought it
was using some tolerance already.

Em sex., 26 de mai. de 2023 19:01, Dr. Erick A. Chacón Montalván <
***@***.***> escreveu:

ErickChacon:
> Maybe we need to review the Base.in implementation for ball? I thought it was using some tolerance already.

Yes, I think that is what we need, no tolerane is being used. I can modify that.

juliohm:
Thanks. I think we have other examples of Base.in with tolerance in the
codebase already.

Em sex., 26 de mai. de 2023 19:14, Dr. Erick A. Chacón Montalván <
***@***.***> escreveu:

ErickChacon:
@juliohm, I modified the `in` method for `Ball`. Test are passsing now.

juliohm (approved):

PR #448, opened 2023-06-06
empty collections

ErickChacon:
Allow empty collections to be able to define an empty pointset as possible output for samping in `PointPattern.jl` as mention in JuliaEarth/PointPatterns.jl/pull/32.

### DIFF
--- a/src/collections.jl
+++ b/src/collections.jl
@@ -8,7 +8,7 @@
 A collection of `items` (points or geometries) seen as a single domain.
 """
 struct Collection{Dim,T,I<:PointOrGeometry{Dim,T}} <: Domain{Dim,T}
-  items::Vector{I}
+    items::Union{Nothing,Vector{I}}
 end
 
 # constructor with iterator of items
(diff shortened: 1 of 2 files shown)


src/collections.jl
```diff
@@ -8,7 +8,7 @@
 A collection of `items` (points or geometries) seen as a single domain.
 """
 struct Collection{Dim,T,I<:PointOrGeometry{Dim,T}} <: Domain{Dim,T}
-  items::Vector{I}
+    items::Union{Nothing,Vector{I}}
```
juliohm, line 11: Can't we somehow retain the type information and avoid the Union with Nothing? After all, we just need to return a zero length elsewhere right?

ErickChacon: Not sure how to avoid the Union with Nothing. Maybe, we can just return an empy `Vector` in `PointPatterns,jl` instead of a empty `PointSet`?

The advantage of using Union, is that we can still use the methods `embeddim`, `coordtype`, `nelements`, `eltype` for 
 a empty `PointSet`.

ErickChacon:
@juliohm, I added some tests in case you think it is the way to go, otherwise we can think of an alternative approach in `PointPatterns.jl`.

src/collections.jl
juliohm: Can't we just write something like `Collection([])` instead to construct an empty vector of items and then adjust the outer constructor accordingly with type annotations to avoid infinite recursions?

I think we need to open a more general discussion on Zulip or as an issue here to figure out how we should safely handle empty Domain types, not just Collection.

juliohm:
I will close this one in favor of a more general discussion around empty domains. Let's brainstorm it more over Zulip and come back here with a more targeted plan.

PR #458, opened 2023-06-19
introduce _sampleweights for developers use

ErickChacon:
Define `_sampleweights` to allow define custom `weights` as needed in `PointPatterns.jl`.

### DIFF
--- a/src/sampling/homogeneous.jl
+++ b/src/sampling/homogeneous.jl
@@ -3,10 +3,11 @@
 # ------------------------------------------------------------------
 
 """
-    HomogeneousSampling(size)
+    HomogeneousSampling(size, [weights])
 
 Generate sample of given `size` from geometric object
-according to a homogeneous density.
+according to a homogeneous density. Optionally, provide `weights`
+to specify custom sampling weights for the elements of a domain.
 """
 struct HomogeneousSampling{W} <: ContinuousSamplingMethod
   size::Int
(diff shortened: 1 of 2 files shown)


ErickChacon:
@juliohm, let me know if you are happy with this to proceed with the changes in `PointPatterns.jl`.

src/sampling/homogeneous.jl
```diff
@@ -12,11 +12,13 @@ struct HomogeneousSampling <: ContinuousSamplingMethod
   size::Int
```
juliohm, line 13: What about a new field `weightfun::F` that is initialized with `measure` by default? Or perhaps a new field `weights::W` that is initialized with `nothing` by default and computed inside the method with `measure`?

I think we can avoid introducing an internal function and instead try to add an option to the struct method.

juliohm (commented):
@ErickChacon can we try to introduce a new field in the struct instead of introducing an internal function?

src/sampling/homogeneous.jl
```diff
@@ -10,13 +10,16 @@ according to a homogeneous density.
 """
 struct HomogeneousSampling <: ContinuousSamplingMethod
   size::Int
+  weights::Union{AbstractVector, Nothing}
```
juliohm:
```suggestion
struct HomogeneousSampling{W} <: ContinuousSamplingMethod
  size::Int
  weights::W
```

juliohm (changes requested):
We need to update the docstring and add tests with custom weights.

ErickChacon:
> We need to update the docstring and add tests with custom weights.

Done!

juliohm (approved):

juliohm:
Tests are failing for another unrelated reason. Merging it.

PR #472, opened 2023-06-22
Add assertion in Box constructor

### LINKED ISSUE #470: Assert that `pmin ⪯ pmax` in Box constructor
Our Box assumes that `pmin \preceq pmax`. We need to add an assertion to avoid undefined behavior downstream.

eliascarv:
closes #470

### DIFF
--- a/src/primitives/box.jl
+++ b/src/primitives/box.jl
@@ -20,7 +20,7 @@ struct Box{Dim,T} <: Primitive{Dim,T}
   max::Point{Dim,T}
 
   function Box{Dim,T}(min, max) where {Dim,T}
-    @assert min ⪯ max "'min' must be less than or equal to 'max'"
+    @assert min ⪯ max "`min` must be less than or equal to `max`"
     new(min, max)
   end
 end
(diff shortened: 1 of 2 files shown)


src/primitives/box.jl
```diff
@@ -11,14 +11,22 @@ See https://en.wikipedia.org/wiki/Hyperrectangle.
...
+Box(Point(0, 0, 0), Point(1, 1, 1)) # unit cube
+Box((0, 0), (1, 1))
 ```
 """
 struct Box{Dim,T} <: Primitive{Dim,T}
   min::Point{Dim,T}
   max::Point{Dim,T}
+
+  function Box{Dim,T}(min, max) where {Dim,T}
+    @assert min ⪯ max "'min' must be less than or equal to 'max'"
```
juliohm:
```suggestion
    @assert min ⪯ max "`min` must be less than or equal to `max`"
```

src/primitives/box.jl
```diff
@@ -11,14 +11,22 @@ See https://en.wikipedia.org/wiki/Hyperrectangle.
 ## Examples
 
 ```julia
-Box(Point(0,0,0), Point(1,1,1)) # unit cube
+Box(Point(0, 0, 0), Point(1, 1, 1)) # unit cube
```
juliohm:
```suggestion
Box(Point(0, 0, 0), Point(1, 1, 1))
```

juliohm (approved):

Issue #488, opened 2023-06-29
Performance improvements 🚀


juliohm:
This issue tracks our efforts to improve the performance of different algorithms and data structures implemented in the project. The public API is stabilizing and there is only one major breaking change planned before a possible v1.0.

Given the solid design that separates user-facing functions from internals, we can easily refactor the code without breaking code downstream. We kindly ask the community to add comments here whenever they encounter a performance problem.

juliohm:
Copying from Zulip:

> One slightly breaking change I would consider is changing the vararg constructors of geometries to use static arrays (possibly mutable). This would make it much easier to write performant code that handles a lot of small geometries (e.g. segments or triangles) and avoid code such as Segment(view(v, [i, i + 1])) that currently exists (which in this case allocates the index vector plus possibly the view, afaik). I think I also have a general expectation that the arguments of a constructor are stored within the struct and not allocated, so having Triangle(a, b, c) allocate [a, b, c] is rather surprising to me and having expressions such as centroid(Segment(a, b)) and area(Triangle(a, b, c)) perform well would be very desirable imo. The only drawback that I see is that the number of points becomes fixed, which is arguably a good thing for Ngon but maybe not in the case of Chain/Bezier (though constructing them with a regular vector would still be possible of course).

comment by @mfsch

juliohm:
That is something we have in mind. In the past we had considered `StaticVector` for the list of vertices inside some of the `Polytope` types, but refactored and erased this. We may need to reintroduce this optimization and use `NTuple` directly whenever possible. Some of our `Polytope` types like `PolyArea` and `Chain`'s subtypes require dynamic vectors due to the use cases.

BloodWorkXGaming:
Related to this, I found another potential bottleneck: segments() function for chains is slower by about a factor of 6 than the "manual" version with StaticArrays:
This small change had an improvement from 30 to 5 seconds in my case:

```
# @views for seg in segments(chain)
@views for (p1, p2) in zip(verts[1:n-1], verts[2:n])
        seg = Segment(SA[p1, p2])
       .... do stuff ....
end
 ```

juliohm:
Thanks we are in the process of rewriting some internals in terms of tuples
and static vectors. Also adding more threads to speed things up.

If you have performance suggestions please submit PRs in places you already
have working code.

Em sex., 7 de jul. de 2023 14:48, Jonas ***@***.***> escreveu:

mfsch:
[...]

PR #501, opened 2023-07-08
WIP: Reduced allocations of boundingbox implementation + Tests for allocations

BloodWorkXGaming:
Hello,

as discussed in  #488 there were a lot of unnecessary allocations in the boundingbox implementations.
I added some custom implementations for some types which removes those allocations.

Furthermore, I added tests to check if the implementations are allocation free. This however required some changes to the setup of the tests due to the f32 and f64 checks. 
The Types T and P1, P2, etc. were global non-const variables and therefore caused allocations due to the type checker.
I've split it up in two modules + one quote for easier setup.

Another addition are the allocation check macros, which might further help verify the future performance improvement efforts.


src/boundingboxes.jl
```diff
@@ -40,6 +40,17 @@ end
 
 boundingbox(g::Grid) = Box(extrema(g)...)
 
+boundingbox(m::SimpleMesh) = boundingbox(vertices(m))
```
juliohm, line 43:
```suggestion
boundingbox(m::Mesh) = boundingbox(vertices(m))
```

src/boundingboxes.jl
```diff
@@ -40,6 +40,17 @@ end
...
+
+function boundingbox(p::PolyArea)
+  box = boundingbox(vertices(p.outer))
+  for inner in p.inners
+      box = boundingbox(vertices(inner), coordinates(box.min), coordinates(box.max))
+  end
+  box
+end
```
juliohm, line 51: boundingbox(p::Polytope) = boundingbox(vertices(p))

src/boundingboxes.jl
```diff
@@ -49,13 +60,12 @@ boundingbox(geoms::AbstractVector{<:Geometry}) = boundingbox(boundingbox.(geoms)
...
+function boundingbox(points::AbstractVector{Point{Dim,T}}, xmin::Vec{Dim, T} = Vec(ntuple(i -> typemax(T), Dim)), xmax::Vec{Dim, T} = Vec(ntuple(i -> typemin(T), Dim))) where {Dim,T}
   for p in points
     x = coordinates(p)
-    @. xmin = min(x, xmin)
-    @. xmax = max(x, xmax)
+    xmin = min.(x, xmin)
+    xmax = max.(x, xmax)
   end
   Box(Point(xmin), Point(xmax))
 end
```
juliohm, line 70: No need to write this method

juliohm (changes requested):
Please split this PR into smaller PRs that do one single thing. Check our contribution guidelines. This PR is adding specialization methods for boundingbox, which is great, but is also doing a great amount of refactoring to the tests.

src/boundingboxes.jl
BloodWorkXGaming: Yes, I am aware of that, however this forces unnecessary allocations, as vertices() of a polyarea has to collect all inner chains into an intermediate long vector of points. This saves these allocations.

Another optimization could be completely ignoring the inner vertices, as they are always inside the outer ring. But this could cause issues on malformed polygons

src/boundingboxes.jl
BloodWorkXGaming: this is just a change of the existing method with less allocations, not a new method
The only addition is, that it allows chaining boundingbox calls without needing a intermediate array for multiple boxes.

src/boundingboxes.jl
juliohm: The optimization with just the outer ring seems reasonable. Can we use that instead? We assume by design that inner rings are in the interior of the region defined by the outer ring.

src/boundingboxes.jl
juliohm: I fully agree with the usage of `@.` to reduce allocations of the existing method, but the change of arguments to accept an extra xmin and xmax seems unnecessary at this point in time.

We can do another future round of review and improvements after we fix the internal representation of vertices, which is more pressing.

juliohm:
Closing this one in favor of #505

PR #502, opened 2023-07-08
Add ray-tetrahedron intersection method

### LINKED ISSUE #290: intersection(::Ray, ::Mesh) broken
As the title says, this method is broken. Below is a MWE:

```Julia
julia> tet = Tetrahedron(Point(0,0,0), Point(1,0,0), Point(0,1,0), Point(1,1,1))
Tetrahedron{3,Float64}
  └─Point(0.0, 0.0, 0.0)
  └─Point(1.0, 0.0, 0.0)
  └─Point(0.0, 1.0, 0.0)
  └─Point(1.0, 1.0, 1.0)

julia> r = Ray(Point(0.5,0.5,-1), Vec(0.,0,2))
Ray{3, Float64}(Point(0.5, 0.5, -1.0), [0.0, 0.0, 2.0])

julia> intersection(r,tet)
ERROR: StackOverflowError:
Stacktrace:
 [1] intersection(f::Function, g1::Ray{3, Float64}, g2::Tetrahedron{3, Float64, StaticArraysCore.SVector{4, Point3}}) (repeats 2 times)
   @ Meshes ~/.julia/dev/Meshes/src/intersections.jl:37
```

I don't know enough to figure out why it says repeats 2 times.

kylebeggs:
Intended to close #290. This really can be used for any Polyhedron, perhaps we should not restrict the type here?

Using `Multi` here gives a stack overflow 🙃 thats why it is commented out for now.


src/intersections/rays.jl
```diff
@@ -205,4 +205,14 @@ function intersection(f, ray::Ray{3,T}, tri::Triangle{3,T}) where {T}
   return @IT Crossing ray(λ) f
 end
 
+function intersection(f, r::Ray{3,T}, p::Tetrahedron{3,T}) where {T}
+  mesh = boundary(p)
+  intersections = intersection.((r,), elements(mesh))
```
juliohm, line 210: Shouldn't the intersection also take into account the interior of the tetrahedron? If a ray passes through a tetrahedron, then we may get a segment? This intersection case is really difficult to express with our current representation. I know that Polyhedra.jl and other packages pursue other representations that facilitate some of these operations.

kylebeggs: Yeah I suppose I was considering the tetrahedron as the surfaces only. I come from the numerical PDEs world where you'd only really care about where the ray would intersect with the faces. Help me consider the cases:

1. ray goes through volume creating a segment (this could also have further cases where if the segment endpoint is on a face, edge, or vertex. How specific do we want to be?)
2. ray colinear with edge
3. ray origin touches face, edge, vertex only
4. ?

juliohm: Let's start with a reduced list of cases, we don't need to differentiate the cases inside case (1), we just need to return the segment.

I am wondering how you would use this intersection result in the PDE world. Maybe that can guide us on the most appropriate cases.

kylebeggs: I suppose what I really meant was in numerical PDES (well, at least mesh-based schemes, especially FV), you only care about the faces and don't ever really consider the cell's volume (other than the literal value of the cell volume). So coming from this mindset (not necessarily right or wrong), naturally I'd expect/want an intersection like this to just give me the points the ray intersects the faces. If you were writing a simple toy FV code, you probably wouldn't be using this type of function at all anyways. Just trying to explain the mindset/expectations/though process, but it could be just me. Daniel VandenHeuvel could give a much better answer (seems he is active on Zulip) as he has developed some nice FV codes, my work is mainly in meshless PDE methods.

kylebeggs: Apologies for dragging my feet on this - I have been slammed busy lately.

juliohm: So what you are saying is that what you had in mind is really the intersection between the boundary of the tetrahedron and the ray. The boundary of the tetrahedron is a SimpleMesh, so maybe we need a method for Ray and SimpleMesh of Triangle instead?

PR #503, opened 2023-07-10
Refactor internal representation of vertices in Polytope types

eliascarv:


src/polytopes/segment.jl
```diff
@@ -11,8 +11,16 @@ The segment can be called as `s(t)` with `t` between
...
-struct Segment{Dim,T,V<:AbstractVector{Point{Dim,T}}} <: Chain{Dim,T}
-  vertices::V
+struct Segment{Dim,T} <: Chain{Dim,T}
+  vertices::NTuple{2,Point{Dim,T}}
+  Segment{Dim,T}(vertices::NTuple{2,Point{Dim,T}}) where {Dim,T} = new(vertices)
+end
+
+function Segment(vertices::AbstractVector{Point{Dim,T}}) where {Dim,T}
+  if length(vertices) ≠ 2
+    throw(ArgumentError("The number of vertices must be 2"))
```
```suggestion
  n = length(vertices)
  if n ≠ 2
    throw(ArgumentError("Invalid number of vertices for Segment. Expected 2, got $n"))
```

src/polytopes/segment.jl
```diff
@@ -11,8 +11,16 @@ The segment can be called as `s(t)` with `t` between
...
+  end
+  Segment{Dim,T}((first(vertices), last(vertices)))
```
```suggestion
  Segment{Dim,T}(Tuple(vertices))
```

juliohm (changes requested):
Review attached.

src/polytopes/segment.jl
```diff
@@ -11,8 +11,17 @@ The segment can be called as `s(t)` with `t` between
...
+  vertices::NTuple{2,Point{Dim,T}}
+  Segment{Dim,T}(vertices::NTuple{2,Point{Dim,T}}) where {Dim,T} = new(vertices)
+end
+
+function Segment(vertices::AbstractVector{Point{Dim,T}}) where {Dim,T}
+  n = length(vertices)
+  if n ≠ 2
+    throw(ArgumentError("Invalid number of vertices for Segment. Expected 2, got $n"))
+  end
+  Segment{Dim,T}(NTuple{2,Point{Dim,T}}(vertices))
```
```suggestion
  N = length(vertices)
  N == 2 || throw(ArgumentError("Invalid number of vertices for Segment. Expected 2, got $N."))
  v = NTuple{2,Point{Dim,T}}(vertices)
  Segment{Dim,T}(v)
```

juliohm (commented):
Now we need to do the same with our Ngon type.

src/polytopes/ngon.jl
```diff
@@ -20,11 +20,19 @@ are `Triangle` (N=3), `Quadrangle` (N=4), `Pentagon` (N=5), etc.
...
-  vertices::V
+struct Ngon{N,Dim,T} <: Polygon{Dim,T}
+  vertices::NTuple{N,Point{Dim,T}}
+  Ngon{N,Dim,T}(vertices::NTuple{N,Point{Dim,T}}) where {N,Dim,T} = new(vertices)
 end
 
-Ngon{N}(vertices::AbstractVector{Point{Dim,T}}) where {N,Dim,T} = Ngon{N,Dim,T,typeof(vertices)}(vertices)
+function Ngon{N,Dim,T}(vertices::AbstractVector{Point{Dim,T}}) where {N,Dim,T}
+  nv = length(vertices)
+  nv == N || throw(ArgumentError("Invalid number of vertices for Ngon. Expected $N, got $nv."))
```
juliohm:
```suggestion
  P = length(vertices)
  P == N || throw(ArgumentError("Invalid number of vertices for Ngon. Expected $N, got $P."))
```

src/polytopes/ngon.jl
```diff
@@ -47,7 +55,9 @@ Base.unique!(ngon::Ngon) = ngon
 nvertices(::Type{<:Ngon{N}}) where {N} = N
 nvertices(ngon::Ngon) = nvertices(typeof(ngon))
 
-rings(ngon::Ngon{N}) where {N} = [Ring(ngon.vertices)]
+vertices(ngon::Ngon) = collect(ngon.vertices)
+
+rings(ngon::Ngon{N}) where {N} = [Ring(vertices(ngon))]
```
juliohm:
```suggestion
rings(ngon::Ngon) = [Ring(vertices(ngon))]
```

juliohm (changes requested):
We need to introduce a macro `@polytope Segment 1 2` that defines the new polytope type `Segment` as a `Polytope{1}` subtype with `2` vertices. This macro can be introduced right after the definition of `Polytope` in our polytopes.jl source file and used extensively in other files.

That way we will be able to iterate over different designs more quickly and will be able to maintain these internal design decisions in a single place.

Let's also try to reduce the number of outer construtors to a minimum. Please report cases where code is depending on either (1) constructors of the form `PL{Dim,T}(vertices)` or (2) `PL(::AbstractVector)`. We will probably need to refactor the code to avoid allocations.

src/polytopes/segment.jl
```diff
@@ -22,6 +22,9 @@ isclosed(::Type{<:Segment}) = false
 isparametrized(::Type{<:Segment}) = true
 
 nvertices(::Type{<:Segment}) = 2
+nvertices(s::Segment) = nvertices(typeof(s))
```
juliohm: Maybe we should replace the fallback definition for Polytope now that most vertices are static. Can you add this implementation of nvertices to Polytope and erase it from here and from Ngon?

juliohm (changes requested):
That is awesome work @eliascarv. Just a small change before we can merge. Excited to continue the refactoring of these internals in a separate PR.

PR #505, opened 2023-07-12
Remove `collect` from `vertices` function

eliascarv:


src/transforms/repair.jl
```diff
@@ -82,7 +82,7 @@ function apply(::Repair{8}, poly::PolyArea)
 end
 
 function apply(::Repair{8}, poly::Ngon)
-  v = poly |> vertices |> repair8
+  v = poly |> vertices |> collect |> repair8
```
juliohm: no need to collect in case of repair8 with Ngon. We need to generalize the repair8 function to receive iterables.

src/pointification.jl
```diff
@@ -27,3 +27,7 @@ pointify(d::Data) = pointify(domain(d))
 
 # fallback with iterator of geometries
 pointify(geoms) = mapreduce(pointify, vcat, geoms)
+
+# utils
+_asvector(itr) = collect(itr)
+_asvector(vector::AbstractVector) = vector
```
juliohm: no need to create this additional function, just use collect directly on vector as well.

eliascarv: This utility function is to avoid unnecessary `collect`s, because the vertices of `PolyArea`, `Ring` and `Rope` are `AbstractVector`s.

src/transforms/repair.jl
```diff
@@ -91,6 +91,7 @@ function apply(::Repair{8}, ring::Ring)
   Ring(v), nothing
 end
 
+repair8(itr) = repair8(collect(itr))
```
juliohm:
```suggestion
repair8(v) = repair8(collect(v))

```

src/boundingboxes.jl
```diff
@@ -46,10 +46,13 @@ boundingbox(g::Grid) = Box(extrema(g)...)
 
 boundingbox(geoms::AbstractVector{<:Geometry}) = boundingbox(boundingbox.(geoms))
 
-boundingbox(boxes::AbstractVector{<:Box{Dim}}) where {Dim} =
-  boundingbox([point for box in boxes for point in extrema(box)])
+boundingbox(boxes::AbstractVector{Box{Dim,T}}) where {Dim,T} =
```
```suggestion
boundingbox(boxes::AbstractVector{<:Box}) =
```

juliohm (approved):

PR #513, opened 2023-07-17
Add generic fallback to `discretize` function

eliascarv:


src/discretization/regular.jl
```diff
@@ -18,11 +18,17 @@ end
...
+function discretize(box::Box, method::RegularDiscretization)
+  sz = fitdims(method.sizes, paramdim(box))
+  ip = isperiodic(box)
   np = @. sz + !ip
 
-  points = sample(geometry, RegularSampling(np))
+  points = sample(box, RegularSampling(np))
   topo = GridTopology(sz, ip)
 
   SimpleMesh(collect(points), topo)
```
juliohm: We need a simple branch here with a generic Geometry:

```julia
if isparametrized(geometry)
  # old implementation
else
  # bounding box method
end
```

juliohm (approved):

PR #518, opened 2023-07-19
Fix method ambiguities

eliascarv:


src/intersections.jl
```diff
@@ -233,6 +229,19 @@ hasintersect(d::Domain, g::Geometry) = hasintersect(d, [g])
...
+
+hasintersect(c::Chain, p::Point) = hasintersect(p, c)
+
+hasintersect(p::Point, m::Multi) = p ∈ m
+
+hasintersect(m::Multi, p::Point) = hasintersect(p, m)
+
+hasintersect(m::Multi, c::Chain) = hasintersect(collect(m), segments(c))
+
+hasintersect(c::Chain, m::Multi) = hasintersect(m, c)
```
juliohm: please flip the order of these two methods

juliohm (approved):

PR #540, opened 2023-08-01
Add comment about predicates being approximations

DanielVandH:
This PR adds a remark to the start of the predicates section that mentions that the predicates are approximations. I also link to ExactPredicates.jl just so that people know that exact predicates do exist in Julia if they do actually need them.


juliohm:
Wonderful! ❤️

PR #544, opened 2023-08-04
Added macro for allocations checking and refactored tests to support it

BloodWorkXGaming:
Hello :)

Akin to my last PR I refactored the test-suite to allow for simpler and more precise allocation testing.

The current test-setup has multiple problems, which causes allocations to be impreceise:
- It acts in the global Main module, which always causes more allocations than necessary (unlike running it in a function)
- It uses non-constant global variables (like T, P1, P2, etc.) which cause runtime dispatches and therefore unncesecary allocations

Keeping the P1, P2, etc. functionality required splitting it into two submodules, which allows const types of the same name.
The quote at the start is simply a shortcut for reducing the codeduplication.

Caveat/Gotcha with the @testnoallocations macro: It will always call the code twice to force a precompile (which again causes irrelevant allocations.). Take this into account for functions with sideeffects or use @testnoallocations_noprecompile

For boundingbox I replaced the old allocation checks with the new marco where possible. I refrained from adding it to PointSets, which seem to cause allocations while calculating.

I hope this PR is better split-up than the last :)


test/runtests.jl
```diff
@@ -1,131 +1,162 @@
...
-    println("Testing $testfile...")
-    include(testfile)
+module TestsDoublePrecision
+  using ..Main: run_setup, DummyData
+
+  eval(run_setup)
+
+  const T::Type = Float64
+  const P1::Type, P2::Type, P3::Type = Point{1,T}, Point{2,T}, Point{3,T}
+  const V1::Type, V2::Type, V3::Type = Vec{1,T}, Vec{2,T}, Vec{3,T}
```
juliohm, line 151:
```suggestion
  const T = Float64
  const P1, P2, P3 = Point{1,T}, Point{2,T}, Point{3,T}
  const V1, V2, V3 = Vec{1,T}, Vec{2,T}, Vec{3,T}
```

test/runtests.jl
```diff
@@ -1,131 +1,162 @@
...
-    println("Testing $testfile...")
-    include(testfile)
+module TestsSinglePrecision
+  using ..Main: run_setup, DummyData
+
+  eval(run_setup)
+
+  const T::Type = Float32
+  const P1::Type, P2::Type, P3::Type = Point{1,T}, Point{2,T}, Point{3,T}
+  const V1::Type, V2::Type, V3::Type = Vec{1,T}, Vec{2,T}, Vec{3,T}
```
juliohm, line 126:
```suggestion
  const T = Float32
  const P1, P2, P3 = Point{1,T}, Point{2,T}, Point{3,T}
  const V1, V2, V3 = Vec{1,T}, Vec{2,T}, Vec{3,T}
```

test/allocation_macros.jl
```diff
@@ -0,0 +1,64 @@
+# Helper Macros for checking that a function doesn't allocate memory
+# https://forem.julialang.org/dpinol/detecting-test-allocated-gotchas-34op
+
+macro testnoallocations(expressions...)
+    # Uncomment the following line if non-const globals should be prohibited
+    _failIfNonConstGlobalsInExpressions(__module__, expressions...)
```
juliohm, line 6: please use our code style for all function names and variables. we don't use camel case, and avoid underscores whenever possible.

test/allocation_macros.jl
```diff
@@ -0,0 +1,64 @@
...
+    return esc(
+        quote
+            !@isCalledFromFunction() &&
```
juliohm, line 9: indentation is 2 spaces

juliohm (changes requested):
Hi @BloodWorkXGaming ,  appreciate your attempt to improve our test suite.

Can you organize these changes a bit better? If you can find a cleaner method to reorganize our test files, or if you can encapsulate the proposed macro into a package, we are happy to import the package and adjust our tests with suggested good practices.

We care a lot about readability of code and tests, and the present PR doesn't meet our standards.

juliohm:
Closing this as it is not compliant with the project standards. We are happy to assess a utility package with the proposed macro, and to use it in our test suite with minimal changes.

PR #546, opened 2023-08-04
Special case issubset for equal geometries

DanielVandH:
This PR special cases `issubset` so that equal geometries are correctly identified as being subsets of each other.


test/predicates.jl
```diff
@@ -70,6 +73,16 @@
...
+
+    p1 = P2(-1.0, 0.0)
+    p2 = P2(0.0, 0.0)
+    p3 = P2(1.0, 0.0)
+    L1 = Line(p1, p3)
+    L2 = Line(p2, p3)
+    @test L1 ⊆ L2
+    @test L2 ⊆ L1 
+    @test L1 ⊆ L1 
+    @test L2 ⊆ L2
```
juliohm:
```suggestion
    l1 = Line(p1, p3)
    l2 = Line(p2, p3)
    @test l1 ⊆ l2
    @test l2 ⊆ l1 
    @test l1 ⊆ l1 
    @test l2 ⊆ l2
```

juliohm (approved):
It makes total sense. I will just try to remove some undesired == methods from the codebase before merging this one.

PR #552, opened 2023-08-07
Move collinear/coplanar predicates from miscellaneous to predicates

DanielVandH:
Makes more sense now that the predicates section exists to have these here.

### DIFF
--- a/docs/src/miscellaneous.md
+++ b/docs/src/miscellaneous.md
@@ -5,7 +5,5 @@ can be useful to end-users.
 
 ```@docs
 householderbasis
-iscollinear
-iscoplanar
 sideof
 ```
--- a/docs/src/predicates.md
+++ b/docs/src/predicates.md
@@ -54,6 +54,18 @@ isclosed
 isconvex
 ```
 
+## iscollinear 
+
+```@docs 
+iscollinear 
+``` 
+
+## iscoplanar 
+
+```@docs 
+iscoplanar 
+```
+
 ## issimple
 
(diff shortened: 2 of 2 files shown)


juliohm:
Good catch! Can you also move the implementations to the predicates folder and move the exports to near the predicate exports?

Maybe these two predicate should show up at the bottom of the list of predicates given that they aren't unary nor binary, but take 3 or more arguments.

DanielVandH:
Done.
e: Or not. Should've properly refreshed the changes when I ran the code again :-)

DanielVandH:
I also moved the tests that were for `iscoplanar` and `iscollinear` into the predicates folder. There was only one test given for these. Perhaps there should be more added at some point (not in this PR)?

src/predicates.jl
```diff
@@ -15,3 +15,7 @@ include("predicates/hasholes.jl")
 include("predicates/in.jl")
 include("predicates/issubset.jl")
 include("predicates/intersects.jl")
+
+# other predicates
```
juliohm:
```suggestion
```

juliohm (approved):

juliohm:
LGTM. I will just wait for the tests to pass.

We should definitely add more tests 👍🏽

Feel free to add them here or in a separate PR.

juliohm:
Sorry, I removed the comment about "other predicates" imagining it was in the main Meshes.jl file. It is perfect, re-added it.

PR #571, opened 2023-08-22
Weiler-Atherton clipping algorithm

ClaroHenrique:
Test images:

<details> <summary> Test 1</summary>

image

</details>

<details> <summary> Test 2</summary>

image

</details>

<details> <summary> Test 3</summary>

image

</details>

<details> <summary> Test 4 </summary>

image

</details>

<details> <summary> Test 5</summary>

image

</details>

<details> <summary> Test 6 </summary>

image

</details>

<details> <summary> Test 7 </summary>

image

</details>

<details> <summary> Test 8</summary>

image

</details>

<details> <summary> Test 9 </summary>

image

</details>

<details> <summary> Test 10 </summary>

image

</details>


src/clipping/weileratherton.jl
```diff
@@ -0,0 +1,146 @@
...
+  r = filter(!isnothing, vcat(c...))
+
+  if isempty(r)
+    nothing
+  elseif length(r) > 1
+    Multi(PolyArea.(r))
+  else
+    PolyArea(r[1])
+  end
+end
```
juliohm: This method is already available in src/clipping.jl, we only need to add the method with rings in this file.

src/clipping/weileratherton.jl
```diff
@@ -0,0 +1,146 @@
...
+
+function clip(ring::Ring{Dim,T}, other::Ring{Dim,T}, ::WeilerAtherton) where {Dim,T}
+  # assumes both ring and other are CCW
+  # TODO: Corner case without intersection
+  # TODO: Corner case with colinear segments with intersections
+  # TODO: Poly point on boundary
+  # TODO: PolyArea with holes
+  # TODO: Optimize findpointindex
+
+  vᵣ, kindᵣ = verticesswithintersections(ring, other)
```
juliohm: rename function to _pairwiseintersections(ring, other)

juliohm (changes requested):

src/clipping/weileratherton.jl
```diff
@@ -0,0 +1,203 @@
...
+
+* Weiler, Kevin and Atherton, Peter. 1977. [Hidden surface removal using polygon area sorting]
+  (https://dl.acm.org/doi/10.1145/563858.563896)
+"""
+struct WeilerAtherton <: ClippingMethod end
+
+function clip(poly::Polygon{Dim,T}, other::Geometry{Dim,T}, method::WeilerAtherton) where {Dim, T}
+  if hasholes(poly)
+    outers = clip(rings(poly)[1], boundary(other), method)
+    inners = vcat([clip(r, boundary(other), method) for r in rings(poly)[2:end]]...)
```
juliohm, line 20: Try to avoid branching here. You can simply apply the method to all rings using a single vector.

Then later you can decide what to do based on the length of the vector of rings.

src/clipping/weileratherton.jl
```diff
@@ -0,0 +1,203 @@
...
+          push!(ringsₒ, i)
+        end
+      end
+      push!(g, PolyArea(ringsₒ))
+    end
+    Multi(g)
+  else
+    outers = clip(rings(poly)[1], boundary(other), method)
+    if isempty(outers)
+      nothing
```
juliohm, line 41: I have the impression that this code could be much cleaner. @eliascarv can you take a look later and help with the cleanup?

src/clipping/weileratherton.jl
```diff
@@ -0,0 +1,203 @@
...
+      Multi(PolyArea.(outers))
+    end
+  end
+end
+
+function clip(ring::Ring{Dim,T}, other::Ring{Dim,T}, ::WeilerAtherton) where {Dim,T}
+  vᵣ = orientation(ring) == CCW ? vertices(ring) : reverse(vertices(ring))
+  vₒ = orientation(other) == CCW ? vertices(other) : reverse(vertices(other))
+
+  I, pᵣ, tᵣ, pₒ, tₒ = _pairwiseintersections(vᵣ, vₒ)
```
juliohm: A function that returns more than 4 variables is an indication of sub-optimal design. Can you use the a different data structure to write this algorithm? Simple lists should be enough? It is hard to grasp the main idea of the algorithm when it has so much indirection into hash tables.

The idea should be as simple as:

1. Find all pairwise intersections and save two circular lists of points.
2. Associate to each list a vector of Bool telling if the point is intersection
3. Close the loops with the algorithm for all rings of the polygon
4. Query inner rings to find which outer loop they belong too

If these for steps are not straightforward to read in the code, it is a sign that we can refactor to make it cleaner and easier to read and maintain.

src/clipping/weileratherton.jl
```diff
@@ -0,0 +1,190 @@
...
+    Multi(g)
+  else
+    outers = clip(rings(poly)[1], boundary(other), method)
+    if isempty(outers)
+      nothing
+    else
+      Multi(PolyArea.(outers))
+    end
+  end
+end
```
eliascarv, line 46: Note: I didn't test this code. @ClaroHenrique, please test before commit.
```suggestion
function clip(poly::Polygon{Dim,T}, other::Geometry{Dim,T}, method::WeilerAtherton) where {Dim, T}
  outers = clip(first(rings(poly)), boundary(other), method)
  isempty(outers) && return nothing
  if hasholes(poly)
    inners = reduce(vcat, clip(r, boundary(other), method) for r in rings(poly)[2:end])
    ps = map(outers) do o
      # find inner rings inside the outer ring o 
      is = filter(i -> any(v -> sideof(v, o) == IN, vertices(i)), inners)
      PolyArea([o; is])
    end
    Multi(ps)
  else
    Multi(PolyArea.(outers))
  end
end
```

juliohm:
Hi @ClaroHenrique , do you think you can finish this PR, or should it be closed?

ClaroHenrique:
> Hi @ClaroHenrique , do you think you can finish this PR, or should it be closed?

Hi @juliohm , I am closing it for now. I couldn't solve some intersections problems, even with the vertice jittering idea.

PR #574, opened 2023-08-29
Remove `Data` trait and `MeshData`

eliascarv:


src/Meshes.jl
```diff
@@ -20,7 +20,6 @@ using NearestNeighbors: KDTree, BallTree, knn, inrange
 import Tables
```
juliohm: Double check if we can get rid of Tables or other dependencies now.

src/sampling.jl
```diff
@@ -16,7 +16,15 @@ Sample elements or points from geometric `object`
...
-sample(object, method) = sample(Random.GLOBAL_RNG, object, method)
+sample(object, method::SamplingMethod) = sample(Random.GLOBAL_RNG, object, method)
+
+"""
+    sampleinds(rng, domain, method)
+
+Sample indices of elements in `domain` with discrete
+sampling `method` and random number generator `rng`.
+"""
+function sampleinds end
```
juliohm: Please keep this in the original section of the code. It is only defined for DiscreteSamplingMethod

src/views.jl
```diff
@@ -16,6 +16,9 @@ struct DomainView{Dim,T,D<:Domain{Dim,T},I} <: Domain{Dim,T}
   inds::I
 end
 
+# specialize constructor to avoid infinite loops
+DomainView(v::DomainView, inds) = DomainView(getfield(v, :domain), getfield(v, :inds)[inds])
+
```
juliohm, line 21: We didn't have this method already?

test/partitioning.jl
```diff
@@ -373,40 +373,6 @@
...
-      PredicatePartition((i, j) -> iseven(i + j)),
-      SpatialPredicatePartition((x, y) -> norm(x + y) < T(5)),
-      ProductPartition(UniformPartition(2), UniformPartition(2)),
-      HierarchicalPartition(UniformPartition(2), UniformPartition(2))
-    ]
-      Π = partition(data, method)
-      inds = reduce(vcat, indices(Π))
-      @test sort(inds) == 1:100
-    end
   end
```
juliohm, line 376: Are these tests already covered in the Domain case?

juliohm (changes requested):

juliohm (approved):

lstagner:
Currently I am dispatching on a MeshData type. What is the recommeneded replacement for MeshData?

juliohm:
@lstagner it lives in GeoTables.jl now by the name GeoTable

PR #585, opened 2023-09-08
implement Base.in for cone

cserteGT3:
I don't think it needs further explanation :) Let me know if you have any comments or issues I need to address!


juliohm (approved):

juliohm:
Beautiful PR 🙂

PR #595, opened 2023-09-12
Add method to `indices` function with `Grid` and `Chain`

eliascarv:


src/viewing.jl
```diff
@@ -49,14 +49,27 @@ function indices(grid::Grid{2}, poly::Polygon{2})
   mask = zeros(Int, dims)
   cpoly = clip(poly, boundingbox(grid), SutherlandHodgman())
```
juliohm:
```suggestion
  cpoly = poly ∩ boundingbox(grid)
```
Let's use the \cap and let's handle the special case where the intersection is nothing. If the intersection is nothing we need to prematurely return an empty list of indices Int[]

juliohm (changes requested):

juliohm (approved):

PR #596, opened 2023-09-12
Add method to `indices` function with `Grid` and `Point`

eliascarv:


src/viewing.jl
```diff
@@ -44,6 +44,24 @@ that intersect with the `geometry`.
...
+function indices(grid::Grid{Dim}, point::Point{Dim}) where {Dim}
+  point ∉ grid && return Int[]
+
+  # grid properties
+  o = minimum(grid)
+  s = spacing(grid)
+  dims = size(grid)
+
+  # integer coordinates
+  coords = ceil.(Int, (point - o) ./ s)
```
juliohm:
```suggestion
  orig = minimum(grid)
  spac = spacing(grid)
  dims = size(grid)

  # integer coordinates
  coords = ceil.(Int, (point - orig) ./ spac)
```

juliohm (approved):

Issue #601, opened 2023-09-18
`Unitful` value is not promoted to floating `Point` and `Vector` automatically


ehgus:
The current code does not promote the integer `Unitful` values to floating point when constructing `Point` and `Vector`.
It's because their constructor only promote values when the values are a subtype of `Integer`, but all `Unitful` values are not:
https://github.com/JuliaGeometry/Meshes.jl/blob/327333f39c9713a4ae1f55206c8f8e7bcdb1c737/src/primitives/point.jl#L49
```Julia repl
julia> using Unitful; x = 1u"m";

julia> @show isa(x,Integer) typeof(x)
x isa Integer = false
typeof(x) = Quantity{Int64, 𝐋, Unitful.FreeUnits{(m,), 𝐋, nothing}}
Quantity{Int64, 𝐋, Unitful.FreeUnits{(m,), 𝐋, nothing}}
```

Fortunately, `Unitful` values allows the promotion using `float` function, we can fix the code without adding tedious constructor for `Unitful`.

juliohm:
Very nice @ehgus , thanks for sharing the issue. Feel free to submit a PR with a proposal fix and we can review right away.

PR #602, opened 2023-09-18
simplify `Vec` and `Point` constructors

ehgus:
follow-up of #601 

Type promotion of `Point` is rely on `Vec`, so I remove the floating-point promotion at a `Point` constructor. `Vec` promotes type `T` into `float(T)`. I listed possible scenarios how the promotion will work:

|T| float(T) |
|---|---|
|Integer|Float64|
|Float64|Float64|
|Float32|Float32|
|Quantity{Int64, 𝐋, Unitful.FreeUnits{(m,), 𝐋, nothing}}|Quantity{Float64, 𝐋, Unitful.FreeUnits{(m,), 𝐋, nothing}}|
|Quantity{Float32, 𝐋, Unitful.FreeUnits{(m,), 𝐋, nothing}}|Quantity{Float32, 𝐋, Unitful.FreeUnits{(m,), 𝐋, nothing}}|
|Quantity{Float64, 𝐋, Unitful.FreeUnits{(m,), 𝐋, nothing}}|Quantity{Float64, 𝐋, Unitful.FreeUnits{(m,), 𝐋, nothing}}|

The table proves that the `Unitful` quantities will be promoted as expected. I have checked with my test code:
```Julia repl
julia> using Meshes

julia> x= 1u"m"
1 m

julia> Point(x,x,x)
Point(1.0 m, 1.0 m, 1.0 m)

julia> y=1f0u"m"
1.0f0 m

julia> Point(x,y,x)
Point(1.0f0 m, 1.0f0 m, 1.0f0 m)
```

### DIFF
--- a/src/primitives/point.jl
+++ b/src/primitives/point.jl
@@ -40,17 +40,15 @@ J = Point3f(1, 2, 3) # explicitly ask for single precision
 """
 struct Point{Dim,T} <: Primitive{Dim,T}
   coords::Vec{Dim,T}
-  Point{Dim,T}(coords::Vec{Dim,T}) where {Dim,T} = new{Dim,T}(coords)
+  Point(coords::Vec{Dim,T}) where {Dim,T} = new{Dim,T}(coords)
 end
 
 # convenience constructors
 Point{Dim,T}(coords...) where {Dim,T} = Point{Dim,T}(coords)
-Point{Dim,T}(coords) where {Dim,T} = Point{Dim,T}(Vec{Dim,T}(coords))
-Point{Dim,T}(coords) where {Dim,T<:Integer} = Point{Dim,Float64}(coords)
+Point{Dim,T}(coords) where {Dim,T} = Point(Vec{Dim,T}(coords))
 
 Point(coords...) = Point(coords)
 Point(coords) = Point(Vec(coords))
-Point(coords::Vec{Dim,T}) where {Dim,T} = Point{Dim,T}(coords)
 
 # coordinate type conversions
 Base.convert(::Type{Point{Dim,T}}, coords) where {Dim,T} = Point{Dim,T}(coords)
--- a/src/primitives/point.jl
+++ b/src/primitives/point.jl
@@ -44,11 +44,8 @@ struct Point{Dim,T} <: Primitive{Dim,T}
 end
 
 # convenience constructors
-Point{Dim,T}(coords...) where {Dim,T} = Point{Dim,T}(coords)
-Point{Dim,T}(coords) where {Dim,T} = Point(Vec{Dim,T}(coords))
-
-Point(coords...) = Point(coords)
-Point(coords) = Point(Vec(coords))
+Point{Dim,T}(coords...) where {Dim,T} = Point(Vec{Dim,T}(coords...))
+Point(coords...) = Point(Vec(coords...))
 
 # coordinate type conversions
 Base.convert(::Type{Point{Dim,T}}, coords) where {Dim,T} = Point{Dim,T}(coords)
--- a/src/vectors.jl
+++ b/src/vectors.jl
@@ -79,13 +79,6 @@ function StaticArrays.similar_type(::Type{<:Vec}, ::Type{T}, ::Size{S}) where {T
   isone(N) && !(T <: Integer) ? Vec{L,T} : SArray{Tuple{S...},T,N,L}
 end
 
-# utils
-function checkdim(::Type{Vec{Dim,T}}, coords) where {Dim,T}
-  if Dim ≠ length(coords)
-    throw(DimensionMismatch("Invalid dimension."))
-  end
-end
-
 """
     ∠(u, v)
 
--- a/src/vectors.jl
+++ b/src/vectors.jl
@@ -43,7 +43,7 @@ Vec3f(1, 2, 3) # explicitly ask for single precision
 """
 struct Vec{Dim,T} <: StaticVector{Dim,T}
   coords::NTuple{Dim,T}
-  Vec(coords::NTuple{Dim,T}) where {Dim,T} = new{Dim,float(T)}(NTuple{Dim,float(T)}(coords))
+  Vec(coords::NTuple{Dim,T}) where {Dim,T} = new{Dim,float(T)}(coords)
 end
 
 # convenience constructors
--- a/src/vectors.jl
+++ b/src/vectors.jl
@@ -56,7 +56,7 @@ function Vec{Dim,T}(coords::Union{Tuple,AbstractVector}) where {Dim,T}
 end
 
 Vec(coords...) = Vec(coords)
-Vec(coords) = Vec(promote(coords...))
+Vec(coords::Tuple) = Vec(promote(coords...))
 
 # StaticVector constructors
 Vec(coords::StaticVector{Dim,T}) where {Dim,T} = Vec{Dim,T}(coords)
--- a/src/vectors.jl
+++ b/src/vectors.jl
@@ -43,24 +43,18 @@ Vec3f(1, 2, 3) # explicitly ask for single precision
 """
 struct Vec{Dim,T} <: StaticVector{Dim,T}
   coords::NTuple{Dim,T}
-  Vec{Dim,T}(coords::NTuple{Dim,T}) where {Dim,T} = new{Dim,T}(coords)
-  Vec{Dim,T}(coords::NTuple{Dim,T}) where {Dim,T<:Integer} = new{Dim,Float64}(coords)
+  Vec(coords::NTuple{Dim,T}) where {Dim,T} = new{Dim,float(T)}(NTuple{Dim,float(T)}(coords))
 end
 
 # convenience constructors
 Vec{Dim,T}(coords...) where {Dim,T} = Vec{Dim,T}(coords)
-function Vec{Dim,T}(coords::Tuple) where {Dim,T}
-  checkdim(Vec{Dim,T}, coords)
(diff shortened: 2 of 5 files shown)


src/primitives/point.jl
```diff
@@ -40,17 +40,15 @@ J = Point3f(1, 2, 3) # explicitly ask for single precision
 """
 struct Point{Dim,T} <: Primitive{Dim,T}
   coords::Vec{Dim,T}
-  Point{Dim,T}(coords::Vec{Dim,T}) where {Dim,T} = new{Dim,T}(coords)
+  Point(coords::Vec{Dim,T}) where {Dim,T} = new{Dim,T}(coords)
```
juliohm, line 43: Do we need this inner constructor or is it automatically generated by the Julia compiler? Can we try to erase it completely here?

src/vectors.jl
```diff
@@ -43,24 +43,18 @@ Vec3f(1, 2, 3) # explicitly ask for single precision
 """
 struct Vec{Dim,T} <: StaticVector{Dim,T}
   coords::NTuple{Dim,T}
-  Vec{Dim,T}(coords::NTuple{Dim,T}) where {Dim,T} = new{Dim,T}(coords)
-  Vec{Dim,T}(coords::NTuple{Dim,T}) where {Dim,T<:Integer} = new{Dim,Float64}(coords)
+  Vec(coords::NTuple{Dim,T}) where {Dim,T} = new{Dim,float(T)}(NTuple{Dim,float(T)}(coords))
```
juliohm: Can we add type parameters for the inner constructor? I think it is the default, we never used inner constructors in Julia without these parameters.

juliohm (changes requested):
This looks great @ehgus , I attached a few comments/suggestions based off a quick chat with @eliascarv. He will also try to review by the end of the day.

src/primitives/point.jl
ehgus: The inner constructor is necessary. The constructor is the inference rule for interpreting the implicit `Point` type to the explicit `Point` type. The constructors below cannot replace this.

ehgus: I deleted the line and tested the code, which failed some tests.

src/vectors.jl
ehgus: The previous inner constructors presume that the type parameter of `Vec` is the same as that of `coords`.
Instead, my code infers the type parameter of `Vec`  from `coords`.
The proposed code will work like the previous one with more relaxed conditions.

eliascarv: I think `NTuple{Dim,float(T)}` is not needed:
```suggestion
  Vec(coords::NTuple{Dim,T}) where {Dim,T} = new{Dim,float(T)}(coords)
```

src/vectors.jl
```diff
@@ -43,24 +43,18 @@ Vec3f(1, 2, 3) # explicitly ask for single precision
...
-  checkdim(Vec{Dim,T}, coords)
-  Vec{Dim,T}(NTuple{Dim,T}(coords))
-end
-function Vec{Dim,T}(coords::AbstractVector) where {Dim,T}
-  checkdim(Vec{Dim,T}, coords)
-  Vec{Dim,T}(NTuple{Dim,T}(coords))
+function Vec{Dim,T}(coords::Union{Tuple, AbstractVector}) where {Dim,T}
+  checkdim(Vec{Dim,float(T)}, coords)
+  Vec(NTuple{Dim,float(T)}(coords))
 end
```
eliascarv: With this, we can remove the `checkdim` function:
```suggestion
function Vec{Dim,T}(coords::Union{Tuple,AbstractVector}) where {Dim,T}
  if Dim ≠ length(coords)
    throw(DimensionMismatch("invalid dimension"))
  end
  Vec(NTuple{Dim,float(T)}(coords))
end
```
Please, along with this change remove the `checkdim` function.

eliascarv (changes requested):

src/vectors.jl
```diff
@@ -43,24 +43,18 @@ Vec3f(1, 2, 3) # explicitly ask for single precision
...
 
 Vec(coords...) = Vec(coords)
-Vec(coords::Tuple) = Vec(promote(coords...))
-Vec(coords::NTuple{Dim,T}) where {Dim,T} = Vec{Dim,T}(coords)
+Vec(coords) = Vec(promote(coords...))
```
eliascarv: The `Vec` constructor must only support tuples.
```suggestion
Vec(coords::Tuple) = Vec(promote(coords...))
```

eliascarv (changes requested):

juliohm:
@ehgus can you please take a look at the updated suggestions?

src/primitives/point.jl
```diff
@@ -40,17 +40,15 @@ J = Point3f(1, 2, 3) # explicitly ask for single precision
...
 
 # convenience constructors
 Point{Dim,T}(coords...) where {Dim,T} = Point{Dim,T}(coords)
-Point{Dim,T}(coords) where {Dim,T} = Point{Dim,T}(Vec{Dim,T}(coords))
-Point{Dim,T}(coords) where {Dim,T<:Integer} = Point{Dim,Float64}(coords)
+Point{Dim,T}(coords) where {Dim,T} = Point(Vec{Dim,T}(coords))
 
 Point(coords...) = Point(coords)
 Point(coords) = Point(Vec(coords))
-Point(coords::Vec{Dim,T}) where {Dim,T} = Point{Dim,T}(coords)
```
eliascarv: I think we can simplify these constructors:
```julia
# convenience constructors
Point{Dim,T}(coords...) where {Dim,T} = Point(Vec{Dim,T}(coords...))
Point(coords...) = Point(Vec(coords...))
```

ehgus: With the change,  `Point` constructors can no longer take `Tuple` as an argument.
I agree that construction like `Point((1,2,3))` may lead to unnecessary complexity in the code, and users can easily handle the change. I want to make sure that other maintainers also agree with that.

src/tolerances.jl
```diff
@@ -13,5 +13,6 @@ source code in calls to the [`isapprox`](@ref) function:
 isapprox(a::T, b::T, atol=atol(T))
 ```
 """
+atol(::Type{T}) where T = zero(T)
```
juliohm: If this change is unrelated to the Point constructor, we should probably add it in a separate PR.

src/primitives/point.jl
juliohm: Which change @ehgus? The one proposed by @eliascarv above?

juliohm (changes requested):
We need additional tests to make sure that the constructors work with units.

src/primitives/point.jl
eliascarv: The constructor I proposed just forwards the arguments from the `Point(args...)` constructor to the `Vec(args...)` constructor. Therefore `Point((1, 2, 3))` would be equivalent to `Point(Vec((1, 2, 3)))`.

eliascarv:
@ehgus, please remove the `checkdim` function:

image

This function is no longer needed in the PR code.

src/tolerances.jl
ehgus: Yes, this change should be in another PR. Thank you!

juliohm:
Thank you @ehgus for improving the PR further. As soon as you add tests I can review and approve. Then, we can wait for @eliascarv 's review and merge it.

src/vectors.jl
```diff
@@ -83,13 +79,6 @@ function StaticArrays.similar_type(::Type{<:Vec}, ::Type{T}, ::Size{S}) where {T
   isone(N) && !(T <: Integer) ? Vec{L,T} : SArray{Tuple{S...},T,N,L}
 end
 
-# utils
```
ehgus, line 86: > @ehgus, please remove the `checkdim` function:
> 
> image
> 
> This function is no longer needed in the PR code.

I removed the unused function

src/primitives/point.jl
ehgus: I should have tested your code before commenting. Thank you @eliascarv. Now I understand your suggestion and have updated the code.

eliascarv (approved):

test/primitives.jl
```diff
@@ -6,6 +6,9 @@
     @test coordtype(Point(1, 1)) == Float64
     @test coordtype(Point(1.0, 1.0)) == Float64
     @test coordtype(Point(1.0f0, 1.0f0)) == Float32
+    @test coordtype(Point(1u"m",1u"m")) == UNITFUL_FLOAT64
+    @test coordtype(Point(1.0u"m",1.0u"m")) == UNITFUL_FLOAT64
+    @test coordtype(Point(1.0f0u"m",1.0f0u"m")) == UNITFUL_FLOAT32
```
juliohm: Instead of checking the coordtype, what about just checking the `unit(coordinates(point))` and comparing with the expected value? The Unitful types are quite verbose and likely include implementation details that may change with time.

eliascarv: Suggestion:
```julia
@test unit(coordtype(point)) == u"m"
@test Unitful.numtype(coordtype(point)) === Float64
```

ehgus: Thank you for the suggestions, @eliascarv and @juliohm. I have been looking for a function like `Unitful.numtype`. I will soon update the test code.

ehgus:
Tests for `Vec` are also added

juliohm (approved):

juliohm:
Thank you @ehgus for this very important contribution. Feel free to submit another PR with the `atol` improvements we discussed.

Issue #616, opened 2023-10-10
Vertices coloring not working


remann2:
```
m = CartisianGrid(10, 10)
viz(m, color=1:nvertices(m))
 ```

will return DimensionMismatch error, is this a bug?

remann2:
My bad, it seems that only subtypes of Mesh work this way while Grid doesn't? 

> The option color can be a single scalar or a vector of scalars. For [Mesh](https://juliageometry.github.io/MeshesDocs/stable/domains/meshes.html#Meshes.Mesh) subtypes, the length of the vector of colors determines if the colors should be assigned to vertices or to elements. (https://juliageometry.github.io/MeshesDocs/stable/visualization.html)

juliohm:
If it is not working, it is a bug. We have Grid <: Mesh.

Would you like to submit a PR fixing the issue? The plotting recipe is in
the ext folder.

Em ter., 10 de out. de 2023 06:20, Steve ***@***.***>
escreveu:

remann2:
Thanks, I will try.

juliohm:
The easiest way is to simply convert(SimpleMesh, grid) and call the viz
function again.

Em ter., 10 de out. de 2023 06:33, Steve ***@***.***>
escreveu:

remann2:
Hi, I've create a pr plz have a look.

PR #617, opened 2023-10-10
fix dimension mismatch when colors=nvertices

remann2:
the bug comes from that size(mesh) returns number of segments in each dimension, which mismatches the number of vertices.

### DIFF
--- a/ext/cartesiangrid.jl
+++ b/ext/cartesiangrid.jl
@@ -84,7 +84,11 @@ function vizgrid2D!(plot)
     nd = embeddim($grid)
     or = coordinates(minimum($grid))
     sp = spacing($grid)
+    
     sz = size($grid)
+    if *(sz...) != length($colorant)
+      sz = sz .+ 1
+    end
 
     xs, ys = cartesiancenters(or, sp, sz, nd)
 
--- a/ext/cartesiangrid.jl
+++ b/ext/cartesiangrid.jl
@@ -15,12 +15,22 @@ function Makie.plot!(plot::Viz{<:Tuple{CartesianGrid}})
   # with 1D, 2D, 3D elements
   if color isa AbstractVector
     # visualize grid as heatmap or volume
-    if ndim == 1
-      vizgrid1D!(plot)
-    elseif ndim == 2
-      vizgrid2D!(plot)
-    elseif ndim == 3
-      vizgrid3D!(plot)
+    if length(color) == nvertices(grid)
+      if ndim == 1
+        vizmesh1D!(plot)
+      elseif ndim == 2
+        vizmesh2D!(plot)
+      elseif ndim == 3
+        vizmesh3D!(plot)    
+      end
+    else
+      if ndim == 1
+        vizgrid1D!(plot)
+      elseif ndim == 2
+        vizgrid2D!(plot)
+      elseif ndim == 3
+        vizgrid3D!(plot)
+      end
     end
   else
     # create the smallest mesh of simplices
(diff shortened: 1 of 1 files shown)


remann2:
this fixes #616

ext/cartesiangrid.jl
```diff
@@ -84,7 +84,11 @@ function vizgrid2D!(plot)
     nd = embeddim($grid)
     or = coordinates(minimum($grid))
     sp = spacing($grid)
+    
     sz = size($grid)
+    if *(sz...) != length($colorant)
+      sz = sz .+ 1
+    end
```
juliohm: This fix should be implemented in the main Makie.plot! method so that it works with 2D and 3D grids.

juliohm:
@remann2 thank you for trying to fix the issue. Can you please refactor the PR to fix the plot in 3D as well? Basically, you need to propose a change to the Makie.plot! method instead of the vizgrid2D function.

remann2:
I was going to do that, but before that I did a test on 1D and 3D grid, where no error occurred and the plots looked correct. So I just modified the 2D function.

juliohm:
Can you please share some plots with the proposed fix? I remember that this branch of the code simply called `Mke.heatmap`. It is a bit surprising that just increasing the `sz` by one will produce an interpolated version.

For instance, can you please plot `Mke.plot(CartesianGrid(10,10), 1:100)` and then `Mke.plot(CartesianGrid(10,10), 1:11*11)`?

remann2:
```
using Meshes
using CairoMakie
m2 = CartesianGrid(2, 2)
m3 = CartesianGrid(2, 2, 2)
```

```viz(m2, color=1:nelements(m2))```
image
```viz(m2, color=1:nvertices(m2))```
image
```viz(m3, color=1:nelements(m3))```
image
```viz(m3, color=1:nvertices(m3))```
image

I didn't pay much attention since the 3D grid plots didn't error, seems like the last plot isn't right?

**Comment by juliohm** [comment]
The 2D plot is not correct either. Notice that the colors should be interpolated when the vector has the length equal to the number of vertices. If you convert the grid to a SimpleMesh as I suggested and then plot with nvertices you will see the difference.

The easiest solution is to adapt the recipe to convert to SimpleMesh whenever the length of colors equals the length of vertices.

**Comment by remann2** [comment]
Indeed! will do that now

**Comment by remann2** [comment]
sorry for the delay. From the vizmesh3D! method in the recipe, it seems that 3D mesh is discretized into small meshes plotted individually so that vertices color interpolation is not supported? 

```
m3 = CartesianGrid(2, 2, 2)
viz(convert(SimpleMesh, m3), color=1:nvertices(m3))
```
image

**Comment by juliohm** [comment]
That is correct. The 3D case is currently implemented with meshscatter by
plotting various boxes. Last time I checked that was the only method to
visualize 3D grids with Makie.jl.

You can concentrate on the 2D case. Later we can review the 3D case.

Em ter., 10 de out. de 2023 12:42, Steve ***@***.***>
escreveu:

**Comment by remann2** [comment]
then there is not much to do. 
the fixed 2d plot:
```
m2 = CartesianGrid(2, 2)
viz(m2, color=1:nvertices(m2), showfacets=true)
```
image

remann2:
I didn't convert the grid to SimpleMesh since I haven't found an easy way to construct Viz{<:Tuple{SimpleMesh}} plot type. I can do viz(grid::Grid, args...) = viz(convert(SimpleMesh, grid), args...) but I am not sure where to put it.
Since the interfaces used in vizmeshXD! are compatible on Grid, I end up passing plot directly to the vizmeshXD! methods.

juliohm:
@remann2 please start a fresh PR with the minimum changes necessary to fix the issue raised. The current PR is changing various other branches of the code. You only need to go inside the ndim == 2 condition and check the length of the colors vectors to switch between vizgrid and vizmesh

PR #623, opened 2023-10-17
don't apply translation to Vec, only to Point; fix #622

### LINKED ISSUE #622: translation messes up Plane
The transformation is applied to `u` and `v` of a plane as well, see the following example:

```julia
julia> p1 = Plane((10, 0, 10), (0, 0, 1))
Plane{3,Float64}
├─ p: Point(10.0, 0.0, 10.0)
├─ u: Vec(1.0, -0.0, -0.0)
└─ v: Vec(-0.0, 1.0, -0.0)

julia> ft = Translate(100, 35, 70)
Translate transform
└─ offsets = (100, 35, 70)

julia> p2 = ft(p1)
Plane{3,Float64}
├─ p: Point(110.0, 35.0, 80.0)
├─ u: Vec(101.0, 35.0, 70.0)
└─ v: Vec(100.0, 36.0, 70.0)
```

I realized this, when I was transforming disks, and looked pretty weird. Turns out, the underlying plane is modified like in the example above.

I assume that this should not behave like this.

cserteGT3:
I'm not sure that this is the right solution for #622, but this is my take on it.
It's based on juliohm's comment:
>Perhaps we should skip the Vec field always, except in a few cases? Our current Point struct holds a Vec but that will change soon. It is the only case where we want to apply the transform to the Vec.

What dou you think?

I'll extend the tests if you agree with this solution.

Edit: the example from the issue looks like this with this PR:

```julia
julia> p1 = Plane((10, 0, 10), (0, 0, 1))
Plane{3,Float64}
├─ p: Point(10.0, 0.0, 10.0)
├─ u: Vec(1.0, -0.0, -0.0)
└─ v: Vec(-0.0, 1.0, -0.0)

julia> ft = Translate(100, 35, 70)
Translate transform
└─ offsets = (100, 35, 70)

julia> ft(p1)
Plane{3,Float64}
├─ p: Point(110.0, 35.0, 80.0)
├─ u: Vec(1.0, -0.0, -0.0)
└─ v: Vec(-0.0, 1.0, -0.0)

julia> d1 = Disk(p1, 15.0)
Disk{3,Float64}
├─ plane: Plane(p: (10.0, 0.0, 10.0), u: (1.0, -0.0, -0.0), v: (-0.0, 1.0, -0.0))
└─ radius: 15.0

julia> ft(d1)
Disk{3,Float64}
├─ plane: Plane(p: (110.0, 35.0, 80.0), u: (1.0, -0.0, -0.0), v: (-0.0, 1.0, -0.0))
└─ radius: 15.0

julia> r1 = Ray((0,0,0), (1,0,0))
Ray{3,Float64}
├─ p: Point(0.0, 0.0, 0.0)
└─ v: Vec(1.0, 0.0, 0.0)

julia> ft(r1)
Ray{3,Float64}
├─ p: Point(100.0, 35.0, 70.0)
└─ v: Vec(1.0, 0.0, 0.0)

julia> t1 = Torus((0.,0.,0.), (1.,0.,0.), (0.,1.,0.), 0.2)
Torus{3,Float64}
├─ center: Point(0.5, 0.5, 0.0)
├─ normal: Vec(0.0, 0.0, 1.0)
├─ major: 0.7071067811865476
└─ minor: 0.2

julia> ft(t1)
Torus{3,Float64}
├─ center: Point(100.5, 35.5, 70.0)
├─ normal: Vec(0.0, 0.0, 1.0)
├─ major: 0.7071067811865476
└─ minor: 0.2
```

Fix #622

### DIFF
--- a/src/primitives/plane.jl
+++ b/src/primitives/plane.jl
@@ -47,7 +47,8 @@ normal(p::Plane) = normalize(p.u × p.v)
 
 Base.isapprox(p₁::Plane{T}, p₂::Plane{T}) where {T} =
   isapprox((p₁.v - p₁.u) ⋅ normal(p₂), zero(T), atol=atol(T)) &&
-  isapprox((p₂.v - p₂.u) ⋅ normal(p₁), zero(T), atol=atol(T))
+  isapprox((p₂.v - p₂.u) ⋅ normal(p₁), zero(T), atol=atol(T)) &&
+  isapprox((p₂(0,0) - p₁(0,0)) ⋅ normal(p₁), zero(T), atol=atol(T))
 
 (p::Plane)(u, v) = p.p + u * p.u + v * p.v
 
--- a/src/primitives/plane.jl
+++ b/src/primitives/plane.jl
@@ -47,8 +47,7 @@ normal(p::Plane) = normalize(p.u × p.v)
 
 Base.isapprox(p₁::Plane{T}, p₂::Plane{T}) where {T} =
   isapprox((p₁.v - p₁.u) ⋅ normal(p₂), zero(T), atol=atol(T)) &&
-  isapprox((p₂.v - p₂.u) ⋅ normal(p₁), zero(T), atol=atol(T)) &&
-  isapprox((p₂(0,0) - p₁(0,0)) ⋅ normal(p₁), zero(T), atol=atol(T))
+  isapprox((p₂.v - p₂.u) ⋅ normal(p₁), zero(T), atol=atol(T))
 
 (p::Plane)(u, v) = p.p + u * p.u + v * p.v
 
--- a/src/primitives/frustum.jl
+++ b/src/primitives/frustum.jl
@@ -3,7 +3,7 @@
 # ------------------------------------------------------------------
 
 """
-  Frustum(bot, top)
+    Frustum(bot, top)
 
 A frustum (truncated cone) with `bot` and `top` disks.
 See https://en.wikipedia.org/wiki/Frustum.
--- a/src/primitives/frustumsurface.jl
+++ b/src/primitives/frustumsurface.jl
@@ -3,7 +3,7 @@
 # ------------------------------------------------------------------
 
 """
-  FrustumSurface(bot, top)
+    FrustumSurface(bot, top)
 
 A frustum (truncated cone) surface with `bot` and `top` disks.
 See https://en.wikipedia.org/wiki/Frustum.
--- a/src/transforms/translate.jl
+++ b/src/transforms/translate.jl
@@ -20,4 +20,5 @@ isinvertible(::Type{<:Translate}) = true
 
 Base.inv(t::Translate) = Translate(-1 .* t.offsets)
 
-applycoord(t::Translate, v::Vec) = v + Vec(t.offsets)
+applycoord(t::Translate, v::Vec) = v
+applycoord(t::Translate, p::Point) = p + Vec(t.offsets)
--- a/test/transforms.jl
+++ b/test/transforms.jl
@@ -240,7 +240,7 @@
     f = Translate(T(0), T(0), T(1))
     g = Plane(P3(0, 0, 0), V3(0, 0, 1))
     r, c = TB.apply(f, g)
-    @test r ≈ Plane(P3(0, 0, 0), V3(0, 0, 1))
+    @test r ≈ Plane(P3(0, 0, 1), V3(0, 0, 1))
     @test TB.revert(f, r, c) ≈ g
 
     # ---------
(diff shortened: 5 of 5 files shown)


juliohm:
That looks great. I would just implement the case with Point without using
internal implementation details: p + Vec(t.offsets)

Em ter., 17 de out. de 2023 08:27, cserteGT3 ***@***.***>
escreveu:

cserteGT3:
In #624 I raised the problem of isapprox for planes. I also did the changes I think would resolve the issue. It is related because using the isapprox function makes easier the testing of the proposed changes of translation implementation.

There is one test that still fails: I'm not really sure what stretching a plane means :/

What do you think of these changes?

juliohm:
These changes look great. Thanks for fixing the isapprox for planes as well. Feel free to reimplement it as you think is appropriate, don't feel constrained to the current implementation.

Let me know when the PR is ready for review :)

cserteGT3:
Thanks! :)

I've added one commit, that is not relevant, but bothered me: I messed up the docstring identation when I added the Frustum and FrustumSurface.

I think I'm satisfied with the current state. The only issue, that I don't understand is how should stretching behave with planes. And because of the isapprox change, now one of its test fails: https://github.com/JuliaGeometry/Meshes.jl/actions/runs/6547627918/job/17780492937?pr=623#step:6:662

Do you have any suggestions how to solve this, or what result should I expect here?

juliohm:
I think stretching a Plane shouldn't cause any effect. If we consider a horizontal plane for simplicity, and try to stretch it along x, y and z axis, the resulting plane will be the same. So I think we need a special method for Stretch with Plane. Agree?

juliohm:
Actually, if the plane is not aligned with the coordinate system, then stretching it will case a change to the u an v basis. The reference point should remain fixed. If you can experiment with the idea geometrically, that would be helpful.

cserteGT3:
Actually I don't have currently ideas for experimenting with stretch. Is it possible track this in a new issue, mark that test broken and proceed with this PR?

juliohm:
Can we just stretch the u and v and retain the p as is in the Stretch method for Plane? I think that is a good definition to fix the tests.

cserteGT3:
> Can we just stretch the u and v and retain the p as is in the Stretch method for Plane? I think that is a good definition to fix the tests.

Something like, this?

```julia
function applycoord(t::Stretch, p::Plane)
  Plane(p(0, 0), applycoord(t, p.u), applycoord(t, p.v))
end
```

Edit:
This works for a plane, but now what should a stretched cylinder be? The example below (result of the above change) doesn't seem right to me. Maybe cylinders should also have their special case for stretching?

```julia
julia> f = Stretch(1, 1, 2)
Stretch transform
└─ factors = (1, 1, 2)

julia> g = Cylinder(1)
Cylinder{3,Float64}
├─ bot: Plane(p: (0.0, 0.0, 0.0), u: (1.0, -0.0, -0.0), v: (-0.0, 1.0, -0.0))
├─ top: Plane(p: (0.0, 0.0, 1.0), u: (1.0, -0.0, -0.0), v: (-0.0, 1.0, -0.0))
└─ radius: 1.0

julia> f(g)
Cylinder{3,Float64}
├─ bot: Plane(p: (0.0, 0.0, 0.0), u: (1.0, -0.0, -0.0), v: (-0.0, 1.0, -0.0))
├─ top: Plane(p: (0.0, 0.0, 1.0), u: (1.0, -0.0, -0.0), v: (-0.0, 1.0, -0.0))
└─ radius: 1.0
```

juliohm:
Stretching a Box leads to translation of the center of the Box. Maybe we
should also translate the point in the Plane. Let's translate all fields
for now.

Try to define the method without using the internals of the Plane.

Em qua., 18 de out. de 2023 03:13, cserteGT3 ***@***.***>
escreveu:

juliohm:
@cserteGT3 did you have some time to finish this PR? It would be really nice to release a hotfix.

cserteGT3:
Unfortunately not. I have to prepare for a conference presentation so I put this on the backlog :(

src/primitives/plane.jl
```diff
@@ -47,7 +47,8 @@ normal(p::Plane) = normalize(p.u × p.v)
 
 Base.isapprox(p₁::Plane{T}, p₂::Plane{T}) where {T} =
   isapprox((p₁.v - p₁.u) ⋅ normal(p₂), zero(T), atol=atol(T)) &&
-  isapprox((p₂.v - p₂.u) ⋅ normal(p₁), zero(T), atol=atol(T))
+  isapprox((p₂.v - p₂.u) ⋅ normal(p₁), zero(T), atol=atol(T)) &&
+  isapprox((p₂(0,0) - p₁(0,0)) ⋅ normal(p₁), zero(T), atol=atol(T))
```
eliascarv: We will leave this change for a future PR.
```suggestion
  isapprox((p₂.v - p₂.u) ⋅ normal(p₁), zero(T), atol=atol(T))
```

juliohm (approved):

eliascarv (approved):

Issue #549, opened 2023-08-06
Fix JarvisMarch


juliohm:
The JarvisMarch algorithm for convex hulls is not working currently. We need a hot-fix for it.

DanielVandH:
What's wrong with it?

juliohm:
Hi Daniel,

It was failing with very basic point sets, probably because of angle
definitions. I'll try to create a MWE, but you can try it with a couple of
random points as in rand(Point2, 100).

Em sex., 20 de out. de 2023 01:36, Daniel VandenHeuvel <
***@***.***> escreveu:

DanielVandH:
I initially tried some random point sets, e.g. I looked at:

```julia
using Meshes, DelaunayTriangulation, Test
DelaunayTriangulation.getx(p::Point2) = p.coords[1]
DelaunayTriangulation.gety(p::Point2) = p.coords[2]
Base.isless(p::Point2, q::Point2) = Base.isless(getxy(p), getxy(q))
⊢(p, q) = DelaunayTriangulation.circular_equality(p, q)

for _ in 1:10000
    n = rand(3:1000)
    points = rand(Point2, n) 
    ch = convex_hull(points).indices
    ch2_p = hull(points, JarvisMarch()).rings[1].vertices.data
    push!(ch2_p, ch2_p[begin])
    ch2 = indexin(ch2_p, points)
    @test ch ⊢ ch2 
end
```

Nothing goes wrong.

juliohm:
I should have saved the MWE. We probably fixed the bug with some recent changes to our 2D angle calculation.

Will close the issue for now. We can reopen it if the bug reappears.

DanielVandH:
Here is one (potential) issue:

```julia
using Meshes, DelaunayTriangulation, Test
DelaunayTriangulation.getx(p::Point2) = p.coords[1]
DelaunayTriangulation.gety(p::Point2) = p.coords[2]
Base.isless(p::Point2, q::Point2) = Base.isless(getxy(p), getxy(q))

points = [Point2(10(i - 1), 10(j - 1)) for i in 1:10, j in 1:10] |> vec
ch = convex_hull(points).indices
ch2_p = hull(points, JarvisMarch()).rings[1].vertices.data
push!(ch2_p, ch2_p[begin])

_points = [p.coords.coords for p in points]
using CairoMakie
fig = Figure()
ax = Axis(fig[1, 1])
scatter!(ax, _points)
[...]

PR #625, opened 2023-10-21
Simplify convex hull from GrahamScan()

DanielVandH:
Here is one way that the convex hull can be simplified in the presence of collinear points, at least for `GrahamScan()`. This is to address https://github.com/JuliaGeometry/Meshes.jl/issues/549#issuecomment-1773743092. The basic idea is to keep popping while making a left turn, but only actually push if the turn is a strictly right turn (not a straight jump). It does require some extra handling for the initialisation, though.

I tried doing the same idea for `JarvisMarch()`, but I am not familiar enough with it to do that. (The algorithm I was using in the linked issue was the monotone chain algorithm.)

### DIFF
--- a/src/hulls/graham.jl
+++ b/src/hulls/graham.jl
@@ -50,9 +50,8 @@ function hull(points, ::GrahamScan)
   # all points are collinear, return segment
   isnothing(idx) && return Segment(O, q[end])
 
-  idx = max(idx, 2) # is this necessary?
-
   # rotational sweep
+  idx = max(idx, 2)
   r = [O, q[idx - 1], q[idx]]
   for B in q[(idx + 1):end]
     while ∠(r[end - 1], r[end], B) > atol(T) && length(r) ≥ 3
--- a/src/hulls/graham.jl
+++ b/src/hulls/graham.jl
@@ -43,13 +43,28 @@ function hull(points, ::GrahamScan)
   θ = [∠(A, O, B) for B in q]
   q = q[sortperm(θ)]
 
+  # initialise and skip collinear points at beginning 
+  r = [O]
+  idx = 1
+  while idx ≤ length(q) && coordinates(q[idx])[2] == coordinates(O)[2]
+    idx += 1
+  end
+  idx = max(idx, 2)
+  push!(r, q[idx - 1])
+  if idx > length(q)
+    return PolyArea(r)
+  else
+    push!(r, q[idx])
+  end
+
   # rotational sweep
-  r = [O, q[1], q[2]]
-  for B in q[3:end]
+  for B in q[(idx + 1):end]
     while ∠(r[end - 1], r[end], B) > atol(T) && length(r) ≥ 3
       pop!(r)
     end
-    push!(r, B)
+    if !iscollinear(r[end - 1], r[end], B)
+      push!(r, B)
+    end
   end
 
   PolyArea(r)
--- a/src/hulls/graham.jl
+++ b/src/hulls/graham.jl
@@ -45,15 +45,15 @@ function hull(points, ::GrahamScan)
 
   # skip collinear points at beginning 
   y(p) = coordinates(p)[2]
-  idx = findfirst(qᵢ -> y(qᵢ) ≠ y(O), q)
+  i = findfirst(qᵢ -> y(qᵢ) ≠ y(O), q)
   
   # all points are collinear, return segment
-  isnothing(idx) && return Segment(O, q[end])
+  isnothing(i) && return Segment(O, q[end])
 
   # rotational sweep
-  idx = max(idx, 2)
-  r = [O, q[idx - 1], q[idx]]
-  for B in q[(idx + 1):end]
+  i = max(i, 2)
+  r = [O, q[i - 1], q[i]]
+  for B in q[(i + 1):end]
     while ∠(r[end - 1], r[end], B) > atol(T) && length(r) ≥ 3
       pop!(r)
     end
--- a/src/hulls/graham.jl
+++ b/src/hulls/graham.jl
@@ -45,17 +45,14 @@ function hull(points, ::GrahamScan)
 
   # initialise and skip collinear points at beginning 
   r = [O]
-  idx = 1
-  while idx ≤ length(q) && coordinates(q[idx])[2] == coordinates(O)[2]
-    idx += 1
+  oy = coordinates(O)[2]
+  idx = findfirst(qq -> coordinates(qq)[2] ≠ oy, q)
+  if isnothing(idx)
+    push!(r, q[end]) # all points are collinear, so just add in the last point 
+    return PolyArea(r)
   end
   idx = max(idx, 2)
-  push!(r, q[idx - 1])
-  if idx > length(q)
-    return PolyArea(r)
-  else
-    push!(r, q[idx])
-  end
+  push!(r, q[idx - 1], q[idx])
 
   # rotational sweep
   for B in q[(idx + 1):end]
--- a/src/hulls/graham.jl
+++ b/src/hulls/graham.jl
@@ -43,18 +43,17 @@ function hull(points, ::GrahamScan)
   θ = [∠(A, O, B) for B in q]
   q = q[sortperm(θ)]
 
-  # initialise and skip collinear points at beginning 
-  r = [O]
-  oy = coordinates(O)[2]
-  idx = findfirst(qq -> coordinates(qq)[2] ≠ oy, q)
-  if isnothing(idx)
-    push!(r, q[end]) # all points are collinear, so just add in the last point 
-    return PolyArea(r)
-  end
-  idx = max(idx, 2)
-  push!(r, q[idx - 1], q[idx])
+  # skip collinear points at beginning 
+  y(p) = coordinates(p)[2]
+  idx = findfirst(qᵢ -> y(qᵢ) ≠ y(O), q)
+  
+  # all points are collinear, return segment
+  isnothing(idx) && return Segment(O, q[end])
+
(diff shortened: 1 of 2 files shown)


juliohm:
Thank you for the suggested improvement @DanielVandH. It simplifies the result but the code gets more complicated, right? Can you please compare performance-wise which implementation wins?

I feel that the original implementation in master is cleaner to understand.

Also feel free to contribute other algorithms to the list of hull algorithms.

DanielVandH:
> Can you please compare performance-wise which implementation wins?

The original implementation will win, of course.  The function is trying to do more than it was before - the efficiency is now in terms of the amount of points that will be returned for these special cases. You could alternatively add a `simplify` keyword just like MATLAB uses if you want to allow for different approach.

There is a simpler way if there is a predicate giving the position of a point relative to a line (and distinguishes between right, left, and collinear), but I didn't see one in this package (and this simpler approach would be more susceptible to robustness issues without using e.g. ExactPredicates.jl, so I'm not going to consider it).

Some benchmarks on this branch. Not a significant difference.

```julia
julia> @benchmark hull(pts, method) ( # collinearities
         setup = begin
             pts = $[
               P2(2, 0),
               P2(4, 0),
               P2(6, 0),
               P2(10, 0),
               P2(12, 1),
               P2(14, 3),
               P2(14, 6),
               P2(14, 9),
               P2(13, 10),
               P2(11, 11),
               P2(8, 12),
               P2(3, 11),
               P2(0, 8),
               P2(0, 7),
               P2(0, 6),
               P2(0, 5),
               P2(0, 4),
               P2(0, 3),
               P2(0, 2),
               P2(1, 0)
             ]
             method=$GrahamScan()
         end
       )
BenchmarkTools.Trial: 10000 samples with 9 evaluations.
 Range (min … max):  2.078 μs …  3.541 ms  ┊ GC (min … max):  0.00% … 99.78%
 Time  (median):     3.267 μs              ┊ GC (median):     0.00%
 Time  (mean ± σ):   3.502 μs ± 35.386 μs  ┊ GC (mean ± σ):  10.09% ±  1.00%

   ▅█                ▄▅▆▅▃
  ▂███▃▂▁▁▁▂▁▁▁▁▁▂▃▅███████▇▄▃▃▃▂▂▂▂▂▂▂▂▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁ ▃
  2.08 μs        Histogram: frequency by time         5.4 μs <

 Memory estimate: 6.80 KiB, allocs estimate: 30.

julia> @benchmark hull(pts, method) ( # random
         setup = begin
           pts = $(rand(P2, 50000))
           method=$GrahamScan()
         end
       )
BenchmarkTools.Trial: 424 samples with 1 evaluation.
 Range (min … max):  10.319 ms … 27.342 ms  ┊ GC (min … max): 0.00% … 56.71%
 Time  (median):     11.418 ms              ┊ GC (median):    0.00%
 Time  (mean ± σ):   11.796 ms ±  2.004 ms  ┊ GC (mean ± σ):  4.26% ±  9.75%

  ▇█▇▃▄▆▆▇▆▆▅▂▂
  █████████████▆▄▄▁▁▁▁▁▁▄▁▄▄▄▆▁▇▁▁▄▇▁▁▆▄▄▁▇▄▄▆▆▄▁▁▆▄▁▁▁▁▁▄▁▁▄ █
  10.3 ms      Histogram: log(frequency) by time      20.4 ms <

 Memory estimate: 12.98 MiB, allocs estimate: 72.
```

On master:

```julia
julia> @benchmark hull(pts, method) ( # collinearities
         setup = begin
             pts = $[
               P2(2, 0),
               P2(4, 0),
               P2(6, 0),
               P2(10, 0),
               P2(12, 1),
               P2(14, 3),
               P2(14, 6),
               P2(14, 9),
               P2(13, 10),
               P2(11, 11),
               P2(8, 12),
               P2(3, 11),
               P2(0, 8),
               P2(0, 7),
               P2(0, 6),
               P2(0, 5),
               P2(0, 4),
               P2(0, 3),
               P2(0, 2),
               P2(1, 0)
             ]
             method=$GrahamScan()
         end
       )
BenchmarkTools.Trial: 10000 samples with 9 evaluations.
 Range (min … max):  2.211 μs …  2.641 ms  ┊ GC (min … max): 0.00% … 99.75%
 Time  (median):     3.411 μs              ┊ GC (median):    0.00%
 Time  (mean ± σ):   3.714 μs ± 26.392 μs  ┊ GC (mean ± σ):  7.09% ±  1.00%

   ▃             ▃▆█▇▅▂
  ▅█▄▂▁▁▁▂▂▂▂▃▄▅███████▇▅▄▄▄▃▂▂▂▂▂▂▁▂▂▂▂▁▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁ ▂
  2.21 μs        Histogram: frequency by time        6.12 μs <

 Memory estimate: 6.83 KiB, allocs estimate: 30.

julia> @benchmark hull(pts, method) ( # random
         setup = begin
           pts = $(rand(P2, 50000))
           method=$GrahamScan()
         end
       )
BenchmarkTools.Trial: 383 samples with 1 evaluation.
 Range (min … max):  10.777 ms … 30.000 ms  ┊ GC (min … max): 0.00% … 58.32%
 Time  (median):     12.557 ms              ┊ GC (median):    0.00%
 Time  (mean ± σ):   13.049 ms ±  1.998 ms  ┊ GC (mean ± σ):  4.33% ±  9.54%

           ▄███▃▁
  ▃▄▄▃▃▁▃▄▇██████▆▄▃▄▃▃▃▁▁▂▁▁▁▁▂▃▁▂▁▂▁▁▁▂▁▁▂▁▁▂▁▃▁▃▃▃▁▃▃▃▃▂▂▂ ▃
  10.8 ms         Histogram: frequency by time        19.6 ms <

 Memory estimate: 12.98 MiB, allocs estimate: 72.
```

(Also: These times also improve if using `sizehint!` on the `r` vector, e.g. `sizehint!(r, max(4, floor(Int, cbrt(n))))`, and some more in-place methods for sorting / `@view`s. Not relevant to the changes in this PR though, just a note.)

> I feel that the original implementation in master is cleaner to understand.

Yes, it's doing less so I'm sure it is easier to understand. I don't think this implementation is too hard to understand either, though.

src/hulls/graham.jl
```diff
@@ -43,13 +43,28 @@ function hull(points, ::GrahamScan)
...
+  while idx ≤ length(q) && coordinates(q[idx])[2] == coordinates(O)[2]
+    idx += 1
+  end
+  idx = max(idx, 2)
+  push!(r, q[idx - 1])
+  if idx > length(q)
+    return PolyArea(r)
+  else
+    push!(r, q[idx])
+  end
```
juliohm: It seems that this code can be simplified with the use of `findfirst`. You can find the first index satisfying a condition and then initialize `r`. If no index satisfies the condition, then the result is `nothing` and you can return the trivial `PolyArea`. Can you please adjust the code accordingly?

juliohm (changes requested):
Sorry I took so long to review this PR.

Regarding the approach with a predicate function, maybe we should add this predicate to the `GrahamScan` struct and use one of our predicates as the default? If I understood correctly, this design would give us all the necessary flexibility to cover all variations of algorithms's results.

src/hulls/graham.jl
DanielVandH: How did I forget that `findfirst` exists... Yes, done.

DanielVandH:
> Regarding the approach with a predicate function, maybe we should add this predicate to the GrahamScan struct and use one of our predicates as the default? If I understood correctly, this design would give us all the necessary flexibility to cover all variations of algorithms's results.

Thinking about it some more, it might be worse off to do it this way. It's probably better if you instead want *all* collinear points (see e.g. [here](https://github.com/DanielVandH/DelaunayTriangulation.jl/blob/main/src/data_structures/convex_hull.jl#L50-L91), where predicates are used to make sure that all collinear points are included when using the monotone chain algorithm [and I've only just noticed that I call it a Graham scan in the docstring...]). Let's not pursue that here.

juliohm:
I've adjusted the code a bit more. Can you please confirm that the `idx = max(idx, 2)` is necessary? Did you notice that our `q` variable already starts from index 2?

DanielVandH:
Yes, it's necessary - one of the tests runs into it. The issue comes when the very next point is not on the same `y`-coordinate, i.e. `q[1]`.

test/hulls.jl
```diff
@@ -71,6 +71,73 @@
       pts = vertices(poly)
       chul = hull(pts, method)
       @test nvertices(chul) < length(pts)
+
+      if method == GrahamScan() # simplification not yet implemented for JarvisMarch()
```
juliohm:
```suggestion
      if method == GrahamScan()
```

test/hulls.jl
```diff
@@ -71,6 +71,73 @@
       pts = vertices(poly)
       chul = hull(pts, method)
       @test nvertices(chul) < length(pts)
+
+      if method == GrahamScan()
+        # simplifying rectangular hull / triangular
+        points = [P2(i - 1, j - 1) for i in 1:11, j in 1:11] |> vec
```
juliohm:
```suggestion
        points = [P2(i - 1, j - 1) for i in 1:11 for j in 1:11]
```
This is a Julia trick to flatten a list comprehension with two or more indices. We just need to add more `for` keywords instead of the comma. Can you please confirm that this is the expected flattening order?

juliohm (approved):

juliohm:
Thank you @DanielVandH 💯

PR #630, opened 2023-11-09
Fix edge case for bridge

### LINKED ISSUE #629: `Bridge` returns `NaN`s for `PolyArea` with intersecting `Ring`s
I was trying to do a geojoin involving United States census block level data, and ran into an error.  I tracked it down to a specific geometry, where calling `simplexify` would introduce `NaN` into the vertices. I have created a minimum working example below.
```julia
julia> ring1 = Ring(
               Point(0., 0.),
               Point(0., 3.),
               Point(2., 3.),
               Point(2., 2.),
               Point(3., 2.),
               Point(3., 0.)
           );

julia> ring2 = Ring(
               Point(1., 1.),
               Point(1., 2.),
               Point(2., 2.),
               Point(2., 1.)
           );

julia> pa = PolyArea(ring1, ring2);

julia> simplexify(pa)
10 SimpleMesh{2,Float64}
  10 vertices
  ├─ Point(0.0, 0.0)
  ├─ Point(0.0, 3.0)
  ├─ Point(2.0, 3.0)
  ├─ Point(NaN, NaN)
  ├─ Point(NaN, NaN)
  ├─ Point(1.0, 2.0)
  ├─ Point(1.0, 1.0)
  ├─ Point(2.0, 1.0)
  ├─ Point(3.0, 2.0)
  └─ Point(3.0, 0.0)
  10 elements
  ├─ Triangle(7, 1, 2)
  ├─ Triangle(6, 2, 3)
  ├─ Triangle(5, 6, 3)
  ├─ Triangle(3, 4, 5)
  ├─ Triangle(6, 7, 2)
  ├─ Triangle(10, 1, 7)
  ├─ Triangle(10, 7, 8)
  ├─ Triangle(10, 8, 5)
  ├─ Triangle(9, 10, 5)
  └─ Triangle(5, 4, 9)
```
Note the NaN's in the list of vertices above.

Here is a visualization of the two rings composing the PolyArea:
mwe

Ethan-Russell:
Fixes #629

If the nearest points to be bridged are identical, I made the outer ring continue by simply using the point rather than creating two shifted versions.  Technically, it makes the resulting polygon intersect itself at that point.  If we require polygons to not intersect themselves, I can amend it so that one of the vertices is shifted, then it continues with the algorithm offsetting the vertices to make a non-intersecting polygon.

### DIFF
--- a/src/transforms/bridge.jl
+++ b/src/transforms/bridge.jl
@@ -68,33 +68,50 @@ function bridge(rings, rinds, δ)
     A = outer[omax]
     B = inner[imax]
 
-    # direction and normal to segment A--B
-    v = B - A
-    u = Vec(-v[2], v[1])
-    n = u / norm(u)
+    # Check for edge case where A and B are equivalent
+    if A == B
+      outer = [
+        outer[begin:(omax - 1)]
+        [A]
+        circshift(inner, -imax + 1)[2:end]
+        [A]
+        outer[(omax + 1):end]
+      ]
+      oinds = [
+        oinds[begin:omax-1]
+        circshift(iinds, -imax + 1)
+        oinds[omax:end]
+      ]
+    else
 
-    # the point A is split into A′ and A′′ and
-    # the point B is split into B′ and B′′ based
-    # on a given bridge width δ
-    A′ = A + (δ / 2) * n
-    A′′ = A - (δ / 2) * n
-    B′ = B + (δ / 2) * n
-    B′′ = B - (δ / 2) * n
+      # direction and normal to segment A--B
+      v = B - A
+      u = Vec(-v[2], v[1])
+      n = u / norm(u)
 
-    # insert hole at closest vertex
-    outer = [
-      outer[begin:(omax - 1)]
-      [A′, B′]
-      circshift(inner, -imax + 1)[2:end]
-      [B′′, A′′]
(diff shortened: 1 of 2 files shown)


juliohm:
That is a very clever fix @Ethan-Russell , thanks for taking the time to read the implementation. I am just concerned that this fix only handles the case where the intersection occurs at a point. What if the inner ring touches the outer ring on a full edge? Maybe we should try to fix the more general case with the "enlargement" idea to fix the issue once and for all?

My suggestion is that we close this PR, and start a fresh one with a new Repair transform that takes a PolyArea and "enlarges" the outer ring to make sure there are no intersections. We could then use this Repair transform inside the Bridge transform like we do with the Repair{9}. What do you think?

The list of currently implemented repairs is here:

https://github.com/JuliaGeometry/Meshes.jl/blob/master/src/transforms/repair.jl

juliohm:
Another reason that I am concerned with this fix here is that discretization algorithms like Dehn1889 assume simple polygons. In particular, they assume that there are no overlapping edges. When we introduce a single point in and out of the hole, we introduce two overlapping edges.

juliohm:
@Ethan-Russell I will close this as discussed above.

Please let us know if you will have some time to dedicate to this fix.

PR #638, opened 2023-11-24
Add method for intersects(::Ray, ::Sphere)

ajahraus:
Add a method for `intersects(::Ray, ::Sphere)`
Implemented via checking if the angle from sphere center, to ray base, to the ray tip is less than the angle from the sphere center, to ray base, to a point tangent to the sphere. If the former is less than the later, they intersect. 

image

### DIFF
--- a/src/predicates/intersects.jl
+++ b/src/predicates/intersects.jl
@@ -37,6 +37,16 @@ intersects(r::Ray, t::Triangle) = !isnothing(r ∩ t)
 
 intersects(t::Triangle, r::Ray) = intersects(r, t)
 
+function intersects(ray::Ray{Dim,T}, sphere::Sphere{Dim,T})::Bool where {Dim,T}
+    sphere_center = center(sphere) - ray(0)
+    r_over_d = radius(sphere) / hypot(sphere_center...)
+    r_over_d > one(T) && return true
+    centered_ray = ray(1) - ray(0)
+    return abs(∠(sphere_center, centered_ray)) < asin(r_over_d)
+end
+
+intersect(sphere::Sphere, ray::Ray) = intersects(ray, sphere)
+
 intersects(p::Point, g::Geometry) = p ∈ g
 
 intersects(g::Geometry, p::Point) = intersects(p, g)
--- a/src/predicates/intersects.jl
+++ b/src/predicates/intersects.jl
@@ -37,7 +37,7 @@ intersects(r::Ray, t::Triangle) = !isnothing(r ∩ t)
 
 intersects(t::Triangle, r::Ray) = intersects(r, t)
 
-function intersects(ray::Ray{Dim,T}, sphere::Sphere{Dim,T})::Bool where {Dim,T}
+function intersects(ray::Ray{Dim,T}, sphere::Union{Ball{Dim,T},Sphere{Dim,T}})::Bool where {Dim,T}
     sphere_center = center(sphere) - ray(0)
     r_over_d = radius(sphere) / hypot(sphere_center...)
     r_over_d > one(T) && return true
@@ -45,7 +45,7 @@ function intersects(ray::Ray{Dim,T}, sphere::Sphere{Dim,T})::Bool where {Dim,T}
     return abs(∠(sphere_center, centered_ray)) < asin(r_over_d)
 end
 
-intersect(sphere::Sphere, ray::Ray) = intersects(ray, sphere)
+intersect(sphere::Union{Ball,Sphere}, ray::Ray) = intersects(ray, sphere)
 
 intersects(p::Point, g::Geometry) = p ∈ g
 
--- a/src/predicates/intersects.jl
+++ b/src/predicates/intersects.jl
@@ -37,15 +37,19 @@ intersects(r::Ray, t::Triangle) = !isnothing(r ∩ t)
 
 intersects(t::Triangle, r::Ray) = intersects(r, t)
 
-function intersects(ray::Ray{Dim,T}, sphere::Union{Ball{Dim,T},Sphere{Dim,T}})::Bool where {Dim,T}
-    sphere_center = center(sphere) - ray(0)
-    r_over_d = radius(sphere) / hypot(sphere_center...)
-    r_over_d > one(T) && return true
-    centered_ray = ray(1) - ray(0)
-    return abs(∠(sphere_center, centered_ray)) < asin(r_over_d)
+function intersects(r::Ray, s::Sphere)
+    s_ = center(s) - r(0)
+    h = hypot(s_...)
+    radius(s) > h && return true
+    r_ = r(1) - r(0)
+    return abs(∠(s_, r_)) < asin(radius(s) / h)
 end
 
-intersect(sphere::Union{Ball,Sphere}, ray::Ray) = intersects(ray, sphere)
+intersects(s::Sphere, r::Ray) = intersects(r, s)
+
+intersects(r::Ray, b::Ball) = intersects(r, boundary(b))
+
+intersects(b::Ball, r::Ray) = intersects(r, b)
 
 intersects(p::Point, g::Geometry) = p ∈ g
 
--- a/src/predicates/intersects.jl
+++ b/src/predicates/intersects.jl
@@ -38,11 +38,11 @@ intersects(r::Ray, t::Triangle) = !isnothing(r ∩ t)
 intersects(t::Triangle, r::Ray) = intersects(r, t)
 
 function intersects(r::Ray, s::Sphere)
-  s₀ = center(s) - r(0)
-  h = hypot(s₀...)
+  u = center(s) - r(0)
+  h = norm(u)
   radius(s) > h && return true
-  r₀ = r(1) - r(0)
-  abs(∠(s₀, r₀)) < asin(radius(s) / h)
+  v = r(1) - r(0)
+  abs(∠(u, v)) < asin(radius(s) / h)
 end
 
 intersects(s::Sphere, r::Ray) = intersects(r, s)
--- a/src/predicates/intersects.jl
+++ b/src/predicates/intersects.jl
@@ -42,7 +42,7 @@ function intersects(r::Ray, s::Sphere)
   h = hypot(s₀...)
   radius(s) > h && return true
   r₀ = r(1) - r(0)
-  return abs(∠(s₀, r₀)) < asin(radius(s) / h)
+  abs(∠(s₀, r₀)) < asin(radius(s) / h)
 end
 
 intersects(s::Sphere, r::Ray) = intersects(r, s)
--- a/src/predicates/intersects.jl
+++ b/src/predicates/intersects.jl
@@ -38,11 +38,11 @@ intersects(r::Ray, t::Triangle) = !isnothing(r ∩ t)
 intersects(t::Triangle, r::Ray) = intersects(r, t)
 
 function intersects(r::Ray, s::Sphere)
-    s_ = center(s) - r(0)
-    h = hypot(s_...)
-    radius(s) > h && return true
-    r_ = r(1) - r(0)
-    return abs(∠(s_, r_)) < asin(radius(s) / h)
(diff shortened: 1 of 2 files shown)


src/predicates/intersects.jl
```diff
@@ -33,6 +33,16 @@ intersects(r::Ray, t::Triangle) = !isnothing(r ∩ t)
 
 intersects(t::Triangle, r::Ray) = intersects(r, t)
 
+function intersects(ray::Ray{Dim,T}, sphere::Union{Ball{Dim,T},Sphere{Dim,T}})::Bool where {Dim,T}
```
juliohm: Let's write a method for `Sphere` and then add a separate method for `Ball` that simply calls `boundary(ball)` and forwards the sphere to the existing method.

Please rename the arguments of the function to single letters as in the other methods. Try to follow the code style of the surrounding code as much as possible: no underscores in variable names, no `return` at the last line, etc.

test/predicates.jl
```diff
@@ -189,6 +189,21 @@
...
+    r =  Ray((0,0),(0,1))
+    @test intersects(r,s) == true
+    @test intersects(s,r) == true
+    s = Sphere(Point3(0,3,0),1)
+    r =  Ray((0,0,0),(1,0,0))
+    @test intersects(r,s) == false
+    @test intersects(s,r) == false
+    r =  Ray((0,0,0),(0,1,0))
+    @test intersects(r,s) == true
+    @test intersects(s,r) == true
```
juliohm: We need more less trivial tests. For example, rays that are not aligned with the coordinate reference system, rays and balls in all 4 quadrants (2D) or 8 octants (3D).

juliohm:
Thank you for the contribution @ajahraus ! Let's start working on the small details to get it merged. :)

test/predicates.jl
ajahraus: I could write a way to generate a collection of tests, i.e.
```julia 
using Rotations: Angle2d

r = Ray((0,0),(1,0))
s_t = Sphere(P2(3, 0), 1)
s_f = Sphere(P2(3, 3), 1)

translations = (Translate(p) for p in [(0,0), (10,0), (0,10), (-10,0), (0,-10))
rotations = (Rotate(Angle2d(theta) for theta in [0 : pi/4 : 7pi/4])
for t in translations, rot in rotations
    r_ =  t(rot(r))
    s_t_ =t(rot(s_t)) 
    s_f_ =t(rot(s_f)) 
    @test intersects(r_, s_t_)
    @test !intersects(r_, s_f_)
end
```
Perhaps that's a bit much, but it would cover all the bases. I can write something similar to for the 3D case, it would just have more translations and rotations to cover all the octants and rotation matrices.

ajahraus: Actually, now that I think about it, that could be made into a function and be used for literally any `intersects` method, as long as you provide a positive example and negative example of the intersections you want to test in various locations and orientations.

juliohm: A function or a loop with tests seems like a good idea. Just make sure you avoid underscores in the variable names according to the code style of the project.

test/predicates.jl
```diff
@@ -130,6 +130,14 @@
   end
 
   @testset "intersects" begin
+    function assertintersecsundertransforms(g1, gt, gf)
```
ajahraus: Typo, should be `assertintersectsundertransforms` ... or something less verbose entirely

ajahraus: The idea behind this is that you provide a geometry `g1`, and one each of another kind of geometry, `gt` and `gf`, where `gt` returns true from `intersects`, and `gf` returns false. 

Do you think this should be type annotated for specifically `Dim = 2`, and `gt` and  `gf` being the same Type?

ajahraus: I guess I would want a similar but distinct implementation for the `Dim == 3` case ...

ajahraus: Oh, and I suppose that `rotations` and `translations` could be defined in their own functions to let them be dispatched independently in the 2 and 3 dimension cases.

src/predicates/intersects.jl
```diff
@@ -33,6 +33,20 @@ intersects(r::Ray, t::Triangle) = !isnothing(r ∩ t)
 
 intersects(t::Triangle, r::Ray) = intersects(r, t)
 
+function intersects(r::Ray, s::Sphere)
+  s₀ = center(s) - r(0)
+  h = hypot(s₀...)
+  radius(s) > h && return true
+  r₀ = r(1) - r(0)
+  return abs(∠(s₀, r₀)) < asin(radius(s) / h)
```
juliohm:
```suggestion
  abs(∠(s₀, r₀)) < asin(radius(s) / h)
```
Is it possible to rewrite this in terms of `atan2`? Not a requirement, but maybe can improve the stability on the edge cases.

test/predicates.jl
```diff
@@ -130,6 +130,14 @@
...
+      translations = Iterators.map(Translate, [(0, 0), (10, 0), (0, 10), (-10, 0), (0, -10)])
+      rotations = Iterators.map(Rotate ∘ Rotations.Angle2d, 0:(π / 4):(7π / 4))
+      transforms = (t ∘ r for (t, r) in Iterators.product(translations, rotations))
+      @test all(intersects(t(g1), t(gt)) for t in transforms)
+      @test all(!intersects(t(g1), t(gf)) for t in transforms)
```
juliohm: Let's rewrite this in terms of a simple for loop:

```julia
for t in Translate.([(), (), ...]), r in Rotate.([...})
  f = t \circ r
  @test intersects(f(g1), f(g2))
end

File: test/predicates.jl (hunk 3 above)
Comment (juliohm) [inline test/predicates.jl]: We can remove this function and copy/paste the for loop directly in the test, after the definition of the two geometries being tested.

File: test/predicates.jl [hunk 6]
```diff
@@ -189,6 +197,11 @@
     @test intersects(t, r1) == true
     @test intersects(r2, t) == false
 
+    r = Ray((0, 0), (1, 0))
+    st = Sphere(P2(3, 0), 1)
+    sf = Sphere(P2(0, 3), 1)
+    assertintersecsundertransforms(r, st, sf)
```
juliohm: Here, you can simply copy/paste the for loop.

src/predicates/intersects.jl
ajahraus: I don't really think so. I mean, geometrically we could, but it would add more complexity to the calculation, and I actually think this simple case nicely encapsulates the question of "what's maximum angular size of a sphere, when measured from the base of the ray?" Perhaps a doc-string explaining the geometry, and perhaps even a link to a diagram (similar to the one I attached) would go a long way to explaining the algorithm in a concise way. 

The edge case of "what if the base of the ray is inside the sphere?" is handled on the early return on line 39, and the only other edge cases I can imagine are if `hypot(s₀...)` or `radius(s)` are somehow negative. But in my estimation, I think we'd want to treat those cases the same as the positive value, which `asin` will do, but `atan2` will not. 

Naturally, my imagination is not boundless, and the possibility of bugs-from-edge-cases always seems to be 😉. I am happy to further arguments in favour of the atan2 case, of course.

test/predicates.jl
ajahraus: If this is the only test case we want to use this formulation for, I'm fine with in-lining it as a for-loop.

But, what are your thoughts on generalizing more of the `intersects` tests using this technique? On one hand, it creates a re-usable function for ensuring that the `intersects` assertions hold true under rigid body transformations, but on the other hand it might prove to be largely redundant and bloat the test suite without providing much additional reliability.

juliohm: In general, we try to copy/paste the tests to facilitate debugging later. When a test fails in a loop inside a function, it is hard to know which iteration failed precisely, so even loops we try to avoid, and instead copy/paste the iterations by hand.

In this case, it suffices to copy/paste the loop in the test directly. Ideally we would just pick example candidates from each octant, orthant.

src/predicates/intersects.jl
juliohm: No problem at all. We can stick to asin as suggested. I just wanted to double check with you first.

src/predicates/intersects.jl
```diff
@@ -33,6 +33,20 @@ intersects(r::Ray, t::Triangle) = !isnothing(r ∩ t)
 
 intersects(t::Triangle, r::Ray) = intersects(r, t)
 
+function intersects(r::Ray, s::Sphere)
+  s₀ = center(s) - r(0)
+  h = hypot(s₀...)
+  radius(s) > h && return true
+  r₀ = r(1) - r(0)
+  abs(∠(s₀, r₀)) < asin(radius(s) / h)
```
juliohm:
```suggestion
  u = center(s) - r(0)
  h = norm(u)
  radius(s) > h && return true
  v = r(1) - r(0)
  abs(∠(u, v)) < asin(radius(s) / h)
```
Can you please double check this suggested change?

test/predicates.jl
ajahraus: Alright, that works for me. I have found it hard to figure out "what exactly went wrong" when parameterizing tests in the past, so I think your logic is very sound.

src/predicates/intersects.jl
ajahraus: Looks good to me 👍

juliohm:
We are almost there! Can you please unroll the loop and pick just a few tricky combinations of ray and sphere in 2D and 3D?

ajahraus:
> We are almost there! Can you please unroll the loop and pick just a few tricky combinations of ray and sphere in 2D and 3D?

I'm afraid I'm not really sure what you mean by "unroll" the loop. Is that a macro? Or is my loop written in a sub-par style? (Tone is hard to convey through text, so I want to emphasize that's a genuine question, not an objection phased like a question 😉)

As for a tricky intersection for 2- and 3-D, I was thinking either the base of the ray could be _very_ far away from the sphere, so that the calculations of the angles are approaching the floating point epsilon; or that the ray intersects the sphere at just _barely_ less than the angle of a ray tangent to the sphere. Thoughts?

juliohm:
> I'm afraid I'm not really sure what you mean by "unroll" the loop. Is that a macro? Or is my loop written in a sub-par style? (Tone is hard to convey through text, so I want to emphasize that's a genuine question, not an objection phased like a question 😉)

No worries :) I meant that we will are better off with an explicit list of examples without the for loop. That way we can visualize each combination of geometries and debug when tests start to fail.

> As for a tricky intersection for 2- and 3-D, I was thinking either the base of the ray could be _very_ far away from the sphere, so that the calculations of the angles are approaching the floating point epsilon; or that the ray intersects the sphere at just _barely_ less than the angle of a ray tangent to the sphere. Thoughts?

I think we can add simple combinations with geometries that are near each other but in different octants and quadrants. Then, we can add more extreme cases where angle calculations approach machine epsilon.

Looking forward to reviewing it again, and merging soon 💯 Thanks for the efforts and patience improving the code!

juliohm:
@ajahraus I fixed another issue, which introduced a conflict with this PR. If you can rebase or resolve the conflicts, and then work on the final details, I'd be happy to review and merge.

juliohm (changes requested):
@ajahraus sorry I wasn't very clear in my last message. I tried to convey the message that we should pick just a few difficult cases in the 4 orthant or 8 octant and write these cases manually. You don't need to literally cover all combinations by hand.

Also notice that you are applying the transform t to both arguments, which is a bit strange. I would expect that a good test set would apply different transformations to the ray and to the sphere, not the same transformation to both.

ajahraus:
Ah, I understand now. The original idea is that, whether or not the ray and sphere intersect originally, the result of the intersection should remain unchanged under any rigid body motion that's applied to both the ray and sphere.

juliohm:
You can define a couple rigid transforms and test that with a clear comment. Something like "result doesn't change under rigid transforms". In that case it makes sense to apply these transforms to both arguments.

After that we need more challenging tests, the ones I actually had in mind. We need different relative configurations between the ray and the sphere.

PR #641, opened 2023-11-29
Specializing `boundingbox` for `Ray` primitives

MachSilva:
Add method `boundingbox(::Ray)` that returns a box that contains the origin of the ray and is unbounded at the direction of the ray to contain all of its points. Since a ray could be seen as a primitive with an "infinite" size, its bounding box should be big enough to contains all of its points. Example:

```julia
julia> ray = Ray(Point(0.41169768366272996, 0.8990554132423699), Vec(0.47249211625247445, 0.2523149692768657))
Ray{2,Float64}
├─ p: Point(0.41169768366272996, 0.8990554132423699)
└─ v: Vec(0.47249211625247445, 0.2523149692768657)

julia> boundingbox(ray)
Box{2,Float64}
├─ min: Point(0.41169768366272996, 0.8990554132423699)
└─ max: Point(Inf, Inf)
```

### DIFF
--- a/src/boundingboxes.jl
+++ b/src/boundingboxes.jl
@@ -34,8 +34,8 @@ function boundingbox(r::Ray{Dim,T}) where {Dim,T}
   upper(p, v) = v > 0 ? typemax(T) : p
   p = r(0)
   v = r(1) - r(0)
-  l = lower.(coordinates(p), coordinates(v))
-  u = upper.(coordinates(p), coordinates(v))
+  l = lower.(coordinates(p), v)
+  u = upper.(coordinates(p), v)
   Box(Point(l), Point(u))
 end
 
--- a/src/boundingboxes.jl
+++ b/src/boundingboxes.jl
@@ -29,6 +29,12 @@ boundingbox(p::Point) = Box(p, p)
 
 boundingbox(b::Box) = b
 
+function boundingbox(r::Ray{Dim,T}) where {Dim,T}
+  lowerbound(x, direction) = direction < 0 ? typemin(T) : x
+  upperbound(x, direction) = direction > 0 ? typemax(T) : x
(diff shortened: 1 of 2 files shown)


src/boundingboxes.jl
```diff
@@ -29,6 +29,12 @@ boundingbox(p::Point) = Box(p, p)
 
 boundingbox(b::Box) = b
 
+function boundingbox(r::Ray{Dim,T}) where {Dim,T}
+  lowerbound(x, direction) = direction < 0 ? typemin(T) : x
+  upperbound(x, direction) = direction > 0 ? typemax(T) : x
+  Box(Point(lowerbound.(r.p.coords, r.v)), Point(upperbound.(r.p.coords, r.v)))
```
```suggestion
  lower(p, v) = v < 0 ? typemin(T) : p
  upper(p, v) = v > 0 ? typemax(T) : p
  p = r(0)
  v = r(1) - r(0)
  l = lower.(coordinates(p), coordinates(v))
  u = upper.(coordinates(p), coordinates(v))
  Box(Point(l), Point(u))
```

juliohm (changes requested):
Thank you very much @MachSilva for the contribution. I've added a few quick changes to guide you in the process. We need to refactor the implementation to avoid using internal fields of structs which are not part of the public api. For example, we cannot access p.coords, instead do coordinates(p).

Can you please refactor the code and ask for review again?

src/boundingboxes.jl
MachSilva: Sure. I'll just check the tests once more.

juliohm (approved):

juliohm:
That is a beautiful first contribution @MachSilva 🎉 

I will just wait for the tests, then merge and release a patch version.

MachSilva:
> That is a beautiful first contribution @MachSilva 🎉
> 
> I will just wait for the tests, then merge and release a patch version.

Thank you for the guidance :)

PR #642, opened 2023-11-30
Fixes #635: search using 3D simplex

### LINKED ISSUE #635: intersects Function Returns True Even When No Intersection Exists
I attempted to assess the intersection of two boundaries (Hexahedron) (there's no 'intersects' available for Hexahedron), specifically, two SimpleMeshes of Quadrangles. The 'intersects' function returns 'true' even when there's no actual intersection in this scenario.
```
julia> h = Hexahedron((0, 0, 0), (4, 0, 0), (4, 4, 0), (0, 4, 0), (1, 1, 2), (3, 1, 2), (3, 3, 2), (1, 3, 2))
julia> h1 =Hexahedron( (1, 1, 1), (3, 1, 1), (3, 3, 1),  (1, 3, 1), (1.5, 1.5, 1.5), (2.5, 1.5, 1.5), (2.5, 2.5, 1.5), (1.5, 2.5, 1.5)) |> Translate(2.6, 0.0, 0.0)

julia> h_bound = boundary(h)
julia> h1_bound = boundary(h1)

julia> intersects(h_bound, h1_bound)
```
immagine_2023-11-22_085523913

MachSilva:
Make `intersects` function to search the intersection with a tetrahedron (3D simplex) instead of a triangle (2D simplex)
when the points are in a 3D euclidean space.

If you can, feel free to add more tests to check if the algorithm is correct.

(Sorry for the misspellings in the commit message)

Fix #635

### DIFF
--- a/src/predicates/intersects.jl
+++ b/src/predicates/intersects.jl
@@ -77,16 +77,19 @@ function intersects(g₁::Geometry{Dim,T}, g₂::Geometry{Dim,T}) where {Dim,T}
     d₂ = simplexify(g₂)
     return intersects(g₁, d₂)
   end
+  return _gjk(g₁, g₂)
+end
 
+function _gjk(g₁::Geometry{2,T}, g₂::Geometry{2,T}) where {T}
   # initial direction
   c₁, c₂ = centroid(g₁), centroid(g₂)
-  d = c₁ ≈ c₂ ? rand(Vec{Dim,T}) : c₂ - c₁
+  d = c₁ ≈ c₂ ? rand(Vec{2,T}) : c₂ - c₁
 
   # first point in Minkowski difference
   P = minkowskipoint(g₁, g₂, d)
 
   # origin of coordinate system
-  O = minkowskiorigin(Dim, T)
+  O = minkowskiorigin(2, T)
 
   # initialize simplex vertices
   points = [P]
@@ -106,8 +109,57 @@ function intersects(g₁::Geometry{Dim,T}, g₂::Geometry{Dim,T}) where {Dim,T}
       AB = B - A
       AO = O - A
       d = AB × AO × AB
-    elseif Dim == 3 == length(points)
-      # triangle case (incomplete simplex)
+    else # simplex case
+      C, B, A = points
+      AB = B - A
+      AC = C - A
+      AO = O - A
+      ABᵀ = AC × AB × AB
+      ACᵀ = AB × AC × AC
+      if ABᵀ ⋅ AO > zero(T)
+        popat!(points, 1) # pop C
+        d = ABᵀ
+      elseif ACᵀ ⋅ AO > zero(T)
+        popat!(points, 2) # pop B
+        d = ACᵀ
+      else
+        return true
+      end
+    end
+  end
+end
+
+function _gjk(g₁::Geometry{3,T}, g₂::Geometry{3,T}) where {T}
+  # initial direction
+  c₁, c₂ = centroid(g₁), centroid(g₂)
+  d = c₁ ≈ c₂ ? rand(Vec{3,T}) : c₂ - c₁
+
+  # first point in Minkowski difference
+  P = minkowskipoint(g₁, g₂, d)
+
+  # origin of coordinate system
+  O = minkowskiorigin(3, T)
+
+  # initialize simplex vertices
+  points = [P]
+
+  # move towards the origin
+  d = O - P
+  while true
+    P = minkowskipoint(g₁, g₂, d)
+    if (P - O) ⋅ d < zero(T)
+      return false
+    end
+    push!(points, P)
+
+    # line segment case
+    if length(points) == 2
+      B, A = points
+      AB = B - A
+      AO = O - A
+      d = AB × AO × AB
+    elseif length(points) == 3
+      # triangle case
       C, B, A = points
       AB = B - A
       AC = C - A
@@ -118,52 +170,36 @@ function intersects(g₁::Geometry{Dim,T}, g₂::Geometry{Dim,T}) where {Dim,T}
         ABCᵀ = -ABCᵀ
       end
       d = ABCᵀ
-    else # simplex case
-      if Dim == 2
-        C, B, A = points
-        AB = B - A
-        AC = C - A
-        AO = O - A
-        ABᵀ = AC × AB × AB
-        ACᵀ = AB × AC × AC
-        if ABᵀ ⋅ AO > zero(T)
-          popat!(points, 1) # pop C
-          d = ABᵀ
-        elseif ACᵀ ⋅ AO > zero(T)
-          popat!(points, 2) # pop B
-          d = ACᵀ
-        else
-          return true
-        end
-      else # if Dim == 3
-        #      A
-        #    / | \
-        #   /  D  \
-        #  / /   \ \
-        # B ------- C
-        # Simplex faces (with normal vectors pointing to the centroid):
-        # ABC, ADB, BDC, ACD
-        # (AXY = AX ⋅ AY)
-        D, C, B, A = points
-        AB = B - A
-        AC = C - A
-        AD = D - A
-        AO = O - A
-        ABCᵀ = AB × AC
-        ADBᵀ = AD × AB
-        ACDᵀ = AC × AD
-        if ABCᵀ ⋅ AO > zero(T)
-          popat!(points, 1) # pop D
-          d = ABCᵀ
-        elseif ADBᵀ ⋅ AO > zero(T)
-          popat!(points, 2) # pop C
-          d = ADBᵀ
-        elseif ACDᵀ ⋅ AO > zero(T)
-          popat!(points, 3) # pop B
-          d = ACDᵀ
-        else
-          return true
-        end
+    else
+      # simplex case
+      #      A
+      #    / | \
+      #   /  D  \
+      #  / /   \ \
+      # B ------- C
+      # Simplex faces (with normal vectors pointing to the centroid):
+      # ABC, ADB, BDC, ACD
+      # (AXY = AX × AY)
+      # ABC normal vector points to vertex D
+      D, C, B, A = points
+      AB = B - A
+      AC = C - A
+      AD = D - A
+      AO = O - A
+      ABCᵀ = AB × AC
+      ADBᵀ = AD × AB
+      ACDᵀ = AC × AD
+      if ABCᵀ ⋅ AO > zero(T)
+        popat!(points, 1) # pop D
+        d = ABCᵀ
+      elseif ADBᵀ ⋅ AO > zero(T)
+        popat!(points, 2) # pop C
+        d = ADBᵀ
+      elseif ACDᵀ ⋅ AO > zero(T)
+        popat!(points, 3) # pop B
+        d = ACDᵀ
+      else
+        return true
       end
     end
   end
--- a/src/predicates/intersects.jl
+++ b/src/predicates/intersects.jl
@@ -155,7 +155,7 @@ function gjk!(O::Point{3,T}, points) where T
     B, A = points
     AB = B - A
     AO = O - A
-    d = AB × AO × AB
+    d = perpendicular(AB, AO)
   elseif length(points) == 3
     # triangle case
     C, B, A = points
@@ -258,4 +258,4 @@ function perpendicular(v::Vec{2,T}, d::Vec{2,T}) where T
   Vec(r[1], r[2])
 end
 
-perpendicular(v::Vec{3}, d::Vec{3}) = a × b × a
+perpendicular(v::Vec{3}, d::Vec{3}) = v × d × v
--- a/src/predicates/intersects.jl
+++ b/src/predicates/intersects.jl
@@ -164,7 +164,7 @@ function gjk!(O::Point{3,T}, points) where T
     AC = C - A
     AO = O - A
     ABCᵀ = AB × AC
-    if ABCᵀ ⋅ AO > 0
+    if ABCᵀ ⋅ AO < 0
       points[1], points[2] = points[2], points[1]
       ABCᵀ = -ABCᵀ
     end
--- a/src/predicates/intersects.jl
+++ b/src/predicates/intersects.jl
@@ -106,21 +106,64 @@ function intersects(g₁::Geometry{Dim,T}, g₂::Geometry{Dim,T}) where {Dim,T}
       AB = B - A
       AO = O - A
       d = AB × AO × AB
-    else # simplex case
+    elseif Dim == 3 == length(points)
+      # triangle case (incomplete simplex)
       C, B, A = points
       AB = B - A
       AC = C - A
       AO = O - A
-      ABᵀ = AC × AB × AB
-      ACᵀ = AB × AC × AC
-      if ABᵀ ⋅ AO > zero(T)
-        popat!(points, 1) # pop C
-        d = ABᵀ
-      elseif ACᵀ ⋅ AO > zero(T)
-        popat!(points, 2) # pop B
-        d = ACᵀ
-      else
-        return true
+      ABCᵀ = AB × AC
+      if ABCᵀ ⋅ AO < 0
+        points[1], points[2] = points[2], points[1]
+        ABCᵀ = -ABCᵀ
+      end
+      d = ABCᵀ
+    else # simplex case
+      if Dim == 2
+        C, B, A = points
+        AB = B - A
+        AC = C - A
+        AO = O - A
+        ABᵀ = AC × AB × AB
+        ACᵀ = AB × AC × AC
+        if ABᵀ ⋅ AO > zero(T)
+          popat!(points, 1) # pop C
+          d = ABᵀ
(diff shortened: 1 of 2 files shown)


juliohm:
It is lovely to see a fix for that! Can you perhaps share a few drawings of the implemented adjustments? After that we can reorganize the code into subfunctions handling the different cases in 2D and 3D to facilitate maintenance and improve readability.

MachSilva:
Yes. I'll look into it as soon as I can.

juliohm:
@MachSilva did you have a chance to take a look into this? The feature in this PR is super useful.

MachSilva:
Yes. I'll share the explanation of the code adjustments soon. I also agree
about splitting the algorithm in two subfunctions.

MachSilva:
The previous version of the `intersects` function handles only geometries in 2D space using the GJK algorithm. Assuming previous knowledge of the GJK algorithm from the reader, we end up with two tasks:
- build a simplex using the points from Minkowski difference
- find a simplex that contains the origin
    - if we find such simplex, an intersection has been found;
    - otherwise, if it is impossible, no intersection exists.

The simplex is represented by the set `points` in the code. The search is done by a while loop.
The loop keeps trying to add a new point to the set to form a simplex and then checks if the simplex contains the origin.
A simplex in 2D is a triangle. A simplex in 3D is a tetrahedron.
The loop starts with one point, so for each iteration we have two options:
- the set is incomplete and does not contain enough vertices to form a simplex;
- or the set does form a simplex (`length(points) == 3` if `Dim` = 2, `length(points) == 4` if `Dim` = 3).

In the first case, we just need to add more points in the right direction. In the last case, we need to check if the simplex contains the origin.  This dimension check was the first modification that I did.

Previously the algorithm handled 3D geometries by performing a search with a triangle. Each edge of the triangle represented a line that divided the 2D plane into to regions (one of them with the origin). However in 3D space, each edge of the triangle represented a plane that divided the space in two volumes, going from -Inf to +Inf in Z-axis. That's the cause of the false positives.

So the second modification was implement the check with a tetrahedron. Just as the triangle case in 2D, that for each edge created a normal vector hoping that the origin was in the same region (of a total of 2) the vector was pointing; I did the same with the tetrahedron: for each face, I found its normal vector and checked if the origin was in the same volume that the vector was pointing.

However, for deriving normal vectors from the faces I needed to define some standards.
So, I defined in the code that the variables of the type AXY would represent the normal vectors of the face AXY and that they should point to the centroid of the tetrahedron. Then I defined that the tetrahedron would be represented by an *ordered* set of four points (D,C,B,A) that has a face ABC whose normal vector prod(AB, AC) would be a vector pointing to the centroid and to the vertex D. I need this to avoid vectors with flipped directions (normal(ABC) = - normal(ACB)).

This is the relevant code section (the order of the points matters):
```julia
D, C, B, A = points
AB = B - A
AC = C - A
AD = D - A
AO = O - A
ABCᵀ = AB × AC
ADBᵀ = AD × AB
ACDᵀ = AC × AD
if ABCᵀ ⋅ AO > zero(T)
  popat!(points, 1) # pop D
  d = ABCᵀ
elseif ADBᵀ ⋅ AO > zero(T)
  popat!(points, 2) # pop C
  d = ADBᵀ
elseif ACDᵀ ⋅ AO > zero(T)
  popat!(points, 3) # pop B
  d = ACDᵀ
else
  return true
end
```

I tried to make some drawings, but I could not think of any helpful drawing. I hope my answer was clear. If not, feel free to ask.

MachSilva:
One way of making the function easier to read is to delegate the algorithm to another function (maybe called `_gjk`) and define two specializations. One for Dim=2 and another one for Dim=3.
```julia
function intersects(g₁::Geometry{Dim,T}, g₂::Geometry{Dim,T}) where {Dim,T}
  # must have intersection of bounding boxes
  intersects(boundingbox(g₁), boundingbox(g₂)) || return false

  # handle non-convex geometries
  if !isconvex(g₁)
    d₁ = simplexify(g₁)
    return intersects(d₁, g₂)
  elseif !isconvex(g₂)
    d₂ = simplexify(g₂)
    return intersects(g₁, d₂)
  end

  return _gjk(g₁, g₂)
end

function _gjk(g₁::Geometry{2,T}, g₂::Geometry{2,T}) where {T}
  # initial direction
  [...]
  # move towards the origin
  d = O - P
  while true
    P = minkowskipoint(g₁, g₂, d)
    if (P - O) ⋅ d < zero(T)
      return false
    end
    push!(points, P)
    [...]
  end
end

function _gjk(g₁::Geometry{3,T}, g₂::Geometry{3,T}) where {T}
  # initial direction
 [...]
  # move towards the origin
  d = O - P
  while true
    P = minkowskipoint(g₁, g₂, d)
    if (P - O) ⋅ d < zero(T)
      return false
    end
    push!(points, P)
    [...]
  end
end
```

juliohm:
Thank you @MachSilva for the detailed explanation 💯 I like approach where we introduce two functions and dispatch on embedding dimension. Could you please update this PR? That we can pinpoint the exact differences between the two algorithms in 2D and 3D.

MachSilva:
Done. I've separated the two functions in a new commit.

juliohm:
Thank you @MachSilva ! Can you please refactor the code so that the common parts of both functions are reused? I see that the codes are very similar with the exception of the inner loop?

juliohm:
I believe that we can keep the `Dim` parameter generic in the main `intersection` function, and then dispatch different _gjk functions in the inner loop. Makes sense?

MachSilva:
I think it is more readable now.

Edit: I triggered the wrong test subset locally before pushing the commit. I'll fix it, then run the full test set.

juliohm:
It is much better now. I think you have a typo in the function name. You can call it `gjk!` without the underscore. The tests are failing because of the typo.

MachSilva:
Actually I've run the wrong test subset locally. Some variables weren't declared. I'll change the name, fix it then run all the tests.

src/predicates/intersects.jl
```diff
@@ -106,22 +107,94 @@ function intersects(g₁::Geometry{Dim,T}, g₂::Geometry{Dim,T}) where {Dim,T}
...
+the direction for searching the next point.
+
+If the simplex is complete, it removes one point from the set to
+make room for the next point. A complete simplex must have `Dim + 1` points. 
+
+See also [`intersects`](@ref).
+"""
+function gjk! end
+
+function gjk!(::Val{2}, origin, points)
```
juliohm: I believe that we can dispatch on the type of `origin::Point{2,T}` and avoid the first argument. Also, the parameter T would be immediately available for use inside the function.

src/predicates/intersects.jl
```diff
@@ -106,22 +107,94 @@ function intersects(g₁::Geometry{Dim,T}, g₂::Geometry{Dim,T}) where {Dim,T}
...
+    # tetrahedron simplex case
+    #      A
+    #    / | \
+    #   /  D  \
+    #  / /   \ \
+    # B ------- C
+    # Simplex faces (with normal vectors pointing to the centroid):
+    # ABC, ADB, BDC, ACD
+    # (AXY = AX × AY)
+    # ABC normal vector points to vertex D
```
juliohm: Nice comment! This will help a lot in the future for anyone interested in contributing.

src/predicates/intersects.jl
```diff
@@ -106,22 +107,94 @@ function intersects(g₁::Geometry{Dim,T}, g₂::Geometry{Dim,T}) where {Dim,T}
...
+    AC = C - A
+    AD = D - A
+    AO = origin - A
+    ABCᵀ = AB × AC
+    ADBᵀ = AD × AB
+    ACDᵀ = AC × AD
+    T = coordtype(origin)
+    if ABCᵀ ⋅ AO > zero(T)
+      popat!(points, 1) # pop D
+      return ABCᵀ
```
juliohm: Our code style avoids `return`, it is only used if we need to return prematurely.

src/predicates/intersects.jl
MachSilva: Yes. That's what I want to do. However, I'll need to modify the functions `minkowskipoint` and `minkowskiorigin` because they make the 2D case dispatch on the `origin::Point{3,T}`.

src/predicates/intersects.jl
MachSilva: Good to know. May you send me any additional code style guides of this repository? I did not find any guide about for `return` statements in the mentioned repository https://github.com/SciML/ColPrac, just about not changing function return types.

Additionally, to make the `gjk!` function dispatch on `origin::Point{2,T}`, I'll need to change the return type from `minkowskipoint(g₁::Geometry{2,T}, g₂::Geometry{2,T}, d)::Point{3,T}` to `minkowskipoint(g₁::Geometry{2,T}, g₂::Geometry{2,T}, d)::Point{2,T}` when Dim=2.

juliohm: We have JuliaFormatter set on this repository. It will fix most of the code style issues for us automatically. The `return` issue is something we can't fix automatically, but it is easy to share with first-time contributors. Thanks for trying to adhere to it.

No problem if you need to change the `minkowskipoint` and `minkowskiorigin`.

src/predicates/intersects.jl
```diff
@@ -168,8 +247,18 @@ intersects(m::Multi, c::Chain) = intersects(c, m)
...
+
+# find a vector perpendicular to `v` using vector `d` as some direction hint
+# expect that `perpendicular(v, d) ⋅ d ≥ 0` or, in other words,
+# that the angle between the result vector and `d` is less or equal than 90º
+function perpendicular(v::Vec{2,T}, d::Vec{2,T}) where T
+  a = Vec(v[1], v[2], zero(T))
+  b = Vec(d[1], d[2], zero(T))
+  r = a × b × a
+  Vec(r[1], r[2])
+end
```
juliohm, line 259: Very good idea 💯

MachSilva: Thank you. I tried to rewrite the function to use only 2D vectors.
It was something like:
```julia
function perpendicular(v::Vec{2,T}, d::Vec{2,T}) where T
  M = @SMatrix T[0 -1; 1 0] # rotation matrix 90º
  M * (sign(det([v d])) * v) # det([v d]) is proportional to sin(angle(v, d))
end
```

It almost worked, but it failed in just one test. So, I didn't use it.

Now, I just need to write some tests.

juliohm: It would be nice to use the function in the 3D case for consistency, even if the implementation differs. We can dispatch on the dimension again to trigger the correct implementation.

src/predicates/intersects.jl
```diff
@@ -93,37 +93,116 @@ function intersects(g₁::Geometry{Dim,T}, g₂::Geometry{Dim,T}) where {Dim,T}
...
+the direction for searching the next point.
+
+If the simplex is complete, it removes one point from the set to
+make room for the next point. A complete simplex must have `Dim + 1` points. 
+
+See also [`intersects`](@ref).
+"""
+function gjk! end
+
+function gjk!(origin::Point{2,T}, points) where T
```
juliohm: let's rename the function argument `origin` to `O` to keep the variable names as similar as possible in the git diff.

src/predicates/intersects.jl
```diff
@@ -168,8 +247,18 @@ intersects(m::Multi, c::Chain) = intersects(c, m)
 function minkowskipoint(g₁::Geometry{Dim,T}, g₂::Geometry{Dim,T}, d) where {Dim,T}
   n = Vec{Dim,T}(d[1:Dim])
   v = supportfun(g₁, n) - supportfun(g₂, -n)
-  Point(ntuple(i -> i ≤ Dim ? v[i] : zero(T), max(Dim, 3)))
+  Point(ntuple(i -> i ≤ Dim ? v[i] : zero(T), Dim))
```
juliohm: I believe that we don't need the ternary operator here anymore given that we are now using `Dim` consistently with the dimension of the coordinates? Perhaps we just need to return `v` now?

juliohm:
@MachSilva did you have some time to finalize this one?

MachSilva:
No, not yet.

2023年12月18日(月) 16:53 Júlio Hoffimann ***@***.***>:

MachSilva:
> @MachSilva did you have some time to finalize this one?

No, not yet. I'll have time again after a few days.

juliohm:
@MachSilva pinging you again on this one :)

src/predicates/intersects.jl
```diff
@@ -93,37 +93,116 @@ function intersects(g₁::Geometry{Dim,T}, g₂::Geometry{Dim,T}) where {Dim,T}
 
   # move towards the origin
   d = O - P
+
```
juliohm:
```suggestion
```

src/predicates/intersects.jl
```diff
@@ -166,10 +244,22 @@ intersects(m::Multi, c::Chain) = intersects(c, m)
 
 # support point in Minkowski difference
 function minkowskipoint(g₁::Geometry{Dim,T}, g₂::Geometry{Dim,T}, d) where {Dim,T}
-  n = Vec{Dim,T}(d[1:Dim])
+  n = Vec{Dim,T}(d)
   v = supportfun(g₁, n) - supportfun(g₂, -n)
```
juliohm:
```suggestion
  v = supportfun(g₁, d) - supportfun(g₂, -d)
```

src/predicates/intersects.jl
```diff
@@ -166,10 +244,22 @@ intersects(m::Multi, c::Chain) = intersects(c, m)
 
 # support point in Minkowski difference
 function minkowskipoint(g₁::Geometry{Dim,T}, g₂::Geometry{Dim,T}, d) where {Dim,T}
```
juliohm:
```suggestion
function minkowskipoint(g₁::Geometry, g₂::Geometry, d)
```

src/predicates/intersects.jl
```diff
@@ -165,11 +243,22 @@ intersects(m::Multi, c::Chain) = intersects(c, m)
...
 
 # support point in Minkowski difference
-function minkowskipoint(g₁::Geometry{Dim,T}, g₂::Geometry{Dim,T}, d) where {Dim,T}
-  n = Vec{Dim,T}(d[1:Dim])
-  v = supportfun(g₁, n) - supportfun(g₂, -n)
-  Point(ntuple(i -> i ≤ Dim ? v[i] : zero(T), max(Dim, 3)))
+function minkowskipoint(g₁::Geometry, g₂::Geometry, d)
+  v = supportfun(g₁, d) - supportfun(g₂, -d)
+  Point(v)
 end
```
juliohm:
```suggestion
minkowskipoint(g₁::Geometry, g₂::Geometry, d) = Point(supportfun(g₁, d) - supportfun(g₂, -d))
```

src/predicates/intersects.jl
```diff
@@ -165,11 +243,19 @@ intersects(m::Multi, c::Chain) = intersects(c, m)
...
+# expect that `perpendicular(v, d) ⋅ d ≥ 0` or, in other words,
+# that the angle between the result vector and `d` is less or equal than 90º
+function perpendicular(v::Vec{2,T}, d::Vec{2,T}) where T
+  a = Vec(v[1], v[2], zero(T))
+  b = Vec(d[1], d[2], zero(T))
+  r = a × b × a
+  Vec(r[1], r[2])
+end
+
+perpendicular(v::Vec{3,T}, d::Vec{3,T}) where T = a × b × a
```
juliohm:
```suggestion
perpendicular(v::Vec{3}, d::Vec{3}) = a × b × a
```

src/predicates/intersects.jl
```diff
@@ -100,30 +100,108 @@ function intersects(g₁::Geometry{Dim,T}, g₂::Geometry{Dim,T}) where {Dim,T}
...
+  d
+end
+
+function gjk!(O::Point{3,T}, points) where T
+  # line segment case
+  if length(points) == 2
+    B, A = points
+    AB = B - A
+    AO = O - A
+    d = AB × AO × AB
```
juliohm:
```suggestion
    d = perpendicular(AB, AO)
```

src/predicates/intersects.jl
```diff
@@ -165,11 +243,19 @@ intersects(m::Multi, c::Chain) = intersects(c, m)
...
+# expect that `perpendicular(v, d) ⋅ d ≥ 0` or, in other words,
+# that the angle between the result vector and `d` is less or equal than 90º
+function perpendicular(v::Vec{2,T}, d::Vec{2,T}) where T
+  a = Vec(v[1], v[2], zero(T))
+  b = Vec(d[1], d[2], zero(T))
+  r = a × b × a
+  Vec(r[1], r[2])
+end
+
+perpendicular(v::Vec{3}, d::Vec{3}) = a × b × a
```
juliohm:
```suggestion
perpendicular(v::Vec{3}, d::Vec{3}) = v × d × v
```

juliohm (approved):
Thank you @MachSilva for updating the PR, and for addressing this long-standing feature request ❤️ I will wait for the tests before merging and releasing a patch.

Appreciate your help here!

PR #645, opened 2023-12-05
Remove the `unview` function in favor of `parent` and `parentindices`

eliascarv:


src/domains.jl
```diff
@@ -47,6 +47,10 @@ Base.eltype(domain::Domain) = eltype([domain[i] for i in 1:nelements(domain)])
 
 Base.keys(domain::Domain) = 1:nelements(domain)
 
+Base.parent(d::Domain) = d
+
+Base.parentindices(d::Domain) = 1:nelements(d)
```
juliohm:
```suggestion
Base.parent(domain::Domain) = domain

Base.parentindices(domain::Domain) = 1:nelements(domain)
```

src/subdomains.jl
```diff
@@ -17,7 +17,7 @@ struct SubDomain{Dim,T,D<:Domain{Dim,T},I<:AbstractVector{Int}} <: Domain{Dim,T}
 end
 
 # specialize constructor to avoid infinite loops
-SubDomain(v::SubDomain, inds::AbstractVector{Int}) = SubDomain(getfield(v, :domain), getfield(v, :inds)[inds])
+SubDomain(v::SubDomain, inds::AbstractVector{Int}) = SubDomain(v.domain, v.inds[inds])
```
juliohm: Let's rename all `v` that previously meant "view" to `d` for domain.

juliohm (changes requested):

juliohm (approved):

PR #647, opened 2023-12-08
Implement the `vcat` function for Domains

eliascarv:


src/subdomains.jl
```diff
@@ -29,6 +29,15 @@ nelements(d::SubDomain) = length(d.inds)
 
 centroid(d::SubDomain, ind::Int) = centroid(d.domain, d.inds[ind])
 
+# specialized for efficiency
+function Base.vcat(d1::SubDomain, d2::SubDomain)
+  if d1.domain === d2.domain
+    SubDomain(d1.domain, vcat(d1.inds, d2.inds))
```
juliohm, line 35: We also need to check that the intersection of d1.inds and d2.inds is empty.

juliohm (approved):

PR #648, opened 2023-12-12
Refactor sideof surface mesh

juliohm:
@kylebeggs some of the tests are failing and I would like to simplify the original code that you submitted some time ago for improved performance. Can you please comment on the different branches of the code? Isn't the ray casting algorithm only concerned with how many times we intersect with triangles, no matter the type of intersection?

Can we simplify these branches only account for the cases IN (including the case ON the surface) and OUT?


juliohm:
PR includes undesired commit. Opening another one.

PR #649, opened 2023-12-12
Refactor sideof surface mesh

juliohm:
@kylebeggs some of the tests are failing and I would like to simplify the original code that you submitted some time ago for improved performance. Can you please comment on the different branches of the code? Isn't the ray casting algorithm only concerned with how many times we intersect with triangles, no matter the type of intersection?

Can we simplify these branches only account for the cases IN (including the case ON the surface) and OUT?

### DIFF
--- a/src/sideof.jl
+++ b/src/sideof.jl
@@ -68,40 +68,71 @@ sideof(points, geom::Geometry) = map(point -> sideof(point, geom), points)
 Determines on which side the `point` is in relation to the surface `mesh`.
 Possible results are `IN`, `OUT`, or `ON` the `mesh`.
 """
-sideof(point::Point{3}, mesh::Mesh{3}) = sideof((point,), mesh) |> first
+sideof(point::Point{3}, mesh::Mesh{3}) = sideof([point], mesh) |> first
 
 function sideof(points, mesh::Mesh{3})
   @assert paramdim(mesh) == 2 "sideof only defined for surface meshes"
+
+  # triangulate mesh if necessary
   (eltype(mesh) <: Triangle) || return sideof(points, simplexify(mesh))
-  map(point -> _sideof(point, mesh), points)
+
+  # SVD basis for projection
+  verts = vertices(mesh)
+  basis = svdbasis(verts)
+
+  # project points and mesh on SVD basis
+  ppts = proj(points, basis)
+  pmsh = SimpleMesh(proj(verts, basis), topology(mesh))
+
+  # bounding box of projected triangles
+  boxs = boundingbox.(pmsh)
+
+  # retreive basis
+  u, v = basis
+
+  # loop over query points
+  map(1:length(points)) do i
+    # find triangles with intersecting bounding box
+    inds = findall(b -> ppts[i] ∈ b, boxs)
+
+    if isempty(inds)
+      OUT # the point is outside the mesh
+    else
+      # filter original triangle mesh
+      fmsh = view(mesh, inds)
+
+      # define ray from SVD basis
+      nray = Ray(points[i], u × v)
+
+      # ray casting algorithm with filtered mesh
+      _raycasting(nray, fmsh)
+    end
+  end
 end
 
-function _sideof(point::Point{3,T}, mesh::Mesh{3,T}) where {T}
-  z = last.(coordinates.(extrema(mesh)))
-  r = Ray(point, Vec(zero(T), zero(T), 2 * (z[2] - z[1])))
-
-  hasintersect = false
-  edgecrosses = 0
-  ps = Point{3,T}[]
-  for t in mesh
-    I = intersection(r, t)
+function _raycasting(ray::Ray{3,T}, tris) where {T}
+  ecross = 0 # number of edge crosses
+  inters = false # do we have intersection?
+  points = Point{3,T}[]
+  for tri in tris
+    I = intersection(ray, tri)
     if type(I) == Crossing
-      hasintersect = !hasintersect
+      inters = !inters
     elseif type(I) ∈ (EdgeTouching, CornerTouching, Touching)
       return ON
     elseif type(I) == EdgeCrossing
-      edgecrosses += 1
+      ecross += 1
     elseif type(I) == CornerCrossing
-      p = get(I)
-      if !any(≈(p), ps)
-        push!(ps, p)
-        hasintersect = !hasintersect
+      point = get(I)
+      if !any(≈(point), points)
+        push!(points, point)
+        inters = !inters
       end
     end
   end
 
   # check how many edges we crossed
-  isodd(edgecrosses ÷ 2) && (hasintersect = !hasintersect)
+  isodd(ecross ÷ 2) && (inters = !inters)
 
-  hasintersect ? IN : OUT
+  inters ? IN : OUT
 end
--- a/test/sideof.jl
+++ b/test/sideof.jl
@@ -1,15 +1,27 @@
 @testset "sideof" begin
+  # -----
+  # LINE
+  # -----
+
   p1, p2, p3 = P2(0, 0), P2(1, 1), P2(0.25, 0.5)
   l = Line(P2(0.5, 0.0), P2(0.0, 1.0))
   @test sideof(p1, l) == LEFT
   @test sideof(p2, l) == RIGHT
   @test sideof(p3, l) == ON
 
+  # -----
+  # RING
+  # -----
+
   p1, p2, p3 = P2(0.5, 0.5), P2(1.5, 0.5), P2(1, 1)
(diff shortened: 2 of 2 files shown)


src/sideof.jl
```diff
@@ -68,40 +68,71 @@ sideof(points, geom::Geometry) = map(point -> sideof(point, geom), points)
...
-  ps = Point{3,T}[]
-  for t in mesh
-    I = intersection(r, t)
+function _raycasting(ray::Ray{3,T}, tris) where {T}
+  ecross = 0 # number of edge crosses
+  inters = false # do we have intersection?
+  points = Point{3,T}[]
+  for tri in tris
+    I = intersection(ray, tri)
     if type(I) == Crossing
```
kylebeggs, line 119: This simply checks if the ray crosses the triangle in the middle as you would expect in this algorithm. This is the main check (vast majority of time the algorithm only hits this branch)

src/sideof.jl
```diff
@@ -68,40 +68,71 @@ sideof(points, geom::Geometry) = map(point -> sideof(point, geom), points)
...
-      hasintersect = !hasintersect
+      inters = !inters
     elseif type(I) ∈ (EdgeTouching, CornerTouching, Touching)
```
kylebeggs, line 121: This branch checks if the query point lies ON the surface (triangle). You can't really get rid of this branch, unless you know of a faster way to check if the query point lies on the surface and do that before starting the loop.

src/sideof.jl
```diff
@@ -68,40 +68,71 @@ sideof(points, geom::Geometry) = map(point -> sideof(point, geom), points)
...
       return ON
     elseif type(I) == EdgeCrossing
```
kylebeggs, line 123: You have to count edges separately because 2 triangles share an edge, this "intersection" will count twice for our purpose (hence the divide by 2 towards the end)

src/sideof.jl
```diff
@@ -68,40 +68,71 @@ sideof(points, geom::Geometry) = map(point -> sideof(point, geom), points)
...
-      edgecrosses += 1
+      ecross += 1
     elseif type(I) == CornerCrossing
```
kylebeggs, line 125: Lastly, you also have the handle if the ray hits a corner separately because you have no guarantees how many triangles are connected to that corner and how many repeated "intersections" to account for (unlike edges where we guaranteed are 2). You could probably handle this and the edges branch better by creating a queue for all the triangles and then when you hit one of these branches you find all connected triangles and dequeue them, so you don't have to have this vector you push to (`points`). But there is no way I'm aware of to avoid these branches altogether without making some assumptions about the mesh. I figured we wanted to keep this function as general as possible to work for any mesh.

kylebeggs (commented):
I added some comments on the branches. I don't think you can do much better to get rid of the branches if you want to keep the algorithm completely robust. If you are OK with an approximation to the `sideof` query, then there are much much faster algorithms. Perhaps you can create a hybrid where you do the fast algorithm first and check the result and if it is close to a boundary, then run this raycasting approach.

juliohm:
> If you are OK with an approximation to the `sideof` query, then there are much much faster algorithms. Perhaps you can create a hybrid where you do the fast algorithm first and check the result and if it is close to a boundary, then run this raycasting approach.

Thank you, that sounds like an interesting approach. Please feel free to share some references on these faster approximations. I would love to take a look and find the time to implement.

juliohm:
@kylebeggs let's assume for the moment that we can store all the intersections between the ray and the triangles in a vector. That is actually the case now given that we are filtering the triangles using their bounding boxes on a 2D projection.

So we have the following vector of intersections:

```julia
ints = [intersection(ray, tri) for tri in tris]
```

Can you rewrite the algorithm you wrote in terms of this vector now? That will help clarify the different criteria, and will potentially help us find optimization paths.

Perhaps we can handle the corner cases in a preprocessing step using only intersection between ray and segments of triangles.

juliohm:
Another question, I noticed that you defined the ray in terms of the bounding box of the mesh using the following:

```julia
z = last.(coordinates.(extrema(mesh)))
r = Ray(point, Vec(zero(T), zero(T), 2 * (z[2] - z[1])))
```

Can we use `Vec(0, 0, 1)` instead? By definition, `Ray(p, v) == Ray(p, 2v) == Ray(p, 3v)`.

kylebeggs:
> Can we use Vec(0, 0, 1) instead? By definition, Ray(p, v) == Ray(p, 2v) == Ray(p, 3v).

Yes, I was wrongly treating `Ray` as a `Segment`.

> ints = [intersection(ray, tri) for tri in tris]
Can you rewrite the algorithm you wrote in terms of this vector now? That will help clarify the different criteria, and will potentially help us find optimization paths.

I think you mean code optimization here? I don't think there is much more you can do regarding algorithmic optimization. There is no way around having to check all the unusual/rare cases like hitting corners/edges. As far as code optimization goes, I think there are potentially better ways to write it but depends on how "cheap" you can find all the triangles that are connected to each other via edges or corners. I'm not aware of that or how fast that would be. Unfortunately, I'm not aware of any way to get rid of these branches other than filtering and counting the vector for the different cases, but I highly doubt that would be faster than what is currently done.

I think the most impact could be made by creating a specific ray-triangle intersection algorithm which cuts down on the number of possible return types. You could get rid of some of the branches in that code, and then you'd be able to get rid of some branches here (mainly the ON cases).

> Thank you, that sounds like an interesting approach. Please feel free to share some references on these faster approximations. I would love to take a look and find the time to implement.

Here is the [paper](https://dl.acm.org/doi/10.1145/3197517.3201337) I was referencing that I think we should implement (I actually already implemented this, but it doesn't include the optimization in section 3.2 of the paper)

juliohm:
> I think you mean code optimization here?

I mean a simple function that takes the list of intersections as input, and returns the 3 cases IN, OUT, ON.

> Here is the [paper](https://dl.acm.org/doi/10.1145/3197517.3201337) I was referencing that I think we should implement (I actually already implemented this, but it doesn't include the optimization in section 3.2 of the paper)

Will take a look, thanks for sharing :)

kylebeggs:
> I mean a simple function that takes the list of intersections as input, and returns the 3 cases IN, OUT, ON.

I don't understand. It is basically `_raycasting` here. This would be it:
```julia
function _check_ints(ints)
  ecross = 0 # number of edge crosses
  inters = false # do we have intersection?
  points = Point{3,T}[]
  for I in ints
    if type(I) == Crossing
      inters = !inters
    elseif type(I) ∈ (EdgeTouching, CornerTouching, Touching)
      return ON
    elseif type(I) == EdgeCrossing
      ecross += 1
    elseif type(I) == CornerCrossing
      point = get(I)
      if !any(≈(point), points)
        push!(points, point)
        inters = !inters
      end
    end
  end

  # check how many edges we crossed
  isodd(ecross ÷ 2) && (inters = !inters)

  inters ? IN : OUT
end
```
The `_raycasting` function just finds the intersection as you loop. I don't see an advantage to materializing those intersetions beforehand. You can't make any more assumptions _about_ the intersections simply by materializing them. You still have to check their type and then do something accordingly.

Another route you could take is only check for `Crossing` types and you could get rid of the other branches, but you treat those degenerate cases by having to run the algorithm a second time for each query and perturb the ray by some number slightly bigger than the data type machine epsilon. If it gives the same result, you're good to go, if it gives a different result, run the full checks. I'm not sure if this would be faster thought because you have to compute the intersections for 2 rays now, although once you do, its just a sum of those intersections, versus having these branches.

Another really cool idea along the previous idea, although would require a ton of code, is to use autodiff to check the sensitivity of the output. If it's really high, you probably hit a degenerate case.

kylebeggs:
See this paper for the latter ideas - https://arxiv.org/abs/2212.08226

juliohm:
Superseeded by Jacbson et al 2013. on the master branch.

Thank you @kylebeggs for the references. Really good stuff.

kylebeggs:
Really nice to see this merged in. What are the speedups compared to the previous implementation? My guess is 20x?

juliohm:
Didn't benchmark it yet. I've implemented the method for meshes with
geodesic triangles (solid angle) by Jacobson, cited by Barill et all. To
implement the method for oriented point clouds we need to first approximate
the triangle areas at the points as they suggest. Something to consider in
the future after we have better integration with octrees and aabbs.

Em ter., 26 de dez. de 2023 12:49, Kyle Beggs ***@***.***>
escreveu:

kylebeggs:
Yeah, the best way of estimating the area represented by the points is ambiguous/tricky. I've given some thought about it, but it basically requires some form of meshing which then somewhat defeats the point of having a point cloud at all (might as well actually mesh it). I've started to take the route to parameterize the point cloud surfaces anyways which makes sense for my specific use cases (solving PDEs)

PR #654, opened 2023-12-18
Implement ParaboloidSurface

ziotom78:
As discussed in Zulip, this PR implements a paraboloidal surface. I chose the name `ParaboloidSurface` instead of `ParaboloidalSurface` in order to match `CylinderSurface`, which I used as a reference.

The paraboloid can be instantiated using this syntax:

```julia
ParaboloidSurface(point, radius, focallength)
```

(The parameter `radius` is used to limit the extents of the surface, otherwise it would extend to infinite in the three dimensions.) It is possible to display the mesh, compute its area, get a bounding box, sampling it and thus it can be visualized using `viz`. Here is an example, taken from the documentation (the paraboloid is shown in green, while the gray disk shows the extents of the rim):

image

There are still a couple of issues with names before the PR is ready:

-   As suggested in Zulip, I renamed the `vertex` function into `point`:
    ```julia
    par = ParaboloidSurface(Point3(1, 3, 2), 5.0, 3.0)

    # This is how it is implemented now, but I believe it is not
    # very readable… There are infinite “points” in a paraboloid
    p = point(par)
    ```

    I still think that `vertex` is less ambiguous:

    ```julia
    # It is obvious what this means
    p = vertex(par)
    ```

-   Similarly, it was suggested to shorten `focallength` to `length`. I did not implement this yet, because I would interpret the code

    ```julia
    l = length(par)
    ```

    as a measurement of some “size” of the parabola (e.g., the diameter of its rim, or its overall height along the z axis). But the focal length has nothing to do with the actual size of the parabola, which is only determined by its rim.

However, I have started using Meshes.jl since a few days and am still learning it, so it is very likely that I am missing some common use case here.

If the PR looks interesting to the community and there is interest in merging it, we should first answer these questions:

1.  Should I keep `point`, or is `vertex` better?
2.  Can `focallength` stay, or is it better to rename it `length`?


juliohm:
This is a very nice PR @ziotom78 , thanks for contributing the new primitive 🥇 

> 1. Should I keep `point`, or is `vertex` better?

You can use `vertex` internally in the field of the struct. We don't need an accessor function given that the surface is parametrized and surf(0, 0) is the vertex. We follow the same pattern in other primitives.

> 2. Can focallength stay, or is it better to rename it length?

Yes, agree that focallength is more explicit. Please go ahead with the name.

docs/src/geometries/primitives.md
```diff
@@ -153,3 +153,29 @@ Torus
...
+
+Surface of a paraboloid with the focal axis aligned along the ``z`` direction and clipped up to some radius on the ``xy`` direction.
+
+The equation of the paraboloid is the following:
+
+```math
+f(x, y) = \frac{(x - x_0)^2 + (y - y_0)^2}{4f} + z_0\qquad\text{for } x^2 + y^2 < r^2,
+```
+where ``(x_0, y_0, z_0)`` is the vertex of the parabola, ``f`` is the focal length, and ``r`` is the clip radius. See also <https://en.wikipedia.org/wiki/Paraboloid>.
+
```
juliohm: Let's try to keep the entire docstring in the source code so that we can easily update it in place in the future. No need to add this extra documentation in the docs file directly.
```suggestion
```

docs/src/geometries/primitives.md
```diff
@@ -153,3 +153,29 @@ Torus
...
+```@docs
+ParaboloidSurface
+```
+
+The following example shows a paraboloid together with a disk on the ``xy`` plane; the disk has the same radius used for the parabola. 
+
```
juliohm:
```suggestion
```

docs/src/geometries/primitives.md
```diff
@@ -153,3 +153,29 @@ Torus
...
+```@example primitives
+f = 0.25
+v = Point3(5, 2, 4)
+r = 1.0
```
juliohm:
```suggestion
v = Point3(5, 2, 4)
r = 1.0
f = 0.25
```

src/boundingboxes.jl
```diff
@@ -46,6 +46,15 @@ function boundingbox(s::Sphere{Dim,T}) where {Dim,T}
   Box(c - r⃗, c + r⃗)
 end
 
+function boundingbox(p::ParaboloidSurface{T}) where {T}
```
juliohm, line 49: Try to use the parameter T in the expressions. This will ensure that people can pass unitful coordinates in the future. Also, try to avoid accessing internal fields of the paralaboid with `p.field` syntax. Instead, use the public functions defined by the primitive.

src/primitives/paraboloidsurface.jl
```diff
@@ -0,0 +1,115 @@
...
+struct ParaboloidSurface{T} <: Primitive{3, T}
+    point::Point{3,T}
+    radius::T
+    focallength::T
+end
+
+function ParaboloidSurface(point::Tuple, radius, focallength)
+    pt = Point(point)
+    ParaboloidSurface(pt, convert(eltype(pt), radius), convert(eltype(pt), focallength))
+end
```
juliohm:
```suggestion
ParaboloidSurface(point::Tuple, radius, focallength) =
    ParaboloidSurface(Point(point), radius, focallength)
```

src/primitives/paraboloidsurface.jl
```diff
@@ -0,0 +1,115 @@
...
+
+ParaboloidSurface(point::Point{3,T}, radius) where {T} = ParaboloidSurface(point, T(radius), one(T))
```
juliohm:
```suggestion
ParaboloidSurface(point::Point{3,T}, radius) where {T} = ParaboloidSurface(point, T(radius), T(1))
```

src/primitives/paraboloidsurface.jl
```diff
@@ -0,0 +1,115 @@
...
+
+ParaboloidSurface(point::Point{3,T}) where {T} = ParaboloidSurface(point, T(1), T(1))
+
+ParaboloidSurface{T}() where {T} = ParaboloidSurface(Point{3, T}(0, 0, 0), T(1), T(1))
+
+ParaboloidSurface() = ParaboloidSurface{Float64}()
```
juliohm:
```suggestion
ParaboloidSurface() = ParaboloidSurface(Point(0, 0, 0), 1, 1)
```

src/primitives/paraboloidsurface.jl
```diff
@@ -0,0 +1,115 @@
...
+
+@doc raw"""
+    point(p::ParaboloidSurface{T})
+
+Return the vertex of the paraboloid as a `Point{3, T}`.
+"""
+point(p::ParaboloidSurface) = p.point
```
juliohm: No need for this one. We can use paraboloid(0, 0) to get this special point.

src/primitives/paraboloidsurface.jl
```diff
@@ -0,0 +1,115 @@
...
+
+@doc raw"""
```
juliohm: try to follow the docstring convention of other primitives, we don't use `@doc raw` at the moment.

src/primitives/paraboloidsurface.jl
```diff
@@ -0,0 +1,115 @@
...
+focallength(p::ParaboloidSurface) = p.focallength
+
+@doc raw"""
+    focallength(p::ParaboloidSurface)
+
+Return the radius of the rim of the paraboloid.
+"""
+radius(p::ParaboloidSurface) = p.radius
+
+paramdim(::Type{<:ParaboloidSurface}) = 2
```
juliohm: This should come after the constructors. It is a more fundamental function.

src/primitives/paraboloidsurface.jl
```diff
@@ -0,0 +1,115 @@
...
+
+function area(p::ParaboloidSurface{T}) where {T}
+    f = p.focallength
+    r = p.radius
+    8π/3 * f^2 * ((1 + r^2/(2f)^2)^(3/2) - 1)
+end
+
+measure(p::ParaboloidSurface) = area(p)
```
juliohm: These should be moved to the measure.jl source file along with other measure methods.

src/primitives/paraboloidsurface.jl
```diff
@@ -0,0 +1,115 @@
...
+
+isparametrized(::Type{<:ParaboloidSurface}) = true
+
+isperiodic(::Type{<:ParaboloidSurface}) = (false, true)
```
juliohm: These should be moved to after the `paramdim` method.

src/primitives/paraboloidsurface.jl
```diff
@@ -0,0 +1,115 @@
...
+    sinθ, cosθ = sincospi(T(2) * θ)
+    x = r * p.radius * cosθ
+    y = r * p.radius * sinθ
+    z = (x^2 + y^2)/4f
+    
+    Point(cx + x, cy + y, cz + z)
+end
+
+discretize(p::ParaboloidSurface) = discretize(p, RegularDiscretization(30))
+simplexify(p::ParaboloidSurface) = discretize(p, RegularDiscretization(30))
```
juliohm: These should be moved to discretization.jl along with the other methods. Also, I think we don't want to add a method to `simplefixy`, it has a fallback already that calls discretize.

juliohm (changes requested):
Very nice work. Just a few details before we can merge it.

src/primitives/paraboloidsurface.jl
ziotom78: Ok, I am fixing this. I always use `@doc raw` because it's nicer if you use LaTeX formulae (as I did in this PR). For the moment, I just doubled all the occurrences of `\`.

src/primitives/paraboloidsurface.jl
```diff
@@ -0,0 +1,101 @@
...
+Base.isapprox(p₁::ParaboloidSurface{T}, p₂::ParaboloidSurface{T}) where {T} =
+    (
+        p₁.vertex ≈ p₂.vertex &&
+        isapprox(p₁.focallength, p₂.focallength, atol=atol(T)) &&
+        isapprox(p₁.radius, p₂.radius, atol=atol(T))
+    )
+
+function (p::ParaboloidSurface{T})(r, θ) where {T}
+    # r is the radial coordinate, θ is the angular coordinate
+    (r < 0 || r > 1) && error("radus r=$r is out of [0, 1]")
```
kylebeggs: I think you should throw a `DomainError` here to be more explicit.

src/primitives/paraboloidsurface.jl
ziotom78: Mmm, with this change `ParaboloidSurface{Float32}()` does no longer work:

```julia
ParaboloidSurface: Error During Test at /home/tomasi/.julia/dev/Meshes/test/primitives.jl:1195
  Got exception outside of a @test
  MethodError: no method matching ParaboloidSurface{Float32}()
  
  Closest candidates are:
    (::Type{ParaboloidSurface{T}} where T)(::Any, ::Any, ::Any)
     @ Meshes ~/.julia/dev/Meshes/src/primitives/paraboloidsurface.jl:36
```

src/boundingboxes.jl
```diff
@@ -46,6 +46,15 @@ function boundingbox(s::Sphere{Dim,T}) where {Dim,T}
...
+    v = p.vertex
+    r = p.radius
+    f = p.focallength
```
juliohm: This implementation is still using internal fields of the struct. Try to use `v = p(0, 0), r = radius(p), f = focallength(p)` instead.

src/discretization.jl
```diff
@@ -119,6 +119,8 @@ discretize(torus::Torus) = discretize(torus, RegularDiscretization(50))
 
 discretize(cylsurf::CylinderSurface) = discretize(cylsurf, RegularDiscretization(50, 2))
 
+discretize(p::ParaboloidSurface) = discretize(p, RegularDiscretization(30))
```
juliohm:
```suggestion
discretize(p::ParaboloidSurface) = discretize(p, RegularDiscretization(50))
```
We use 50 by default in all geometries

src/measures.jl
```diff
@@ -57,6 +57,12 @@ function measure(c::CylinderSurface{T}) where {T}
   (norm(t(0, 0) - b(0, 0)) + r) * 2 * r * T(π)
 end
 
+function measure(p::ParaboloidSurface{T}) where {T}
+    f = p.focallength
+    r = p.radius
+    8π/3 * f^2 * ((1 + r^2/(2f)^2)^(3/2) - 1)
```
juliohm: needs to use T and avoid internal field access

src/primitives.jl
```diff
@@ -54,3 +54,4 @@ include("primitives/conesurface.jl")
 include("primitives/frustum.jl")
 include("primitives/frustumsurface.jl")
 include("primitives/torus.jl")
+include("primitives/paraboloidsurface.jl")
```
juliohm: let's include the file after the cylindersurface.jl, all the methods you added happened after cylindersurface

src/primitives/paraboloidsurface.jl
juliohm: Do we need this constructor with the explicit coordinate type? Our primitives don't have this special case. Whenever someone needs a special coordinate type, they can obtain it with the other constructors that rely on a specific point type.

src/primitives/paraboloidsurface.jl
```diff
@@ -0,0 +1,103 @@
...
+focallength(p::ParaboloidSurface) = p.focallength
+
+"""
+    focallength(p::ParaboloidSurface)
+
+Return the radius of the rim of the paraboloid.
+"""
+radius(p::ParaboloidSurface) = p.radius
+
+@doc raw"""
```
juliohm:
```suggestion
"""
```

src/boundingboxes.jl
ziotom78: Oh sorry, I missed that.

src/measures.jl
ziotom78: Gulp, this was not super-easy, hope I did it correctly…

juliohm:
A few inconsistencies that we need to address:

1. The docstring refers to apex, vertex, point, ... we need to pick one name and stick to it everywhere in the source. What do you think of apex? We could even add a `apex` method given that this function already exists in the code.
2. We need to make sure that we are adding methods in the right section of the code. Across all files, make sure that the methods for ParalaboloidSurface are added right after CylinderSurface. I know this is a bit pedantic, but it helps a lot in the maintenance of such large codebase.
3. All tests of functions like discretize, boundingbox, etc should be added in the corresponding test file.

src/Meshes.jl
```diff
@@ -151,6 +151,7 @@ export
   Frustum,
   FrustumSurface,
   Torus,
+  ParaboloidSurface,
```
juliohm: should come after cylindersurface if the source file is included after cylindersurface.jl

src/boundingboxes.jl
```diff
@@ -46,6 +46,15 @@ function boundingbox(s::Sphere{Dim,T}) where {Dim,T}
   Box(c - r⃗, c + r⃗)
 end
 
+function boundingbox(p::ParaboloidSurface{T}) where {T}
+    v = p(T(0), T(0))
+    r = radius(p)
+    f = focallength(p)
+
+    _, cy, _ = coordinates(v)
+    Box(v + Vec(-r, -r, T(0)), v + Vec(r, r, T(r^2 / (4f))))
```
```suggestion
    Box(v + Vec(-r, -r, T(0)), v + Vec(r, r, (r^2 / (4f)))
```

src/measures.jl
```diff
@@ -57,6 +57,12 @@ function measure(c::CylinderSurface{T}) where {T}
   (norm(t(0, 0) - b(0, 0)) + r) * 2 * r * T(π)
 end
 
+function measure(p::ParaboloidSurface{T}) where {T}
+    f = focallength(p)
+    r = radius(p)
+    T(8π/3) * f^2 * ((T(1) + r^2/(2f)^2)^T(3/2) - T(1))
```
juliohm:
```suggestion
    T(8π/3) * f^2 * ((T(1) + r^2/(2f)^2)^(3/2) - T(1))
```

src/primitives/paraboloidsurface.jl
```diff
@@ -0,0 +1,103 @@
...
+The axis is always aligned with the z direction.
+"""
+axis(p::ParaboloidSurface) = Line(p.vertex, p.vertex + Vec(0, 0, p.focallength))
+
+Base.isapprox(p₁::ParaboloidSurface{T}, p₂::ParaboloidSurface{T}) where {T} =
+    (
+        p₁.vertex ≈ p₂.vertex &&
+        isapprox(p₁.focallength, p₂.focallength, atol=atol(T)) &&
+        isapprox(p₁.radius, p₂.radius, atol=atol(T))
+    )
```
juliohm:
```suggestion
Base.isapprox(p₁::ParaboloidSurface{T}, p₂::ParaboloidSurface{T}) where {T} =
  p₁.vertex ≈ p₂.vertex &&
  isapprox(p₁.focallength, p₂.focallength, atol=atol(T)) &&
  isapprox(p₁.radius, p₂.radius, atol=atol(T))
```

src/primitives/paraboloidsurface.jl
```diff
@@ -0,0 +1,103 @@
...
+
+function (p::ParaboloidSurface{T})(r, θ) where {T}
+    # r is the radial coordinate, θ is the angular coordinate
+    (r < 0 || r > 1) && throw(DomainError((r, θ), "radius r=$r is out of [0, 1]"))
```
juliohm:
```suggestion
  if (r < 0 || r > 1)
    throw(DomainError((r, θ), "radius r=$r is out of [0, 1]"))
  end
```

juliohm (changes requested):
Also consider the review comments above.

src/boundingboxes.jl
ziotom78: I'm a bit puzzled here… Since `r` and `f` are already of type `T`, isn't `r^2 / 4r` automatically of type `T`?

```julia
julia> r = Float32(1.0)
1.0f0

julia> f = Float32(4.0)
4.0f0

julia> fn(r, f) = r^2 / (4f)
fn (generic function with 1 method)

julia> fn(r, f)
0.0625f0

julia> typeof(fn(r, f))
Float32

julia> @code_typed fn(r, f)
CodeInfo(
1 ─ %1 = Base.mul_float(r, r)::Float32
│   %2 = Base.mul_float(4.0f0, f)::Float32
│   %3 = Base.div_float(%1, %2)::Float32
└──      return %3
) => Float32
```

Or do you encourage types to be explicit in any case in the Meshes.jl codebase?

src/Meshes.jl
ziotom78: Done.

src/primitives/paraboloidsurface.jl
ziotom78: I removed the failing test and reverted your change.

ziotom78:
>The docstring refers to apex, vertex, point, ... we need to pick one name and stick to it everywhere in the source. What do you think of apex? We could even add a `apex` method given that this function already exists in the code.

I like it! I added `apex` and changed all the occurrences. I also fixed the docstrings so that now only the term «apex» is used.

>We need to make sure that we are adding methods in the right section of the code. Across all files, make sure that the methods for ParalaboloidSurface are added right after CylinderSurface. I know this is a bit pedantic, but it helps a lot in the maintenance of such large codebase.

Done

>All tests of functions like discretize, boundingbox, etc should be added in the corresponding test file.

Done.

The last thing that is left is about the use of `T()` in a context where I was assuming it was superfluous.

src/boundingboxes.jl
juliohm: Exactly, if the inputs are of type T then the expression will be of type T if the literals are integer. We are assuming that T <: Real or a unitful type on top of Real.

src/primitives/paraboloidsurface.jl
```diff
@@ -0,0 +1,108 @@
...
+"""
+apex(p::ParaboloidSurface) = p.apex
+
+"""
+    axis(p::ParaboloidSurface)
+
+Return the focal axis, connecting the focus with the apex of the paraboloid.
+The axis is always aligned with the z direction.
+"""
+axis(p::ParaboloidSurface) = Line(p.apex, p.apex + Vec(0, 0, p.focallength))
```
juliohm:
```suggestion
axis(p::ParaboloidSurface{T}) where {T} = Line(p.apex, p.apex + Vec(T(0), T(0), p.focallength))
```

src/measures.jl
```diff
@@ -57,6 +57,12 @@ function measure(c::CylinderSurface{T}) where {T}
   (norm(t(0, 0) - b(0, 0)) + r) * 2 * r * T(π)
 end
 
+function measure(p::ParaboloidSurface{T}) where {T}
+    f = focallength(p)
+    r = radius(p)
```
juliohm:
```suggestion
    r = radius(p)
    f = focallength(p)
```

src/boundingboxes.jl
```diff
@@ -46,6 +46,15 @@ function boundingbox(s::Sphere{Dim,T}) where {Dim,T}
   Box(c - r⃗, c + r⃗)
 end
 
+function boundingbox(p::ParaboloidSurface{T}) where {T}
+    v = apex(p)
+    r = radius(p)
+    f = focallength(p)
+
+    _, cy, _ = coordinates(v)
```
```suggestion
```

src/primitives/paraboloidsurface.jl
```diff
@@ -0,0 +1,108 @@
...
+end
+
+ParaboloidSurface(apex::Tuple, radius, focallength) =
+    ParaboloidSurface(Point(apex), radius, focallength)
+
+ParaboloidSurface(apex::Point{3,T}, radius) where {T} = ParaboloidSurface(apex, T(radius), T(1))
+
+ParaboloidSurface(apex::Point{3,T}) where {T} = ParaboloidSurface(apex, T(1), T(1))
+
+ParaboloidSurface() = ParaboloidSurface(Point(0, 0, 0), 1.0, 1.0)
```
juliohm:
```suggestion
ParaboloidSurface() = ParaboloidSurface(Point(0, 0, 0))
```

src/primitives/paraboloidsurface.jl
```diff
@@ -0,0 +1,108 @@
...
+end
+
+ParaboloidSurface(apex::Tuple, radius, focallength) =
+    ParaboloidSurface(Point(apex), radius, focallength)
+
+ParaboloidSurface(apex::Point{3,T}, radius) where {T} = ParaboloidSurface(apex, T(radius), T(1))
+
+ParaboloidSurface(apex::Point{3,T}) where {T} = ParaboloidSurface(apex, T(1), T(1))
+
+ParaboloidSurface() = ParaboloidSurface(Point(0, 0, 0))
```
juliohm: Please follow the same pattern of the constructors in CylinderSurface, both with Tuple and Point

src/boundingboxes.jl
```diff
@@ -46,6 +46,13 @@ function boundingbox(s::Sphere{Dim,T}) where {Dim,T}
   Box(c - r⃗, c + r⃗)
 end
 
+function boundingbox(p::ParaboloidSurface{T}) where {T}
+    v = apex(p)
+    r = radius(p)
+    f = focallength(p)
+    Box(v + Vec(-r, -r, T(0)), v + Vec(r, r, (r^2 / (4f)))
```
```suggestion
    Box(v + Vec(-r, -r, T(0)), v + Vec(r, r, r^2 / (4f)))
```

juliohm:
Thank you for addressing the issues. Left two final review comments.

src/primitives/paraboloidsurface.jl
ziotom78: Done

juliohm (approved):

src/primitives/paraboloidsurface.jl
```diff
@@ -0,0 +1,112 @@
...
+Same as above, but here the apex is at `Apex(0, 0, 0)`.
+
+See also <https://en.wikipedia.org/wiki/Paraboloid>.
+"""
+struct ParaboloidSurface{T} <: Primitive{3, T}
+    apex::Point{3,T}
+    radius::T
+    focallength::T
+end
+
```
juliohm, line 40:
```suggestion

ParaboloidSurface(apex::Point{3,T}, radius, focallength) where {T} = ParaboloidSurface{T}(apex, radius, focallength)

```

juliohm:
Thank you @ziotom78 ! That is a very nice first contribution! Thanks for putting the time to fix the small details, they help a lot in the long run 💯

PR #661, opened 2023-12-20
Remove `cart2vert` in favor of `vertex`

eliascarv:


src/mesh.jl
```diff
@@ -173,21 +173,23 @@ function XYZ end
 
 Base.size(g::Grid) = size(topology(g))
 
-Base.minimum(g::Grid{Dim}) where {Dim} = cart2vert(g, ntuple(i -> 1, Dim))
-Base.maximum(g::Grid{Dim}) where {Dim} = cart2vert(g, size(g) .+ 1)
+Base.minimum(g::Grid{Dim}) where {Dim} = vertex(g, ntuple(i -> 1, Dim))
+Base.maximum(g::Grid{Dim}) where {Dim} = vertex(g, size(g) .+ 1)
 Base.extrema(g::Grid{Dim}) where {Dim} = minimum(g), maximum(g)
 
+vertex(grid::Grid, ijk::Tuple) = vertex(grid,  LinearIndices(size(grid) .+ 1)[ijk...])
```
eliascarv: Maybe we can use `Dims{Dim}` to be safer?
```suggestion
vertex(grid::Grid{Dim}, ijk::Dims{Dim}) where {Dim} = vertex(grid,  LinearIndices(size(grid) .+ 1)[ijk...])
```

juliohm: Agree, and this fallback should show up right after the first appearance of the function

src/mesh/cartesiangrid.jl
```diff
@@ -108,7 +108,7 @@ CartesianGrid(dims::Dims{Dim}) where {Dim} = CartesianGrid{Float64}(dims)
 
 CartesianGrid(dims::Vararg{Int,Dim}) where {Dim} = CartesianGrid{Float64}(dims)
 
-cart2vert(g::CartesianGrid, ijk::Tuple) = Point(coordinates(g.origin) .+ (ijk .- g.offset) .* g.spacing)
+vertex(g::CartesianGrid, ijk::Tuple) = Point(coordinates(g.origin) .+ (ijk .- g.offset) .* g.spacing)
```
juliohm: here we can also use Dims

juliohm (approved):

PR #662, opened 2023-12-20
Change required methods to implement `Mesh` interface

eliascarv:
Now, the required method to implement `Mesh` interface is `vertex`, instead of `vertices`.


eliascarv:
Do we still need this method? I do not think so.
Captura de tela de 2023-12-20 14-06-48

juliohm (approved):

PR #669, opened 2023-12-26
RectilinearGrid: Implement multidimensional indexing

eliascarv:


src/mesh/rectilineargrid.jl
```diff
@@ -56,3 +56,11 @@ function centroid(g::RectilinearGrid, ind::Int)
   p2 = vertex(g, ijk .+ 1)
   Point((coordinates(p1) + coordinates(p2)) / 2)
 end
+
+function Base.getindex(g::RectilinearGrid{Dim}, I::CartesianIndices{Dim}) where {Dim}
+  dims = size(I)
+  start = Tuple(first(I))
+  stop = Tuple(last(I)) .+ 1
+  xyz = ntuple(i -> g.xyz[i][start[i]:stop[i]], Dim)
```
eliascarv, line 64: Do we need to use `view(xyz[i], range)`?

PR #685, opened 2024-01-12
Add `Affine` transform

eliascarv:


juliohm (changes requested):
Let's rename Affine --> RotateTranslate because there are other types of affine transforms (e.g. Strecth)

juliohm:
Perhaps we should simply rename it to `MulAdd`, and allow any matrix A in Ax + b instead of just the rotation R.

juliohm:
Thinking about this again, the name `Affine` is fine, we just need to replace the rotation matrix by a generic matrix A. The `Affine(A, b)` will have this single constructor.

src/transforms/affine.jl
```diff
@@ -0,0 +1,60 @@
+# ------------------------------------------------------------------
+# Licensed under the MIT License. See LICENSE in the project root.
+# ------------------------------------------------------------------
+
+"""
+    Affine(A, b)
+
+Rotate geometry or mesh with rotation matrix `A` and translate
+their coordinates with translation vector `b`.
```
juliohm:
```suggestion
Affine transform `Ax + b` with matrix `A` and vector `b`.
```

src/transforms/affine.jl
```diff
@@ -0,0 +1,60 @@
...
+  Dim = first(sz)
+  if Dim ≠ length(b)
+    throw(ArgumentError("`A` and `b` must have the same dimension"))
+  end
+  Affine(_assmatrix(Dim, A), _assvector(Dim, b))
+end
+
+isrevertible(::Type{<:Affine}) = false
+
+isinvertible(::Type{<:Affine}) = false
```
juliohm: The transform is invertible if the matrix A is invertible.

src/transforms/affine.jl
```diff
@@ -0,0 +1,60 @@
...
+    throw(ArgumentError("`A` must be a square matrix"))
+  end
+  Dim = first(sz)
+  if Dim ≠ length(b)
+    throw(ArgumentError("`A` and `b` must have the same dimension"))
+  end
+  Affine(_assmatrix(Dim, A), _assvector(Dim, b))
+end
+
+isrevertible(::Type{<:Affine}) = false
```
juliohm: It is always revertible if you use the pseudo inverse of A. Given `y = Ax + b`, the revert expression is `x = A \ (y - b)`

PR #689, opened 2024-01-17
Add recipe for `Grid`

eliascarv:


ext/fallbacks.jl
```diff
@@ -11,10 +11,9 @@ Makie.convert_arguments(::Type{<:Viz}, mesh::Mesh) = (convert(SimpleMesh, mesh),
 
 # skip conversion for these types
 Makie.convert_arguments(::Type{<:Viz}, gset::GeometrySet) = (gset,)
+Makie.convert_arguments(::Type{<:Viz}, grid::Grid) = (grid,)
 Makie.convert_arguments(::Type{<:Viz}, mesh::SimpleMesh) = (mesh,)
```
juliohm:
```suggestion
Makie.convert_arguments(::Type{<:Viz}, mesh::SimpleMesh) = (mesh,)
Makie.convert_arguments(::Type{<:Viz}, grid::Grid) = (grid,)
```

ext/grid.jl
```diff
@@ -0,0 +1,158 @@
...
+  elseif Dim == 3
+    vizgrid3D!(T, plot, colorant, nv, nc)
+  end
+end
+
+vizgrid1D!(::Type{<:Grid}, plot, colorant, nv, nc) = vizmesh1D!(plot)
+
+vizgrid2D!(::Type{<:Grid}, plot, colorant, nv, nc) = vizmesh2D!(plot)
+
+function vizgrid2D!(::Type{<:Union{CartesianGrid,RectilinearGrid}}, plot, colorant, nv, nc)
```
juliohm: Can we refactor all the vizgrid methods to accept only the grid type and the plot? I think it would be more useful to keep the colorant and nv and nc inside as an implementation detail. That way we can reuse these methods in other places if necessary.

juliohm (changes requested):

PR #692, opened 2024-01-19
type-paramterize Segment eval

kylebeggs:
This breaks if you have the type of the Segment (points) be something that doesn't work with comparing to Int literals, such as Unitful.jl. By using `zero(T)` and `one(T)`, this is fixed.

I wasn't sure how to go about testing this, if you could provide advice here? Obvious way is to load Unitful, but I think you don't want to add that as a test dependency. I can't think of another type to use without adding a dependency. Maybe we can make some dummy type?

### DIFF
--- a/src/polytopes/segment.jl
+++ b/src/polytopes/segment.jl
@@ -30,7 +30,7 @@ end
 Base.isapprox(s₁::Segment, s₂::Segment; kwargs...) =
   all(isapprox(v₁, v₂; kwargs...) for (v₁, v₂) in zip(s₁.vertices, s₂.vertices))
 
-function (s::Segment{Dim,T})(t) where {Dim,T}
+function (s::Segment)(t)
   if t < 0 || t > 1
     throw(DomainError(t, "s(t) is not defined for t outside [0, 1]."))
   end
--- a/src/polytopes/segment.jl
+++ b/src/polytopes/segment.jl
@@ -26,8 +26,8 @@ center(s::Segment{Dim,T}) where {Dim,T} = s(T(0.5))
 Base.isapprox(s₁::Segment, s₂::Segment; kwargs...) =
   all(isapprox(v₁, v₂; kwargs...) for (v₁, v₂) in zip(s₁.vertices, s₂.vertices))
 
-function (s::Segment)(t)
-  if t < 0 || t > 1
+function (s::Segment{Dim,T})(t) where {Dim,T}
+  if t < zero(T) || t > one(T)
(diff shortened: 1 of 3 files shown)


juliohm:
Thank you @kylebeggs ! That sounds great. Feel free to add Unitful.jl as a test dependency, that is no problem.

We will add Unitful.jl/DynamicQuantities.jl soon to the stack.

kylebeggs:
This is trickier than I originally thought. I have a real love/hate relationship with Unitful.jl. I think here `t` should always be a `Real`, right? In that case, what we really need is to convert it to the *precision* we want, which isn't simply using `T` from `Segment`. To do this, I implemented some `get_precision` methods that work with all the Julia Base types and Unitful.jl types as well. It's just one idea, I think there are a few ways to go about this.

kylebeggs:
Or maybe the real solution is making sure `T` in `Segment`, `Point`, and basically everything else is specifying the precision (always `Float64` or `Float32`) rather than the actual type that it contains?

juliohm:
Looking again into this, isn`t `t` a unitless parameter that lives in [0, 1]? The reult of the segment(t) is a point with unitful coordinates though.

kylebeggs:
Yes, that is right. The question is if the `Segment` is made of  `Float32`s and the user inputs `t::Float64` then you would get `Point{Dim,Float64}`. The previous implementation for `center(::Segment)` handled this correctly by converting `t` to match `Segment`, but that way of converting d0oesn't work when you have Unitful.jl coordinates. I essentially generalized that conversion to work with Unitful.jl coords too.

juliohm:
Suppose that `a::Float32` and `b::Float32` are the end points of the `Segment`. Are you proposing that we should fight the Julia's type promotion rules in the expression `a + t*(b-a)` and return `Float32` even when `t::Float64`?

I think we should aim for a design that works naturally with Julia's built-in number rules. If a user is really interested in preserving the `Float32` precision, then he/she should use `t::Float32`. Do you agree?

kylebeggs:
Yes, I agree with that. I got thrown off by the implementation for `center` thinking that was what you were trying to achieve, but after a second look I see since no `t` is supplied, it makes sense to return the same type of `Segment`. It is `center` that was really giving the issue with Unitful.jl, so I just added a method for that.

juliohm:
Yes, we should probably rewrite the `center` definition for `Segment` to remove that `T` from there too, and use a definition in terms of the end points `a` and `b`. The idea in general is that the parameter `t` or `u,v` or `u,v,w` used in parametrizations is unitless.

I will close this PR for now. We can improve the situation in another fresh PR.

PR #700, opened 2024-01-22
Use the `Makie.image` function to plot the `CartesianGrid`

eliascarv:
GeoStatsVizTests.jl testset:
```
     Testing Running tests...
Test Summary:       | Pass  Total     Time
GeoStatsVizTests.jl |  221    221  1m55.9s
     Testing GeoStatsVizTests tests passed
 ```

### DIFF
--- a/ext/grid.jl
+++ b/ext/grid.jl
@@ -4,22 +4,87 @@
 
 function Makie.plot!(plot::Viz{<:Tuple{Grid}})
   grid = plot[:object][]
-  T = typeof(grid)
   Dim = embeddim(grid)
   if Dim == 1
-    vizgrid1D!(T, plot)
+    vizgrid1D!(plot)
   elseif Dim == 2
-    vizgrid2D!(T, plot)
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: ext/grid.jl [hunk 1]
```diff
@@ -4,22 +4,87 @@
...
+
+    if showfacets[]
+      tup = Makie.@lift xysegments(0:$sz[1], 0:$sz[2])
+      x, y = Makie.@lift($tup[1]), Makie.@lift($tup[2])
+      Makie.lines!(plot, x, y, color=facetcolor, linewidth=segmentsize)
+    end
+
+    # adjust spacing and origin
+    spx, spy = sp[]
+    orx, ory = or[]
```
juliohm, line 72: do we need to unpack the observables here? Can't we just forward them to Makie.scale! and makie.translate!?

eliascarv: Unfortunately no, because `Makie.scale!` and `Makie.translate!` don't accept observables.

PR #715, opened 2024-01-28
Clamping points to boxes

### LINKED ISSUE #698: Clamping points to boxes and/or bounding boxes
As part of a simulation project I'm working with points, grids, and boxes in two dimensions. At different points I need to clamp point locations that lie outside of a box by moving them to the box boundary. This is straightforward with the `clamp` and `clamp!` functions from `Base.Math`. You just have to clamp both dimensions of the point, or however many dimensions you have.

I see the clipping functionality but no clamping. New methods could look something like

```julia
clamp(Point{Dim,T}, Box{Dim,T}) where {Dim,T}

clamp(PointSet{Dim,T}, Box{Dim,T}} where {Dim,T}
```

Happy to contribute if it makes sense.

markmbaum:
This small PR adds methods to the built-in [`clamp`](https://docs.julialang.org/en/v1/base/math/#Base.Math.clamp) function.
* Two new methods clamp points and pointsets to boxes.
* The methods are tested and documented.
  * I put documentation under Algorithms/Clamping and built the docs locally to make sure the new page renders correctly.
  * If the page should go somewhere else, or if a new page for Clamping is unnecessary, just let me know. It could also go under Transforms, I suppose.
* The package version is also incremented from 0.40.0 to 0.41.0. 

The new methods are very simple, but further updates could potentially
* Allow for clamping points and boxes with different numbers of dimensions. It would only operate on the shared dimensions.
* Allow for clamping points to shapes other than boxes, which is more involved.

closes #698


src/clamping.jl
```diff
@@ -0,0 +1,17 @@
+"""
+    clamp(point, box)
+
+Clamp the coordinates of a [`Point`](@ref) to the edges of a [`Box`](@ref). For each dimension, coordinates outside of the box are moved to the nearest edge of the box. The point and box must have an equal number of dimensions."""
+function Base.clamp(point::Point{Dim,T}, box::Box{Dim,T})::Point{Dim,T} where {Dim,T}
+  ntuple(Dim) do i
+    Base.clamp(point.coords[i], box.min.coords[i], box.max.coords[i])
```
juliohm: We try to not use internal fields in implementations to facilitate the maintenance and future refactoring of the code. Can you please replace `point.coords` by `coordinates(point)` and `box.min` by `minimum(box)` etc?

Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "Meshes"
 uuid = "eacbb407-ea5a-433e-ab97-5258b1ca43fa"
 authors = ["Júlio Hoffimann and contributors"]
-version = "0.40.0"
+version = "0.41.0"
```
juliohm:
```suggestion
version = "0.40.1"
```
We are currently using minor releases for breaking changes (we are still v0.y.z)

src/clamping.jl
markmbaum: Yes no problem

markmbaum:
Made the suggested changes and included a new test for `PointSet` clamping so that coverage will not go down.

juliohm (approved):

juliohm:
Thank you @markmbaum ❤️

PR #717, opened 2024-01-29
Coordinate types

eliascarv:


src/coordinates/basic.jl
```diff
@@ -0,0 +1,64 @@
+# ------------------------------------------------------------------
+# Licensed under the MIT License. See LICENSE in the project root.
+# ------------------------------------------------------------------
+
+"""
+    Cartesian(x₁, x₂, ..., xₙ)
+
+N-Dimensional Cartesian coordinate with coordinates `x₁, x₂, ..., xₙ`.
```
juliohm:
```suggestion
N-dimensional Cartesian coordinates `x₁, x₂, ..., xₙ`.
```

src/coordinates/basic.jl
```diff
@@ -0,0 +1,64 @@
...
+
+Cartesian(coords::NTuple{N,T}) where {N,T} = Coordinates{N,T}(coords)
+Cartesian(coords::Tuple) = Coordinates(promote(coords...))
+Cartesian(coords...) = Cartesian(coords)
+
+"""
+    Polar(ρ, ϕ)
+
+Polar coordinate with radial distance `ρ`
+and angular coordinate `ϕ` (in radians).
```
juliohm:
```suggestion
Polar coordinates with radius `ρ` and angle `ϕ`.
```

src/coordinates/basic.jl
```diff
@@ -0,0 +1,64 @@
...
+end
+
+Polar(ρ::T, ϕ::T) where {T} = Polar{T}(ρ, ϕ)
+Polar(ρ, ϕ) = Polar(promote(ρ, ϕ)...)
+
+"""
+    Cylindrical(ρ, ϕ, z)
+
+Cylindrical coordinate with radial distance `ρ`,
+angular coordinate `ϕ` (in radians) and height `z`.
```
juliohm:
```suggestion
Cylindrical coordinates with radius `ρ`, angle `ϕ` and height `z`.
```

src/coordinates/basic.jl
```diff
@@ -0,0 +1,64 @@
...
+end
+
+Cylindrical(ρ::T, ϕ::T, z::T) where {T} = Cylindrical{T}(ρ, ϕ, z)
+Cylindrical(ρ, ϕ, z) = Cylindrical(promote(ρ, ϕ, z)...)
+
+"""
+    Spherical(r, θ, ϕ)
+
+Spherical coordinate with radial distance `r`,
+polar angle `θ` (in radians) and azimuthal angle `ϕ` (in radians).
```
juliohm:
```suggestion
Spherical coordinates with radius `r`, polar angle `θ` and azimuth angle `ϕ`.
```

src/coordinates/gis.jl
```diff
@@ -0,0 +1,93 @@
+# ------------------------------------------------------------------
+# Licensed under the MIT License. See LICENSE in the project root.
+# ------------------------------------------------------------------
+
+"""
+    LatLon(lat, lon)
+
+Geographic coordinate with latitude `lat` (in degrees)
+and longitude `lon` (in degrees).
```
juliohm:
```suggestion
Latitude and longitude in degrees.
```

src/coordinates/gis.jl
```diff
@@ -0,0 +1,93 @@
...
+LatLon(lat::T, lon::T) where {T<:Quantity} = LatLon{T}(lat, lon)
+LatLon(lat::Quantity, lon::Quantity) = LatLon(promote(lat, lon)...)
+LatLon(lat::Number, lon::Number) = LatLon(lat * u"°", lon * u"°")
+
+"""
+    LatLonAlt(lat, lon, alt)
+
+Geographic coordinate with latitude `lat` (in degrees),
+longitude `lon` (in degrees) and altitude `alt`.
+The altitude coordinate unit must be a length unit, which by default is meter.
```
juliohm:
```suggestion
Latitude and longitude in degrees and altitude in length units (default to meter).
```

src/coordinates/gis.jl
```diff
@@ -0,0 +1,93 @@
...
+LatLonAlt(lat::T, lon::T, alt::A) where {T<:Quantity,A<:Quantity} = LatLonAlt{T,A}(lat, lon, alt)
+LatLonAlt(lat::Quantity, lon::Quantity, alt::Quantity) = LatLonAlt(promote(lat, lon)..., alt)
+LatLonAlt(lat::Number, lon::Number, alt::Number) = LatLonAlt(lat * u"°", lon * u"°", alt * u"m")
+
+"""
+    EastNorth(east, north)
+
+Geographic coordinate with distance `east` to the prime meridian
+and distance `north` to the equator.
+The coordinate units must be length units, which by default is meter.
```
juliohm:
```suggestion
East and North in length units (default to meter).
```

src/coordinates/gis.jl
```diff
@@ -0,0 +1,93 @@
...
+
+EastNorth(east::T, north::T) where {T<:Quantity} = EastNorth{T}(east, north)
+EastNorth(east::Quantity, north::Quantity) = EastNorth(promote(east, north)...)
+EastNorth(east::Number, north::Number) = EastNorth(east * u"m", north * u"m")
+
+"""
+    WebMercator(x, y)
+
+Web Mercator coordinate with coordinates `x` and `y`.
+The coordinate units must be length units, which by default is meter.
```
juliohm:
```suggestion
WebMercator coordinates in length units (default to meter).
```

src/coordinates/gis.jl
```diff
@@ -0,0 +1,87 @@
...
+end
+
+LatLonAlt(lat::T, lon::T, alt::A) where {T<:Quantity,A<:Quantity} = LatLonAlt{T,A}(lat, lon, alt)
+LatLonAlt(lat::Quantity, lon::Quantity, alt::Quantity) = LatLonAlt(promote(lat, lon)..., alt)
+LatLonAlt(lat::Number, lon::Number, alt::Number) = LatLonAlt(lat * u"°", lon * u"°", alt * u"m")
+
+"""
+    EastNorth(east, north)
+
+East and North in length units (default to meter).
```
juliohm:
```suggestion
East and north coordinates in length units (default to meter).
```

src/coordinates/basic.jl
```diff
@@ -0,0 +1,98 @@
...
+"""
+    Cylindrical(ρ, ϕ, z)
+
+Cylindrical coordinates with radius `ρ`, angle `ϕ` and height `z`.
+
+## Unique coordinates
+
+* radius: `ρ ∈ [0,∞)`
+* angle: `ϕ ∈ [0,2π)`
+* height: `z ∈ [0,∞)`
```
juliohm:
```suggestion
Cylindrical coordinates with radius `ρ ∈ [0,∞)`, angle `ϕ ∈ [0,2π)` and height `z ∈ [0,∞)`.
```

src/coordinates/basic.jl
```diff
@@ -0,0 +1,98 @@
...
+"""
+    Spherical(r, θ, ϕ)
+
+Spherical coordinates with radius `r`, polar angle `θ` and azimuth angle `ϕ`.
+
+## Unique coordinates
+
+* radius: `r ∈ [0,∞)`
+* polar angle: `θ ∈ [0,π]`
+* azimuth angle: `ϕ ∈ [0,2π)`
```
juliohm:
```suggestion
Spherical coordinates with radius `r ∈ [0,∞)`, polar angle `θ ∈ [0,π]` and azimuth angle `ϕ ∈ [0,2π)`.
```

src/coordinates/basic.jl
```diff
@@ -0,0 +1,98 @@
...
+
+"""
+    Polar(ρ, ϕ)
+
+Polar coordinates with radius `ρ` and angle `ϕ`.
+
+## Unique coordinates
+
+* radius: `ρ ∈ [0,∞)`
+* angle: `ϕ ∈ [0,2π)`
```
eliascarv:
```suggestion
Polar coordinates with radius `ρ ∈ [0,∞)` and angle `ϕ ∈ [0,2π)`.
```

src/coordinates.jl
```diff
@@ -0,0 +1,59 @@
...
+
+# ------------
+# CONVERSIONS
+# ------------
+
+# Cartesian <--> Polar
+Base.convert(::Type{<:Cartesian}, (; ρ, ϕ)::Polar) = Cartesian(ρ * cos(ϕ), ρ * sin(ϕ))
+function Base.convert(::Type{<:Polar}, (; coords)::Cartesian{2})
+  x, y = coords
+  Polar(sqrt(x^2 + y^2), atan(y / x))
```
juliohm: please use the atan method with 2 arguments

src/coordinates.jl
```diff
@@ -0,0 +1,59 @@
...
+end
+
+# Cartesian <--> Cylindrical
+Base.convert(::Type{<:Cartesian}, (; ρ, ϕ, z)::Cylindrical) = Cartesian(ρ * cos(ϕ), ρ * sin(ϕ), z)
+function Base.convert(::Type{<:Cylindrical}, (; coords)::Cartesian{3})
+  x, y, z = coords
+  Cylindrical(sqrt(x^2 + y^2), atan(y / x), z)
```
juliohm: same here

src/coordinates.jl
```diff
@@ -0,0 +1,59 @@
...
+end
+
+# Cartesian <--> Spherical
+Base.convert(::Type{<:Cartesian}, (; r, θ, ϕ)::Spherical) = Cartesian(r * sin(θ) * cos(ϕ), r * sin(θ) * sin(ϕ), r * cos(θ))
+function Base.convert(::Type{<:Spherical}, (; coords)::Cartesian{3})
+  x, y, z = coords
+  Spherical(sqrt(x^2 + y^2 + z^2), atan(sqrt(x^2 + y^2) / z), atan(y / x))
```
juliohm: same here

src/Meshes.jl
```diff
@@ -123,6 +126,16 @@ export
   ⋅,
   ×,
 
+  # coordinates
```
juliohm, line 129:
```suggestion
  # coordinates
  Coordinates,
```

mikeingold:
I just noticed this PR. Nice work! I made a similar implementation a while back (https://github.com/mikeingold/UnitfulCoordinateSystems.jl) and this got me thinking back through some lessons learned along the way. Feel free to plunder any code from that package as you see fit; my goal is to deprecate it in favor of the coordinate implementation here.
- I originally used the same convention of using Unicode symbols for struct fields but found that some people have strong opinions about being required to use Unicode symbols. I don't know if it was the best solution, but I decided later to stick to ASCII-based symbols by default and provide Unicode symbols as aliases for convenience. Within the package code itself I used a lot of Unicode for brevity/readability, but end-users can stick to whichever they're most comfortable with.
- Following common Julia conventions, the abstract type should probably be defined as `AbstractCoordinate{N}`.
- For convenience, I implemented a collection of constructors to perform conversions between the coordinate types, e.g. things like `Spherical(::Cylindrical)`. Something like this, or maybe `convert` methods, are handy when needing to perform coordinate transformations. I implemented conversion methods for all 3D-to-3D cases, and for 2D-to-3D cases where the source `z` is assumed to be zero.
- For the geographic types, do these only work with Quantity from Unitful.jl or have you also tested it with the ones from DynamicQuantities.jl? I don't have a strong argument that it should/must, but I try to write code generic enough that works for both implementations where possible.

juliohm:
> * I originally used the same convention of using Unicode symbols for struct fields but found that some people have strong opinions about being required to use Unicode symbols. I don't know if it was the best solution, but I decided later to stick to ASCII-based symbols by default and provide Unicode symbols as aliases for convenience. Within the package code itself I used a lot of Unicode for brevity/readability, but end-users can stick to whichever they're most comfortable with.

We are following the symbols and units in the ISO. The docstrings will be updated soon.

> * Following common Julia conventions, the abstract type should probably be defined as `AbstractCoordinate{N}`.

We removed the `Abstract` prefix in our stack, it usually just leads to more typing. We are familiar with the abstract types as developers, and users can benefit from a more succinct name in IO methods sometimes.

> * For convenience, I implemented a collection of constructors to perform conversions between the coordinate types, e.g. things like `Spherical(::Cylindrical)`. Something like this, or maybe `convert` methods, are handy when needing to perform coordinate transformations. I implemented conversion methods for all 3D-to-3D cases, and for 2D-to-3D cases where the source `z` is assumed to be zero.

That is something that we are trying to decide yet, if we should support the conversion in the constructors. The main drawback is that we would need to add constructors for every pair, and that is not very good for maintenance. We could rely on a macro for that, but also not ideal.

> * For the geographic types, do these only work with Quantity from Unitful.jl or have you also tested it with the ones from DynamicQuantities.jl? I don't have a strong argument that it should/must, but I try to write code generic enough that works for both implementations where possible.

Initially we are considering only Unitful.jl given that it allows compile-time dispatch and specializations. It also has the benefit of retaining the same memory footprint of raw float numbers. I understand that DynamicQuantities.jl stores the units in the value, and therefore requires more memory to represent the same number of points.

mikeingold:
> We are following the symbols and units in the ISO. The docstrings will be updated soon.

I don't mean to change the naming so much as to use the phonetic expansions of Greek symbols, e.g. `phi` vs `ϕ`.

> That is something that we are trying to decide yet, if we should support the conversion in the constructors. The main drawback is that we would need to add constructors for every pair, and that is not very good for maintenance.

The solution I came up with here was to use a single constructor e.g. `Coordinate(::AbstractCoordinate)` for each (see [here](https://github.com/mikeingold/UnitfulCoordinateSystems.jl/blob/bd1b1a9e601af087bab08c0652cb6fc46ac3bb9b/src/UnitfulCoordinateSystems.jl#L58) where these generic component functions are defined for each coordinate type (see [here](https://github.com/mikeingold/UnitfulCoordinateSystems.jl/blob/master/src/components.jl)). It's a bit of code, but all functions are fairly simple so shouldn't require much in the way of maintenance.

juliohm:
> I don't mean to change the naming so much as to use the phonetic expansions of Greek symbols, e.g. `phi` vs `ϕ`.

That seems reasonable. We like unicode though, and use it everywhere in the codebase already. We will see if use cases of these coords can benefit from more verbose field names.

> The solution I came up with here was to use a single constructor e.g. `Coordinate(::AbstractCoordinate)` for each (see [here](https://github.com/mikeingold/UnitfulCoordinateSystems.jl/blob/bd1b1a9e601af087bab08c0652cb6fc46ac3bb9b/src/UnitfulCoordinateSystems.jl#L58) where these generic component functions are defined for each coordinate type (see [here](https://github.com/mikeingold/UnitfulCoordinateSystems.jl/blob/master/src/components.jl)). It's a bit of code, but all functions are fairly simple so shouldn't require much in the way of maintenance.

That is a nice approach, assuming that we can find a general set of accessor functions. What we started to realize is that we won't need to call these `convert` methods directly, and instead will provide high-level abstractions to convert entire datasets from one set of coordinates to another.

src/coordinates.jl
```diff
@@ -0,0 +1,81 @@
...
+# UTILITIES
+# ----------
+
+const Len{T} = Quantity{T,u"𝐋"}
+const Rad{T} = Quantity{T,NoDims,typeof(u"rad")}
+const Deg{T} = Quantity{T,NoDims,typeof(u"°")}
+
+# only add the unit if the argument is not a quantity
+addunit(x::Number, u) = x * u
+addunit(x::Quantity, u) = throw(ArgumentError("invalid units for coordinate type, please check the docstring"))
```
juliohm:
```suggestion
addunit(x::Quantity, u) = throw(ArgumentError("invalid units for coordinates, please check the documentation"))
```

juliohm (approved):

PR #706, opened 2024-01-24
remove T from center(::Segment) method

kylebeggs:
Addresses #702 . Not sure if the tests are the right way to go about it.

Issue #702, opened 2024-01-23
allowing `center(::Segment{Dim,T})` when `T` is not `<:Real` such as a `Unitful.Quantity`


kylebeggs:
Opening a new issue although discussed in #692 so we can brainstorm a solution before making a PR. The immediate issue is that this implementation does not work when the data type of `Segment` is not a `Real`, as then the check to make sure the unitless parameter `t` is between 0 and 1 errors.

Here is the current implementation:
```julia
center(s::Segment{Dim,T}) where {Dim,T} = s(T(0.5))

function (s::Segment)(t)
  if t < 0 || t > 1
    throw(DomainError(t, "s(t) is not defined for t outside [0, 1]."))
  end
  a, b = s.vertices
  a + t * (b - a)
end
```

A few options:
- convert 0 and 1 in the check for `t` for match the data type of `Segment`. This is what I did initially, but I don't think this makes sense because `t` should always be a unitless real number, and then now you are constrained to this only working for data types which have comparison operators defined.
- A solution that works for `Unitful.Quantity`s would be to just convert 0.5 to the underlying data type of the `Unitful.Quantity` which would give you the same type as the `Points` of `Segment` which is the goal. This was the last state of #692 but also is not ideal because then you'd need to define helper methods to "get the precision".
- Perhaps just calculate the mean directly without using `t` and generic implementation for `Segment`?
```julia
function center(s::Segment)
  a, b = coordinates.(s.vertices)
  return (a + b) ./ 2
end
```

juliohm:
The best approach is the third one where we redefine the center in terms of a and b directly, and divide by the integer 2 instead of multiply by 0.5.

juliohm:
Thanks @kylebeggs , fixed on master.

PR #718, opened 2024-01-30
generalize center(::Segment)

kylebeggs:
This PR is a follow-up of #706 which addresses #702 .

In the previous review, @juliohm commented on `@test coordtype(center(s))` stating

> The correct test here is one that checks the unit of the coordtype is u"m". We don't need to store TT to check the exact Unitful.jl type.

but I think it is important to test for the exact type, because I would expect that same type of `Point` as in the input `Segment` back when doing `center(::Segment)`.

### DIFF
--- a/src/polytopes/segment.jl
+++ b/src/polytopes/segment.jl
@@ -21,7 +21,10 @@ Base.maximum(s::Segment) = s.vertices[2]
 
 Base.extrema(s::Segment) = s.vertices[1], s.vertices[2]
 
-center(s::Segment{Dim,T}) where {Dim,T} = s(T(0.5))
+function center(s::Segment)
+  a, b = coordinates.(extrema(s))
+  return Point((a + b) ./ 2)
+end
 
 Base.isapprox(s₁::Segment, s₂::Segment; kwargs...) =
   all(isapprox(v₁, v₂; kwargs...) for (v₁, v₂) in zip(s₁.vertices, s₂.vertices))
--- a/test/polytopes.jl
+++ b/test/polytopes.jl
@@ -54,6 +54,15 @@
     @test center(s) == P3(0.5, 0.5, 0.5)
     @test coordtype(center(s)) == T
 
+    x1 = T(0)u"m"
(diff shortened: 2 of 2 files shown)


src/polytopes/segment.jl
```diff
@@ -21,7 +21,10 @@ Base.maximum(s::Segment) = s.vertices[2]
 
 Base.extrema(s::Segment) = s.vertices[1], s.vertices[2]
 
-center(s::Segment{Dim,T}) where {Dim,T} = s(T(0.5))
+function center(s::Segment)
+  a, b = coordinates.(extrema(s))
+  return Point((a + b) ./ 2)
```
```suggestion
  a, b = extrema(s)
  Point((coordinates(a) + coordinates(b)) / 2)
```

test/polytopes.jl
```diff
@@ -54,6 +54,15 @@
...
     @test coordtype(center(s)) == T
 
+    x1 = T(0)u"m"
+    x2 = T(1)u"m"
+    s = Segment(Point(x1, x1, x1), Point(x2, x2, x2))
+    @test boundary(s) == Multi([Point(x1, x1, x1), Point(x2, x2, x2)])
+    @test perimeter(s) == 0u"m"
+    x_mid = T(0.5)u"m"
+    @test center(s) == Point(x_mid, x_mid, x_mid)
+    @test coordtype(center(s)) == typeof(x_mid)
```
```suggestion
    xm = T(0.5)u"m"
    @test center(s) == Point(xm, xm, xm)
    @test coordtype(center(s)) == typeof(xm)
```

test/polytopes.jl
```diff
@@ -54,6 +54,15 @@
     @test center(s) == P3(0.5, 0.5, 0.5)
     @test coordtype(center(s)) == T
 
+    x1 = T(0)u"m"
```
juliohm, line 58:
```suggestion
    # unitful coordinates
    x1 = T(0)u"m"
```

juliohm (changes requested):

juliohm (approved):

PR #732, opened 2024-02-06
Add GIS Conversions

eliascarv:


src/coordinates/conversions.jl
```diff
@@ -0,0 +1,57 @@
...
+  λ = x / a
+  ϕ = atan(sinh(y / a))
+  LatLon(rad2deg(ϕ) * u"°", rad2deg(λ) * u"°")
+end
+
+# adjust negative angles
+function atanpos(y, x)
+  α = atan(y, x)
+  ifelse(α ≥ zero(α), α, α + oftype(α, 2π))
+end
```
juliohm: move to coordinates.jl as this is an auxiliary function

juliohm (approved):

PR #742, opened 2024-02-08
Implement IndexedAdjacenciesTopology

RomeoV:
The current topologies, namely SimpleTopology, GridTopology, and HalfEdgeTopology, are not well suited for high-dimensional meshes.

In this PR we add the IndexedAdjacenciesTopologies (IA) described in
- Paoluzzi, A., Bernardini, F., Cattani, C., & Ferrucci, V. (1993). Dimension-independent modeling with simplicial complexes. ACM Transactions on Graphics (TOG), 12(1), 56-102.
and compared in
De Floriani, L., & Hui, A. (2007). Shape Representations Based on Simplicial and Cell Complexes. In Eurographics (State of the Art Reports) (pp. 63-87).

In particular, we assume that every element of the mesh is a Simplex, i.e. the simplest object with a given rank (segment for K=1, triangle for K=2, tetrahedon for K=3, etc) and implement basic topology relations.

Currently supported are
- a new data type `Ksimplex{K, dim, T}`;
- `Adjacency{K}` relations, e.g. finding adjacent triangles of a triangle;
- `Coboundary{0,k}`, i.e. finding the elements that are touching a vertex;
- `normal` for any simplex in any dimension;
- `measure` computing the (hyper-)volume in any dimension;
- `viz` just works (!) by way of retrieving the `elements`.

It would be nice to also implement all the other topological relations at some point, but the papers are not so clear how to do that, and they're not required for many use cases.

Happy to receive any feedback on the status quo.


juliohm:
That is amazingly fast work @RomeoV , thanks for putting it together 🚀 

We need to split the PR into smaller self-contained PRs. The way I would proceed here:

1. PR with WingedTopology (I think the authors of the paper call it Winged instead of IndexedAdjacencies)
2. PR with topological relations for WingedTopology (These relations need to use the public api of WingedTopology)
3. PR with KSimplex geometry (Let's discuss this further, it seems useful for theoretical work)

Can we close this PR and start with (1)? Notice that we have a code style in place (we don't use underscores, etc), it would be nice to follow the code style as much as possible.

RomeoV:
I don't like the name winged `WingedTopology` for two reasons:
- First, imo the name is quite opaque if you're not familiar with the topic;
- second, if you search for `Winged Topology` you get several entries about the [Winged Edge](https://en.wikipedia.org/wiki/Winged_edge) data structure, which is another data structure (for 2D representations only), but is just close enough to be confusing. To be fair, though, searching for "Indexed Adjacencies Topology" doesn't yield much better. Not sure which direction to go here.

RomeoV:
About the PR order:
IndexedAdjacenciesTopology relies on the `Ksimplex` class, so I think we'd need to merge that first.

The definition is currently

``` julia
struct IndexedAdjacenciesTopology{K, dim, K_} <: Topology
    simplicies::AbstractVector{Connectivity{Ksimplex{K, dim}, K_}}
    neighboring_simplicies::AbstractVector{NTuple{K_, Int}}
    R_star_relations::AbstractVector{Int}  # see L De Floriani, A Hui, 2007; Section 5.3
    # [...] constructor
end
```


As an aside, do you prefer `KSimplex` or `Ksimplex`? I saw that it's `Ngon` not `NGon`, but I suppose the word also works different ("gon" isn't a name by itself).

RomeoV:
Re code style: Is that defined somewhere or just implicit? I can't seem to find anything, either on the Meshes.jl->Contributions page, or on the ColPrac.jl page linked from the Readme.

juliohm:
> I don't like the name winged `WingedTopology` for two reasons:
> 
> * First, imo the name is quite opaque if you're not familiar with the topic;
> * second, if you search for `Winged Topology` you get several entries about the [Winged Edge](https://en.wikipedia.org/wiki/Winged_edge) data structure, which is another data structure (for 2D representations only), but is just close enough to be confusing. To be fair, though, searching for "Indexed Adjacencies Topology" doesn't yield much better. Not sure which direction to go here.

What about `PaoluzziTopology` to refer to the first author? We follow this naming convention sometimes where there is ambiguity in the literature. It is also nice to acknowledge their research.

> IndexedAdjacenciesTopology relies on the `Ksimplex` class, so I think we'd need to merge that first.

That is a design issue. We need to decouple the geometry from the topology, the current `KSimplex` stores vertices with materialized coordinates. You can see that our `SimpleTopology`, `GridTopology`, `HalfEdgeTopology` ... only store indices into the geometric objects. Can you think of a refactor that fully decouples the topology from the vertices?

> Re code style: Is that defined somewhere or just implicit? I can't seem to find anything, either on the Meshes.jl->Contributions page, or on the ColPrac.jl page linked from the Readme.

We try to be explicit, but it is so tedious to document every detail of the style that we follow some implicit conventions too. If you can use the other topology types as inspiration, you will see that we favor short variable names without underscores. We use symbols instead of full words when the context is clear, etc.

RomeoV:
> We need to decouple the geometry from the topology.

The difference from the new topology and the implemented topologies is that is only works with simplices, so it seems sensible to force that in the topology.
We could consider making a new abstract type `AbstractKSimplex <: Polytope` and then only forcing that in the struct definition, but we can not allow general geometric objects.

RomeoV:
> We try to be explicit [...]

Just to be clear, there is no official style guide or something like that, correct? Just trying to minimize the number of changes I'm making ;)
Isn't there any naming variables etc style guide in the SciML community? EDIT: I guess there is this: https://github.com/SciML/SciMLStyle I'll have a look.

RomeoV:
> What about `PaoluzziTopology`.

I think that could work. But can you explain to me what you don't like about `IndexedAdjacenciesTopology`? I find it a very fitting name that's also used in the literature (in the paper you sent me originally).

RomeoV:
> the current KSimplex stores vertices with materialized coordinates

That's true, but the `IndexedAdjacenciesTopology` only stores `Connectivity{Ksimplex}`, so it doesn't store the materialized coordinates.
This is modeled after the way `SimpleTopology` does it.

juliohm:
> The difference from the new topology and the implemented topologies is that is only works with simplices, so it seems sensible to force that in the topology.
> We could consider making a new abstract type `AbstractKSimplex <: Polytope` and then only forcing that in the struct definition, but we can not allow general geometric objects.

Got it, so maybe we should focus on the `Simplex` type first following your suggestion. Ideally we will be able to refactor `Triangle` and `Tetrahedron` to be `Simplex` too.

> Just to be clear, there is no official style guide or something like that, correct?

We do have a bot that formats the code in the master branch. The explicit part of the code style is here: https://github.com/JuliaGeometry/Meshes.jl/blob/master/.JuliaFormatter.toml

> I think that could work. But can you explain to me what you don't like about `IndexedAdjacenciesTopology`?

It is supeeeerrr looooong to type =)

Jokes aside, I think we could work with an abbreviation like `IndAdjTopology`? What do you think? Any shorter variation would be an improvement.

RomeoV:
I personally don't think it's too long, but I think `IndAdjTopology` is a good compromise.

I agree we can probably refactor Triangle etc as a `KSimplex`, but perhaps we can do that in a later PR. I think it would be great if we can merge something so we don't run out of steam.
FYI I'm trying to finish these PRs today because my schedule is not looking good for next week ;p

RomeoV:
The three PRs are done and updated, naturally relying on one another, with the order
- add `Ksimplex` (lmk if you prefer `KSimplex`)
- add `IndAdjTopology`
- add toporelations for the new topology.

Let me know if you have any major concerns at the moment, but as the changes affect the other parts of the library I hope we can move forward swiftly, and maybe improve things later.

Do you want me to include some versions bumps for each?

RomeoV:
Note btw that my use of underscores in variables names should at least be consistent with the SciML guide on variable names: https://github.com/SciML/SciMLStyle?tab=readme-ov-file#general-naming-principles

RomeoV:
Closed in favor of #744 #746 #747 .

PR #744, opened 2024-02-08
Create Simplex data type.

RomeoV:
## Summary
A K-simplex is a "simplical" polytope that occupies space in K dimensions using K+1 points,
and is embedded in some dimension `Dim`.
For example, a line segment in 3-D space would be Simplex{1, 3} as it occupies a single dimension,
is made up of two points, and lies in 3d space.

## More detailed rationale
For many applications, meshing 3d space (e.g. into tetraheda) is sufficient, and topologies like `SimpleTopology` or `HalfEdgeTopology` can be used to represent meshes and support relational lookups (adjacent vertices, faces, edges, etc).
However, some applications require meshing in higher dimensions. For example, we may mesh the domain of a multivariate probability distribution, which may be high dimensional, into appropriate "cells", and can not simply rely on a grid structure.

A problem in higher dimensions is that for general polytopes, seemingly simple properties like the (hyper)-area become computationally infeasible, and it becomes more unclear how to compute a normal vector.
Similarly, neighboring face lookup must be supported by an efficient data structure.
However, if we only consider a mesh of simplices, these relations become more clear again.
(For example, a closed form solution to the normal vector and volume exist, and there is a exactly known number of neighbors.)

We therefore start by introducing the `Simplex`, a data type that represents a specific type of polytope with the minimal number of vertices, and implements efficient computational methods for computing commonly used properties.
The `Simplex` type may therefore be used to construct high dimensional meshes (something that we are doing in another package but are currently not including here).

Once the `Ksipmlex` is defined, we can define a topology data structure over simplicies of a given rank and embedding dimension, and compute efficient relational lookups.

## Relation to existing simplices
As noted, there are already simplices defined for lower dimensions, namely `Segment`, `Triangle`, and `Tetrahedon`; each of which subtype from `Polygon <: Polytope`.
Supposing the existence of `Simplex`, one can wonder whether we should redefine the existing simplices as a subtype. However, we run into a type system connundrum:
- Triangle should be both subtyping from `Polygon` and `Simplex`, as they provide distinct properties.
  Namely: `Polygon` defines a rank=2 structure made up of some number of points. `Simplex` defines a rank=K structure made up out of exactly K+1 points. In that sense, a triangle is both a `Polygon` and a `Simplex` at once (in some sense it is the intersection of the two types). However, Julia doesn't support multiple inheritance style type systems.
  
## Other approaches.
It's noteworthy that we already have an abstract type `Polytope`, and a way to instantiate specific polytopes, like `@polytope Tetrahedron 3 4`.
We may therefore also define any given Simplex as `@polytope My_k_4_dim_5_simplex 4 5` and additionally define `issimplex(::My_k_4_dim_5_simplex) = true`, and similarly for `measure`, `nvertices`, etc etc
We could even move this into a new macro `@simplex` that defines all the variables.
However, the way types are defined in the macro still stays inflexible for e.g. redefinition on the fly, and is generally more opaque.
Instead, we could simply redefine `const Segment = Simplex{1}`, and similarly for triangle and tetrahedon.

## Implementation details:

We provide:
- Efficient storage
- a bunch of constructors

Notice that we introduce an auxiliary type `N==(K+1)`. Currently Julia doesn't really have computed type capabilities, so the constructor has to take care of computing `N` correctly.
Check https://discourse.julialang.org/t/addition-to-parameter-of-parametric-type/20059/6 for some discussion.

This PR takes care of the first part from #742. Additional utilities for the `Simplex` type are introduced in #764.


juliohm:
Currently we define two types of simplex: Triangle and Tetrahedron, which as you know have fixed dimension. These two geometry types support the `issimplex(g)` trait that is resolved in compile time.

I understand that you are suggesting a new `Polytope` type that generalizes the `Triangle` and `Tetrahedron` to N-dimensions. Perhaps we could call it `Simplex`? If we add this geometry type, then we will need to add a method of `issimplex`. I am just wondering if this geometry type is really necessary, or if it an auxiliary construct for the topology data structured in #742 ?

If it is really necessary, then I think we should consider refactoring `Triangle` and `Tetrahedron` as subtypes of this general `Simplex`. This will require more work than what we have currently in this PR.

I suggest that we focus on the topology type first (decoupling indices from vertices) and then we can come back here to reassess the need of the `Simplex` as a geometry.

RomeoV:
@juliohm Please find a more detailed rationale in the header of the PR.

RomeoV:
I am also noticing that I keep writing `KSimplex` instead of `Ksimplex`, so I have come to the conclusion that is the better name.

juliohm:
Thank you for the detailed comments @RomeoV , they are very helpful.

Based on your investigation, I agree that we should not try to subtype the fixed-dimension simplices (Segment, Triangle, Tetrahedron). Instead, the `Simplex` could be an independent geometry type for n-dimensional applications.

A couple of improvements before we can merge this work:

1. Let's use the simpler name `Simplex` instead of `KSimplex` or `Ksimplex`. The simpler the better.
2. Please only rely on the public api of the objects, replace instances like `getfield.(vertices(splx), :coords)` by `@. coordinates(vertices(splx))`
3. Remove commented lines and avoid undercores in variable names

Looking forward to take another look into the PR after this first round of review 👍🏽

RomeoV:
@juliohm I think the PR is ready for another round of review.

src/connectivities.jl
```diff
@@ -94,7 +94,7 @@ end
 Materialize a face using the `connec` list and a global vector of `points`.
 """
 materialize(connec::Connectivity{PL,N}, points::AbstractVector{P}) where {PL<:Polytope,N,P<:Point} =
-  PL(ntuple(i -> @inbounds(points[connec.indices[i]]), N))
+  PL(ntuple(i -> @inbounds(points[connec.indices[i]]), N)...)
```
juliohm: This seems unrelated to the Simplex addition. Can you please revert? Splatting leads to slow performance sometimes.

src/polytopes.jl
```diff
@@ -189,6 +189,12 @@ include("polytopes/tetrahedron.jl")
 include("polytopes/hexahedron.jl")
 include("polytopes/pyramid.jl")
 
+# -----------------------
+# K-SIMPLEX
+# -----------------------
+include("polytopes/simplex.jl")
```
```suggestion
# -----------------------
# K-SIMPLEX
# -----------------------

include("polytopes/simplex.jl")
```

src/polytopes/simplex.jl
```diff
@@ -0,0 +1,69 @@
+"""
+    Simplex(p1, p2, ..., pK, pK+1)
+
+A K-simplex is a "simplical" polytope that occupies space in K dimensions using K+1 points,
+and is embedded in some dimension `Dim`.
+For example, a line segment in 3-D space would be Simplex{1, 3} as it occupies a single dimension,
+is made up of two points, and lies in 3d space.
```
juliohm:
```suggestion
A K-simplex is a polytope with parametric dimension K and K+1 vertices.
It generalizes the [`Segment`](@ref), the  [`Triangle`](@ref) and the
[`Tetrahedron`](@ref) to higher dimensions (K > 3).

See also [`issimplex`](@ref).
```

src/polytopes/simplex.jl
```diff
@@ -0,0 +1,69 @@
...
+
+Note that we only allow constructors of the form `Simplex(p1,p2,...,pk+1)`, not `Simplex(Tuple(...))`.
```
juliohm:
```suggestion
```

juliohm: We need to follow the same convention of Triangle, Tetrahedron, etc. If they can be constructed with Tuple, then Simplex should be too.

src/polytopes/simplex.jl
```diff
@@ -0,0 +1,69 @@
...
+"""
+struct Simplex{K,Dim,T,K_} <: Polytope{K,Dim,T}
+    # Notice that this class has no default constructor (on purpose), since `K` can't be determined.
+    # Instead we provide different constructors that also include some dimensionality checking.
+    vertices::NTuple{K_, Point{Dim, T}}
```
juliohm: identation is 2 whitespaces

src/polytopes/simplex.jl
```diff
@@ -0,0 +1,69 @@
...
+end
+
+# ---------------------
+# CONSTRUCTORS
+# ---------------------
+# Include some dimensionality checking.
```
juliohm:
```suggestion
```

src/polytopes/simplex.jl
```diff
@@ -0,0 +1,69 @@
...
+Simplex{K,Dim}(vertices::Vararg{Point{Dim,T},K_})    where {K,Dim,T,K_   } = let
+    K+1 == K_ || throw(ArgumentError("Num vertices must be rank K plus one."))
+    K<=Dim || throw(ArgumentError("Rank K must be less or equal to embedding dimension Dim."))
+    Simplex{K_-1,Dim,T,K_}(vertices)
+end
+Simplex{K,Dim,T}(vertices::Vararg{Point{Dim,T′},K_}) where {K,Dim,T,K_,T′} = let
+    K+1 == K_ || throw(ArgumentError("Num vertices must be rank K plus one"))
+    K<=Dim || throw(ArgumentError("Rank K must be less or equal to embedding dimension Dim."))
+    Simplex{K_-1,Dim,T,K_}(vertices)
+end
```
juliohm: We need better error messages for end-users. Try to avoid abbreviations, use plain English.

src/polytopes/simplex.jl
```diff
@@ -0,0 +1,69 @@
...
+
+"Compute the measure (i.e. hyper-volume) in any dimension using the Cayley-Menger Determinant."
+function measure(splx::Simplex{K,Dim,T}) where {K, Dim, T<:Real}
+    # https://en.wikipedia.org/wiki/Cayley%E2%80%93Menger_determinant
+    Ds_ = pairwise(SqEuclidean(), coordinates.(vertices(splx)))
+    Ds = [Ds_ ones(size(Ds_, 1), 1);
+          ones(1, size(Ds_, 2)) 0]
+    factor = (-1)^(K+1)/(factorial(K)^2*2^K)
+    return sqrt(factor * det(Ds))
+end
```
juliohm: Please make sure that these functions are defined in the correct files. Meshes.jl is such a large project that these small organizational details make a huge difference in maintenance.

test/polytopes.jl
```diff
@@ -870,4 +870,49 @@
       └─ Point(0.0, 0.0, 1.0)"""
     end
   end
+
+  @testset "Simplex" begin
+    @testset "construction" begin
+        pts = (Point(1., 2., 3.), Point(1., 2., 4.));
+        pts32 = convert.(Point{3,Float32}, pts)
```
juliohm: Please use the T, P2, P3 constants defined in the tests. The tests are run with single and double precision if you use these constants.

test/polytopes.jl
```diff
@@ -870,4 +870,49 @@
...
+
+        @test Simplex(pts...).vertices == pts
+        @test Simplex{1}(pts...).vertices == pts
+        @test Simplex{1,3}(pts...).vertices == pts
+        @test Simplex{1,3,Float64}(pts...).vertices == pts
+        @test Simplex{1,3,Float32}(pts...).vertices == pts32
+        @test Simplex(pts32...).vertices == pts32
```
juliohm: Try to focus on end-user constructors. These internal constructors are an implementation detail most times, and can change in future refatorings.

src/polytopes/simplex.jl
eliascarv, line 40: Parametric constructors are not intended for end users in Meshes.jl. Only the `Simplex((pts...))` and `Simplex(pts...)` constructors are required:

```suggestion
function Simplex(vertices::NTuple{N,Point{Dim,T}}) where {Dim,T,N}
  K = N - 1
  K ≤ Dim || throw(ArgumentError("simplex rank (number of vertices - 1) must be less or equal to embedding dimension"))
  Simplex{K,Dim,T,N}(vertices)
end

Simplex(vertices::Point{Dim,T}...) where {Dim,T} = Simplex(vertices)
Simplex(vertices::Tuple...) = Simplex(Point.(vertices))
```

eliascarv (changes requested):

src/polytopes/simplex.jl
```diff
@@ -0,0 +1,69 @@
...
+For example, a line segment in 3-D space would be Simplex{1, 3} as it occupies a single dimension,
+is made up of two points, and lies in 3d space.
+
+Note that we only allow constructors of the form `Simplex(p1,p2,...,pk+1)`, not `Simplex(Tuple(...))`.
+"""
+struct Simplex{K,Dim,T,K_} <: Polytope{K,Dim,T}
+    # Notice that this class has no default constructor (on purpose), since `K` can't be determined.
+    # Instead we provide different constructors that also include some dimensionality checking.
+    vertices::NTuple{K_, Point{Dim, T}}
+end
```
eliascarv: Please, use `N` instead of `K_`.
```suggestion
struct Simplex{K,Dim,T,N} <: Polytope{K,Dim,T}
  vertices::NTuple{N,Point{Dim,T}}
end
```

src/polytopes/simplex.jl
```diff
@@ -0,0 +1,69 @@
...
+A K-simplex is a "simplical" polytope that occupies space in K dimensions using K+1 points,
+and is embedded in some dimension `Dim`.
+For example, a line segment in 3-D space would be Simplex{1, 3} as it occupies a single dimension,
+is made up of two points, and lies in 3d space.
+
+Note that we only allow constructors of the form `Simplex(p1,p2,...,pk+1)`, not `Simplex(Tuple(...))`.
+"""
+struct Simplex{K,Dim,T,K_} <: Polytope{K,Dim,T}
+    # Notice that this class has no default constructor (on purpose), since `K` can't be determined.
+    # Instead we provide different constructors that also include some dimensionality checking.
```
eliascarv: I think this comment is not necessary. What do you think @juliohm?

src/polytopes/simplex.jl
eliascarv: Parametric constructors are not intended for end users in Meshes.jl. Only the `Simplex((pts...))` and `Simplex(pts...)` constructors are required:

```suggestion
function Simplex(vertices::NTuple{N,Point{Dim,T}}) where {Dim,T,N}
  K = N - 1
  K ≤ Dim || throw(ArgumentError("simplex rank (number of vertices - 1) must be less or equal to embedding dimension"))
  Simplex{K,Dim,T,N}(vertices)
end

Simplex(vertices::Point{Dim,T}...) where {Dim,T} = Simplex(vertices)
Simplex(vertices::Tuple...) = Simplex(Point.(vertices))
```

RomeoV:
Thanks for all the feedback @juliohm @eliascarv. All your suggestions should be included now. 
@eliascarv, you suggested to remove the parametric constructors. However, they are required for `materialize`, see f74d34e4dcc28f7df4c1b5db43285cd5689a0649.

I have also added a few more predicates. ~However, currently the `Base.in` predicate has an issue, which I have to find still. If that is the only blocker, I will revert the addition of `Base.in` and `Meshes.sideof` so we can proceed.~ The rest should be good as is, and you're welcome to add some more feedback. Thanks!

EDIT: The `Base.in` is fixed.

src/measures.jl
```diff
@@ -83,6 +83,23 @@ function measure(t::Tetrahedron)
...
 
+"Compute the measure (i.e. hyper-volume) for a simplex in any dimension using the Cayley-Menger Determinant."
+function measure(splx::Simplex{K,Dim,T,N}) where {K, Dim, T<:Real, N}
+  distancebuf = Matrix{T}(undef, N+1, N+1)
+  measure!(distancebuf, splx)
+end
+
+"In-place verion of [`measure`](@ref)."
+function measure!(distancebuf::AbstractMatrix{T}, splx::Simplex{K,Dim,T,N}) where {K, Dim, T<:Real, N}
+  size(distancebuf) == (N+1, N+1) || throw(ArgumentError("Buffer dimensions must be (N+1) times (N+1)."))
```
eliascarv: In the Julia documentation, the recommendation is that error messages start with a lowercase letter, unless the argument or type name is written with a capital letter:
image

src/connectivities.jl
```diff
@@ -83,6 +83,11 @@ function connect(indices::Tuple, ::Type{Ngon})
   Connectivity{Ngon{N},N}(indices)
 end
 
+function connect(indices::Tuple, ::Type{Simplex})
+  N = length(indices)
+  Connectivity{Simplex{N-1},N}(indices)
+end
+
```
juliohm: I don't think this method is necessary given that we already have a general method for Polytope?

src/measures.jl
```diff
@@ -83,6 +83,23 @@ function measure(t::Tetrahedron)
...
+
+"In-place verion of [`measure`](@ref)."
+function measure!(distancebuf::AbstractMatrix{T}, splx::Simplex{K,Dim,T,N}) where {K, Dim, T<:Real, N}
+  size(distancebuf) == (N+1, N+1) || throw(ArgumentError("buffer dimensions must be (N+1) times (N+1)"))
+  # https://en.wikipedia.org/wiki/Cayley%E2%80%93Menger_determinant
+  pairwise!(SqEuclidean(), @view(distancebuf[1:end-1, 1:end-1]), coordinates.(vertices(splx)))
+  distancebuf[end, 1:end-1] .= 1; distancebuf[1:end-1, end] .= 1; ; distancebuf[end, end] = 0
+  factor = (-1)^(K+1)/(factorial(K)^2*2^K)
+  return sqrt(factor * det(distancebuf))
+end
```
juliohm: Please avoid introducing a new function `measure!`. Our interface for measure takes a geometry and returns a scalar, there is no need for a variant with mutation.

src/polytopes.jl
```diff
@@ -189,6 +189,13 @@ include("polytopes/tetrahedron.jl")
 include("polytopes/hexahedron.jl")
 include("polytopes/pyramid.jl")
 
+# -----------------------
+# K-SIMPLEX
```
```suggestion
# SIMPLEX
```

src/polytopes/simplex.jl
```diff
@@ -0,0 +1,57 @@
+"""
+    Simplex(p1, p2, ..., pK, pK+1)
+
+A K-simplex is a polytope with parametric dimension K and K+1 vertices.
+It generalizes the [`Segment`](@ref), the  [`Triangle`](@ref) and the
+[`Tetrahedron`](@ref) to higher dimensions (K > 3).
+
+See also [`issimplex`](@ref).
+
+Note that we only allow constructors of the form `Simplex(p1,p2,...,pk+1)`, not `Simplex(Tuple(...))`.
```
juliohm:
```suggestion
```

src/polytopes/simplex.jl
```diff
@@ -0,0 +1,57 @@
...
+"""
+struct Simplex{K,Dim,T,N} <: Polytope{K,Dim,T}
+  vertices::NTuple{N,Point{Dim,T}}
+end
+
+
```
juliohm, line 26:
```suggestion
```

src/polytopes/simplex.jl
```diff
@@ -0,0 +1,57 @@
...
+end
+Simplex{K}(vertices::Vararg{Point,N}) where {K,N} = let
+  N == K+1 || throw(ArgumentError("number of vertices must be rank K plus one"))
+  Simplex(vertices...)
+end
+
+# ---------------------
+# HIGH-LEVEL INTERFACE
+# ---------------------
+
```
juliohm:
```suggestion
```

juliohm (changes requested):
Thank you @RomeoV for the continuous improvements. After seeing this new round of updates, I think we need to split this work further into smaller pieces that are guaranteed to meet the high standards of the project.

Can you remove all the Base.in, normal, sideof, measure implementations from this PR, and focus on writing just the type with constructors, tests of construction and a section in the docs?

After this one is merged, we can work on additional PRs with one function implementation at a time, with more tests and comments.

I am starting to feel that this new `Simplex` type is basically functioning as a `Mesh` of simplices in dimensions higher than 3. We need to be careful to not implement the wrong interface (`Geometry` vs. `Mesh`).

Thanks for the patience, I am sure that we will converge on this, and that you will be more used to the standards adopted in this project.

RomeoV:
Thanks. I understand the need for improving the code to fit the quality standards of the library. Unfortunately, I will need to delegate this to another time as my other responsibilities are taking over, and we are still quite far from merging the other functionality (`IndexedAdjacenciesTopology`) which will make the `Simplex` useful for higher dimensions. I will see if I can get back to it by the end of the month.

RomeoV:
> I am starting to feel that this new Simplex type is basically functioning as a Mesh of simplices in dimensions higher than 3.

I don't agree with this. A `Simplex` is a `Geometry` that consists of a bunch of fully connected vertices and doesn't really relate to being a `Mesh`, which assembles geometric objects with non-trivial connections between them. Of course later the `Simplex` type is used to build a mesh in higher dimensions (the point of implementing `Simplex` in the first place), i.e. #746, but in this PR there shouldn't be anything related to a mesh just yet.

Maybe you can elaborate on your point to help me understand your concerns.

juliohm:
> I don't agree with this. A `Simplex` is a `Geometry` that consists of a bunch of fully connected vertices and doesn't really relate to being a `Mesh`, which assembles geometric objects with non-trivial connections between them. Of course later the `Simplex` type is used to build a mesh in higher dimensions (the point of implementing `Simplex` in the first place), i.e. #746, but in this PR there shouldn't be anything related to a mesh just yet.
> 
> Maybe you can elaborate on your point to help me understand your concerns.

I don't disagree with this, I am just saying that it is good to try to differentiate these two carefully moving forward, as many methods of one could apply to another.

We can continue working on this when you have more time to improve the PR. Looking forward to it 👍🏽

RomeoV:
I split the PR as requested, the other part is in #764 . We should be able to proceed to just review the Simplex type now.

Note that I also ran the Julia Formatter on all the files (I think) so hopefully that should fix some of the nits.

src/polytopes.jl
```diff
@@ -189,6 +189,12 @@ include("polytopes/tetrahedron.jl")
 include("polytopes/hexahedron.jl")
 include("polytopes/pyramid.jl")
 
+# -----------------------
+# SIMPLEX
+# -----------------------
```
```suggestion
# --------
# SIMPLEX
# --------
```

src/polytopes/simplex.jl
```diff
@@ -0,0 +1,41 @@
...
+nvertices(::Type{<:Simplex{K}}) where {K} = K + 1
+
+function Base.isapprox(p₁::SimplexT, p₂::SimplexT; kwargs...) where {SimplexT<:Simplex}
+  all(isapprox(v₁, v₂; kwargs...) for (v₁, v₂) in zip(vertices(p₁), vertices(p₂)))
+end
```
juliohm:
```suggestion
function Base.isapprox(s₁::Simplex, s₂::Simplex; kwargs...)
  all(isapprox(v₁, v₂; kwargs...) for (v₁, v₂) in zip(vertices(s₁), vertices(s₂)))
end
```

src/connectivities.jl
```diff
@@ -83,6 +83,11 @@ function connect(indices::Tuple, ::Type{Ngon})
   Connectivity{Ngon{N},N}(indices)
 end
 
+function connect(indices::Tuple, ::Type{Simplex})
+  N = length(indices)
+  Connectivity{Simplex{N - 1},N}(indices)
+end
+
```
juliohm, line 90: Please update the docstring as well:

```
The type `PL` can be a [`Ngon`](@ref) or [`Simplex`](@ref) in which case the length of
the indices is used to identify the actual polytope type.
```

src/polytopes/simplex.jl
```diff
@@ -0,0 +1,41 @@
...
+
+# Parametric constructor required for [`materialize`](@ref).
+function Simplex{K}(vertices::NTuple{N}) where {K,N}
+  N == K + 1 || throw(ArgumentError("Number of vertices must be rank K plus one."))
+  Simplex(vertices...)
+end
+function Simplex{K}(vertices::Vararg{Point,N}) where {K,N}
+  N == K + 1 || throw(ArgumentError("Number of vertices must be rank K plus one."))
+  Simplex(vertices...)
+end
```
juliohm: @eliascarv can you please double check that these constructors are following our conventions? I.e., they are defined in terms of the appropriate methods, avoiding unnecessary splatting, and parametric constructors coming first.

src/polytopes/simplex.jl
```diff
@@ -0,0 +1,41 @@
...
+function Simplex(vertices::NTuple{N,Point{Dim,T}}) where {Dim,T,N}
+  K = N - 1
+  K ≤ Dim || throw(ArgumentError("Simplex rank (number of vertices - 1) must be less or equal to embedding dimension."))
+  Simplex{K,Dim,T,N}(vertices)
+end
+
+Simplex(vertices::Point{Dim,T}...) where {Dim,T} = Simplex(vertices)
+Simplex(vertices::Tuple...) = Simplex(Point.(vertices))
+
+# Parametric constructor required for [`materialize`](@ref).
```
juliohm:
```suggestion
```

src/polytopes/simplex.jl
```diff
@@ -0,0 +1,41 @@
...
+function Simplex{K}(vertices::NTuple{N}) where {K,N}
+  N == K + 1 || throw(ArgumentError("Number of vertices must be rank K plus one."))
+  Simplex(vertices...)
+end
```
juliohm:
```suggestion
end

```

test/polytopes.jl
```diff
@@ -873,4 +873,51 @@
       └─ Point(0.0, 0.0, 1.0)"""
     end
   end
+  @testset "Simplex" begin
```
```suggestion

  @testset "Simplex" begin
```

test/polytopes.jl
```diff
@@ -873,4 +873,51 @@
...
+    @testset "construction" begin
```
juliohm: @RomeoV please remove these sub-testsets and try to follow the same format of the other tests for Chain, Polygon, Polyhedra

juliohm (changes requested):
Review is attached.

src/polytopes/simplex.jl
eliascarv: Create a main constructor and make the others call it.
```suggestion
function Simplex{K}(vertices::NTuple{N,Point{Dim,T}}) where {K,Dim,T,N}
  N == K + 1 || throw(ArgumentError("number of vertices must be equal to rank K plus one"))
  K ≤ Dim || throw(ArgumentError("rank K must be less or equal to embedding dimension"))
  Simplex{K,Dim,T,N}(vertices)
end

Simplex{K}(vertices::Point{Dim,T}...) where {K,Dim,T} = Simplex{K}(vertices)
Simplex{K}(vertices::Tuple...) where {K} = Simplex{K}(Point.(vertices))
```

eliascarv (changes requested):

src/polytopes/simplex.jl
eliascarv: I think it could be more compact:
```julia
Base.isapprox(s₁::Simplex, s₂::Simplex; kwargs...) =
  all(isapprox(v₁, v₂; kwargs...) for (v₁, v₂) in zip(vertices(s₁), vertices(s₂)))
```

juliohm (approved):

juliohm:
Merging as the failed test is random and not related.

PR #746, opened 2024-02-08
Create IndexedAdjacenciesTopology

RomeoV:
- may rename still
- relies on #744
- see #742 for some more discussion.


juliohm:
As discussed in #764 , I am closing this PR.

It would be nice to add this topological data structure for dim = 1, 2, 3 though. In case you think it is useful to your work in the future, please feel free to submit an updated PR with this limited scope, without the `Simplex` type.

PR #783, opened 2024-02-22
Implement inverse projection for `EqualAreaCylindrical`

eliascarv:


src/crs/eqareacylindrical.jl
```diff
@@ -99,13 +99,36 @@ function Base.convert(::Type{EqualAreaCylindrical{latₜₛ,Datum}}, coords::Lat
...
+  k₀ = cos(ϕₜₛ) / sqrt(1 - e² * sin(ϕₜₛ)^2)
+  # same formula as q, but ϕ = 90°
+  qₚ = ome² * (1 / ome² - (1 / 2e) * log((1 - e) / (1 + e)))
+  
+  λ = λ₀ + x / (a * k₀)
+  q = 2y * k₀ / a
+  β = asin(q / qₚ)
+
+  auth = AuthalicLatLon{Datum}(rad2deg(β) * u"°", rad2deg(λ) * u"°")
+  convert(LatLon{Datum}, auth)
```
eliascarv, line 133: It can also be written like this:
```suggestion
  λ = λ₀ + x / (a * k₀)
  q = 2y * k₀ / a
  β = asin(q / qₚ)
  ϕ = auth2geod(β, e²)

  LatLon{Datum}(rad2deg(ϕ) * u"°", rad2deg(λ) * u"°")
```
Which would be better?

src/crs/eqareacylindrical.jl
```diff
@@ -99,13 +107,36 @@ function Base.convert(::Type{EqualAreaCylindrical{latₜₛ,Datum}}, coords::Lat
...
```
eliascarv: It can also be written like this:
```suggestion
  λ = λ₀ + x / (a * k₀)
  q = 2y * k₀ / a
  β = asin(q / qₚ)
  ϕ = auth2geod(β, e²)

  LatLon{Datum}(rad2deg(ϕ) * u"°", rad2deg(λ) * u"°")
```
Which would be better?

Issue #789, opened 2024-02-28
Parameterization of a Torus with `minor > major`


mikeingold:
I stumbled onto this while testing a new surface integral function:

- A `Torus` can be constructed whose `minor` radii is greater than its `major` radii. Based on a quick reading of the Wikipedia page (IANA-geometer), it sounds like this is valid but sort of a special case.
- For such a `Torus`, the current parameterization code will error. Line 55 essentially takes `sqrt(t.major^2 - t.minor^2)`, which throws a `DomainError` since the argument is negative but also `<:Real`.

https://github.com/JuliaGeometry/Meshes.jl/blob/d52cc5e850a773d28e60bd43269ea33699de8774/src/primitives/torus.jl#L47-L63

I don’t know enough about how the parameterization here is derived to know whether the absolute value of the interior term could be taken first, or if parameterizing `Torus` with this property is a harder problem. It’s also non-obvious to me whether this type of `Torus` was even intended to be supported.

juliohm:
@stla could you please take a look in this issue? Do I remember correctly that you added the Torus primitive last year?

stla:
Yes, that's me. In fact, I implemented a non-usual parameterization of the torus, valid only for major > minor. I thought it was interesting because it is a conformal parameterization, but actually this is useless.

juliohm:
What do you suggest in this case @stla? Should we adopt a different parametrization?

stla:
Yes, we should adopt the classical parameterization (certainly given on Wikipedia).

juliohm:
Appreciate if you can take a look into it. Otherwise, we can try to fix
ourselves when we find time. Do you have a link to the wiki page and
formula? That can speed things up.

Em qua., 28 de fev. de 2024 11:53, stla ***@***.***> escreveu:

stla:
[Here](https://en.wikipedia.org/wiki/Torus#Geometry). I could do the modifications but not so soon.

mikeingold:
This seems like it should be easy enough to re-implement. I’ll make an attempt at a PR for this.

juliohm:
Fixed on master.

PR #792, opened 2024-03-11
New parameterization function for Torus

mikeingold:
Status: work in progress.

This PR aims to address issue #789 where current parameterization function for Torus fails when `major<minor`.

The commits added so far seem to produce a valid parameterization of a `Torus`, but results disagree with the previous function. The difference seems to simply be with respect to alignment of the domains, where both produce valid results. I'll be investigating further. If this turns out to be the case, we can either
- Perform another domain transformation on `u` and `v` to match old results: non-breaking but this transform might seem very arbitrary in the future.
- Leave results as-is and update unit tests to reflect changes: requires a breaking change.


juliohm:
Thank you @mikeingold , that looks much simpler. Appreciate if you can update the tests accordingly. We can make it a breaking release after it is merged.

mikeingold:
I just updated the tests to reflect the new parameterization outputs and all tests pass on my local machine.

juliohm (approved):

juliohm:
Thank you @mikeingold !

PR #795, opened 2024-03-13
WIP: Implementing a general parameterization for `Cylinder`

mikeingold:
Changes (checked box indicates implemented and passes tests):
- [x] Parameterization function for `Cylinder`
- [x] New helper function `_hassaferadius(::Cylinder)` to validate that planes don't intersect
- [x] Method for `intersection(f, ::Plane, ::Plane)`
- [x] Update to `axis(::Cylinder)` that eliminates intermediate allocation
- [x] Distance method for `evaluate(Euclidean(), ::Line, ::Line)`

### DIFF
--- a/src/distances.jl
+++ b/src/distances.jl
@@ -6,6 +6,37 @@
 evaluate(d::PreMetric, g::Geometry, p::Point) = evaluate(d, p, g)
 
 """
+    evaluate(Euclidean(), line1, line2)
+
+Evaluate the minimum Euclidean distance between `line1` and `line2`.
+"""
+function evaluate(::Euclidean, line1::Line, line2::Line)
+  # Inspired by: John Alexiou, Find shortest distance between lines in 3D
+  #   URL (version 2022-11-04) https://math.stackexchange.com/q/2217845
+
+  e1 = line1(1) - line1(0)
+  e2 = line2(1) - line2(0)
+  n = e1 × e2
+  nn = n ⋅ n
+  r = line2(0) - line1(0)
+
+  # Find the Point's on each line where they are closest
+  t1 = ((e2 × n) ⋅ r) / nn
+  t2 = ((e1 × n) ⋅ r) / nn
+  p1 = line1(t1)
+  p2 = line2(t2)
+
+  if iszero(nn)
+    # lines are parallel, so pick an arbitrary Point on line1 and
+    #   find distance from it to line2
+    return evaluate(Euclidean(), line1(0), line2)
+  else
+    # find distance between closest Point's
+    return evaluate(Euclidean(), p1, p2)
+  end
+end
+
+"""
     evaluate(Euclidean(), point, line)
 
 Evaluate the Euclidean distance between `point` and `line`.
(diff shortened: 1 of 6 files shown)


mikeingold:
Still working on this but looks like something is off with the current parameterization. It works for the end points of the axis but seems wrong for other points. It's also generating a `StackOverflowError` in `intersection(f, ::Line, ::Plane)` for the case I added to the unit tests :thinking:.

src/distances.jl
```diff
@@ -5,6 +5,37 @@
 # flip arguments so that points always come first
 evaluate(d::PreMetric, g::Geometry, p::Point) = evaluate(d, p, g)
 
+"""
+    evaluate(Euclidean(), line1, line2)
+
+Evaluate the minimum Euclidean distance between `line1` and `line2`.
+"""
+function evaluate(::Euclidean, line1::Line, line2::Line)
```
juliohm, line 13: Perhaps this method can leverage the `intersectparameters` utility?

juliohm (commented):
Should we split this PR into smaller ones? I believe it touches different parts of the codebase:

1. new intersection between line-line and line-plane
2. fixes to Cylinder and CylinderSurface

mikeingold:
> Should we split this PR into smaller ones? I believe it touches different parts of the codebase:
> 
>     1. new intersection between line-line and line-plane
> 
>     2. fixes to Cylinder and CylinderSurface

If you feel really strongly about PR's being narrowly-defined, I can try to split this one but it will probably take some time. As-is, this PR does roll in updates that touch different parts of the code base, but all the updates are in service of the primary goal of parameterizing `Cylinder`. The helper function is intended to be used for checking a `::Cylinder`, either at time of construction or utilization, to ensure that the parameterization doesn't have inverted line segments due to intersecting planes. The line-line distance and plane-plane intersection calculations are used to facilitate that determination and it made more sense to just implement those missing methods than calculate it internally as a one-off.

Thanks to some inspiration from your recent commits to master, I figured out that the `StackOverflowError` that was occurring was because the argument types were unintentionally altering the geometry types inside the function, and `intersect(f, ::Line{Float64}, ::Plane{Float32})` isn't currently defined since the `T`s aren't identical and they don't auto-promote.

I just got the parameterization function working and the tests passing. Let me know @juliohm how you want to proceed. I'll either need to start new branches/PR's, or just merge these changes into the latest master updates.

juliohm:
Hi @mikeingold , I think it is a good idea to split it up this PR into smaller ones. That way we can add more tests, reuse internal utilities such as `intersectparamaters` and review more carefully. It is also easy to undo if a new bug is found in the future.

mikeingold:
> Hi @mikeingold , I think it is a good idea to split it up this PR into smaller ones. That way we can add more tests, reuse internal utilities such as `intersectparamaters` and review more carefully. It is also easy to undo if a new bug is found in the future.

I'm splitting this into multiple PR's that I'll submit in sequence:
- Distance between two lines
- Intersection between two planes
- Helper function to check whether a `Cylinder` has intersecting planes
- Parameterization of a `Cylinder`

PR #797, opened 2024-03-14
format urls in docstrings

cserteGT3:
I think I found all urls, that were not formatted.
Sorry for the two PRs,  I just realized, that I could have checked all.


juliohm (approved):
Thanks for fixing all these links! ❤️

PR #804, opened 2024-03-17
Add parameterization function to `Cylinder`

mikeingold:
Summary
- Adds a parameterization function to `Cylinder` that works for cylinders with non-parallel top and bottom planes
- Adds a helper function to inspect a `Cylinder` and determine whether its top and bottom planes will intersect
- ~~Updates the `axis(::Cylinder)` method to produce the desired `Line` directly, rather than transform the `Cylinder` into a `CylinderSurface` and then run the same code on that.~~

Feel free to change the name of the helper function if you can think of a better one.


mikeingold:
I just updated the tests to address the line that Codecov reported as missing coverage, plus a couple of other lines in the same file. Two of these new tests involving constructing a `Cylinder` with tuples instead of `Point`s currently fail. This seems like maybe a type issue where one is defaulting to `Float64` vs `Float32`, but my first attempt at addressing that didn't work. I'll take another look at it tomorrow, but might just need to drop that change from this PR if I can't resolve quickly since it's not central to this PR's focus, just a nice-to-have while I was already adding tests for this source file.

juliohm:
We adopted a convention in which the methods are implemented in the `*Surface` type when possible to avoid code repetition. In that sense, we should implement the methods in `CylinderSurface` and call in `Cylinder`.

To check whether the top and bottom planes intersect, we can

```julia
intersectsplanes(c) = isnothing(c.top ∩ c.bot) && radius_check
```

where `radius_check` is the one involving the distance to the cylinder axis.

mikeingold:
> To check whether the top and bottom planes intersect, we can
> 
> ```julia
> intersectsplanes(c) = isnothing(c.top ∩ c.bot) && radius_check
> ```
> 
> where `radius_check` is the one involving the distance to the cylinder axis.

I'm not sure I'm following this exactly. How about I split the functionality into something like
- `intersectradius(c)` which yields the radius at which the intersection occurs (`Inf` if planes are parallel)
- `hasintersectingplanes(c) = c.radius > intersectradius(c)` to determine whether the planes do intersect for this cylinder

mikeingold:
~~Are you able to re-trigger the CI run @juliohm? Looks like the tests all pass, but 3 of the CI runs seem to have just failed to run at all with no error output.~~ I believe this PR is otherwise good-to-go pending any feedback you might have.

src/primitives/cylinder.jl
```diff
@@ -59,7 +59,36 @@ axis(c::Cylinder) = axis(boundary(c))
 
 isright(c::Cylinder) = isright(boundary(c))
 
+intersectradius(c::Cylinder) = intersectradius(boundary(c))
+
```
juliohm:
```suggestion
```

juliohm:
Hi @mikeingold , added you to the meshes team so that you can re-trigger these bots in the future.

If the tests pass here, we can go ahead and merge.

Issue #780, opened 2024-02-20
Measure of CylinderSurface assumes right-cylinder


mikeingold:
The `CylinderSurface` type currently imposes no requirement that the top and bottom planes be parallel.

https://github.com/JuliaGeometry/Meshes.jl/blob/82f5921d94742ee082f1fcecc5af49b10ceccae0/src/primitives/cylindersurface.jl#L27-L31

However, the current `measure` method for this type seems to assume that `isright(c)` is true. The following expression seems to be mathematically-equivalent to $(2\pi r ) h + 2 (\pi r^2)$, i.e. the cylinder walls plus two circular disks.

https://github.com/JuliaGeometry/Meshes.jl/blob/82f5921d94742ee082f1fcecc5af49b10ceccae0/src/measures.jl#L53-L58

As discussed on Zulip, ongoing work in MeshIntegrals.jl may be able to calculate the `measure` for the general case here, but isn't currently ready for prime-time. A short-term fix might be to check `isright` and generate an error condition otherwise, unless somebody can whip-up a closed-form expression for this general case, but I'm not sure how Meshes users feel about not-implemented-style errors.

juliohm:
@mikeingold your recent work on Cylinder/CylinderSurface provides a fix for this or we really need numerical integration in the oblique case?

mikeingold:
I suspect that this existing symbolic function actually works correctly for right cylinders and also cylinders with parallel planes.

For the case of non-parallel planes, numerical integration could be a valid, if expensive, fallback method. There might also be a closed-form general solution to calculate this directly, if someone wants to work on deriving it.

mikeingold:
I had a few spare minutes this evening to think about this more. I think I was able to sketch up the math for how to compute this in the general case without integration so long as the planes don’t intersect.

mikeingold:
I'm going to have to re-work #806 slightly, but I think have a closed-form solution/algorithm that works in all but the intersecting-planes case (inspired by [https://mathworld.wolfram.com/CylindricalSegment.html](https://mathworld.wolfram.com/CylindricalSegment.html))

For `V = measure(c::Cylinder)`
1. If `c` is a right-cylinder, then use the current solution: $V = h ~ \pi r^2$
2. If `c` has intersecting planes ... throw a error indicating this case is currently unimplemented.
3. Otherwise
[...]

PR #806, opened 2024-03-18
WIP: Generalized `measure(::Cylinder)`

mikeingold:
Addresses #780


mikeingold:
I need to think about this one a little more. I think this version makes an implicit assumption about the alignment of the planes that won't always be valid.

mikeingold:
I'm going to stop work on this branch for now. As described in #780, there turned out to be some cases where the algorithm I came up fails to work. Our current longer-term plan is to use numerical integration methods, like those currently in MeshIntegrals.jl, resulting in something like

```julia
function measure(c::Cylinder{T}) where {T}
    if isright(c)
        h = norm(axis(c)(T(1)) - axis(c)(T(0)))
        h * T(pi) * r^2
    else
        integral(p -> T(1), c)
    end
end
```

PR #829, opened 2024-04-11
Make `showsegments` observable in `Polygon` viz

eliascarv:


eliascarv:
Closed in favor of #831

PR #837, opened 2024-04-25
WIP: Implement macro for automatically defining commutative methods

mikeingold:
This change would eliminate the generic argument-switching method that causes `StackOverflow`s to occur for non-implemented methods.

### DIFF
--- a/src/intersections.jl
+++ b/src/intersections.jl
@@ -97,7 +97,6 @@ return types, Julia is able to optimize the branches of the code
 and generate specialized code. This is not the case when
 `f === identity`.
 """
-intersection(f, g₁, g₂) = intersection(f, g₂, g₁)
 intersection(g₁, g₂) = intersection(identity, g₁, g₂)
 
 # ----------------
--- a/src/intersections/domains.jl
+++ b/src/intersections/domains.jl
@@ -11,7 +11,7 @@
   end
 end
 
-@commutativef intersection(f, dom::Domain, pset::PointSet) = intersection(f, Multi(collect(dom)), pset)
+intersection(f, dom::Domain, pset::PointSet) = intersection(f, Multi(collect(dom)), pset)
 
 function intersection(f, dom₁::Domain{Dim,T}, dom₂::Domain{Dim,T}) where {Dim,T}
   # loop over all geometries
--- a/src/intersections/domains.jl
+++ b/src/intersections/domains.jl
@@ -2,7 +2,7 @@
 # Licensed under the MIT License. See LICENSE in the project root.
 # ------------------------------------------------------------------
 
-function intersection(f, geom::Geometry, pset::PointSet)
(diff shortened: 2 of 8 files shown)


mikeingold:
@juliohm I need to work out a git merge conflict but the new macro seems to work well. I’m planning to add tests to verify that deliberately-unsupported intersecting methods throw method errors and not stack overflows. Let me know what you think of the idea.

juliohm:
@mikeingold it is an interesting idea. Does it handle methods with repeated types? Say `f(a1::A, a2::A)`? In this case the attempt to commute the function will lead to method redefinition. Did you check that in the macro?

mikeingold:
The macro doesn't currently act on the argument types, the current logic is
basically:
- make sure the provided expression is a regular or generic function
definition
- locate the sub-expression containing the call structure, i.e. `f(a::A,
b::B)`
- duplicate the original expression but then swap the two args to transform
it to the form `f(b::B, a::A)`
- return both the original method and the commuted method definitions

There's actually two slightly-different versions of the macro added in this
PR. I wrote the ***@***.***` macro to perform exactly the above
transformation, but then realized that `intersection` actually has the
structure `intersection(f, a, b)`. Because of this, I added a variant
***@***.***` that performs the same transformation in the presence of an
extra leading argument.

Explicitly checking for this same-typed condition is possible but might not
be worth the added complexity. The macro would be a non-exported utility
with a fairly straight-forward docstring explaining that the functionality
is limited to transforming `f(a::A, b::B) -> f(a::A, b::B); f(b::B, a::A)`.
If you're writing a new `intersection` method and it has the same arg
types, then no additional commutative method definition would be required
since it's already commutative by definition. If you apply the macro
anyway, you'll get a precompilation error during testing with a message
indicating that its usage caused a re-definition.

I think the trickier possible issue might be method ambiguities. Applying
this to `intersection`, I went through and tagged all methods with
different first and second argument types, but found in testing that this
generated an ambiguity for `intersection(f, ::PointSet, ::PointSet)` since
apparently both `intersection(f, ::Domain, ::PointSet)` and
`intersection(f, ::PointSet, ::Domain)` could be applied to it. I don't
have much experience with those types, so I wanted to ask your opinion on
that particular case.

juliohm:
Exactly, my main concern is related to dispatch with arguments that are
equal or subtypes of each other. The current design seems safer in the
sense that we have full control where to remove ambiguity.

The idea in this PR is worth exploring. Maybe we can get a working version
to assess the benefits? If it is taking more time than what you expected,
no worries. Macros are always tricky to get right and we avoid them as much
as possible. If it addresses the specific  issue here in a safe manner, we
can take a closer look.

Em qui., 25 de abr. de 2024, 10:16, Michael Ingold ***@***.***>
escreveu:

juliohm:
@mikeingold I am closing this given that the new master branch is completely refactored, and that we didn't find much benefit in the macro so far.

PR #838, opened 2024-04-25
Export and document inverse function for transforms

cserteGT3:
As discussed on zulip, this PR:
* exports `inverse`
* also mentions the function in the docs

I'm not sure about the docs, but giving that I was looking for it and couldn't figure out myself, I am strongly in favor of putting it somewhere somehow.

### DIFF
--- a/docs/src/transforms.md
+++ b/docs/src/transforms.md
@@ -14,6 +14,13 @@ GeometricTransform
 CoordinateTransform
 ```
 
+Inverse transforms that are `isinvertible` can be computed with the `inverse` function
+(reexported by Meshes.jl).
+
+```@docs
+inverse
+```
+
 ## Rotate
 
 ```@docs
--- a/docs/src/transforms.md
+++ b/docs/src/transforms.md
@@ -14,11 +14,12 @@ GeometricTransform
 CoordinateTransform
 ```
 
-Inverse transforms that are `isinvertible` can be computed with the `inverse` function
-(reexported by Meshes.jl).
+Some transforms have an inverse that can be created with the [`inverse`](@ref) function.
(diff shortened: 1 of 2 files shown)


juliohm:
Thank you! We will see what is the best place in the stack to reexport it. The function is introduced in TransformsBase.jl, so maybe we need to export it there.

juliohm:
@cserteGT3 we didn't forget about this PR. Will take a look into it right after we finish some downstream updates for the new release of the project.

docs/src/transforms.md
```diff
@@ -14,6 +14,13 @@ GeometricTransform
 CoordinateTransform
 ```
 
+Inverse transforms that are `isinvertible` can be computed with the `inverse` function
+(reexported by Meshes.jl).
```
juliohm:
```suggestion
Some transforms have an inverse that can be created with the [`inverse`](@ref) function.
The function [`isinvertible`](@ref) can be used to check if a transform is invertible.
```

docs/src/transforms.md
```diff
@@ -14,6 +14,13 @@ GeometricTransform
...
+
+```@docs
+inverse
```
juliohm, line 21:
```suggestion
inverse
isinvertible
```

src/Meshes.jl
```diff
@@ -509,6 +509,7 @@ export
   LaplaceSmoothing,
   TaubinSmoothing,
   isaffine,
+  inverse,
```
juliohm, line 513:
```suggestion
  isinvertible,
  inverse,
```

juliohm (approved):

eliascarv (approved):

juliohm:
Thank you @cserteGT3 ❤️

Issue #839, opened 2024-04-26
Miscellaneous issues to fix after major CRS refactor


juliohm:
## Urgent:

- [x] `normal` must be reviewed to return normalized normals for all geometries
- [x] review the role of `atol` in `Vec` `isapprox` comparisons
- [x] default tolerances in `PartitionMethod` should be based on `atol`
- [x] `show` of `BezierCurve` should be incorporated into generic `show`
- [x] review fields of structs to make sure they all have units when appropriate
- [x] review occurrences of `zero(lentype(p))^2`
- [x] `coordinates` returns a `Vec`, but should return the coordinates instead
- [x] fix occurrences of `Point(Vec(...))` in some algorithms

## Not urgent:

- [x] `rand` of `Point` should produce concrete types
- [x] `to` should work with any crs, converting to `Cartesian`
- [x] more flexible crs in `Grid` subtypes
- [x] propagate crs information in some algorithms
- [x] fix `Cone` and `ConeSurface` constructors with non-unitful args
- [x] review constructors of `Point`, `Vec` and `CRS` with `Tuple` and `SVector`
- [x] review distance interface in distances.jl
- [x] replace `@assert` by appropriate exception type
- [x] review occurrences of `SVector` in fields

cserteGT3:
Just out of curiosity can I ask why was `@assert` replaced by the `assertion` function?

juliohm:
From the docstring of `@assert`, the macro is not intended for logical checks in implementations, but merely for debugging purposes. One can disable it with runtime flags, but we don't want that. We are making sure that the `AssertioError` is thrown if a condition is not satisfied.

cserteGT3:
Thank you for the explanation!

juliohm:
All fixed on master.

PR #861, opened 2024-06-03
Add support for using custom `Datum` in `RectilinearGrid` and `StructuredGrid`

eliascarv:
Part of #839

### DIFF
--- a/src/boundingboxes.jl
+++ b/src/boundingboxes.jl
@@ -73,11 +73,9 @@ boundingbox(g::CartesianGrid) = Box(extrema(g)...)
 
 boundingbox(g::RectilinearGrid) = Box(extrema(g)...)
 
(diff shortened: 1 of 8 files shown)


src/boundingboxes.jl
```diff
@@ -76,7 +76,7 @@ boundingbox(g::RectilinearGrid) = Box(extrema(g)...)
 boundingbox(g::TransformedGrid{Dim,<:CartesianGrid{Dim}}) where {Dim} =
   boundingbox(parent(g)) |> transform(g) |> boundingbox
 
-boundingbox(g::TransformedGrid{Dim,<:RectilinearGrid{Dim}}) where {Dim} =
+boundingbox(g::TransformedGrid{Dim,<:RectilinearGrid{Datum,Dim}}) where {Datum,Dim} =
```
juliohm: I believe that we can omit the type parameters in these methods, since they are not used.

juliohm (approved):

PR #885, opened 2024-06-13
Add `Shadow` transform

### LINKED ISSUE #884: Add `Shadow` transform to project onto lower-dimensional spaces
Given a 3D domain, we should be able to quickly produce shadows with `Shadow(:xy)`, `Shadow("yz")`, etc. This is not a `CoordinateTransform` because it changes the underlying topology. For instance, a 3D CartesianGrid can become 2D.

eliascarv:
Fix #884


src/transforms/shadow.jl
```diff
@@ -0,0 +1,82 @@
...
+"""
+    Shadow(dims)
+
+Project the geometry or domain onto the given `dims`,
+producing a "shadow" of the original object.
+
+## Examples
+
+```julia
+Shadow(:xy)
```
juliohm, line 26: missing examples with other types of dims

src/transforms/shadow.jl
```diff
@@ -0,0 +1,82 @@
...
+Shadow("xz")
+```
+"""
+struct Shadow{Dim} <: GeometricTransform
+  dims::Dims{Dim}
+end
+
+Shadow(dims::Int...) = Shadow(dims)
+
+function _index(d)
```
juliohm: should move to before the Shadow struct as an auxiliary function

src/transforms/shadow.jl
```diff
@@ -0,0 +1,82 @@
...
+
+Shadow(dims::AbstractString) = Shadow(Dims(_index(d) for d in dims))
+
+Shadow(dims::Symbol) = Shadow(string(dims))
+
+parameters(t::Shadow) = (; dims=t.dims)
+
+apply(t::Shadow, v::Vec) = _shadow(v, _sorteddims(t.dims)), nothing
+
+apply(t::Shadow, g::GeometryOrDomain) = _shadow(g, _sorteddims(t.dims)), nothing
```
juliohm: _sorteddims --> _sort

juliohm (approved):

PR #905, opened 2024-06-20
Fix `VoronoiTesselation` implementation

eliascarv:


src/tesselation/voronoi.jl
```diff
@@ -20,13 +20,19 @@ end
...
   coords = map(p -> ustrip.(to(p)), pset)
   triang = triangulate(coords, rng=method.rng)
   vorono = voronoi(triang, clip=true)
 
   # mesh with all (possibly unused) points
-  points = get_polygon_points(vorono)
+  points = map(get_polygon_points(vorono)) do (x, y)
+    coords = Cartesian{d}(x * u, y * u)
+    Point(coords)
+  end
```
juliohm, line 35: this assumes Cartesian, we need something more general

juliohm:
Merging this in the meantime. We need a more general solution in a separate PR.

PR #909, opened 2024-06-27
Fix `point in polyarea`

eliascarv:


juliohm (approved):

juliohm:
This PR fixes  the behavior, but introduces a regression in terms of memory usage. We probably need a separate method just for `LatLon` coordinates, including `GeodeticLatLon`, `GeocentricLatLon` and `AuthalicLatLon`.

test/winding.jl
```diff
@@ -11,6 +11,17 @@
   @test winding(p, reverse(c)) ≈ T(-2)
   @test winding([p, p], c) ≈ T[2, 2]
 
+  p = Point(LatLon(T(0.5), T(0.5)))
```
juliohm, line 14: perhaps we should introduce a `latlon` helper function in our tests, along with a `pointlatlon(lat, lon) = Point(latlon(lat, lon))`?

test/winding.jl
```diff
@@ -11,6 +11,17 @@
...
+  c = Ring(Point.([LatLon(T(0), T(0)), LatLon(T(1), T(0)), LatLon(T(1), T(1)), LatLon(T(0), T(1))]))
+  @test winding(p, c) ≈ T(1)
+  @test winding(p, reverse(c)) ≈ T(-1)
+  @test winding([p, p], c) ≈ T[1, 1]
+
+  # error: both arguments must have the same CRS
+  p = Point(LatLon(T(0.5), T(0.5)))
+  c = Ring(point.([(0, 0), (1, 0), (1, 1), (0, 1), (0, 0), (1, 0), (1, 1), (0, 1)]))
```
juliohm, line 22: similarly, `point` is an alias to `Point` with `Cartesian` coords in the tests. Maybe we can explore the idea of introducing a `cart` helper with the appropriate machine type, and a `pointcart(x, y) = Point(cart(x, y))` to make it more explicit for readers.

juliohm (approved):
I will merge this one, and we can work on separate PRs to improve the helper functions in tests.

PR #926, opened 2024-07-09
Fix constructor of `PointSet` taking `Vector{Vector}`

JoshuaLampert:
Since `Point([1,2,3])` fails with 
```julia
julia> Point([1,2,3])
ERROR: MethodError: no method matching (CoordRefSystems.Cartesian{CoordRefSystems.NoDatum})(::Vector{Int64})
[...]
```
also the constructor calls `PointSet([[1,2,3], [4,5,6]])` and `PointSet([1,2,3], [4,5,6])`, which are suggested in the docs, fail. This PR fixes this issue and adds tests for this.

### DIFF
--- a/src/refinement/trisubdivision.jl
+++ b/src/refinement/trisubdivision.jl
@@ -3,15 +3,15 @@
 # ------------------------------------------------------------------
 
 """
-  TriSubdivision()
+    TriSubdivision()
 
 Refinement of a mesh by preliminarly triangulating it if needed and
 then subdividing each triangle into four triangles.
 
 ## References
 
-* Charles Loop. 1987. [Smooth subdivision surfaces based on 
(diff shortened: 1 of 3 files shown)


src/refinement/trisubdivision.jl
```diff
@@ -3,15 +3,15 @@
 # ------------------------------------------------------------------
 
 """
-  TriSubdivision()
+    TriSubdivision()
```
JoshuaLampert, line 6: This is unrelated, but I noticed that this is not rendered nicely in the docs.

juliohm:
That is really good catch @JoshuaLampert , thank you. We will merge the PR, and review if this method should actually exist. It looks like a left-over from the major refactoring we did last months.

PR #941, opened 2024-07-16
Refactor the implementation of the `rand` function

eliascarv:


src/Meshes.jl
```diff
@@ -62,6 +62,9 @@ include("vectors.jl")
 # geometries
 include("geometries.jl")
 
+# rand implementation
+include("rand.jl")
```
juliohm: this should go in the algorithms section

src/rand.jl
```diff
@@ -0,0 +1,138 @@
...
+# HELPER FUNCTIONS
+# -----------------
+
+_pointvec(rng::Random.AbstractRNG, CRS, n) = [_rand(rng, Point, CRS) for _ in 1:n]
+
+_pointtup(rng::Random.AbstractRNG, CRS, n) = ntuple(_ -> _rand(rng, Point, CRS), n)
+
+_randvec(rng::Random.AbstractRNG, CRS) = rand(rng, Vec{CoordRefSystems.ndims(CRS),Met{Float64}})
+
+_randlen(rng::Random.AbstractRNG) = rand(rng, Met{Float64})
```
juliohm: _rvector, _rtuple, _rvec, _rlen

docs/src/rand.md
```diff
@@ -0,0 +1,42 @@
...
+It's possible to change the CRS using the `crs` keyword argument:
+
+```@example rand
+rand(Point, crs=Cartesian2D)
+```
+
+A vector of geometries can be generated by passing the number of elements as the second argument:
+
+```@example rand
+rand(Line, 5, crs=LatLon)
```
juliohm:
```suggestion
rand(Segment, 5, crs=LatLon)
```

juliohm (approved):

PR #947, opened 2024-07-19
Add support for non-Euclidean manifolds

### LINKED ISSUE #116: Parametrize on the metric space
Originating from #98, it was discussed that we probably should parametrize the library according to a particular metric. Even if not all algorithms may work with arbitrary metrics, it would surely be nice to just have to set a global parameter or package option to have some algorithms work with non-Euclidean metrics.

Concretely, I think we need to:
1. implement custom metrics or reuse those from Distances.jl
2. expose a metric parameter
3. replace all occurrences of distance calculations to use the metric directly
4. find everything that breaks because of implicit assumptions on an Euclidean metric (e.g., that two parallel lines stay parallel *ad vitam eternam*) and deal with it

I am in favor of adding custom metrics with a custom type hierarchy since `Distances.jl` was made for a particular use of optimizing distance computations between iterators. A first PR could take care of that, and export them as part of the API.
2. can be as simple as `const metric = Ref{MetricType}(Euclidean())`. It will need extra thinking to make sure performance stays as good as before the parametrization.

But there is more. The introduction of non-Euclidean metrics gives rise to non-Euclidean geometry. From [Wikipedia](https://en.wikipedia.org/wiki/Non-Euclidean_geometry):

> non-Euclidean geometry arises by either relaxing the metric requirement, or replacing the parallel postulate with an alternative

where the parallel postulate says that two parallel lines stay parallel (not true on, e.g., spherical geometry where they must intersect at some point).

For example, in spherical geometry, a line is a circle and a plane is a sphere.

Because of that, 4. is probably the tricky part: some algorithms may not work with weird metrics (I'm thinking of the line intersection algorithm, for example, and of any algorithm that assumes that the angles of a triangle sum up to 180° - which is for spherical geometry, a "nice" metric).

However, I think we can, for the moment, just ignore these non-Euclidean geometries. We'd need to clearly state what is supposed to work with custom metrics, and add support for what doesn't later. I am convinced that there is a clean and efficient way to deal with this, we'll just need some time to figure it out.

juliohm:
Fix #116


src/primitives/point.jl
```diff
@@ -33,16 +33,25 @@ Point(1u"m", 2u"m", 3u"m") # integer is converted to float by design
...
 end
 
+Point{M}(coords::C) where {M<:AbstractManifold,C<:CRS} = Point{M,C}(coords)
+
+Point(coords::CRS) = Point{_manifold(coords)}(coords)
+
+_manifold(coords::CRS) = 𝔼{CoordRefSystems.ndims(coords)}
+_manifold(::LatLon) = 🌐
+_manifold(::GeocentricLatLon) = 🌐
+_manifold(::AuthalicLatLon) = 🌐
```
juliohm: let's move these methods to the begin or end of file

src/primitives/paraboloidsurface.jl
```diff
@@ -86,7 +86,7 @@ apex(p::ParaboloidSurface) = p.apex
 Return the focal axis, connecting the focus with the apex of the paraboloid.
 The axis is always aligned with the z direction.
 """
-axis(p::ParaboloidSurface{P,ℒ}) where {P,ℒ} = Line(p.apex, p.apex + Vec(ℒ(0), ℒ(0), p.focallength))
+axis(p::ParaboloidSurface{M,C,ℒ}) where {M,C,ℒ} = Line(p.apex, p.apex + Vec(ℒ(0), ℒ(0), p.focallength))
```
juliohm:
```suggestion
axis(p::ParaboloidSurface{C,Mₚ,ℒ}) where {C,Mₚ,ℒ} = Line(p.apex, p.apex + Vec(ℒ(0), ℒ(0), p.focallength))
```

src/primitives/ray.jl
```diff
@@ -9,8 +9,8 @@ A ray originating at point `p`, pointed in direction `v`.
 It can be called as `r(t)` with `t > 0` to cast it at
 `p + t * v`.
 """
-struct Ray{C<:CRS,V<:Vec} <: Primitive{C}
-  p::Point{C}
+struct Ray{M<:𝔼,C<:CRS,V<:Vec} <: Primitive{M,C}
+  p::Point{M,C}
```
juliohm: please follow the same pattern of other geometries: p with Mp and Ray with M = E{Dim}

src/primitives/point.jl
```diff
@@ -100,13 +104,20 @@ at a reference (or start) point `A`.
...
 ≥(A::Point, B::Point) = all(x -> x ≥ zero(x), A - B)
 <(A::Point, B::Point) = all(x -> x > zero(x), B - A)
 >(A::Point, B::Point) = all(x -> x > zero(x), A - B)
 
+≤(A::Point{🌐}, B::Point{🌐}) = _lat(A) ≤ _lat(B)
+≥(A::Point{🌐}, B::Point{🌐}) = _lat(A) ≥ _lat(B)
+<(A::Point{🌐}, B::Point{🌐}) = _lat(A) < _lat(B)
+>(A::Point{🌐}, B::Point{🌐}) = _lat(A) > _lat(B)
+
+_lat(P) = convert(LatLon, P.coords).lat
```
juliohm: move this down to the paragraph of helper functions, and implement it in terms of the coords

ext/grid.jl
```diff
@@ -22,7 +28,7 @@ function vizgrid!(plot, ::Val{3}, ::Val{3})
   if nv[] == nc[]
     error("not implemented")
```
juliohm: isn't this case covered by the vizmesh! branch too?

ext/MeshesMakieExt.jl
```diff
@@ -44,8 +44,7 @@ include("utils.jl")
 
 # viz recipes
 include("grid.jl")
-include("simplemesh.jl")
-include("subcartesiangrid.jl")
+include("mesh.jl")
```
juliohm, line 46: mesh.jl then grid.jl

ext/fallbacks.jl
```diff
@@ -9,12 +9,11 @@ Makie.plottype(::AbstractVector{<:Vec}) = Viz{<:Tuple{AbstractVector{<:Vec}}}
...
 Makie.convert_arguments(::Type{<:Viz}, geom::Geometry) = (GeometrySet([geom]),)
 Makie.convert_arguments(::Type{<:Viz}, domain::Domain) = (GeometrySet(collect(domain)),)
-Makie.convert_arguments(::Type{<:Viz}, mesh::Mesh) = (convert(SimpleMesh, mesh),)
 Makie.convert_arguments(::Type{<:Viz}, vec::Vec) = ([vec],)
 
 # skip conversion for these types
 Makie.convert_arguments(::Type{<:Viz}, gset::GeometrySet) = (gset,)
-Makie.convert_arguments(::Type{<:Viz}, mesh::SimpleMesh) = (mesh,)
+Makie.convert_arguments(::Type{<:Viz}, mesh::Mesh) = (mesh,)
 Makie.convert_arguments(::Type{<:Viz}, grid::Grid) = (grid,)
```
juliohm: Mesh, Grid, GeometrySet

ext/fallbacks.jl
```diff
@@ -9,12 +9,11 @@ Makie.plottype(::AbstractVector{<:Vec}) = Viz{<:Tuple{AbstractVector{<:Vec}}}
...
 Makie.convert_arguments(::Type{<:Viz}, grid::SubCartesianGrid) = (grid,)
```
juliohm: Do we need this still?

eliascarv (approved):

PR #963, opened 2024-07-28
Update mention of ExactPredicates.jl to also mention AdaptivePredicates.jl

DanielVandH:
I updated this a while ago to mention ExactPredicates.jl, so it's probably also appropriate to mention AdaptivePredicates.jl. I also reworded the sentence a bit


juliohm (approved):

juliohm:
Nice improvement!

PR #965, opened 2024-07-29
Use julia-format action

JoshuaLampert:
As discussed in https://github.com/JuliaGeometry/Meshes.jl/pull/964#issuecomment-2254636469 this uses the julia-format action, which makes suggestions in a PR based on the formatter instead of creating PRs. This should reduce the noise and guarantees that no non-formatting PRs are merged.
Note, however, that clickable suggestions in the PR are only made for PRs that don't come from forks. For PRs from forks only warnings with the corresponding diff from the formatter are shown (due to privilege issues).


JoshuaLampert:
Hm, I wanted to check if the annotations work, but unfortunately it looks like there are no annotations. I don't know why, especially because reviewdog [promised to do so](https://github.com/JuliaGeometry/Meshes.jl/actions/runs/10147851489/job/28059227207?pr=965#step:2:509):
>  reviewdog: This GitHub Token doesn't have write permission of Review API [1],
so reviewdog will report results via logging command [2] and create annotations similar to
github-pr-check reporter as a fallback.

src/Meshes.jl
```diff
@@ -49,7 +49,7 @@ import TransformsBase: parameters, preprocess
 import CoordRefSystems: lentype
 
 # unit utils
-include("units.jl")
+  include("units.jl")
```
juliohm: Is this an accidental change @JoshuaLampert ?

.github/workflows/Format-check.yml
```diff
@@ -0,0 +1,11 @@
+name: Format suggestions
```
juliohm: Should we simply replace the FormatPR.yml action?

src/Meshes.jl
JoshuaLampert: No, this was on purpose to see if the suggestions work. See my comment above. I would have hoped that there appears a suggestion reverting this change.

.github/workflows/Format-check.yml
JoshuaLampert: I don't understand. This PR already replaces it, doesn't it?

src/Meshes.jl
juliohm: I see it now, good idea.

.github/workflows/Format-check.yml
juliohm: Maybe we should just replace the contents of the FormatPR.yml file instead of creating a new filename.

JoshuaLampert: Ah, yes, I can do that. I read the name "FormatPR.yml" more as "create a format PR" rather than "do format a PR", which is why I thought this name is not appropriate anymore, but now that I think about it, the name is fine, I think. See [3775f44](https://github.com/JuliaGeometry/Meshes.jl/pull/965/commits/3775f44aca64e9d3c64484d81559893290b33177).

juliohm:
@JoshuaLampert I see the format error showing up nicely.

Should we fix the code style and see the action turning green again?

JoshuaLampert:
> @JoshuaLampert I see the format error showing up nicely.

...wait what exactly do you see? Do you see an annotation like here?
https://github.com/Astroshaper/AsteroidThermoPhysicalModels.jl/pull/107/files#diff-a17b585bea8f2c467215bd1821733e1580649126c74de4ff139ce6f019964762

Because I don't see that, but I would like to see it.

juliohm:
When I click on the Action that failed with the red cross mark, I see this:

image

It is bit hidden though.

JoshuaLampert:
Yes, I would have hoped for a bit more visibility and easier to read suggestions. In principle this should work like in the PR I linked above, but I don't know why this is not the case here. Maybe the bot needs some privileges?
But if you're fine with the current version, I can fix the code style again and we can merge.

juliohm:
I think this is already an improvement. We can experiment the new action and figure out missing permissions later. I will fix the code style and wait the tests again.

src/Meshes.jl
```suggestion
include("units.jl")
```

PR #971, opened 2024-08-01
Refactor `RectilinearGrid`

eliascarv:


src/mesh/rectilineargrid.jl
```diff
@@ -20,36 +20,36 @@ julia> y = [0.0, 0.1, 0.3, 0.7, 0.9, 1.0]
...
+  Dim = length(xyz)
+  C = Cartesian{NoDatum,Dim,Met{Float64}}
+  RectilinearGrid{𝔼{Dim},C}(args...)
+end
 
-RectilinearGrid(args...) = RectilinearGrid{NoDatum}(args...)
+RectilinearGrid(xyz...) = RectilinearGrid(xyz)
 
-vertex(g::RectilinearGrid{Datum}, ijk::Dims) where {Datum} = Point(Cartesian{Datum}(getindex.(g.xyz, ijk)))
+vertex(g::RectilinearGrid{M,C}, ijk::Dims) where {M,C} = Point(CoordRefSystems.reconstruct(C, getindex.(g.xyz, ijk)))
```
juliohm: let's break this into multiple lines, one defining `raw = getindex....` another doing reconstruction, and finally the point

src/mesh/rectilineargrid.jl
```diff
@@ -60,13 +60,13 @@ function centroid(g::RectilinearGrid, ind::Int)
   withcrs(g, (to(p1) + to(p2)) / 2)
 end
 
-function Base.getindex(g::RectilinearGrid{Datum}, I::CartesianIndices) where {Datum}
+function Base.getindex(g::RectilinearGrid{M,C}, I::CartesianIndices) where {M,C}
```
juliohm, line 72: we don't need M,C here

src/mesh/rectilineargrid.jl
juliohm: no need for M,C here either, you can crs(g) inside the reconstruct

src/mesh/rectilineargrid.jl
```diff
@@ -20,36 +20,45 @@ julia> y = [0.0, 0.1, 0.3, 0.7, 0.9, 1.0]
...
+  RectilinearGrid{𝔼{Dim},C}(coords)
+end
 
-vertex(g::RectilinearGrid{Datum}, ijk::Dims) where {Datum} = Point(Cartesian{Datum}(getindex.(g.xyz, ijk)))
+RectilinearGrid(xyz...) = RectilinearGrid(xyz)
+
+function vertex(g::RectilinearGrid, ijk::Dims)
+  C = crs(g)
+  raw = getindex.(g.xyz, ijk)
+  coords = CoordRefSystems.reconstruct(C, raw)
```
juliohm, line 57:
```suggestion
  raw = getindex.(g.xyz, ijk)
  coords = CoordRefSystems.reconstruct(crs(g), raw)
```

PR #972, opened 2024-08-01
escape underscore in docstring link

disberd:
Simply escape underscore in docstring link to avoid markdown messing the displayed link

### DIFF
--- a/src/transforms.jl
+++ b/src/transforms.jl
@@ -36,7 +36,7 @@ end
     CoordinateTransform
 
 A method to transform the coordinates of objects.
-See [https://en.wikipedia.org/wiki/List_of_common_coordinate_transformations]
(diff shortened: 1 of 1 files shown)


src/transforms.jl
```diff
@@ -36,7 +36,7 @@ end
     CoordinateTransform
 
 A method to transform the coordinates of objects.
-See [https://en.wikipedia.org/wiki/List_of_common_coordinate_transformations]
+See [https://en.wikipedia.org/wiki/List\\_of\\_common\\_coordinate\\_transformations]
 (https://en.wikipedia.org/wiki/List_of_common_coordinate_transformations).
```
juliohm:
```suggestion
See <https://en.wikipedia.org/wiki/List_of_common_coordinate_transformations>.
```

Let's try to fix it with this alternative syntax.

Issue #974, opened 2024-08-02
Preserve order between input points and output polygons in VoronoiTessellation


disberd:
Hi,

I was playing around with the Voronoi tessellation and noticed that despite having one polygon generated for each point in the provided initial vector, the order of the output polygons is completely disconnected from the order of the input points.

I have checked the code internally and this would seem to be quite trivial to solve (either by default, or hopefully via a flag/kwarg to `tessellate` if for some reason an ordered output is considered not a good default).

The problem seems to stem from line 42 in the relevant `tessellate` function body:
https://github.com/JuliaGeometry/Meshes.jl/blob/05cbd2060d69941ba9094f6f629b82b4b5849093/src/tesselation/voronoi.jl#L34-L43

which simply calls `values(vorono.polygons)`, with `vorono.polygons` being a `Dict` with the indices of the starting points as `Key` and the indices of the polygon vertices as values.

Despite the order information being present the original `Dict`, the value iterator coupled with the unordered nature of `Dict` creates the specific issue.

Would you be willing to accpet a PR to either change the default to have ordered output or at least add a flag to enable it?

The solution I tried with a flag was to simply add the following two lines after line 43 in the code above
```julia
 if sorted
  	original_idxs = keys(vorono.polygons) |> collect
	invpermute!(tuples, original_idxs)
 end
```

In an actual PR it would be best to directly import and use the `each_polygon_index` function from the `DelaunayTriangulation` package rather than doing `keys(vorono.polygons)`.

I think this sorting will necessarily add a few allocations but `tessellate` already allocates quite a lot already and the impact in timing seemed to be negligible  from my very limited test

EDIT: preallocating the tuples vector and directly iterating over the dict pairs to assign the polygon vertices to the correct index will likely make this sorted output have no more allocations than the current approach

juliohm:
Hi @disberd this fix would be very welcome. It is good to retain order of points as much as possible.

juliohm:
Fixed on master.

PR #978, opened 2024-08-03
Make VoronoiTessellation respect input order and improve performance

disberd:
This PR ensures that the order of the polygons in the mesh returned by `tessellate(pts, VoronoiTessellation)` respects the order of the inputs points `pts`, closing #974.

In the process of modifying the function, I also tried some benchmarks and did some slight modifications to the code to reduce un-necessaries allocation. 

The function now preallocates a vector of Connectivity objects and fill it in the correct order inside a for loop.
One modification that significantly reduced allocations was to use the `connect(idxs, NGon)` function rather than the generic `connect(idxs)` one, as I believe you would always have NGons as output of the voronoi tessellation.

Another small improvements in allocations was to use `ntuple` instead of `Tuple(idxs)` to create the temporary indices tuple as the latter allocates (I guess due to type instability since it can't know the tuple length just with the type if called with a vector as input).

I also added a test to verify that each point in the input is `inside` the respective polygon based on index.
Unfortunately I realized that probably due to floating point precision, some of the points would not fall inside the respective polygons. In all these cases, the culprit point seems to be in principle just on the polygons border, and it's considered `OUT` just because of floating point precision.
For this reason I added the second condition to ignore the result if the point is actually not inside any of the polygons of the resulting mesh.

If you have ideas on how to test this better (I do not seem to have found a way to consider tolerance for the `in` function) I believe that would actually be the best approach.

## Benchmarks
I have performed some benchmarks in Julia 1.11-rc2 under windows with the following code:
```julia
using Meshes
using BenchmarkTools
using Random
using StableRNGs

randpoint2(n) = [Point(ntuple(i -> rand(Float64), 2)) for _ in 1:n]

bench(ns) = for n in ns
    Random.seed!(n) # Ensure same points are generated for each n
    pts = randpoint2(n)
    b = @benchmark tesselate($pts, VoronoiTesselation(StableRNG(2024)))
    println("Results for $n points:")
    println()
    display(b)
    println()
end
```

which gives the following results:

### Current version (v0.47.8)
```julia-repl
julia> bench([10, 100, 1000, 10000])
Results for 10 points:

BenchmarkTools.Trial: 10000 samples with 1 evaluation.
 Range (min … max):  197.500 μs …  12.322 ms  ┊ GC (min … max): 0.00% … 96.35%
 Time  (median):     206.600 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   229.758 μs ± 234.335 μs  ┊ GC (mean ± σ):  3.52% ±  3.66%

  ▅█▇▅▄▄▄▄▃▂▂▁▁▁▁▁                                              ▂
  █████████████████████▇▆▆▇▆▆▇▆▆▇▇▇▆▇▆▆▆▅▅▅▆▇▇▅▅▅▆▃▅▅▅▅▆▅▅▅▅▄▄▄ █
  198 μs        Histogram: log(frequency) by time        435 μs <

 Memory estimate: 146.22 KiB, allocs estimate: 1954.

Results for 100 points:

BenchmarkTools.Trial: 2157 samples with 1 evaluation.
 Range (min … max):  2.046 ms …  12.608 ms  ┊ GC (min … max): 0.00% … 79.06%
 Time  (median):     2.131 ms               ┊ GC (median):    0.00%
 Time  (mean ± σ):   2.311 ms ± 740.016 μs  ┊ GC (mean ± σ):  4.57% ±  9.95%

  █▇▅▄▃▂▁▁▁
  ██████████▇▇▇▆▆▅▃▅▃▅▅▅▅▅▅▁▅▃▁▃▁▁▁▁▁▁▁▁▁▁▃▁▁▁▁▁▁▁▁▁▁▁▄▅▇▇▅▅▅ █
  2.05 ms      Histogram: log(frequency) by time      6.07 ms <

 Memory estimate: 1.20 MiB, allocs estimate: 13821.

Results for 1000 points:

BenchmarkTools.Trial: 187 samples with 1 evaluation.
 Range (min … max):  22.325 ms … 44.705 ms  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     25.416 ms              ┊ GC (median):    0.00%
 Time  (mean ± σ):   26.812 ms ±  4.288 ms  ┊ GC (mean ± σ):  3.52% ± 6.40%

  ▇▅▄▃▆█▄█▂        ▂
  █████████▇▇█▃▇▆▆▇█▆▅▇▄▅▄▄▆▅▃▁▃▄▃▁▃▄▃▅▁▃▃▁▁▃▃▃▃▁▁▁▁▁▃▃▁▃▃▁▁▃ ▃
  22.3 ms         Histogram: frequency by time        40.9 ms <

 Memory estimate: 10.52 MiB, allocs estimate: 136009.

Results for 10000 points:

BenchmarkTools.Trial: 3 samples with 1 evaluation.
 Range (min … max):  2.066 s …  2.083 s  ┊ GC (min … max): 0.44% … 0.42%
 Time  (median):     2.069 s             ┊ GC (median):    0.42%
 Time  (mean ± σ):   2.073 s ± 9.602 ms  ┊ GC (mean ± σ):  0.29% ± 0.25%

  █        █                                             █
  █▁▁▁▁▁▁▁▁█▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁█ ▁
  2.07 s        Histogram: frequency by time        2.08 s <

 Memory estimate: 106.51 MiB, allocs estimate: 1360405.
```

### After this PR
```julia-repl
julia> bench([10, 100, 1000, 10000])
Results for 10 points:

BenchmarkTools.Trial: 10000 samples with 1 evaluation.
 Range (min … max):  143.500 μs …  17.926 ms  ┊ GC (min … max): 0.00% … 97.78%
 Time  (median):     150.100 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   167.550 μs ± 236.447 μs  ┊ GC (mean ± σ):  4.01% ±  3.31%

  ▆█▇▅▃▃▄▄▃▂▂▁▁▁                                                ▁
  ███████████████▇█▇▇▇▇▆▇▇▆▆▆▆▆▆▅▅▅▅▄▅▅▄▅▅▄▄▅▅▅▄▅▅▅▅▅▅▅▅▅▅▅▅▅▅▅ █
  144 μs        Histogram: log(frequency) by time        336 μs <

 Memory estimate: 124.20 KiB, allocs estimate: 1476.

Results for 100 points:

BenchmarkTools.Trial: 2602 samples with 1 evaluation.
 Range (min … max):  1.637 ms …  19.856 ms  ┊ GC (min … max): 0.00% … 89.87%
 Time  (median):     1.711 ms               ┊ GC (median):    0.00%
 Time  (mean ± σ):   1.918 ms ± 930.405 μs  ┊ GC (mean ± σ):  6.26% ± 10.52%

  █▆▄▃▂▂▁
  █████████▇█▇▆▅▇▁▄▁▃▃▁▁▁▃▃▁▁▁▁▁▁▁▁▃▁▁▁▁▁▁▃▃▅▅▆▃▅▃▁▅▃▁▃▁▃▁▁▃▅ █
  1.64 ms      Histogram: log(frequency) by time      7.05 ms <

 Memory estimate: 1021.17 KiB, allocs estimate: 9469.

Results for 1000 points:

BenchmarkTools.Trial: 242 samples with 1 evaluation.
 Range (min … max):  18.067 ms … 33.481 ms  ┊ GC (min … max): 0.00% … 42.22%
 Time  (median):     19.552 ms              ┊ GC (median):    0.00%
 Time  (mean ± σ):   20.603 ms ±  2.643 ms  ┊ GC (mean ± σ):  4.60% ±  8.39%

     ▂▇█▂▅▅▁
  ▆▆▆███████▅▅▄▅▅▅▄▃▃▃▃▃▅▃▄▃▃▃▄▁▁▃▁▁▃▄▁▃▃▃▁▁▃▃▄▄▃▃▁▁▃▃▁▃▁▁▁▃▃ ▃
  18.1 ms         Histogram: frequency by time        29.3 ms <

 Memory estimate: 8.35 MiB, allocs estimate: 87556.

Results for 10000 points:

BenchmarkTools.Trial: 3 samples with 1 evaluation.
 Range (min … max):  2.053 s …   2.115 s  ┊ GC (min … max): 0.00% … 0.51%
 Time  (median):     2.064 s              ┊ GC (median):    0.00%
 Time  (mean ± σ):   2.077 s ± 32.766 ms  ┊ GC (mean ± σ):  0.17% ± 0.30%

  █         █                                             █
  █▁▁▁▁▁▁▁▁▁█▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁█ ▁
  2.05 s         Histogram: frequency by time        2.11 s <

 Memory estimate: 84.92 MiB, allocs estimate: 862476.
```

### DIFF
--- a/src/tesselation/voronoi.jl
+++ b/src/tesselation/voronoi.jl
@@ -39,9 +39,12 @@ function tesselate(pset::PointSet, method::VoronoiTesselation)
     coords = CoordRefSystems.reconstruct(C, T.(xy))
     Point(coords)
   end
-  polygs = each_polygon(vorono)
-  tuples = [Tuple(inds[begin:(end - 1)]) for inds in polygs]
-  connec = connect.(tuples)
+  polygs = get_polygons(vorono)
+  connec = Vector{Connectivity}(undef, length(polygs))
+  for (idx, poly) in polygs
+    tup = ntuple(i -> poly[i], length(poly) - 1)
+    connec[idx] = connect(tup, Ngon)
+  end
   mesh = SimpleMesh(points, connec)
 
   # remove unused points
--- a/src/tesselation/voronoi.jl
+++ b/src/tesselation/voronoi.jl
@@ -41,9 +41,9 @@ function tesselate(pset::PointSet, method::VoronoiTesselation)
   end
   polygs = get_polygons(vorono)
   connec = Vector{Connectivity}(undef, length(polygs))
-  for (idx, poly) in polygs
-    tup = ntuple(i -> poly[i], length(poly) - 1)
-    connec[idx] = connect(tup, Ngon)
+  for (order_idx, vertices_inds) in polygs
+    tup = ntuple(i -> vertices_inds[i], length(vertices_inds) - 1)
+    connec[order_idx] = connect(tup, Ngon)
   end
   mesh = SimpleMesh(points, connec)
 
--- a/src/tesselation/voronoi.jl
+++ b/src/tesselation/voronoi.jl
@@ -41,9 +41,9 @@ function tesselate(pset::PointSet, method::VoronoiTesselation)
   end
   polygs = get_polygons(vorono)
   connec = Vector{Connectivity}(undef, length(polygs))
-  for (order_idx, vertices_inds) in polygs
-    tup = ntuple(i -> vertices_inds[i], length(vertices_inds) - 1)
-    connec[order_idx] = connect(tup, Ngon)
+  for (i, inds) in polygs
+    tup = ntuple(j -> inds[j], length(inds) - 1)
+    connec[i] = connect(tup, Ngon)
   end
   mesh = SimpleMesh(points, connec)
 
--- a/test/tesselation.jl
+++ b/test/tesselation.jl
@@ -45,5 +45,12 @@
     pts = randpoint3(10)
     pset = PointSet(pts)
     @test_throws AssertionError tesselate(pset, VoronoiTesselation(StableRNG(2024)))
+
+    # Test polygon order is the same as input points order
+    pts = randpoint2(10)
+    mesh = tesselate(pts, VoronoiTesselation(StableRNG(2024)))
+    @test all(zip(pts, mesh)) do (pt, poly)
+        pt in poly || pt ∉ mesh # The pt ∉ mesh is because some points are on the border of the respective polygon and return false from `pt in poly` due to floating point precision. So if that happens we check that the point is not in any other polygon
+    end
   end
 end
--- a/test/tesselation.jl
+++ b/test/tesselation.jl
@@ -50,7 +50,7 @@
     pts = randpoint2(10)
     mesh = tesselate(pts, VoronoiTesselation(StableRNG(2024)))
     @test all(zip(pts, mesh)) do (pt, poly)
-        pt in poly || pt ∉ mesh # The pt ∉ mesh is because some points are on the border of the respective polygon and return false from `pt in poly` due to floating point precision. So if that happens we check that the point is not in any other polygon
+      pt in poly || pt ∉ mesh # The pt ∉ mesh is because some points are on the border of the respective polygon and return false from `pt in poly` due to floating point precision. So if that happens we check that the point is not in any other polygon
     end
   end
 end
--- a/test/tesselation.jl
+++ b/test/tesselation.jl
@@ -50,7 +50,13 @@
     pts = randpoint2(10)
     mesh = tesselate(pts, VoronoiTesselation(StableRNG(2024)))
     @test all(zip(pts, mesh)) do (pt, poly)
-      pt in poly || pt ∉ mesh # The pt ∉ mesh is because some points are on the border of the respective polygon and return false from `pt in poly` due to floating point precision. So if that happens we check that the point is not in any other polygon
+      pt in poly && return true
+      # Point is not in poly, might be to a rounding error. We check if the target polygon's centroid is the closest of all
+      centroid_dists = map(mesh) do element
+        norm(centroid(element) - pt)
+      end
+      this_dist = norm(centroid(poly) - pt)
+      all(<=(this_dist), centroid_dists)
     end
   end
 end
--- a/test/tesselation.jl
+++ b/test/tesselation.jl
@@ -36,7 +36,7 @@
     mesh = tesselate(pset, VoronoiTesselation(StableRNG(2024)))
     @test crs(mesh) === crs(pset)
 
-    coords = [LatLon(rand(-80:T(0.1):80), rand(-180:T(0.1):180)) for _ in 1:10]
+    coords = [LatLon(rand(-20:T(0.1):20), rand(-180:T(0.1):180)) for _ in 1:10]
     pset = PointSet(Point.(coords))
     mesh = tesselate(pset, VoronoiTesselation(StableRNG(2024)))
     @test crs(mesh) === crs(pset)
--- a/test/tesselation.jl
+++ b/test/tesselation.jl
@@ -47,8 +47,9 @@
     @test_throws AssertionError tesselate(pset, VoronoiTesselation(StableRNG(2024)))
 
     # Test polygon order is the same as input points order
+    rng = StableRNG(2024)
     pts = randpoint2(10)
-    mesh = tesselate(pts, VoronoiTesselation(StableRNG(2024)))
+    mesh = tesselate(pts, VoronoiTesselation(rng))
     @test all(zip(pts, mesh)) do (pt, poly)
       pt in poly || pt ∉ mesh # The pt ∉ mesh is because some points are on the border of the respective polygon and return false from `pt in poly` due to floating point precision. So if that happens we check that the point is not in any other polygon
     end
--- a/test/tesselation.jl
+++ b/test/tesselation.jl
@@ -44,6 +44,6 @@
     # Test polygon order is the same as input points order
     pts = randpoint2(10)
     mesh = tesselate(pts, VoronoiTesselation(StableRNG(2024)))
-    @test issameorder(pts, mesh)
+    @test all(sideof(p, boundary(poly)) ∈ (IN, ON) for (p, poly) in zip(pts, mesh))
   end
 end
--- a/test/tesselation.jl
+++ b/test/tesselation.jl
@@ -56,7 +56,7 @@
         norm(centroid(element) - pt)
       end
       this_dist = norm(centroid(poly) - pt)
-      all(<=(this_dist), centroid_dists)
+      all(>=(this_dist), centroid_dists)
     end
   end
 end
--- a/test/tesselation.jl
+++ b/test/tesselation.jl
@@ -47,9 +47,8 @@
     @test_throws AssertionError tesselate(pset, VoronoiTesselation(StableRNG(2024)))
 
     # Test polygon order is the same as input points order
-    rng = StableRNG(2024)
     pts = randpoint2(10)
-    mesh = tesselate(pts, VoronoiTesselation(rng))
+    mesh = tesselate(pts, VoronoiTesselation(StableRNG(2024))
     @test all(zip(pts, mesh)) do (pt, poly)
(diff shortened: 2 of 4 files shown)


src/tesselation/voronoi.jl
```diff
@@ -39,9 +39,12 @@ function tesselate(pset::PointSet, method::VoronoiTesselation)
...
   end
-  polygs = each_polygon(vorono)
-  tuples = [Tuple(inds[begin:(end - 1)]) for inds in polygs]
-  connec = connect.(tuples)
+  polygs = get_polygons(vorono)
+  connec = Vector{Connectivity}(undef, length(polygs))
+  for (order_idx, vertices_inds) in polygs
+    tup = ntuple(i -> vertices_inds[i], length(vertices_inds) - 1)
+    connec[order_idx] = connect(tup, Ngon)
+  end
```
juliohm: Perhaps we can avoid the abstract `Connectivity` type here with an extra allocation of inds, followed by a conversion to tuple with `ntuple`. I know in this case the number of vertices in each voronoi cell is unknown, which probably leads to `Connectivity` eltype at the end anyways.

@eliascarv what do you think?

eliascarv: Variable name suggestion:
```suggestion
  polygs = get_polygons(vorono)
  connec = Vector{Connectivity}(undef, length(polygs))
  for (i, inds) in polygs
    tup = ntuple(j -> inds[j], length(inds) - 1)
    connec[i] = connect(tup, Ngon)
  end
```

test/tesselation.jl
```diff
@@ -45,5 +45,12 @@
     pts = randpoint3(10)
     pset = PointSet(pts)
     @test_throws AssertionError tesselate(pset, VoronoiTesselation(StableRNG(2024)))
+
+    # Test polygon order is the same as input points order
+    pts = randpoint2(10)
+    mesh = tesselate(pts, VoronoiTesselation(StableRNG(2024)))
```
juliohm, line 46:
```suggestion
    rng = StableRNG(2024)
    pts = randpoint2(10)
    mesh = tesselate(pts, VoronoiTesselation(rng))
```

test/tesselation.jl
```diff
@@ -45,5 +45,13 @@
     pts = randpoint3(10)
     pset = PointSet(pts)
     @test_throws AssertionError tesselate(pset, VoronoiTesselation(StableRNG(2024)))
+
+    # Test polygon order is the same as input points order
+    rng = StableRNG(2024)
+    pts = randpoint2(10)
+    mesh = tesselate(pts, VoronoiTesselation(rng))
```
juliohm:
```suggestion
    pts = randpoint2(10)
    mesh = tesselate(pts, VoronoiTesselation(StableRNG(2024))
```

Sorry, didn't notice that other tests were using this pattern. Undoing it.

eliascarv (approved):

test/tesselation.jl
```diff
@@ -45,5 +45,12 @@
     pts = randpoint3(10)
     pset = PointSet(pts)
     @test_throws AssertionError tesselate(pset, VoronoiTesselation(StableRNG(2024)))
+
+    # Test polygon order is the same as input points order
+    pts = randpoint2(10)
+    mesh = tesselate(pts, VoronoiTesselation(StableRNG(2024))
+    @test all(zip(pts, mesh)) do (pt, poly)
+      pt in poly || pt ∉ mesh # The pt ∉ mesh is because some points are on the border of the respective polygon and return false from `pt in poly` due to floating point precision. So if that happens we check that the point is not in any other polygon
```
juliohm: I think we can remove the randomness of this test to avoid this floating point side effect. Can you please construct a deterministic set of points where the `point in poly` predicate is true?

juliohm: Alternatively you can try to add a test that checks that the distance from the point to the `centroid(poly)` is smaller or equal to any other centroid.

disberd: Went with the second appraoch

disberd:
Tests now seem to fail due to the re-introduction of the latitude check in CoordRefSystems, on test lines not really impacted by the contents of this PR.

At the same time i makes sense that tests now fail as the resulting latitude of some points after tessellation was >100°

test/tesselation.jl
```diff
@@ -45,5 +45,18 @@
...
     pset = PointSet(pts)
     @test_throws AssertionError tesselate(pset, VoronoiTesselation(StableRNG(2024)))
+
+    # Test polygon order is the same as input points order
+    pts = randpoint2(10)
+    mesh = tesselate(pts, VoronoiTesselation(StableRNG(2024)))
+    @test all(zip(pts, mesh)) do (p, poly)
+      p ∈ poly && return true
+      # point is not in poly due to a floating point error.
+      # check if the target polygon's centroid is the closest
```
juliohm:
```suggestion
      # check if the target polygon's centroid is the closest
```

Let's simplify the test to avoid these subtle floating point issues.

juliohm (approved):

test/tesselation.jl
```diff
@@ -36,7 +36,7 @@
     mesh = tesselate(pset, VoronoiTesselation(StableRNG(2024)))
     @test crs(mesh) === crs(pset)
 
-    coords = [LatLon(rand(-90:T(0.1):90), rand(-180:T(0.1):180)) for _ in 1:10]
+    coords = [LatLon(rand(-80:T(0.1):80), rand(-180:T(0.1):180)) for _ in 1:10]
```
juliohm:
```suggestion
    coords = [LatLon(rand(-20:T(0.1):20), rand(-180:T(0.1):180)) for _ in 1:10]
```

juliohm:
The Voronoi tesselation algorithm from DelaunayTriangulation.jl is blind to CRS information, and it is apparently not easy to control the range of latitude values it produces. We need to handle this properly in a future PR. For now, I will just comment out the test that is failing with `LatLon`.

test/tesselation.jl
```diff
@@ -36,14 +36,20 @@
...
+    # Test polygon order is the same as input points order
+    pts = randpoint2(10)
+    mesh = tesselate(pts, VoronoiTesselation(StableRNG(2024)))
+    @test all(zip(pts, mesh)) do (p, poly)
+      # check if the target polygon's centroid is the closest
+      dist(e) = evaluate(Euclidean(), p, centroid(e))
+      all(mesh) do elem
+        dist(elem) ≥ dist(poly)
+      end
+    end
```
eliascarv: @disberd, defining functions inside tests does not work in some cases.
Try moving this test to a helper function like this:
```julia
# check if polygon order is the same as input points order
function issameorder(pts, mesh)
  all(zip(pts, mesh)) do (p, poly)
    # check if the target polygon's centroid is the closest
    dist(e) = evaluate(Euclidean(), p, centroid(e))
    all(dist(elem) ≥ dist(poly) for elem in mesh)
  end
end
```
Test helper functions lives in `test/testutils.jl`.
After doing this, use the function:
```julia
@test issameorder(pts, mesh)
```

eliascarv: @disberd, I'll do this to help you with your PR.

juliohm: @eliascarv I introduced this function, so it is my fault. Please avoid adding it to testutils.jl, it is only going to be used here. Refactor the code without the function if necessary.

juliohm:
The problem is caused by the fact that the `point in poly` predicate with `Ngon` is not using the more robust algorithm implemented in the general case:

https://github.com/JuliaGeometry/Meshes.jl/blob/430b38f46c31b98388a41e75ed9502ef236edb5b/src/predicates/in.jl#L140-L151

We need to benchmark these two algorithms later in a separate PR.

juliohm:
Thank you @disberd 💯 ❤️

PR #999, opened 2024-08-12
Weiler atherton clipping method

jnemanja:
In draft because testing should have more cases and documentation probably needs some improvements. An extra commit is added with informal testing used during development, which is to be discarded before merge.


src/intersections/polygons.jl
```diff
@@ -3,14 +3,7 @@
...
 function intersection(f, poly₁::Polygon, poly₂::Polygon)
-  # TODO: use Weiler-Atherton or other more general clipping method
-  clipped = if isconvex(poly₂)
-    clip(poly₁, poly₂, SutherlandHodgman())
-  elseif isconvex(poly₁)
-    clip(poly₂, poly₁, SutherlandHodgman())
-  else
-    throw(ErrorException("intersection not implemented between two non-convex polygons"))
-  end
+  clipped = clip(poly₂, poly₁, WeilerAtherton())
```
juliohm:
```suggestion
  clipped = clip(poly₁, poly₂, WeilerAtherton())
```

test/intersections.jl
```diff
@@ -1188,7 +1189,7 @@
     quad = Quadrangle(cart(0, 0), cart(0.1, 0.0), cart(0.1, 0.1), cart(0.0, 0.1))
     poly = PolyArea(cart(0, 0), cart(2, 0), cart(1, 1), cart(1, 0.5))
     @test intersection(quad, poly) |> type == Intersecting
-    @test all(vertices(quad ∩ poly) .≈ [cart(0, 0), cart(0.1, 0), cart(0.1, 0.05)])
+    @test all(vertices(quad ∩ poly) .≈ [cart(0.1, 0.05), cart(0, 0), cart(0.1, 0)])
```
juliohm: Please make sure that the result is returned in CCW orientation.

test/scratchbooks/weileratherton.jl
```diff
@@ -0,0 +1,138 @@
...
+end
+##
+clip_and_show(r, t |> Scale(0.8) |> Translate(0.25, 0.1))
+clip_and_show(r, t |> Scale(0.8) |> Translate(-0.25, 0.1))
+clip_and_show(r, t |> Scale(0.8) |> Translate(0.0, 0.3))
+clip_and_show(r, t |> Scale(0.8) |> Translate(0.0, 0.0))
+clip_and_show(r, t |> Scale(0.8) |> Translate(0.1, 0.0))
+clip_and_show(r, t |> Scale(0.8) |> Translate(0.2, 0.3))
+clip_and_show(r, t |> Scale(0.8) |> Translate(-0.2, 0.3))
+clip_and_show(r, t |> Scale(0.8) |> Translate(-0.2, -0.3))
```
juliohm: Appreciate if you can remove this file from the PR, and add some of the resulting visualizations as comments instead. You can also share this script in the comments for future reproduction.

docs/src/algorithms/clipping.md
```diff
@@ -32,4 +32,31 @@ viz(poly)
...
+clipped = clip(poly, other, WeilerAtherton())
+
+viz(poly)
+viz!(other, color = :black, alpha = 0.2)
+viz!(boundary(clipped), color = :red, segmentsize = 3)
+Mke.current_figure()
+```
+
+Algorithm is by default limited with a maximum number of vertices of 50 for each ring of the resulting polygon.
+The limit can be adjusted using a ScopedValue WEILER_ATHERTON_MAX_VERTICES_NUM.
```
juliohm: Can you please comment on why we need this limit?

juliohm (changes requested):
That is an amazing contribution @jnemanja, thanks for putting the time into it. We will need to adjust the code over the week to match the style of the project, and to make sure that the behavior is correct.

Can you please address this first round of review below? Thank you!

src/intersections/polygons.jl
jnemanja: Fixed.

test/intersections.jl
jnemanja: As far as I can tell, the method returns the correct orientations, as long as the input polygons are correctly oriented. The changes here are just changing the first vertex from which the new polygon starts. This is to be expected because the Weiler-Atherton searches for the first "Entering" vertex, while Sutherland-Hodgman starts from the first vertex of the to be clipped polygon. Algorithm does not try to ensure correct orientations for the input or output polygon rings, as in the cases for the clipping polygons with the holes, it relies on the output following the orientation of the input, because the algorithm is executed on outer and inner rings separately. In all my tests it looks to be working correctly.

test/scratchbooks/weileratherton.jl
jnemanja: Removed.

docs/src/algorithms/clipping.md
jnemanja: This was something needed during the development, when traversing the data structure to collect new output rings the algorithm could end up in infinite cycles. This should not happen anymore. I did not test on broken and self intersecting polygons, but it should not happen there either. I don't know why I added a documentation for it instead of removing it in the first place. Now it is removed.

test/intersections.jl
```diff
@@ -1161,22 +1161,23 @@
...
+    @test length(out) == 2
+    @test all(out[1] .≈ [cart(5, 3), cart(5, 4), cart(3, 4), cart(4, 3)])
+    @test all(out[2] .≈ [cart(2, 0), cart(5, 0), cart(5, 2), cart(4, 1), cart(2, 1)])
 
     # inside
     poly = Quadrangle(cart(1, 0), cart(1, 1), cart(0, 1), cart(0, 0))
     other = Quadrangle(cart(5, 0), cart(5, 4), cart(0, 4), cart(0, 0))
     @test intersection(poly, other) |> type == Intersecting
-    @test all(vertices(poly ∩ other) .≈ vertices(poly))
+    poly_vertices = vertices(poly)
```
```suggestion
```
unused

jnemanja: Removed.

src/clipping/weileratherton.jl
```diff
@@ -0,0 +1,397 @@
...
+    WeilerAtherton()
+
+The Wieler-Atherton algorithm for clipping polygons.
+
+## References
+
+* Weiler, K., & Atherton, P. 1977. [Hidden surface removal using polygon area sorting]
+  (https://dl.acm.org/doi/pdf/10.1145/563858.563896)
+"""
+
```
juliohm:
```suggestion
```

src/clipping/weileratherton.jl
```diff
@@ -0,0 +1,397 @@
...
+  for v in vs[2:end]
+    new = RingVertex{Normal}(v)
+    append_vertices!(start, new)
+    start = new
+  end
+
+  return start.left
+end
+
+is_normal(_::RingVertex{Normal}) = true
```
juliohm: please avoid underscore in function and variable names

src/clipping/weileratherton.jl
```diff
@@ -0,0 +1,397 @@
...
+    l_current = measure(Segment(segment_start.point, getfield(vertex, origin_side).point))
+    if (l_new < l_current) || (getfield(vertex, origin_side) == segment_end)
+      next_vertex = getfield(vertex, origin_side)
+      setfield!(intersection, side, next_vertex)
+      setfield!(vertex, origin_side, intersection)
+      break
+    end
+    vertex = getfield(vertex, origin_side)
+  end
+end
```
juliohm: Please follow the code style. For example, `segment_start` can be replaced by `start`. You can then add comments to the function body to explain what is happening.

src/clipping/weileratherton.jl
```diff
@@ -0,0 +1,397 @@
...
+
+# Traversing and selecting following vertices for collecting of the rings after clipping depends
+# on the vertex type. Helper next_vertex follows the correct path.
+next_vertex(v::RingVertex{Normal}) = v.right
+next_vertex(v::RingVertex{Entering}) = v.right
+next_vertex(v::RingVertex{Exiting}) = v.left
+
+function collect_clipped(entering)
+  rings = []
```
juliohm: If you know the eltype of these buffers, try to specify it. This way the Julia compiler will have a greater chance to optimize the code. It is fine to use `Any` when the type is super complicated and the resulting array will be wrapped in a constructor later that will infer the type.

juliohm (changes requested):
Another round of review related to code style before we can start reviewing the algorithm itself.

src/clipping/weileratherton.jl
jnemanja: Fixed.

src/clipping/weileratherton.jl
jnemanja: Underscores removed.

src/clipping/weileratherton.jl
jnemanja: Tried to simplify variable namings everywhere.

src/clipping/weileratherton.jl
jnemanja: Done.

docs/src/algorithms/clipping.md
```diff
@@ -28,6 +28,30 @@ other = Box((0,1), (3,7))
 # clipped polygon
 clipped = clip(poly, other, SutherlandHodgman())
 
+viz(poly)
+viz!(other, color = :black, alpha = 0.2)
+viz!(boundary(clipped), color = :red, segmentsize = 3)
+Mke.current_figure()
+```
+
+## WeilerAtherton
```
juliohm:
```suggestion
## Weiler-Atherton
```

docs/src/algorithms/clipping.md
```diff
@@ -28,6 +28,30 @@ other = Box((0,1), (3,7))
...
+
+```@docs
+WeilerAtherton
```
juliohm:
```suggestion
WeilerAthertonClipping
```

docs/src/algorithms/clipping.md
```diff
@@ -28,6 +28,30 @@ other = Box((0,1), (3,7))
...
+# polygon to clip
+outer = Ring((8, 0), (4, 8), (2, 8), (-2, 0), (0, 0), (1, 2), (5, 2), (6, 0))
+inner = Ring((4, 4), (2, 4), (3, 6))
+poly = PolyArea([outer, inner])
+
+# clipping geometry can be a polygon with holes
+other = poly |> Rotate(pi) |> Translate(8.0, 8.0)
+
+# clipped polygon
+clipped = clip(poly, other, WeilerAtherton())
```
juliohm:
```suggestion
clipped = clip(poly, other, WeilerAthertonClipping())
```

src/Meshes.jl
```diff
@@ -430,6 +430,7 @@ export
   # clipping
   ClippingMethod,
   SutherlandHodgman,
+  WeilerAtherton,
```
```suggestion
  WeilerAthertonClipping,
```

src/clipping/weileratherton.jl
```diff
@@ -0,0 +1,398 @@
+# ------------------------------------------------------------------
+# Licensed under the MIT License. See LICENSE in the project root.
+# ------------------------------------------------------------------
+
+"""
+    WeilerAtherton()
```
juliohm:
```suggestion
    WeilerAthertonClipping()
```

juliohm (changes requested):
@jnemanja can you please rebase this PR to the latest master branch, and rename `WeilerAtherton` to `WeilerAthertonClipping` to be more explicit?

As we are preparing for a v1.0, we will be more careful with naming methods. We already renamed `SutherlandHodgman` to `SutherlandHodgmanClipping` for clariry.

src/clipping/weileratherton.jl
jnemanja: All done, rebased and fixed the method naming.

juliohm: Thank you @jnemanja ❤️ will reserve some time to review the PR carefully now that it is cleaned and rebased. We will need some additional iterations, but the code is getting much easier to digest now.

jnemanja:
Added tests to reach full coverage. Also, in the process I noticed that the order of rings in the polygons with holes is not always proper, so added a fix for that.

src/clipping/weileratherton.jl
```diff
@@ -0,0 +1,408 @@
...
+        return nothing
+      end
+    end
+    return poly
+  else
+    return clip(poly, b, method)
+  end
+end
+
+function clip(gs::GeometrySet, other::Geometry, method::WeilerAthertonClipping)
```
jnemanja: Shall this be moved to clipping.jl and change method type to ClippingMethod?

jnemanja:
And rebased to latest again.

src/clipping/weileratherton.jl
juliohm: That is a good idea. I would even generalize this method to `Domain` instead of `GeometrySet`.

docs/src/algorithms/clipping.md
```diff
@@ -28,6 +28,30 @@ other = Box((0,1), (3,7))
...
+## Weiler-Atherton
+
+```@docs
+WeilerAthertonClipping
+```
+
+```@example clipping
+# polygon to clip
+outer = Ring((8, 0), (4, 8), (2, 8), (-2, 0), (0, 0), (1, 2), (5, 2), (6, 0))
+inner = Ring((4, 4), (2, 4), (3, 6))
```
juliohm, line 46: the inner ring should have CW orientation in the PolyArea constructor

docs/src/algorithms/clipping.md
```diff
@@ -28,6 +28,30 @@ other = Box((0,1), (3,7))
...
+poly = PolyArea([outer, inner])
+
+# clipping geometry can be a polygon with holes
+other = poly |> Rotate(pi) |> Translate(8.0, 8.0)
```
juliohm:
```suggestion
other = poly |> Rotate(π) |> Translate(8, 8)
```

test/intersections.jl
```diff
@@ -1149,22 +1149,22 @@ end
...
+  @test length(out) == 2
+  @test all(out[1] .≈ [cart(3, 4), cart(4, 3), cart(5, 3), cart(5, 4)])
+  @test all(out[2] .≈ [cart(5, 2), cart(4, 1), cart(2, 1), cart(2, 0), cart(5, 0)])
 
   # inside
   poly = Quadrangle(cart(1, 0), cart(1, 1), cart(0, 1), cart(0, 0))
   other = Quadrangle(cart(5, 0), cart(5, 4), cart(0, 4), cart(0, 0))
   @test intersection(poly, other) |> type == Intersecting
-  @test all(vertices(poly ∩ other) .≈ vertices(poly))
+  @test all(vertices(poly ∩ other) .≈ [cart(1, 0), cart(1, 1), cart(0, 1), cart(0, 0)])
```
```suggestion
  @test poly ∩ other) ≗ poly
```

Maybe we should adjust some of these tests to use `\circeq` between polygons.

src/clipping/weileratherton.jl
```diff
@@ -0,0 +1,408 @@
...
+  intersection = RingVertex{vtype}(point)
+  insertintersection!(clipping, intersection, :left)
+  insertintersection!(clipped, intersection, :right)
+
+  if vtype == Entering
+    push!(entering, intersection)
+  end
+end
+
+function clip(poly::Polygon, ring::Ring, _::WeilerAthertonClipping)
```
juliohm: let's move this function to be the first function in the file for improved clarity

juliohm:
```suggestion
function clip(poly::Polygon, ring::Ring, ::WeilerAthertonClipping)
```

src/clipping/weileratherton.jl
```diff
@@ -0,0 +1,408 @@
...
+  end
+
+  n = length(polys)
+  out = n == 0 ? nothing : (n == 1 ? polys[1] : GeometrySet(polys))
+  return out
+end
+
+function clip(poly::Polygon, other::Geometry, method::WeilerAthertonClipping)
+  b = boundary(other)
+  if typeof(b) <: Multi
```
juliohm:
```suggestion
  if b isa Multi
```

src/clipping/weileratherton.jl
```diff
@@ -0,0 +1,408 @@
...
+      push!(polys, PolyArea(newpolyrings))
+    end
+  end
+
+  n = length(polys)
+  out = n == 0 ? nothing : (n == 1 ? polys[1] : GeometrySet(polys))
+  return out
+end
+
+function clip(poly::Polygon, other::Geometry, method::WeilerAthertonClipping)
```
juliohm: could this work for any clipping method with a `Multi`? should we simply add a more general method in clipping.jl that handles multi using the `parent`?

src/clipping/weileratherton.jl
```diff
@@ -0,0 +1,408 @@
...
+      end
+    end
+    return poly
+  else
+    return clip(poly, b, method)
+  end
+end
+
+function clip(gs::GeometrySet, other::Geometry, method::WeilerAthertonClipping)
+  polys::Vector{PolyArea} = []
```
juliohm:
```suggestion
  polys = PolyArea[]
```

src/clipping/weileratherton.jl
```diff
@@ -0,0 +1,408 @@
...
+      end
+      return PolyArea(collectedrings)
+    end
+    if all([v ∈ PolyArea(polyrings[1]) for v in vertices(ring)])
+      push!(collectedrings, ring)
+    end
+  end
+
+  # Handle all the non-intersected rings. Add them if they are contained in the clipping ring.
+  polys::Vector{PolyArea} = []
```
juliohm:
```suggestion
  polys = PolyArea[]
```

src/clipping/weileratherton.jl
```diff
@@ -0,0 +1,408 @@
...
+      end
+    end
+
+    clipping = clipping.left
+    if clipping.left == startclipping.left
+      break
+    end
+  end
+
+  collectedrings::typeof(polyrings) = collectclipped(entering)
```
juliohm:
```suggestion
  collectedrings = collectclipped(entering)
```

we don't use type annotations in algorithms

src/clipping/weileratherton.jl
```diff
@@ -0,0 +1,408 @@
...
+  clippedrings = [gettraversalring(r) for r in polyrings]
+  startclipping = gettraversalring(ring)
+
+  # For keeping track of intersected rings, as the non-intersected ones need additional
+  # processing at the end.
+  intersected = zeros(Bool, length(clippedrings))
+
+  # For collecting all entering vertices, as they are used as starting points for collection
+  # of the output rings.
+  entering::Vector{RingVertex{Entering}} = []
```
juliohm:
```suggestion
  entering = RingVertex{Entering}}[]
```

src/clipping/weileratherton.jl
```diff
@@ -0,0 +1,408 @@
...
+    end
+
+    if length(ring) > 2
+      push!(rings, Ring([r.point for r in ring]...))
+    end
+  end
+  return rings
+end
+
+function decide_direction(clipped₁, clipped₂, clipping₁, clipping₂)
```
juliohm: appreciate if you can explain the main goal of these auxiliary functions so that we can reuse existing functionality. for example, we have the `\angle` between rays and points which should be used instead of explicit calls to `atan`, see angles.jl

src/clipping/weileratherton.jl
```diff
@@ -0,0 +1,408 @@
...
+    # after the current one.
+    clippingsegment = Segment(clipping.left.point, clipping.left.left.point)
+
+    for (k, startclipped) in enumerate(clippedrings)
+      clipped = startclipped
+      while true
+        # Like for the clipping, the clipped also uses three consecutive vertices.
+        clippedsegment = Segment(clipped.left.point, clipped.left.left.point)
+
+        vertex_type = nothing
```
juliohm: still not full adherence to code style, we don't use underscore unless strictly needed

juliohm (changes requested):
@jnemanja thank you for the great work here, and for fixing the orientation of polygons in other unrelated tests.

In this second round of review I took a closer look into the implementation. I think we can simplify it a bit more, and reuse functionality in the auxiliary functions.

src/clipping/weileratherton.jl
jnemanja: Done.

docs/src/algorithms/clipping.md
jnemanja: To me, this looks CW.

docs/src/algorithms/clipping.md
jnemanja: Fixed.

test/intersections.jl
jnemanja: Done. Used it also in some other places where I could see it applicable. Probably not an exhaustive fix.

src/clipping/weileratherton.jl
jnemanja: Done, but the functions related to the RingVertex data type are still above, close to the data type definition. Some other function reordering also done.

jnemanja: Fixed. Applied the same in some other places.

src/clipping/weileratherton.jl
jnemanja: Fixed.

src/clipping/weileratherton.jl
jnemanja: Moved this one as well. With slight change to sutherlandhodgeman.jl, seems to cooperate with it finely, but I did not see additional benefits to it in my tests, some malformed results were obtained. But then again the tests did use a bit more complex geometries that should not be expected to work with SutherlandHodgman. In theory, it might help with some non self-intersecting multi geometries, but couldn't confirm.

src/clipping/weileratherton.jl
jnemanja: Done.

src/clipping/weileratherton.jl
jnemanja: Done.

src/clipping/weileratherton.jl
jnemanja: This was there because it didn't work nicely with the PolyArea constructor without it. Now it works because the array is expanded in the constructor. Still, this might point to a need to add some corrections to the PolyArea constructors to work better with less specific arrays.

src/clipping/weileratherton.jl
jnemanja: Done.

src/clipping/weileratherton.jl
jnemanja: I hope the explanations are sufficient. Changed the explicit usage of atan to use the \angle instead. I still had to convert from -pi:pi to 0:2pi range. One thing that surfaced is that the precision was lower than with my original atan solution, which in turn showed that the isapprox is not properly using atol with radians. So I had to handle atols explicitly.

Also, this change shown that zero length segments are being passed to the function, which then used them incorrectly to somehow get the correct results. Now this is circumvented outside code where it is ensured that in the cases where it matters, the zero  length segments are not passed. Still, sometimes zero length segments are passed in situations where it should not matter and then the function just returns nothing, signaling to the caller to drop the vertex. Maybe this should be also handled outside, but this way it is only handled in one place.

src/clipping/weileratherton.jl
jnemanja: Now this should be corrected everywhere.

jnemanja:
And rebased to the newest version.

src/clipping.jl
```diff
@@ -16,8 +16,36 @@ Clip the subject `geometry` with `other` geometry using clipping `method`.
...
+  polys = PolyArea[]
+  for i in 1:nelements(gs)
+    poly = clip(element(gs, i), other, method)
+    if !isnothing(poly)
+      push!(polys, poly)
+    end
+  end
+  n = length(polys)
+  return n == 0 ? nothing : (n == 1 ? polys[1] : GeometrySet(polys))
+end
```
juliohm: this could be written as

```julia
clipped = filter(!isnothing, [clip(geom, other, method) for geom in dom])
isempty(clipped) ? nothing : GeometrySet(clipped)
```

but I am starting to think that we should delay these higher-level methods to a separate PR.  We can even consider a new `Clip` transform that handles these higher-level calls.

src/clipping.jl
```diff
@@ -16,8 +16,36 @@ Clip the subject `geometry` with `other` geometry using clipping `method`.
...
+    for r in parent(b)
+      poly = clip(poly, r, method)
+      if isnothing(poly)
+        return nothing
+      end
+    end
+    return poly
+  else
+    return clip(poly, b, method)
+  end
```
juliohm: This code pattern can be rewritten as two separate methods that dispatch on `Multi` or another geometry type:

```
_clip(poly, multi::Multi, method) = ...
_clip(poly, other, method) = ...
```

but I also think that we can delay this to a separate PR where we have more freedom to adjust the generic fallbacks that are unrelated to the specific clipping method.

src/clipping/sutherlandhodgman.jl
```diff
@@ -18,10 +18,17 @@ The Sutherland-Hodgman algorithm for clipping polygons.
...
+  # r = [r for r in c if !isnothing(r)]
+  # isempty(r) ? nothing : PolyArea(r)
+# end
+
+
+function clip(poly::Polygon, other::Ring, method::SutherlandHodgmanClipping)
+ c = [clip(ring, other, method) for ring in rings(poly)]
+ r = [r for r in c if !isnothing(r)]
+ isempty(r) ? nothing : PolyArea(r)
 end
```
juliohm: Based on the latest comments above, I believe that the safest strategy in this PR is to confine the changes to the new clipping method, with minimum changes to other methods, and no addition of higher-level methods.

Can we rework the PR so that it only adds methods with types `Ring` and `Polygon`?  If a method with `Multi` is needed to handle some specific feature of `WeilerAthertonClipping`, that is fine too.

That way we can work on a separate PR later to better organize the dispatch with different clipping methods and geometries. Appreciate your patience improving the situation here :)

src/clipping.jl
jnemanja: Applied the suggestion, function moved to weileratherton.jl.

src/clipping.jl
jnemanja: Applied the suggestion, function moved to weileratherton.jl. I am not sure this change is adding anything and even looks less clear to me. But I assume it can be the proper way to handle if further extensions are to be done.

src/clipping/sutherlandhodgman.jl
jnemanja: Reverted the sutherlandhodgman.jl. I moved the functions from clipping.jl to weileratherton.jl, I think it is still good to have them there as they allow to have shorter tests that achieve a bit more coverage. Tests are hard to write and quite error prone with the need to hand write the vertices. I dislike the idea of redoing tests at this stage.

test/intersections.jl
```diff
@@ -1149,22 +1149,22 @@ end
...
 
   # octagon
   poly = Octagon(cart(8, -2), cart(8, 5), cart(2, 5), cart(4, 3), cart(6, 3), cart(4, 1), cart(2, 1), cart(2, -2))
   other = Quadrangle(cart(5, 0), cart(5, 4), cart(0, 4), cart(0, 0))
   @test intersection(poly, other) |> type == Intersecting
-  @test all(
-    vertices(poly ∩ other) .≈
-    [cart(3, 4), cart(4, 3), cart(5, 3), cart(5, 2), cart(4, 1), cart(2, 1), cart(2, 0), cart(5, 0), cart(5, 4)]
-  )
+  out = vertices.(poly ∩ other)
```
juliohm: if the intersection result is made of multiple parts, then it should be a `Multi`-geometry. It is seen as a single entity as opposed to a `GeometrySet`, which is seen as a collection.

src/clipping/weileratherton.jl
```diff
@@ -0,0 +1,431 @@
...
+      return nothing
+    end
+  end
+  poly
+end
+
+function clip(dom::Domain, other::Geometry, method::WeilerAthertonClipping)
+  clipped = filter(!isnothing, [clip(geom, other, method) for geom in dom])
+  isempty(clipped) ? nothing : (length(clipped) == 1 ? clipped[1] : GeometrySet(clipped))
+end
```
juliohm: This method is not specific to WeilerAtherton, please remove it. We will refactor the dispatch with Domain later in a separate PR.

src/clipping/weileratherton.jl
```diff
@@ -0,0 +1,431 @@
...
+
+function clip(poly::Polygon, multi::Multi, method::WeilerAthertonClipping)
+  for r in parent(multi)
+    poly = clip(poly, r, method)
+    if isnothing(poly)
+      return nothing
+    end
+  end
+  poly
+end
```
juliohm: Same here. This method is not specific to WeilerAtherton. In order to facilitate the review, please remove it.

src/clipping/weileratherton.jl
```diff
@@ -0,0 +1,431 @@
...
+isnormal(::RingVertex) = false
+
+function appendvertices!(v₁::RingVertex{Normal}, v₂::RingVertex{Normal})
+  v₂.left = v₁.left
+  v₁.left = v₂
+  v₂.right = v₁.right
+  v₁.right = v₂
+end
+
+function clip(poly::Polygon, ring::Ring, ::WeilerAthertonClipping)
```
juliohm, line 95: Please break this function into smaller functions that do one thing. The Julia compiler doesn't work well with large functions containing multiple branches. It is also hard to review the logic within multiple nesting levels.

Any function call in Julia creates a "barrier" where the compiler has a new chance to infer concrete argument types.

src/Meshes.jl
```diff
@@ -106,9 +106,9 @@ include("winding.jl")
 include("sideof.jl")
 include("orientation.jl")
 include("merging.jl")
-include("clipping.jl")
 include("clamping.jl")
 include("intersections.jl")
+include("clipping.jl")
```
juliohm: is this change necessary?

src/clipping/weileratherton.jl
```diff
@@ -0,0 +1,417 @@
...
+    # Before converting, check if any leftover non intersected rings
+    # should be included in this PolyArea.
+    newpolyrings = perhapsaddnonintersected!(r, polyrings, intersected)
+    if !isnothing(newpolyrings)
+      push!(polys, PolyArea(newpolyrings))
+    end
+  end
+
+  n = length(polys)
+  n == 0 ? nothing : (n == 1 ? polys[1] : Multi(GeometrySet(polys)))
```
juliohm:
```suggestion
  n == 0 ? nothing : (n == 1 ? polys[1] : Multi(polys))
```

src/clipping/weileratherton.jl
```diff
@@ -0,0 +1,417 @@
...
+end
+
+insertintersections!(vertices::Array, clipping, clipped, entering) =
+  any(insertintersections!.(vertices, Ref(clipping), Ref(clipped), Ref(entering)))
+
+# Traversing and selecting following vertices for collecting of the rings after clipping depends
+# on the vertex type. Helper nextvertex follows the correct path.
+nextvertex(v::RingVertex{Normal}) = v.right
+nextvertex(v::RingVertex{Entering}) = v.right
+nextvertex(v::RingVertex{Exiting}) = v.left
```
juliohm: This is part of the RingVertex api, please move it to after the RingVertex definition. We will take a look later if these new VertexType and RingVertex are really necessary or if we can rely on simple integers/enums.

juliohm:
@jnemanja did you have a chance to take a look into this PR?

jnemanja:
> @jnemanja did you have a chance to take a look into this PR?

I am not aware of any additional issues I should address. If I missed something, could you point it out.

juliohm:
I believe that tests will fail with the latest commits, but I did trigger the bot again to be sure.

jnemanja:
> I believe that tests will fail with the latest commits, but I did trigger the bot again to be sure.

The tests were successful, but the docs failed. Fixed that now and rebased to the newest release.

src/Meshes.jl
juliohm: can you please move this include to its original line?

test/intersections.jl
```diff
@@ -1149,22 +1149,22 @@ end
...
   other = Quadrangle(cart(5, 0), cart(5, 4), cart(0, 4), cart(0, 0))
   @test intersection(poly, other) |> type == Intersecting
-  @test all(vertices(poly ∩ other) .≈ [cart(5, 3), cart(4, 4), cart(2, 4), cart(0, 2), cart(5, 2)])
+  @test all(vertices(poly ∩ other) .≈ [cart(2, 4), cart(0, 2), cart(5, 2), cart(5, 3), cart(4, 4)])
 
   # octagon
   poly = Octagon(cart(8, -2), cart(8, 5), cart(2, 5), cart(4, 3), cart(6, 3), cart(4, 1), cart(2, 1), cart(2, -2))
   other = Quadrangle(cart(5, 0), cart(5, 4), cart(0, 4), cart(0, 0))
   @test intersection(poly, other) |> type == Intersecting
+  out = vertices.(poly ∩ other)
```
```suggestion
  out = vertices(poly ∩ other)
```

src/Meshes.jl
juliohm: @jnemanja this comment is not addressed

src/intersections/polygons.jl
```diff
@@ -3,14 +3,7 @@
...
 
 function intersection(f, poly₁::Polygon, poly₂::Polygon)
-  # TODO: use Weiler-Atherton or other more general clipping method
-  clipped = if isconvex(poly₂)
-    clip(poly₁, poly₂, SutherlandHodgmanClipping())
-  elseif isconvex(poly₁)
-    clip(poly₂, poly₁, SutherlandHodgmanClipping())
-  else
-    throw(ErrorException("intersection not implemented between two non-convex polygons"))
-  end
```
juliohm: In order to control risk of new bugs, we need to be extra careful with modifications that have wide impact in the project. Our contribution guidelines explain that a PR should do one thing and one thing only to facilitate review and reduce side effects.

Please remove this change in the intersection function, and focus on the new `clip` method with tests.

test/clipping.jl
```diff
@@ -74,3 +74,118 @@
...
+      cart(6, 5),
+      cart(10 / 3, 11 / 3),
+      cart(4.0, 3.0),
+      cart(3.0, 3.0),
+      cart(3.0, 3.5)
+    ]
+  )
+
+  # Tolerances are not properly retrieved for Float32 types, so need to pass them explicitly.
+  atol_el = coords(cart(0.0)).x
```
juliohm: @jnemanja could you please elaborate on this tolerance issue? Why can't we use the default `atol` in the following test?

src/clipping/weileratherton.jl
```diff
@@ -0,0 +1,450 @@
...
+  left::RingVertex{<:VertexType,M,C}
+  right::RingVertex{<:VertexType,M,C}
+
+  function RingVertex{VT,M,C}(point) where {VT<:VertexType,M<:Manifold,C<:CRS}
+    v = new(point)
+    v.left = v
+    v.right = v
+    v
+  end
+end
```
juliohm:
```suggestion
mutable struct RingVertex{VT<:VertexType,P<:Point}
  point::P
  left::RingVertex{<:VertexType,P}
  right::RingVertex{<:VertexType,P}

  function RingVertex{VT,P}(point) where {VT<:VertexType,P<:Point}
    v = new(point)
    v.left = v
    v.right = v
    v
  end
end
```

@jnemanja could this be simplified further as above?

PR #1007, opened 2024-08-16
Update `Proj` implementation

eliascarv:
This PR handles cases that occur in projections between different Manifolds.

### DIFF
--- a/src/transforms/proj.jl
+++ b/src/transforms/proj.jl
@@ -35,6 +35,10 @@ applycoord(::Proj{CRS}, p::Point) where {CRS} = Point(convert(CRS, coords(p)))
 # convert the CRS and preserve the manifold
 applycoord(::Proj{CRS}, p::Point{M}) where {CRS<:Basic,M<:🌐} = Point{M}(convert(CRS, coords(p)))
 
+# fix ambiguities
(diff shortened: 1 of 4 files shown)


src/transforms/proj.jl
```diff
@@ -29,15 +29,21 @@ Proj(code::Type{<:ESRI}) = Proj{CoordRefSystems.get(code)}()
 
 parameters(::Proj{CRS}) where {CRS} = (; CRS)
 
+# converts the CRS and uses an appropriate Manifold
```
juliohm:
```suggestion
# convert the CRS and use the default manifold
```

src/transforms/proj.jl
```diff
@@ -29,15 +29,21 @@ Proj(code::Type{<:ESRI}) = Proj{CoordRefSystems.get(code)}()
...
 applycoord(::Proj{CRS}, p::Point) where {CRS} = Point(convert(CRS, coords(p)))
 
+# converts the CRS and maintains the Manifold
```
juliohm:
```suggestion
# convert the CRS and preserve the manifold
```

juliohm (approved):

PR #1014, opened 2024-08-20
Handle different CRS and length units in `sideof`

### LINKED ISSUE #1010: `Point` in `PolyArea` different results for different units
There seems to be an issue when converting units in the point in polyarea alg. This is in v0.49.0
```julia-repl
julia> poly = Meshes.PolyArea(Point(-0.5mm, -1mm), Point(0.5mm, -1mm),Point(0.5mm, 1mm),Point(-0.5mm, 1mm))
PolyArea
  outer
  └─ Ring((x: -0.5 mm, y: -1.0 mm), ..., (x: -0.5 mm, y: 1.0 mm))

julia> Point(0.95mm, 0.45mm) ∈ poly
false

julia> Point(uconvert(m, 0.95mm), uconvert(m, 0.45mm)) ∈ poly
true
```

kylebeggs:
PR intended to close #1010 . This works for Cartesian, but fails for Mercator. I'm not familiar with Mercator so I don't know what to do with the error message:
```julia
ArgumentError: the length unit of Mercator cannot be changed
```


juliohm:
Maybe we should use `Proj` here instead of `LengthUnit`. It will handle these nuances of different CRS.

juliohm:
And I think the convention can be adjusted to `Proj` the simplest geometry to the `CRS` of the other. In this case we want to `Proj` the `point` into the `CoordRefSystems.constructor(crs(ring))`. Can you check if that works?

juliohm:
Handling CRS and units is really tricky sometimes. Maybe we will need a combination of `LengthUnit` and `Proj`.

To give you some additional context. There are `Basic` CRS like `Cartesian`, `Polar`, etc which accept any length unit, and there are `Geographic` and `Projected` CRS which only accept the units of their EPSG specification.

If that is something that you don't have time to dive in, we can take a look tomorrow.

kylebeggs:
It seems using `Proj` works. All tests pass now, including the new one.

src/sideof.jl
```diff
@@ -58,6 +58,7 @@ Possible results are `IN`, `OUT` or `ON` the `ring`.
 """
 function sideof(point::Point, ring::Ring)
   assertion(CoordRefSystems.ncoords(crs(point)) == 2, "points must have 2 coordinates")
+  point = Proj(crs(point))(point)
```
juliohm:
```suggestion
  point = Proj(crs(ring))(point)
```
I believe we want to project to the crs of the other geometry?

kylebeggs: Yes, that was a mistake

test/sideof.jl
```diff
@@ -15,6 +15,13 @@
   pts = [p1, p2, p3]
   @test sideof(pts, c) == [IN, OUT, ON]
 
+  c_mm = LengthUnit(Unitful.mm)(c)
+  @test sideof(p1, c) == IN
+  @test sideof(p2, c) == OUT
+  @test sideof(p3, c) == ON
+  pts = [p1, p2, p3]
+  @test sideof(pts, c) == [IN, OUT, ON]
```
juliohm: tests are still using the old `c` variable instead of `c_mm`

kylebeggs: dang, trying to go too fast with this PR. fixed.

src/sideof.jl
```diff
@@ -58,6 +58,7 @@ Possible results are `IN`, `OUT` or `ON` the `ring`.
 """
 function sideof(point::Point, ring::Ring)
   assertion(CoordRefSystems.ncoords(crs(point)) == 2, "points must have 2 coordinates")
+  point = Proj(crs(ring))(point)
```
juliohm:
```suggestion
  point′ = point |> Proj(crs(ring))
```

we don't usually overwrite variables when the type changes because that leads to type instability

src/sideof.jl
```diff
@@ -58,6 +58,7 @@ Possible results are `IN`, `OUT` or `ON` the `ring`.
...
   if nvertices(ring) ≤ 1000 || Threads.nthreads() == 1
     _sideofserial(point, ring)
```
juliohm, line 63:
```suggestion
    _sideofserial(point′, ring)
```

test/sideof.jl
```diff
@@ -15,6 +15,13 @@
   pts = [p1, p2, p3]
   @test sideof(pts, c) == [IN, OUT, ON]
 
+  c_mm = LengthUnit(Unitful.mm)(c)
```
juliohm:
```suggestion
  c′ = c |> LengthUnit(Unitful.mm)
```

juliohm (approved):

juliohm:
I think we will need a helper function that combines `Proj` and `LengthUnit`. Something like:

```julia
fixed1, fixed2, ... = adjust(geom1, geom2, ...)
```

that converts all geometries to the `CRS` and lenght unit of the `geom1`.

juliohm:
Alternatively, this could go in CoordRefSystems.jl as a new `crspromote`. We will think about it @kylebeggs, thanks for reporting the issue and attempting a fix.

kylebeggs:
It seems with your changes, `point` does not get converted to the same units as `ring`. Maybe `constructor` is to blame?

juliohm:
Yes, we realized that we need an additional function in CoordRefSystems.jl to promote different CRS correctly. After we finish the necessary work in that package, we can come back here to fix this issue more generally.

We believe that this type of promotion will be needed in various places in the future.

kylebeggs:
Is there any way to push a fix now and then refactor once you add the necessary changes in CoordRefSystems.jl? I think since this can be classified as a bug, it should be fixed ASAP. I actually caught this bug through unit tests in some packages at my work so this has become a blocker for me. I guess I can do the unit promotion myself... but I believe the lowest level packages should be responsible for the fix.

juliohm:
We will take a look at it tomorrow ASAP. Top priority.

Em qui., 22 de ago. de 2024, 16:38, Kyle Beggs ***@***.***>
escreveu:

juliohm:
@kylebeggs this PR should be ready now. We just need to wait for the CRS.jl version to propagate through the General registry. The `Proj` now accepts concrete CRS types.

src/utils/crs.jl
```diff
@@ -9,9 +9,8 @@ Point at the end of the vector `v` with the same CRS of `g`.
 """
 function withcrs(g::GeometryOrDomain, v::StaticVector)
   C = crs(g)
-  cart = Cartesian{datum(C)}(Tuple(v))
-  ctor = CoordRefSystems.constructor(C)
-  Point(convert(ctor, cart))
+  D = datum(C)
+  Point(convert(C, Cartesian{D}(v...)))
```
eliascarv, line 13: Maybe like this?
```suggestion
  coords = Cartesian{datum(C)}(Tuple(v))
  Point(convert(C, coords))
```

eliascarv (approved):

PR #1020, opened 2024-08-23
New parameterization functions for Cone and ConeSurface

### LINKED ISSUE #1011: Incorrect parameterization of a `ConeSurface`
Calculating the `measure` of a `ConeSurface` seems to produce incorrect results. These results appear to be invariant w.r.t. the base radius but approximate the value expected for a cone whose base has unit radius.

## MWE

```julia
origin = Point(0, 0, 0)
z_axis = Vec(0, 0, 1)
xy_plane = Plane(origin, z_axis)
apex = Point(0, 0, 1)

# Test ConeSurface with radius 1.0
base_disk_10 = Disk(xy_plane, 1.0)
cs_10 = ConeSurface(base_disk_10, apex)

# Test ConeSurface with radius 2.5
base_disk_25 = Disk(xy_plane, 2.5)
cs_25 = ConeSurface(base_disk_25, apex)

# Analytic solution for surface area of a cone (including base)
# let s^2 = r^2 + h^2
#   A = (π * r^2) + (π * r * s)
cs_10_area = 3.142 + 4.443
  # == 7.5844
cs_25_area = 19.635 + 21.147
  # == 40.782

measure(cs_10)
  # == 7.568912593985476 m^2

measure(cs_25)
  # == 7.568912593985476 m^2
```

mikeingold:
These additions are intended to fix #1011 and simultaneously implement the missing parameterization function for `::Cone`. Both of these functions are practically identical, with the `Cone` function slightly more general to handle the radius argument. Essentially, they parametricly locate a Point `c` on the cone's base, project a line segment from the apex to that Point, and then select the appropriate Point along that line segment. This format should also be relatively easy to write AD rules for.


mikeingold:
Just need to add some relevant tests (planning for tomorrow) and I'll push this one for Review.

test/primitives.jl
```diff
@@ -1057,6 +1058,9 @@
     @test paramdim(c) == 3
     @test crs(c) <: Cartesian{NoDatum}
     @test Meshes.lentype(c) == ℳ
+    @test c(0, 0, 0) ≈ Point(a)
+    @test c(0, 0, 1) ≈ p.p
```
juliohm:
```suggestion
    @test c(0, 0, 1) ≈ centroid(p)
```

test/primitives.jl
```diff
@@ -1048,6 +1048,7 @@
     @test crs(c) <: Cartesian{NoDatum}
     @test Meshes.lentype(c) == ℳ
     @test boundary(c) == ConeSurface(d, a)
+    @test_throws DomainError c(0, 0, nextfloat(1.0))
```
juliohm:
```suggestion
    @test_throws DomainError c(T(0), T(0), nextfloat(T(1.0)))
```

test/primitives.jl
```diff
@@ -1057,6 +1058,9 @@
     @test paramdim(c) == 3
     @test crs(c) <: Cartesian{NoDatum}
     @test Meshes.lentype(c) == ℳ
+    @test c(0, 0, 0) ≈ Point(a)
+    @test c(0, 0, 1) ≈ p.p
+    @test c(T(0.25), 1, 1) ≈ cart(0, 2, 0)
```
juliohm:
```suggestion
    @test c(T(0.25), T(1.0), T(1.0)) ≈ cart(0, 2, 0)
```

test/primitives.jl
```diff
@@ -1122,6 +1126,7 @@
     @test crs(s) <: Cartesian{NoDatum}
     @test Meshes.lentype(s) == ℳ
     @test isnothing(boundary(s))
+    @test_throws DomainError s(0, nextfloat(1.0))
```
juliohm:
```suggestion
    @test_throws DomainError s(T(0.0), nextfloat(T(1.0)))
```

src/geometries/primitives/conesurface.jl
```diff
@@ -31,17 +31,19 @@ apex(c::ConeSurface) = c.apex
...
   isapprox(c₁.base, c₂.base; atol, kwargs...) && isapprox(c₁.apex, c₂.apex; atol, kwargs...)
 
-function (c::ConeSurface)(φ, h)
-  T = numtype(lentype(c))
-  if (φ < 0 || φ > 1) || (h < 0 || h > 1)
-    throw(DomainError((φ, h), "c(φ, h) is not defined for φ, h outside [0, 1]²."))
+function (conesurface::ConeSurface)(uφ, uh)
+  T = numtype(lentype(conesurface))
+  if (uφ < 0 || uφ > 1) || (uh < 0 || uh > 1)
+    throw(DomainError((uφ, uh), "c(φ, h) is not defined for φ, h outside [0, 1]²."))
```
juliohm:
```suggestion
function (c::ConeSurface)(φ, h)
  T = numtype(lentype(c))
  if (φ < 0 || φ > 1) || (h < 0 || h > 1)
    throw(DomainError((φ, h), "c(φ, h) is not defined for φ, h outside [0, 1]²."))
```

src/geometries/primitives/conesurface.jl
```diff
@@ -31,17 +31,19 @@ apex(c::ConeSurface) = c.apex
...
+
+  a = conesurface.apex
+  R = conesurface.base.radius
+  b = conesurface.base.plane.p
+  û = normalize(conesurface.base.plane.u)
+  v̂ = normalize(conesurface.base.plane.v)
+  sφ, cφ = sincospi(2uφ)
+  c = b + Vec(R * cφ * û) + Vec(R * sφ * v̂)
+  h̄ = uh * (c - a)
+  a + h̄
```
juliohm: Please use the public api of the geometries instead of accessing fields directly as they may change in the future. Given that this is a function of the `ConeSurface`, you can do `c.base`, but `c.base.radius` is accessing an implementation detail of the `base`. You can write `radius(c.base)` instead.

src/geometries/primitives/cone.jl
```diff
@@ -34,3 +34,21 @@ halfangle(c::Cone) = atan(radius(base(c)), height(c))
 
 Base.isapprox(c₁::Cone, c₂::Cone; atol=atol(lentype(c₁)), kwargs...) =
   isapprox(boundary(c₁), boundary(c₂); atol, kwargs...)
+
+function (cone::Cone)(uφ, ur, uh)
+  T = numtype(lentype(cone))
+  if (uφ < 0 || uφ > 1) || (ur < 0 || ur > 1) || (uh < 0 || uh > 1)
+    throw(DomainError((uφ, ur, uh), "c(φ, r, h) is not defined for φ, r, h outside [0, 1]³."))
+  end
```
juliohm:
```suggestion
function (c::Cone)(φ, r, h)
  T = numtype(lentype(c))
  if (φ < 0 || φ > 1) || (r < 0 || r > 1) || (h < 0 || h > 1)
    throw(DomainError((φ, r, h), "c(φ, r, h) is not defined for φ, r, h outside [0, 1]³."))
  end
```

src/geometries/primitives/cone.jl
```diff
@@ -34,3 +34,21 @@ halfangle(c::Cone) = atan(radius(base(c)), height(c))
...
+  a = cone.apex
+  R = cone.base.radius
+  b = cone.base.plane.p
+  û = normalize(cone.base.plane.u)
+  v̂ = normalize(cone.base.plane.v)
+  sφ, cφ = sincospi(2uφ)
+  r = R * ur
+  c = b + Vec(r * cφ * û) + Vec(r * sφ * v̂)
+  h̄ = uh * (c - a)
+  a + h̄
```
juliohm: same comment here about internal details. please use the public api of the geometries that are not `Cone`

juliohm (changes requested):
Thank you @mikeingold ❤️ Super helpful fix.

test/primitives.jl
```diff
@@ -1057,6 +1058,9 @@
     @test paramdim(c) == 3
     @test crs(c) <: Cartesian{NoDatum}
     @test Meshes.lentype(c) == ℳ
+    @test c(0, 0, 0) ≈ Point(a)
+    @test c(0, 0, 1) ≈ centroid(p)
```
juliohm:
```suggestion
    @test c(T(0), T(0), T(0)) ≈ Point(a)
    @test c(T(0), T(0), T(1)) ≈ centroid(p)
```

test/primitives.jl
```diff
@@ -1131,6 +1136,9 @@
     @test paramdim(c) == 2
     @test crs(c) <: Cartesian{NoDatum}
     @test Meshes.lentype(c) == ℳ
+    @test c(0, 0) ≈ Point(a)
+    @test c(0, 1) ≈ cart(2, 0, 0)
+    @test c(T(0.25), 1) ≈ cart(0, 2, 0)
```
juliohm:
```suggestion
    @test c(T(0), T(0)) ≈ Point(a)
    @test c(T(0), T(1)) ≈ cart(2, 0, 0)
    @test c(T(0.25), T(1)) ≈ cart(0, 2, 0)
```

mikeingold:
I had a few minutes to pop in and commit a number of the simpler suggestions. For the others I’ll need a few more minutes to verify that all the names are updated as needed, hopefully  by later today.

src/geometries/primitives/conesurface.jl
juliohm: Here you can use the parametrization of the `Disk` as the public api to generate points.

mikeingold: So, long-term I really want to get AD rules defined for all of these parametric functions. From what I'm reading and a little bit of experimentation, it looks like we're going to have to define explicit AD rules for all of these functions to get that working.

One of my goals with the approach I used here was to keep the math concrete and simple to code up analytic derivation rules for. By contrast, if we rely on abstract API's and don't control for lower-level implementation details, then we'd need to propagate a more complicated chain-rules analysis (beyond my AD skills here).

juliohm: > So, long-term I really want to get AD rules defined for all of these parametric functions

Very nice 👍🏽 

> From what I'm reading and a little bit of experimentation, it looks like we're going to have to define explicit AD rules for all of these functions to get that working.

Can you please clarify how using the public API interferes with AD rules? The function calls should produce the exact same inline code if I understand it correctly.

juliohm:
@mikeingold we merged another PR that refactored the entire test infrastructure to facilitate future contributions. Can you please resolve the conflicts in this PR or start a fresh one with the suggested fixes?

Running specific tests on VSCode is much nicer now.

src/geometries/primitives/conesurface.jl
mikeingold: I'm not an AD SME, so if a simpler solution exists I'm all for it. I've yet to get any AD package to calculate derivatives for any of these parametric functions out-of-the-box, even for a simple one like `(s::Segment)(t)` whose derivative should just be `(b - a)::Vec`.

It seems like we need to define `ChainRulesCore.rrule`s for these functions with analytic solutions. Under that scenario, the function above is trivially derived w.r.t. its arguments, but if we have to rely on other parametric functions and treat them as public API, then we need to figure out significantly more complicated rules.

I'm just not sure that the theoretical maintenance burden vs perfect abstractions is worth the added complexity, at least in this particular case.

juliohm: From what I understand, if we express `cylinder(u,v,w)` in terms of `disk(u, v)` plus other expressions, it should be easy to define the chain rules for the disk parametrization and reuse in the cylinder parametrization.

Let's stay consistent with the public API and revisit this issue in the future when you start working with AD definitions. The entire Meshes.jl project is written that way to facilitate fixes and refactorization. If functions in different source files start to use internal fields of structs defined elsewhere, things can get out of control very quickly.

mikeingold:
Parameterization changes implemented. I reverted the new tests so that I could merge with the latest updates. Just need to re-add tests.

mikeingold:
@juliohm All tests pass except for "Format suggestions" which outputs the error text
```
Some files have not been formatted !!!
test/primitives.jl
```

I'm not familiar enough with this job to know what this indicates.

juliohm:
Thank you @mikeingold ! I adjusted the code to start with `r` in the `Cone` to match the convention in other primitive geometries. I also reversed the `h` axis so that the orientation of normals is correct in the discretization routines.

PR #1056, opened 2024-09-11
Add `showsegments` support for `vizgridfallback!`

eliascarv:


ext/MeshesMakieExt.jl
```diff
@@ -42,6 +42,8 @@ Makie.args_preferred_axis(g::Geometry) = embeddim(g) === 3 ? Makie.LScene : Maki
 Makie.args_preferred_axis(d::Domain) = embeddim(d) === 3 ? Makie.LScene : Makie.Axis
 Makie.args_preferred_axis(::AbstractVector{<:Vec{Dim}}) where {Dim} = Dim === 3 ? Makie.LScene : Makie.Axis
 
+function vizfacets! end
+
```
juliohm: no need to add this empty definition here, it is an internal function

ext/geometryset.jl
```diff
@@ -52,14 +52,20 @@ function vizgset!(plot, ::Type{<:𝔼}, ::Val{0}, ::Val, geoms::ObservableVector
 end
 
 function vizgset!(plot, ::Type{<:𝔼}, ::Val{1}, ::Val, geoms, colorant)
+  showpoints = plot[:showpoints]
+
   meshes = Makie.@lift discretize.($geoms)
   vizmany!(plot, meshes, colorant)
-  showfacets1D!(plot, geoms)
+
+  if showpoints[]
```
juliohm, line 35: are you positive that adding this new conditional retains the same behavior as before?

juliohm (approved):

PR #1057, opened 2024-09-12
Add a heuristic to use uv coords in `vizgridfallback!`

eliascarv:


ext/grid.jl
```diff
@@ -47,24 +47,30 @@ function vizgridfallback!(plot, M, pdim, edim)
...
+  colorant = Makie.@lift process($color, $colormap, $colorrange, $alpha)
+
+  # number of vertices, elements and colors
+  nverts = Makie.@lift nvertices($grid)
+  nelems = Makie.@lift nelements($grid)
+  ncolor = Makie.@lift $colorant isa AbstractVector ? length($colorant) : 1
+
+  # visualize quadrangle mesh with texture using uv coords
+  # plots with uv coords are always interpolated,
+  # so it is only used in the case ncolor == nelems
```
juliohm:
```suggestion
  # so it is only used in the case ncolor == nverts
```

ext/grid.jl
```diff
@@ -47,24 +47,30 @@ function vizgridfallback!(plot, M, pdim, edim)
...
+  # when there is a large number of elements
```
juliohm:
```suggestion
  # or when there is a large number of elements
```

juliohm (approved):

PR #1058, opened 2024-09-12
`GeometrySet`: handle heterogeneous `CRS`

### LINKED ISSUE #948: Review `GeometrySet` to handle heterogeneous `CRS` case
To avoid inference issues, and to make sure that all geometries share a common `CRS` after construction.

eliascarv:
closes #948

### DIFF
--- a/src/domains/sets.jl
+++ b/src/domains/sets.jl
@@ -21,17 +21,14 @@ end
 
 # constructor with iterator of geometries
 function GeometrySet(geoms)
-  geoms′ = map(identity, geoms) # narrow types
-
   # project all geometries to the same CRS if necessary
-  gs = if allequal(crs(g) for g in geoms′)
-    geoms′
+  fun = if allequal(crs(g) for g in geoms)
+    identity # narrow types
   else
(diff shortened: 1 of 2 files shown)


src/domains/sets.jl
```diff
@@ -20,7 +20,19 @@ struct GeometrySet{M<:Manifold,C<:CRS,G<:Geometry{M,C}} <: Domain{M,C}
...
+
+  # project all geometries to the same CRS if necessary
+  gs = if allequal(crs(g) for g in geoms′)
+    geoms′
+  else
+    proj = Proj(crs(first(geoms′)))
+    map(proj, geoms′)
+  end
+
+  GeometrySet(gs)
```
juliohm:
```suggestion
  # project all geometries to the same CRS if necessary
  gs = if allequal(crs(g) for g in geoms)
    map(identity, geoms) # narrow types
  else
    proj = Proj(crs(first(geoms)))
    map(proj, geoms)
  end

  GeometrySet(gs)
```

juliohm (approved):

PR #1069, opened 2024-09-18
Add method for `measure` of `Segment` with Ellipsoid manifold

eliascarv:


src/measures.jl
```diff
@@ -75,7 +75,18 @@ function measure(t::Torus)
...
+# TODO: Haversine returns the shortest distance between two points
+# this is not always equal to the distance between two directed points
+function measure(s::Segment{<:🌐})
+  T = numtype(lentype(s))
+  🌎 = ellipsoid(datum(crs(s)))
+  r = numconvert(T, majoraxis(🌎))
+
+  A, B = extrema(s)
+  evaluate(Haversine(r), A, B)
+end
```
juliohm:
```suggestion
# TODO: replace Haversine by an appropriate geodesic distance
# that considers the west-east orientation of segments
function measure(s::Segment{<:🌐})
  T = numtype(lentype(s))
  🌎 = ellipsoid(datum(crs(s)))
  r = numconvert(T, majoraxis(🌎))

  evaluate(Haversine(r), extrema(s)...)
end
```

juliohm (approved):

PR #1075, opened 2024-09-19
Add `MaxLengthDiscretization` for `Geometry`

eliascarv:


src/discretization/maxlength.jl
```diff
@@ -27,6 +27,15 @@ end
 discretize(chain::Chain, method::MaxLengthDiscretization) =
   mapreduce(s -> discretize(s, method), merge, segments(chain))
 
+function discretize(polygon::Polygon, method::MaxLengthDiscretization)
+  needrefine(e) = perimeter(e) > method.length * nvertices(e)
```
juliohm:
```suggestion
  iscoarse(e) = perimeter(e) > method.length * nvertices(e)
```

PR #1078, opened 2024-09-20
Update the method definition of `centroid` of grids

eliascarv:


src/utils.jl
```diff
@@ -2,9 +2,11 @@
 # Licensed under the MIT License. See LICENSE in the project root.
 # ------------------------------------------------------------------
 
-# auxiliary type for dispatch purposes
+# auxiliary types for dispatch purposes
 const GeometryOrDomain = Union{Geometry,Domain}
```
```suggestion
```

juliohm (approved):

PR #1093, opened 2024-09-29
Add `ParametrizedCurve`

JoshuaLampert:
This PR adds another primitive for a parametrized curve $\gamma: [a, b]\to\mathbb{R}^d$ , where the parametrization comes from a user-defined function. Please tell me if I missed any method, which should be added. I'll add tests later after the interface is ironed out.


src/geometries/primitives/parametrizedcurve.jl
```diff
@@ -0,0 +1,41 @@
...
+end
+
+paramdim(::ParametrizedCurve) = 1
+start_parameter(curve::ParametrizedCurve) = curve.a
+start_point(curve::ParametrizedCurve) = curve(curve.a)
+end_parameter(curve::ParametrizedCurve) = curve.b
+end_point(curve::ParametrizedCurve) = curve(curve.b)
+
+function (curve::ParametrizedCurve)(t)
+    if t < start_parameter(curve) || t > end_parameter(curve)
```
JoshuaLampert: I'm not sure here. It seems like when calling `viz` on a `curve` values in `[0, 1]` are plugged in. What is the best way to deal with general start and end parameters? We could also remove `a` and `b`  from the struct and tell the user to always give a curve defined on `[0, 1]`, but I would like to keep the flexibility of a general interval for convenience.

src/geometries/primitives/parametrizedcurve.jl
```diff
@@ -0,0 +1,41 @@
...
+"""
+    ParametrizedCurve(a, b, func)
+
+A parametrized curve is a curve defined by a function `func` that maps a parameter `t` to a `Point` in space.
+The parameter `t` is defined in the interval `[a, b]`.
+
+## Examples
+
+```julia
+ParametrizedCurve(0.0, 1.0, t -> Point(cospi(2t), sinpi(2t)))
```
juliohm: Do you envision other use cases that involve non-Cartesian coordinates or this primitive is mostly for parametrizations of the form t -> x(t), y(t), z(t)?

Regarding the range of t, do you plan use cases with intervals other than [0,1]? Classically, a parametrized curve alpha is a function alpha: [0,1] --> R^n. If it is differentiable, it is often useful to reparametrize the curve by the arc length. See Differential Geometry of Curves and Surfaces by Manfredo (https://www.amazon.com/Differential-Geometry-Curves-Surfaces-Mathematics/dp/0486806995).

I wonder what would be the most ergonomic definition of constructors for the applications you have in mind. Maybe users would prefer to simply define the functions x(t), y(t), ... directly instead of manually handing the `Point` and CRS types?

```julia
x(t) = ...
y(t) = ...

c = ParametrizedCurve(x, y)
```

src/geometries/primitives/parametrizedcurve.jl
```diff
@@ -0,0 +1,41 @@
...
+    p = func(a)
+    M = manifold(p)
+    C = crs(p)
+    new{M,C,T,typeof(func)}(a, b, func)
+  end
+end
+
+paramdim(::ParametrizedCurve) = 1
+startparameter(curve::ParametrizedCurve) = curve.a
+startpoint(curve::ParametrizedCurve) = curve(curve.a)
```
juliohm:
```suggestion
minimum(curve::ParametrizedCurve) = curve(curve.a)
```

We use `minimum`, `maximum` and `extrema` in other primitives to retrieve the extrema. Maybe we should use the same names here.

src/geometries/primitives/parametrizedcurve.jl
```diff
@@ -0,0 +1,41 @@
...
+  function ParametrizedCurve(a::T, b::T, func) where {T<:Real}
+    p = func(a)
+    M = manifold(p)
+    C = crs(p)
+    new{M,C,T,typeof(func)}(a, b, func)
+  end
+end
+
+paramdim(::ParametrizedCurve) = 1
+startparameter(curve::ParametrizedCurve) = curve.a
```
juliohm: If we always assume [0,1], we remove the need for startparameter and endparameter

juliohm (commented):
Appreciate if you can elaborate on the use cases of this primitive so that we can fine tune the best constructors.

If we restrict t in [0,1] and constrain the type to Cartesian parametrizations, we can think of constructing the curve from parametric x(t), y(t) (and possibly z(t)) alone. Otherwise, we can consider other intervals for t and other CRS like Polar, Spherical, Cylindrical coordinates in which case, the parametrizations would be r(t), theta(t), etc.

JoshuaLampert:
Thanks a lot for your feedback @juliohm. I structure your comments into two points:
1. The parameter range: I see that always having $t\in[0, 1]$ is simple, convenient for the implementation, and often also the standard. However, often I also find a curve defined on a general interval $[a, b]$ (I think I've also learned it that way at university or see e.g. https://en.wikipedia.org/wiki/Line_integral or also https://en.wikipedia.org/wiki/Curve). Of course, you can always simply transform every interval $[a, b]$ into $[0, 1]$ pretty easily, but it adds an inconvenience if one happens to have a curve on another interval, which one wants to define quickly without thinking of a transformation first. One alternative to simply removing the general start and end parameter I thought of is that evaluating the curve is only defined for $t\in[0,1]$, i.e. `curve(0.0)` would give `curve.func(a)` for instance. In other words we perform the transformation internally. On the other hand this is also not optimal as it might be confusing that you provide start and end parameters for the parameter range, but in the end the curve can still only be evaluated in $[0, 1]$. If you also don't see any better option, I would be fine with both versions, which means I let you decide.
2. The CRS: I do plan to use other CRS than `Cartesian`, see also my use case I have in mind below. Thus, I would like the user to explicitly create a `Point`, such that we know the CRS. Having said that, we can of course think of adding another convenience constructor taking functions `x(t)`, `y(t)` and potentially `z(t)`, which internally creates a `Point` in `Cartesian` coordinates by default.

One use case I have in mind is the following. I would like to compute integrals along non-trivial curves in 2D like the one in Figure 7 of https://academic.oup.com/imajna/article/43/3/1616/6586010. Looking at [the source code](https://github.com/jglaubitz/positive_CFs/blob/main/generate_points.m#L179) of this paper the curve can be described by the following curve in polar coordinates: $\gamma: [-\pi, \pi] \to\mathbb{R}^2, t\mapsto [\rho(t), \phi(t)]^T = [1 - \frac{1}{3} \sin(2t)^2, t]^T$. It would be very nice if this could simply be implemented as
```julia
f(theta) = 1 - 1/3 * sin(2*theta)^2
curve = ParametrizedCurve(-pi, pi, t  -> Point(Polar(f(t), t)))
```
The following already works with the current implementation, which is good (but not as nice as the above, see point 1):
```julia
curve = ParametrizedCurve(0.0, 1.0, t -> Point(Polar(f(2pi * t - pi), 2pi * t - pi)))
```
E.g. `viz(curve)` shows
curve

src/geometries/primitives/parametrizedcurve.jl
JoshuaLampert: Done in e61f8fc378be948a9c60955013eef815e82a2cf3.

mikeingold:
> 1. One alternative to simply removing the general start and end parameter I thought of is that evaluating the curve is only defined for $t ∈ [ 0, 1]$ i.e. `curve(0.0)` would give `curve.func(a)` for instance. In other words we perform the transformation internally. On the other hand this is also not optimal as it might be confusing that you provide start and end parameters for the parameter range, but in the end the curve can still only be evaluated in $[0,1]$. If you also don't see any better option, I would be fine with both versions, which means I let you decide.

This is a tricky question. All of Meshes.jl's current parametric functions, and other features that utilize them, assume an interface with domain $[0,1]^n$. However, I think it would be more user-friendly to allow arbitrary end-points, i.e $u = [a, b]$.

If we maintained support for `curve(t)` where $t ∈ [ 0, 1]$ but translated this internally to use another domain, i.e. $u = [a, b]$ this could somewhat bridge the gap. Maybe we just need another way to call the curve directly with `u` coordinates? Brainstorming:
- Maybe a new function, e.g. `parametric(curve, u)`
- Maybe dispatch differently, e.g. `curve(u, ::Symbol)`?

juliohm:
The current approach where the interval [a,b] is handled internally in the struct is flexible and doesn't compromise the assumption that uv coords are unitless coordinates in the range [0,1]. Perhaps we just need to make this interval [a,b] something of secondary importance with [0,1] as the default. What about the following constructor:

```julia
ParametrizedCurve(f, ab=(0.0, 1.0))
```

That way you could write the example as

```julia
f(theta) = 1 - 1/3 * sin(2*theta)^2

# explicit ab interval
ParametrizedCurve(t  -> Point(Polar(f(t), t)), (-pi, pi))

# default interval
ParametrizedCurve(t -> Point(Polar(f(2pi * t - pi), 2pi * t - pi)))
```

JoshuaLampert:
Sounds good @juliohm. I pushed your suggestion.

src/geometries/primitives.jl
```diff
@@ -31,3 +31,4 @@ include("primitives/frustum.jl")
 include("primitives/frustumsurface.jl")
 include("primitives/paraboloidsurface.jl")
 include("primitives/torus.jl")
+include("primitives/parametrizedcurve.jl")
```
juliohm: please move this include to after bezier.jl

src/geometries/primitives/parametrizedcurve.jl
```diff
@@ -0,0 +1,48 @@
+# ------------------------------------------------------------------
+# Licensed under the MIT License. See LICENSE in the project root.
+# ------------------------------------------------------------------
+
+"""
+    ParametrizedCurve(func, ab = (0.0, 1.0))
+
+A parametrized curve is a curve defined by a function `func` that maps a parameter `t` to a `Point` in space.
+The parameter `t` is defined in the interval `[a, b]`. The curve can only be evaluated for `t` in the
+interval `[0, 1]`.
```
juliohm:
```suggestion
A curve defined by a function `func` that maps a parameter `t` in the real line to a
`Point` in space. The parameter `range` can be specified, and defaults to `[0,1]` otherwise.
```

src/geometries/primitives/parametrizedcurve.jl
```diff
@@ -0,0 +1,48 @@
+# ------------------------------------------------------------------
+# Licensed under the MIT License. See LICENSE in the project root.
+# ------------------------------------------------------------------
+
+"""
+    ParametrizedCurve(func, ab = (0.0, 1.0))
```
juliohm:
```suggestion
    ParametrizedCurve(func, range = (0.0, 1.0))
```

src/geometries/primitives/parametrizedcurve.jl
```diff
@@ -0,0 +1,48 @@
...
+## Examples
+
+```julia
+ParametrizedCurve(t -> Point(cospi(2t), sinpi(2t)))
+```
+"""
+struct ParametrizedCurve{M<:Meshes.Manifold,C<:Meshes.CRS,T<:Real,F<:Function} <: Primitive{M,C}
+  a::T
+  b::T
+  func::F
```
juliohm: Let's store the fields `func::F` and `range::R` with `F<:Function` and `R<:Tuple` in that order, to match the constructor.

src/geometries/primitives/parametrizedcurve.jl
```diff
@@ -0,0 +1,48 @@
...
+    p = func(a)
+    M = manifold(p)
+    C = crs(p)
+    T = typeof(a)
+    new{M,C,T,typeof(func)}(a, b, func)
+  end
+end
+
+paramdim(::ParametrizedCurve) = 1
+startparameter(curve::ParametrizedCurve) = curve.a
```
juliohm: startparameter and endparameter need to be exposed as part of the public API? I was thinking that once created, the curve would be evaluated with the unit interval range.

If these are needed, we can consider adding a method `Base.range(c::Curve) = c.range[1]:c.range[2]` instead.

src/geometries/primitives/parametrizedcurve.jl
```diff
@@ -0,0 +1,48 @@
...
+Base.extrema(curve::ParametrizedCurve) = curve(0.0), curve(1.0)
+
+function (curve::ParametrizedCurve)(t)
+  if t < 0.0 || t > 1.0
+    throw(DomainError(t, "c(t) is not defined for t outside [0, 1]."))
+  end
+  a = startparameter(curve)
+  b = endparameter(curve)
+  θ = a + t * (b - a)
+  curve.func(θ)
```
juliohm:
```suggestion
  a, b = curve.range
  curve.func(a + t * (b - a))
```

juliohm (changes requested):
Just a few minor suggestions to consider before we start adding tests and docs.

mikeingold:
Not to overly complicate this suggestion, but I'll just throw this out
there...

If we want to define a one-dimensional geometry defined by user-supplied
parametric functions, it seems logical that we might consider extending
this in the future to also accommodate parametric surfaces and volumes. If
we were to go that far, would we need to define new types, e.g.
ParametricSurface, or could these all fall under the same basic type, e.g.
Box?

JoshuaLampert:
I'll address your suggestions once I am back at my workstation @juliohm.
Thanks for starting the discussion about higher dimensional user-supplied parametric functions @mikeingold. I also thought about that, but assumed we would need different types for that. So I wanted to postpone that for now. But you might be right that it could be good to just have the same struct for all parametric dimensions, which means we should discuss it already in this PR to avoid breaking changes. I would be interested in your thoughts, @juliohm.

mikeingold:
Agreed. I don't want this suggestion to halt progress on this effort, but
if we're interested in extending this work to more dimensions then it might
be worth having a roadmap in mind to avoid breaking changes.

It's not unprecedented for one name to operate in different numbers of
dimensions, e.g. Box, Sphere, and Ball, that might be used for inspiration.
The only real difference between a parametric curve and surfaces/volumes
would be the number of input arguments to the parametric function, and the
sizes of `a` and `b`.

src/geometries/primitives/parametrizedcurve.jl
JoshuaLampert: Reading that as a user, I would assume I can evaluate my curve in the interval [a, b] if I use a custom `range`, but that's not true. I would like to make clear that the curve can only be evaluated in the default range [0, 1].

src/geometries/primitives.jl
JoshuaLampert: Done in [9822d71](https://github.com/JuliaGeometry/Meshes.jl/pull/1093/commits/9822d71f3cd116dce52f02a96632646a726713de).

src/geometries/primitives/parametrizedcurve.jl
JoshuaLampert: Done in [e830105](https://github.com/JuliaGeometry/Meshes.jl/pull/1093/commits/e830105ee365337755550969f06c79166b245051).

src/geometries/primitives/parametrizedcurve.jl
JoshuaLampert: I think they are not needed. I removed them in [e830105](https://github.com/JuliaGeometry/Meshes.jl/pull/1093/commits/e830105ee365337755550969f06c79166b245051).

juliohm:
> If we want to define a one-dimensional geometry defined by user-supplied parametric functions, it seems logical that we might consider extending this in the future to also accommodate parametric surfaces and volumes. If we were to go that far, would we need to define new types, e.g. ParametricSurface, or could these all fall under the same basic type, e.g. Box?

Very good point. I wonder what is the best design here. We are trying to achieve a continuous representation with any number of parametric coordinates `func = (u, v, w, ...) -> Point(f(u, v, w...), g(u, v, w...), h(u, v, w...))`. The number of parameters determines the parametric dimension as usual (1 = curve, 2 = surface, 3 = volume). The only thing that is unique about `func` is that it must return a `Point`. Other than that, it is a normal Julia `Function`.

If we think of the ranges of the parameters `(u, v, w...)` as sides of a `Box`, then we can interpret the `Function` as a mapping from a n-dimensional `Box` to a n-dimensional shape. `TransformedGeometry(geom, transform)` is a related concept that stores the result of the `transform` on the `geom` lazily. Transforms are not the same thing as functions though.

juliohm:
I think I have an idea. The `TransformedGeometry` handles the lazy transformation of a simpler geometry. We can achieve a representation of curves, surfaces etc by mapping `Box` with a new `Morphological(func)` (or similar name) transform. The `func` in this case has a simpler form, it takes a coordinates object and returns another coordinates object:

```julia
function f(c::Cartesian)
  x, y = c.x, c.y
  Polar(x, x / y)
end
```

The `Morphological` coordinate transform will simply wrap the result into a new `Point`. After we have this general machinery to convert `Box` into other n-dimensional shapes, we can think of syntax sugar and helper functions to facilitate the creation of `TransformedGeometry(geom, Morphological(func))`.

Maybe that will cover the use cases you have in mind? Should we experiment with the `Morphological` transform in a separate PR and come back here later?

JoshuaLampert:
Sounds interesting. I'll need to think about it and experiment a bit (I'm currently on vacation, so I might need a bit of time).

JoshuaLampert:
In #1100 we came to the conclusion that having a new `Morphological` transform is not suitable to tackle this problem. Thus, let's come back to adding a new primitive as proposed in this PR. I am still wondering whether it makes sense to have a `ParametrizedCurve` type and then possible also define `ParametrizedSurface` and `ParametrizedVolume` types later or if we could define a new primitive, which is dimension independent (something like `ParametrizedGeometry`). This means the user can provide a function taking any number of parameters, which defines the parametrization of the resulting geometry. While many aspects of this PR are probably generalizable to arbitrary dimensions, I'm not sure everything is so straightforward. I am specifically thinking of a general version of `boundary`. I'm not sure how to define that for a general unknown parametrization in unknown dimension. If you think it would be possible to define something like this, I think a general version would be nice. Otherwise if it is too hard, I would opt for defining `ParametrizedCurve` for 1D etc.

src/geometries/primitives/parametrizedcurve.jl
```diff
@@ -0,0 +1,42 @@
+# ------------------------------------------------------------------
+# Licensed under the MIT License. See LICENSE in the project root.
+# ------------------------------------------------------------------
+
+"""
+    ParametrizedCurve(func, range = (0.0, 1.0))
+
+A parametrized curve is a curve defined by a function `func` that maps a parameter `t` to a `Point` in space.
+The parameter `t` is defined in the interval `[a, b]`. The curve can only be evaluated for `t` in the
+interval `[0, 1]`.
```
juliohm:
```suggestion
A parametrized curve is a curve defined by a function `func` that maps a
(unitless) parameter `t` in the given `range` to a `Point` in space.

## Example

```julia
ParametrizedCurve(t -> Point(cos(t), sin(t)), (0, 2pi))
```
```

This is a more succinct docstring for the type.

The fact that geometries can be evaluated at uv coords in [0,1]^n is not specific to a ParametrizedCurve.

File: src/geometries/primitives/parametrizedcurve.jl [hunk 12]
```diff
@@ -0,0 +1,42 @@
...
+ParametrizedCurve(t -> Point(cospi(2t), sinpi(2t)))
+```
+"""
+struct ParametrizedCurve{M<:Manifold,C<:CRS,F<:Function,R<:Tuple} <: Primitive{M,C}
+  func::F
+  range::R
+
+  function ParametrizedCurve(func, range=(0.0, 1.0))
+    a = first(range)
+    p = func(a)
```
juliohm: I wonder if we should pass the range as an actual Julia range `a:b` to avoid situations where the `typeof(a) != typeof(b)` and where the inferred `typeof(p)` is not adequate to store all points along the range.

Alternatively, we need to `promote` the passed range values to a common type before inferring `p`.

src/geometries/primitives/parametrizedcurve.jl
```diff
@@ -0,0 +1,42 @@
...
+    M = manifold(p)
+    C = crs(p)
+    new{M,C,typeof(func),typeof(range)}(func, range)
+  end
+end
+
+paramdim(::ParametrizedCurve) = 1
+Base.minimum(curve::ParametrizedCurve) = curve(0.0)
+Base.maximum(curve::ParametrizedCurve) = curve(1.0)
+Base.extrema(curve::ParametrizedCurve) = curve(0.0), curve(1.0)
```
juliohm:
```suggestion
Base.minimum(curve::ParametrizedCurve) = curve.func(curve.a)
Base.maximum(curve::ParametrizedCurve) = curve.func(curve.b)
Base.extrema(curve::ParametrizedCurve) = minimum(curve), maximum(curve)
```

For better type inference.

src/geometries/primitives/parametrizedcurve.jl
```diff
@@ -0,0 +1,42 @@
...
+  function ParametrizedCurve(func, range=(0.0, 1.0))
+    a = first(range)
+    p = func(a)
+    M = manifold(p)
+    C = crs(p)
+    new{M,C,typeof(func),typeof(range)}(func, range)
+  end
+end
+
+paramdim(::ParametrizedCurve) = 1
```
juliohm:
```suggestion
paramdim(::Type{<:ParametrizedCurve}) = 1
```

This information is known at the type level.

src/geometries/primitives/parametrizedcurve.jl
```diff
@@ -0,0 +1,42 @@
...
+Base.minimum(curve::ParametrizedCurve) = curve(0.0)
+Base.maximum(curve::ParametrizedCurve) = curve(1.0)
+Base.extrema(curve::ParametrizedCurve) = curve(0.0), curve(1.0)
+
+function (curve::ParametrizedCurve)(t)
+  if t < 0.0 || t > 1.0
```
juliohm:
```suggestion
  if t < 0 || t > 1
```

juliohm (changes requested):
@JoshuaLampert I share your feeling here that it would be hard to generalize these algorithms to n-dimensions and that it is safer to define new primitives one by one per dimension. This PR is almost ready for the `ParametrizedCurve`.

src/geometries/primitives/parametrizedcurve.jl
JoshuaLampert: Where is it documented that the coords are always in [0,1]^n? I see that for `BezierCurve` it is stated explicitly in the docstring. What is different for a `ParametrizedCurve` compared to the other geometries is that it already comes with a range for the parameter, which can be something else than [0, 1]. Thus, to avoid confusion I think it would be nice to explicitly state it in the docstring (I would even do it if is documented clearly for all geometries since it is a bit counterintuitive for me).

src/geometries/primitives/parametrizedcurve.jl
JoshuaLampert: A range `a:b` is a `StepRangeLen`, i.e. it has a step length (defaulting to 1.0, i.e. `0:2pi` becomes `0.0:1.0:6.0` for instance), which seems inappropriate here. I made a proposal in [fefc71e](https://github.com/JuliaGeometry/Meshes.jl/pull/1093/commits/fefc71e55930aa4804eac41e67f7bc9dce016096) to `promote` the range values.

src/geometries/primitives/parametrizedcurve.jl
juliohm: > Where is it documented that the coords are always in [0,1]^n?

The *uv* coords of any geometry are in [0,1]^n by definition. If this feature is not documented yet, then we need to add it to a dedicated section of the Documenter.jl docs instead of repeating it over the docstrings of individual geometries. I am trying to say that this is not the right place to document the fact that Meshes.jl geometries can be called like g(u, v, w, ...). This feature is independent of the construction of the ParametrizedCurve itself.

It is an accident that the docstring of `BezierCurve` mentions it.

src/geometries/primitives/parametrizedcurve.jl
```diff
@@ -0,0 +1,47 @@
...
+  func::F
+  range::R
+
+  function ParametrizedCurve(func, range=(0.0, 1.0))
+    a, b = promote(range...)
+    _range = (a, b)
+    p = func(a)
+    M = manifold(p)
+    C = crs(p)
+    new{M,C,typeof(func),typeof(_range)}(func, _range)
```
juliohm:
```suggestion
    r = (a, b)
    p = func(a)
    M = manifold(p)
    C = crs(p)
    new{M,C,typeof(func),typeof(r)}(func, r)
```

juliohm (approved):

juliohm:
Noticed now that we still need tests and a mention in the docs after the `BezierCurve`.

src/geometries/primitives/parametrizedcurve.jl
```diff
@@ -0,0 +1,47 @@
...
+"""
+    ParametrizedCurve(func, range = (0.0, 1.0))
+
+A parametrized curve is a curve defined by a function `func` that maps a
+(unitless) parameter `t` in the given `range` to a `Point` in space.
+
+## Example
+
+```julia
+ParametrizedCurve(t -> Point(cos(t), sin(t)), (0, 2pi))
```
JoshuaLampert:
```suggestion
```
There already was an `# Example` section.

src/geometries/primitives/parametrizedcurve.jl
JoshuaLampert: In the end you decide. I can only say that I would be confused as a user to see that I need to evaluate a curve in another parameter range as I explicitly provided and I do think this is a special case compared to other geometries. I do not suggest to document this for every geometry. But if you decide not to document it here, it's fine for me.

> If this feature is not documented yet, then we need to add it to a dedicated section of the Documenter.jl docs

I have not found this in the documentation. I agree it would be good to add (if I haven't missed it).

src/geometries/primitives/parametrizedcurve.jl
```diff
@@ -0,0 +1,43 @@
...
+    M = manifold(p)
+    C = crs(p)
+    new{M,C,typeof(func),typeof(r)}(func, r)
+  end
+end
+
+paramdim(::Type{<:ParametrizedCurve}) = 1
+Base.minimum(curve::ParametrizedCurve) = curve.func(first(curve.range))
+Base.maximum(curve::ParametrizedCurve) = curve.func(last(curve.range))
+Base.extrema(curve::ParametrizedCurve) = minimum(curve), maximum(curve)
```
juliohm:
```suggestion
paramdim(::Type{<:ParametrizedCurve}) = 1

Base.minimum(curve::ParametrizedCurve) = curve.func(first(curve.range))

Base.maximum(curve::ParametrizedCurve) = curve.func(last(curve.range))

Base.extrema(curve::ParametrizedCurve) = minimum(curve), maximum(curve)
```

JoshuaLampert:
I'll add tests tomorrow.

juliohm:
Thank you @JoshuaLampert, these are great additions.

I will try to find some time to write a section in the docs regarding the uv coords of geometries.

JoshuaLampert:
That would be great! Thanks a lot for your valuable input and review.

JoshuaLampert:
I added tests. The only [test failure](https://github.com/JuliaGeometry/Meshes.jl/actions/runs/11257039084/job/31300346247?pr=1093#step:6:707) seems unrelated to this PR.

test/crs.jl
```diff
@@ -7,6 +7,8 @@
   @test crs(g) <: Mercator{WGS84Latest}
   g = BezierCurve(merc(0, 0), merc(1, 1), merc(2, 0))
   @test crs(g) <: Mercator{WGS84Latest}
+  g = ParametrizedCurve(t -> merc(cos(t), sin(t)), (0, 2π))
```
juliohm:
```suggestion
  g = ParametrizedCurve(t -> merc(cos(t), sin(t)), (T(0), T(2π)))
```

test/discretization.jl
```diff
@@ -327,6 +327,13 @@ end
   @test eltype(mesh) <: Segment
   @test nvertices.(mesh) ⊆ [2]
 
+  parametrizedcurve = ParametrizedCurve(t -> cart(cos(t), sin(t)), (0, 2π))
+  mesh = discretize(parametrizedcurve, RegularDiscretization(10))
```
juliohm:
```suggestion
  curve = ParametrizedCurve(t -> cart(cos(t), sin(t)), (T(0), T(2π)))
  mesh = discretize(curve, RegularDiscretization(10))
```

test/discretization.jl
```diff
@@ -570,6 +577,11 @@ end
   @test eltype(msh) <: Segment
   @test nvertices(msh) == nelements(msh) + 1
 
+  par = ParametrizedCurve(t -> cart(cos(t), sin(t)), (0, 2π))
+  msh = simplexify(par)
```
juliohm:
```suggestion
  curve = ParametrizedCurve(t -> cart(cos(t), sin(t)), (T(0), T(2π)))
  msh = simplexify(curve)
```

test/sampling.jl
```diff
@@ -82,6 +82,13 @@ end
     @test p ≈ t
   end
 
+  c = ParametrizedCurve(t -> cart(cos(t), sin(t)), (0, 2π))
```
juliohm:
```suggestion
  c = ParametrizedCurve(t -> cart(cos(t), sin(t)), (T(0), T(2π)))
```

juliohm (changes requested):
Minor adjustments regarding machine type of literals.

src/geometries/primitives/parametrizedcurve.jl
```diff
@@ -0,0 +1,45 @@
...
+
+```julia
+ParametrizedCurve(t -> Point(cos(t), sin(t)), (0, 2π))
+```
+"""
+struct ParametrizedCurve{M<:Manifold,C<:CRS,F<:Function,R<:Tuple} <: Primitive{M,C}
+  func::F
+  range::R
+
+  function ParametrizedCurve(func, range=(0.0, 1.0))
```
eliascarv: Can you please rename the parameter and field `func` to `fun`? `fun` is used to name functions in the project.

src/geometries/primitives/parametrizedcurve.jl
```diff
@@ -0,0 +1,45 @@
...
+    a, b = promote(range...)
+    r = (a, b)
+    p = func(a)
+    M = manifold(p)
+    C = crs(p)
+    new{M,C,typeof(func),typeof(r)}(func, r)
+  end
+end
```
eliascarv: For consistency with other constructors:
```suggestion
struct ParametrizedCurve{M<:Manifold,C<:CRS,F<:Function,R<:Tuple} <: Primitive{M,C}
  fun::F
  range::R
  ParametrizedCurve{M,C}(fun::F, range::R) where {M<:Manifold,C<:CRS,F<:Function,R<:Tuple} = new{M,C,F,R}(fun, range)
end

function ParametrizedCurve(fun, range=(0.0, 1.0))
  a, b = promote(range...)
  r = (a, b)
  p = fun(a)
  ParametrizedCurve{manifold(p),crs(p)}(fun, r)
end
```

src/geometries/primitives/parametrizedcurve.jl
JoshuaLampert: Done in [71da5f8](https://github.com/JuliaGeometry/Meshes.jl/pull/1093/commits/71da5f875f6d30dbcfedeed73092e896f02090e7).

juliohm (approved):

eliascarv (approved):

juliohm:
Thank you @JoshuaLampert ! Releasing a patch...

chakravala:
I have already fully implemented a fully dimension independent form of this, which works with arbitrary Grassmann.jl meshes.

My implementation is far superior, because it is built up from Grassmann.jl and has far more capability than Meshes.jl can have, since you are much more limited in what you can express here.

My version of this already is dimension independent and already does a lot more than what you can plan to  do here, because you are not based on Grassmann.jl technology here.

JoshuaLampert:
> I will try to find some time to write a section in the docs regarding the uv coords of geometries.

Now I see it's documented in https://juliageometry.github.io/MeshesDocs/stable/predicates/#Meshes.isparametrized, but a more prominent documentation could still be helpful (but is not urgent).
Note that it is also not always exactly true that every parametrized geometry can be called with local coordinates in [0,1]^n. Exceptions are `Triangle` and `Tetrahedron` (you have the additional restriction of $u + v (+ w) \le 1$) and `Line`, `Plane`, and `Ray`, where some of the arguments can go to infinity in one or both directions.

PR #1098, opened 2024-10-04
Update parameterization implementation of `Cylinder` and `CylinderSurface` & Simplification of `Sphere` parameterization

### LINKED ISSUE #1096: Review parametrization of geometries
The parametrizations were written before explicit support for CRS. We can probably do better now to construct the appropriate CRS type for the geometry and then call geometric transforms to position the points.

For example, in a Cylinder we can use Cylindrical coords, Rotation and Translation to find the final point.

eliascarv:
Fix #1096

### DIFF
--- a/src/geometries/primitives/cylinder.jl
+++ b/src/geometries/primitives/cylinder.jl
@@ -91,8 +91,8 @@ function (c::Cylinder)(ρ, φ, z)
   # project a parametric segment between the top and bottom planes
   ρ′ = T(ρ) * r
   φ′ = T(φ) * 2 * T(π) * u"rad"
-  p₁ = Point(convert(crs(c)), Cylindrical(ρ′, φ′, zero(ℒ)))
-  p₂ = Point(convert(crs(c)), Cylindrical(ρ′, φ′, h))
+  p₁ = Point(convert(crs(c), Cylindrical(ρ′, φ′, zero(ℒ))))
+  p₂ = Point(convert(crs(c), Cylindrical(ρ′, φ′, h)))
   p₁′ = p₁ |> Affine(Q, to(o))
   p₂′ = p₂ |> Affine(Q, to(o))
   l = Line(p₁′, p₂′)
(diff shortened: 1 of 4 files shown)


src/geometries/primitives/cylinder.jl
```diff
@@ -89,10 +89,13 @@ function (c::Cylinder)(ρ, φ, z)
...
   # project a parametric segment between the top and bottom planes
-  lsφ, lcφ = T(ρ) * r .* sincospi(2 * T(φ))
-  p₁ = o + Q * Vec(lcφ, lsφ, zero(ℒ))
-  p₂ = o + Q * Vec(lcφ, lsφ, h)
-  l = Line(p₁, p₂)
+  ρ′ = T(ρ) * r
+  φ′ = T(φ) * 2 * T(π) * u"rad"
+  p₁ = Point(convert(crs(c), Cylindrical(ρ′, φ′, zero(ℒ))))
+  p₂ = Point(convert(crs(c), Cylindrical(ρ′, φ′, h)))
+  p₁′ = p₁ |> Affine(Q, to(o))
```
juliohm: please create the Affine once

src/geometries/primitives/cylindersurface.jl
```diff
@@ -83,39 +83,7 @@ Base.isapprox(c₁::CylinderSurface, c₂::CylinderSurface; atol=atol(lentype(c
...
-  rsφ, rcφ = r .* sincospi(2 * T(φ))
-  zᵦ = -h / 2 - (rcφ * nᵦ[1] + rsφ * nᵦ[2]) / nᵦ[3]
-  zₜ = +h / 2 - (rcφ * nₜ[1] + rsφ * nₜ[2]) / nₜ[3]
-  pᵦ = Point(rcφ, rsφ, zᵦ)
-  pₜ = Point(rcφ, rsφ, zₜ)
-
-  p = pᵦ + T(z) * (pₜ - pᵦ)
-  o + Q' * to(p)
-end
+(c::CylinderSurface)(φ, z) = Cylinder(bottom(c), top(c), radius(c))(1, φ, z)
```
juliohm, line 86: please double check that the literal 1 doesn't change the resulting machine type. it is probably safer to use one(T)

juliohm (approved):

PR #1100, opened 2024-10-07
Add `Morphological` transform

JoshuaLampert:
This is a first draft for the alternative implementation for parametrized geometries given by user-defined parametrizations discussed in https://github.com/JuliaGeometry/Meshes.jl/pull/1093#issuecomment-2388314433. In your comment you suggested that the user-defined function should get a `Cartesian`, but I found that interface a bit cumbersome because, e.g., it does not supporting indexing, which I find more convenient to use. I would also find it more intuitive if the parameters would be unitless such that something like `cospi(v[1])` simply works without needing to manually `ustrip` in the user-defined function. I thus went with a function that gets an `SVector` instead. With the current version the following already works quite nicely:
```julia
box = Box((-pi,), (pi,))
f(theta) = 1 - 1/3 * sin(2*theta)^2
tr = TransformedGeometry(box, Morphological(v -> Polar(f(v[1]), v[1])))
```
I would like to see your feedback on this initial draft @juliohm before I continue.

TODO:
- [x] `viz` does not work yet.
- [x] Add tests
- [x] Add docs


juliohm:
Hi @JoshuaLampert thank you for the PR.

> In your comment you suggested that the user-defined function should get a `Cartesian`, but I found that interface a bit cumbersome because, e.g., it does not supporting indexing, which I find more convenient to use.

The `Morphological` transform can take any function `CRS -> CRS`. The example I gave with `Cartesian` is not general. You could define a function in terms of `Polar` CRS for example:

```julia
function f(c::Polar)
  ρ, ϕ = c.ρ, c.ϕ
  Cartesian(..., ..., ...)
end
```

We cannot rely on the `Cartesian` interface alone because it doesn't cover all use cases.

> I would also find it more intuitive if the parameters would be unitless such that something like `cospi(v[1])` simply works without needing to manually `ustrip` in the user-defined function.

You don't need to call `ustrip` because Unitful.jl defines methods for unitful angles:

```julia
julia> using Unitful

julia> cos(1u"rad")
0.5403023058681398

julia> cospi(1u"rad")
-1.0
```

That way you can be sure that angles in degrees and radians will produce the correct result.

> I thus went with a function that gets an `SVector` instead

Unfortunately this is the API we've been fighting against. Treating coords as raw vectors is a step backwards. It is much safer to write algorithms in terms of coord types with specific coordinate names. To give a practical example where this is problematic, consider `LatLon` coords. We advocate for explicit code that looks like this:

```julia
function f(c::LatLon)
  lat, lon = c.lat, c.lon
  ....
end
```

The `lat` and `lon` are angles in degrees. If you start treating `lat, lon = c[1], c[2]` as unitless numbers, you will quickly bump into problems where file formats store (lat, lon) vs (lon, lat) in different order, and you will perform computation with degrees as if they were lengths (1 degree in longitude at the equator is equal to 111km length).

Some algorithms are agnostic to CRS, and so we provide internal helper functions in meshes to "extract" tuples of coords with or without units. But end-users should never touch these low-level details.

juliohm:
Notice that we can brainstorm something later to account for uv parametric coords. These are unitless in the range [0,1]. This could be syntactic sugar over the `Morphological` transform proposed.

JoshuaLampert:
> Notice that we can brainstorm something later to account for uv parametric coords. These are unitless in the range [0,1]. This could be syntactic sugar over the `Morphological` transform proposed.

So for my use case of a parametrized curve (surface, volume, ...) we want our geometry, which is transformed by the `Morphological` transform to be a `box`, which describes at the same time our parametric coords. So we need a dimensionless `Box`, right? Maybe a stupid question, but how do we create that? I guess, I'm just not seeing how the syntactic sugar is supposed to look like. I would like to first clarify how to create something like a parametrized curve with this approach in order to see if it fulfills my needs in a satisfiable way.

juliohm:
Maybe I am missing something, but I understood that you need a transformed geometry that is parametrized. That should already work with unitless uv coordinates in the range [0,1]:

https://github.com/JuliaGeometry/Meshes.jl/blob/a7802878459f568cb48cd950973e8b86a0c9dd28/src/geometries/transformedgeom.jl#L63

We propagate the parametrization of the base geometry through the coordinate transform. CoordinateTransform's change one point at a time, so that holds.

Assuming you transformed a 2D box, you can do something like:

```julia
g = TransformedGeometry(Box((a, b), (c, d)), Morphological(f))

g(0.4, 0.3) # uv (unitless) coords on transformed box
```

We can think of syntactic sugar that takes `(a, b), (c, d), f` as input and produces the transformed box as the curve, surface, volume. Does it make sense?

We can also consider adding the Jacobian of the function `f` if that makes more sense.

JoshuaLampert:
I would like that the code you post above would work, but unfortunately it does not as I would wish in my current implementation because `Box((a, b), (c, d))` defines a space with units meter and not unitless. If we plug this into a function expecting unitless quantities it fails. My concrete issue is:

```julia
julia> box = Box((0.0,), (1.0,))
Box
├─ min: Point(x: 0.0 m)
└─ max: Point(x: 1.0 m)

julia> f(c) = Cartesian(cospi(2*c.x), sinpi(2*c.x))
f (generic function with 1 method)

julia> t = Morphological(f)
Morphological transform
└─ func: f

julia> tr = TransformedGeometry(box, t)
ERROR: MethodError: no method matching cospi(::Unitful.Quantity{Float64, 𝐋, Unitful.FreeUnits{(m,), 𝐋, nothing}})

Closest candidates are:
  cospi(::Missing)
   @ Base math.jl:1580
  cospi(::BigFloat)
   @ Base mpfr.jl:801
  cospi(::ForwardDiff.Dual{T}) where T
   @ ForwardDiff ~/.julia/packages/ForwardDiff/PcZ48/src/dual.jl:238
  ...

Stacktrace:
 [1] f(c::CoordRefSystems.Cartesian{CoordRefSystems.NoDatum, 1, Unitful.Quantity{Float64, 𝐋, Unitful.FreeUnits{(m,), 𝐋, nothing}}})
   @ Main ./REPL[3]:1
 [2] applycoord(t::Morphological{typeof(f)}, p::Point{𝔼{1}, CoordRefSystems.Cartesian{CoordRefSystems.NoDatum, 1, Unitful.Quantity{…}}})
   @ Meshes ~/.julia/dev/Meshes/src/transforms/morphological.jl:31
 [3] apply(t::Morphological{typeof(f)}, g::Point{𝔼{1}, CoordRefSystems.Cartesian{CoordRefSystems.NoDatum, 1, Unitful.Quantity{…}}})
   @ Meshes ~/.julia/dev/Meshes/src/transforms.jl:60
 [4] (::Morphological{typeof(f)})(object::Point{𝔼{1}, CoordRefSystems.Cartesian{CoordRefSystems.NoDatum, 1, Unitful.Quantity{Float64, 𝐋, Unitful.FreeUnits{…}}}})
   @ TransformsBase ~/.julia/packages/TransformsBase/w6gll/src/interface.jl:124
 [5] TransformedGeometry(g::Box{𝔼{1}, CoordRefSystems.Cartesian{CoordRefSystems.NoDatum, 1, Unitful.Quantity{…}}}, t::Morphological{typeof(f)})
   @ Meshes ~/.julia/dev/Meshes/src/geometries/transformedgeom.jl:22
 [6] top-level scope
   @ REPL[5]:1
```
That's why I initially wanted to `ustrip` the `Point` returned by the `box`, but I see that this is not a good approach. Above you say Unitful supports `cospi` for radians. That's true, but we don't have radians. We get values in m if we use a plain `Box`, this giving us the error above. That's why I would like something like a unitless box. Am I missing something?

src/transforms/morphological.jl
```diff
@@ -0,0 +1,32 @@
...
+"""
+struct Morphological{F<:Function} <: CoordinateTransform
+  func::F
+  function Morphological(func::F) where {F}
+    new{typeof(func)}(func)
+  end
+end
+
+applycoord(t::Morphological, p::Point) = Point(t.func(coords(p)))
+applycoord(t::Morphological, v::Vec) = Point(t.func(Point(v...)))
```
juliohm: I believe that this transform shouldn't act on `Vec` as it requires a function `f` that acts on CRS, which is only present in geometries like `Point`, `Triangle`, ...

juliohm:
> ```julia
> f(c) = Cartesian(cospi(2*c.x), sinpi(2*c.x))
> ```

I see your point now. You want to define a function from a unit-less non-geometry box to a geometry because in that case you can feed whatever values into functions like `cos` and `sin`. The suggested approach with `TransformedGeometry` and a new `Morphological` transform doesn't allow for that because it operates on fully constructed geometries which have a CRS, and thus coordinates with proper units.

It seems your previous approach with new `Primitive` types (`ParametricCurve`, `ParametricSurface` and `ParametricVolume`) is more adequate for your use case. Should we close this PR and continue the work in #1093?

JoshuaLampert:
Ok, let's do that. However, I do think it might still be interesting to add something like a `Morphological` transform. It just doesn't seem to be the best approach for the `ParametrizedCurve` problem, but allowing user-defined morphological transforms on actual (dimensional) geometries might still be of interest. This is already possible with the current implementation. However, I don't need that for my work (yet) and I don't have the time to finish this PR. But if someone else wants to finish this, this would be really nice. What do you think, @juliohm?

juliohm:
That is a very good point @JoshuaLampert , let's keep it open to finish later.

JoshuaLampert:
Thanks @eliascarv for continuing this! This looks pretty nice.

PR #1106, opened 2024-10-12
Add `isparametrized` for `Cone`

JoshuaLampert:
`Cone` is parametrized, but `isparametrized(cone)` gave `false`. This PR fixes this by adding the missing method.

### DIFF
--- a/test/predicates.jl
+++ b/test/predicates.jl
@@ -127,7 +127,7 @@ end
   @test isparametrized(Cylinder)
   @test isparametrized(CylinderSurface)
   @test isparametrized(ConeSurface)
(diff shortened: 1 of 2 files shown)


test/predicates.jl
```diff
@@ -127,6 +127,7 @@ end
   @test isparametrized(Cylinder)
   @test isparametrized(CylinderSurface)
   @test isparametrized(ConeSurface)
+  @test isparametrized(Cone)
```
juliohm:
```suggestion
  @test isparametrized(Cone)
  @test isparametrized(ConeSurface)
```

juliohm (approved):

PR #1121, opened 2024-10-25
Add `MinDistanceSampling` for 3D Ball

JoshuaLampert:
Currently, `MinDistanceSampling` for a 3D Ball is not supported. To support it, there are two different options: 1. Properly support `RegularDiscretization` for a 3D Ball by adding `Pyramid` elements in the center similarly as in [here](https://github.com/JuliaGeometry/Meshes.jl/blob/ed1d9c9b96a48caeb5b02cee5e662e628bfd9c23/src/discretization/regular.jl#L46) and then use the [fallback](https://github.com/JuliaGeometry/Meshes.jl/blob/master/src/sampling.jl#L59) or 2. add a specialized method for `Ball`. This PR follows the second approach as it is quite straightforward to add, but allowing support for `RegularDiscretization` of a 3D Ball would still be nice (s.t. `discretize(ball::Ball{𝔼{3})` could use that).

TODO:
- [x] Add test

### DIFF
--- a/src/sampling/mindistance.jl
+++ b/src/sampling/mindistance.jl
@@ -37,7 +37,11 @@ MinDistanceSampling(α, ρ, δ, metric) = MinDistanceSampling(addunit(α, u"m"),
 
 MinDistanceSampling(α::T; ρ=T(0.65), δ=100, metric=Euclidean()) where {T} = MinDistanceSampling(α, ρ, δ, metric)
 
-function sample(rng::AbstractRNG, obj::GeometryOrDomain, method::MinDistanceSampling)
+sample(rng::AbstractRNG, d::Domain, method::MinDistanceSampling) = _sample(rng, d, method)
+
+sample(rng::AbstractRNG, b::Ball, method::MinDistanceSampling) = _sample(rng, b, method)
+
+function _sample(rng::AbstractRNG, obj, method::MinDistanceSampling)
   # retrieve parameters
   α = method.α
   ρ = method.ρ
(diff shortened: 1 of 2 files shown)


src/sampling/mindistance.jl
```diff
@@ -37,15 +37,15 @@ MinDistanceSampling(α, ρ, δ, metric) = MinDistanceSampling(addunit(α, u"m"),
 
 MinDistanceSampling(α::T; ρ=T(0.65), δ=100, metric=Euclidean()) where {T} = MinDistanceSampling(α, ρ, δ, metric)
 
-function sample(rng::AbstractRNG, d::Domain, method::MinDistanceSampling)
+function sample(rng::AbstractRNG, obj::GeometryOrDomain, method::MinDistanceSampling)
```
JoshuaLampert:
```suggestion
function sample(rng::AbstractRNG, obj::Union{Domain,Ball}, method::MinDistanceSampling)
```
There is a [method ambiguity](https://github.com/JuliaGeometry/Meshes.jl/actions/runs/11523593652/job/32081938951?pr=1121#step:6:621). Is there a better choice than `Union{Domain,Ball}`?

JoshuaLampert: What's the advantage of your last [commit](https://github.com/JuliaGeometry/Meshes.jl/pull/1121/commits/7764ce9e6684c59669e5e820ec2fec08eed12cd3) compared to the `Union{Domain,Ball}` solution?

juliohm (approved):

JoshuaLampert:
Would you mind creating a new release, @juliohm?

juliohm:
Will do in a few seconds 👍🏽  Finishing up another PR with the `getindex` for topology as we discussed on Zulip.

PR #1122, opened 2024-10-25
Implement indexable api for topology

juliohm:
As discussed on Zulip

### DIFF
--- a/src/topologies.jl
+++ b/src/topologies.jl
@@ -150,7 +150,7 @@ Base.iterate(t::Topology, state=1) = state > nelements(t) ? nothing : (t[state],
 
 Base.eltype(t::Topology) = eltype([t[i] for i in 1:nelements(t)])
 
-Base.keys(t::Topology) = 1:nelements(t)
+Base.keys(t::Topology) = Base.OneTo(nelements(t))
 
 # ----------------
 # IMPLEMENTATIONS
--- a/src/topologies.jl
+++ b/src/topologies.jl
@@ -132,6 +132,26 @@ Return the number of facets of the `topology`.
 """
 function nfacets(::Topology) end
 
+# ----------
+# FALLBACKS
+# ----------
+
+Base.getindex(t::Topology, ind::Int) = element(t, ind)
+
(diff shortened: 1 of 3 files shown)


src/topologies.jl
```diff
@@ -132,6 +132,26 @@ Return the number of facets of the `topology`.
...
+
+Base.lastindex(t::Topology) = nelements(t)
+
+Base.length(t::Topology) = nelements(t)
+
+Base.iterate(t::Topology, state=1) = state > nelements(t) ? nothing : (t[state], state + 1)
+
+Base.eltype(t::Topology) = eltype([t[i] for i in 1:nelements(t)])
+
+Base.keys(t::Topology) = 1:nelements(t)
```
JoshuaLampert:
```suggestion
Base.keys(t::Topology) = Base.OneTo(nelements(t))
```
I think, often these kind of index ranges are a `Base.OneTo`, which behaves very similar to `1:`, but can sometimes be slightly more efficient?

src/topologies.jl
```diff
@@ -132,6 +132,26 @@ Return the number of facets of the `topology`.
 """
 function nfacets(::Topology) end
 
+# ----------
+# FALLBACKS
+# ----------
+
+Base.getindex(t::Topology, ind::Int) = element(t, ind)
+
+Base.getindex(t::Topology, inds::AbstractVector) = [element(t, ind) for ind in inds]
```
JoshuaLampert:
```suggestion
Base.getindex(t::Topology, inds::AbstractVector) = [t[ind] for ind in inds]
```

JoshuaLampert (commented):
Looks good to me. I have only two minor suggestions.

src/topologies.jl
juliohm: Interesting. Didn't know there was a difference in performance.

JoshuaLampert (approved):

juliohm:
I've also implemented the suggested optimizations for Domain for consistency. Waiting for tests before merging and releasing a patch.

JoshuaLampert:
Thanks!

PR #1124, opened 2024-11-06
Replace Base.OneTo(n) by 1:n for consistency

juliohm:
@JoshuaLampert I believe that these are the only two places in the code base where this internal function from Base is used. In order to be 100% consistent, I am proposing that we stick to the range syntax used elsewhere. Is that ok?


eliascarv (approved):

JoshuaLampert (approved):
Fine with me.

Issue #578, opened 2023-09-03
Add more clipping methods


juliohm:
- [ ] Weiler-Atherton: https://dl.acm.org/doi/pdf/10.1145/563858.563896
- [ ] Greiner-Hormann: https://dl.acm.org/doi/pdf/10.1145/274363.274364
- [ ] Vatti: https://dl.acm.org/doi/pdf/10.1145/129902.129906
- [ ] Martínez: https://www.sciencedirect.com/science/article/abs/pii/S0098300408002793

PR #1125, opened 2024-11-08
Adding Greiner Hormann Method for polygon Clipping

souma4:
This is a pull request to add the [up-to-date](https://www.sciencedirect.com/science/article/pii/S259014861930007X) [Greiner Hormann](https://www.inf.usi.ch/hormann/papers/Greiner.1998.ECO.pdf) method for clipping polygons as part of [Issue#578](https://github.com/JuliaGeometry/Meshes.jl/issues/578).

Currently, it is under a HEAVY DRAFT. All three phases are in need of heavy work, and there is ghost code from false starts. This draft is simply to put this code on the open pull requests so that others may contribute if they want, or to correct false starts.

There may be a need to update the `src/intersections.jl` IntersectionTypes. The existing [reference C++ code](https://www.inf.usi.ch/hormann/polyclip/) includes different IntersectionTypes. However, the current `Meshes.jl` implementation for overlapping intersections may not necessitate a change to these types. 

Additionally, I created an internal struct for this method because tracking neighbors across polygons and adjacent vertices is critical, and keeping it under one struct is much more readable.

IMO, the first goal is to get this to run. Second goal is to match with the codebase. Then we can include type definitions/safety.

### DIFF
--- a/src/clipping/greinerhormann.jl
+++ b/src/clipping/greinerhormann.jl
@@ -1,0 +1,883 @@
+# ------------------------------------------------------------------
+# Licensed under the MIT License. See LICENSE in the project root.
+# ------------------------------------------------------------------
+
+"""
+    GreinerHormannClipping()
+
+The Greiner-Hormann algorithm for clipping polygons.
+
+## References
+
+* Greiner, G. & Hormann, K. 1998. [Efficient Clipping of Arbitrary Polygons]
+  (https://dl.acm.org/doi/pdf/10.1145/274363.274364)
+
+### Notes
+
+* The Algorithm applies to any arbitrary polygon.
+"""
+
+
+@enum IntersectionLabel begin
+  na
+  crossing
+  bouncing
+  left_on
+  right_on
+  on_on
+  on_left
+  on_right
+  delayed_crossing
+  delayed_bouncing
+end
+
+@enum RelativePositionType begin
+  left
+  right
+  is_pm
+  is_pp
+end
+
+@enum EntryExit begin
+  exit
+  entry
+  neither
+end
+
+struct GreinerHormannClipping <: ClippingMethod end
+
+mutable struct GreinerVertex
+  coords::Point
+  prev::Int # this should point to previous vertex
+  next::Int # point to next vert
+  neighbour::Int # point to neighbour vertex for intersection
+  source::Bool
+  intersection::Bool
+  alpha::Float64
+  label::IntersectionLabel
+  enex::EntryExit
+
+  GreinerVertex(coords::Point, prev::Int, next::Int, neighbour::Int, source::Bool, intersection::Bool, alpha::Float64, label::IntersectionLabel, enex::EntryExit) = new(
+    coords,
+    prev,
+    next,
+    neighbour,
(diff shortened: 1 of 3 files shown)


src/clipping/greinerhormann.jl
```diff
@@ -0,0 +1,883 @@
...
+@enum EntryExit begin
+  exit
+  entry
+  neither
+end
+
+struct GreinerHormannClipping <: ClippingMethod end
+
+mutable struct GreinerVertex
+  coords::Point
```
juliohm, line 50: if this field is a point, better rename it accordingly

src/clipping/greinerhormann.jl
```diff
@@ -0,0 +1,883 @@
...
+  end
+
+  prev_index = ring[index].prev
+  next_index = ring[index].next
+
+  ring[prev_index].next = next_index
+  ring[next_index].prev = prev_index
+
+  deleteat!(ring, index)
+  return ring
```
juliohm, line 175: we only use `return` in premature termination

src/clipping/greinerhormann.jl
```diff
@@ -0,0 +1,883 @@
...
+end
+
+function pointInPoly(R::Point, poly::Ring)
```
juliohm, line 178: here you mean `point \in poly` maybe? we already have a super efficient (multi-threaded) implementation for this

juliohm (commented):
Thank you @souma4 ! These are very nice, non-trivial contributions! ❤️ 

A few comments:

- Perhaps we can reuse existing algorithms like `point in poly` here, given that our current implementation is extremely efficient and reasonably tested.
- If you find out that a function is missing that is independent of this clipping algorithm, consider adding it in a separate self-contained PR before this one.

A few questions:

1. Are you aware of the software license of the reference code? We need to be careful to not base our implementation in GPL for example.
2. Can you comment on the time complexity of the algorithm used to find segment-segment intersections? If we could rely on an efficient [sweep-line algorithm](https://en.wikipedia.org/wiki/Bentley%E2%80%93Ottmann_algorithm), that would be perfect. This video series provides a really nice exposition of the algorithm: https://youtu.be/nNtiZM-j3Pk?si=eCsujt3J6EkHy8tS

Regarding the sweep-line algorithm, I drafted an initial version but got stuck due to DataStructures.jl limitations. I will open a separate issue with this partial implementation, and will cc you there.

souma4:
Honestly, I need to do a mass check of all functions in the GeoStats ecosystem because I keep finding existing functions that can help. Point in poly is an example. Like duh it exists, but brain be weird.

The current implementation for the intersection phase is O(nm), so a sweep line algorithm would be great. They mention that using such an approach is a natural extention. I'll give the draft some checks when I get the chance. 

That's true that the MIT license needs to be preserved. The reference code actually doesn't seem to have a specific license tied to it. The paper is CC BY-NC-ND 4.0 and the supplementals are CC BY 4.0. 🤷. With a plane sweep algorithm, a labeling phase that can interface with that, all in Julia, it should be fine, but let me know what makes you feel most comfortable.

Issue #1131, opened 2024-11-12
Introduce a `eachvertex` function


juliohm:
The API currently defines `vertices`, which guarantees an `AbstractVector`. We need a `eachvertex` function to return iterators instead for use cases that need to avoid allocations.

juliohm:
Fixed on master.

Issue #1127, opened 2024-11-11
Eliminate allocations of `vertices` in `Multi` geometries


juliohm:
The current implementation currently allocates:

https://github.com/JuliaGeometry/Meshes.jl/blob/42f8543991b0da3442cbe7accebc89a2ac558b0e/src/geometries/multigeom.jl#L53

We need a quick fix for that.

ghyatzo:
assuming homogeneous multigeometries (multiple geometries all of the same type, like a triangle mesh)
we can do something like this:
```julia
julia> function myvertices(m::Multi) #with some kind of parametric type constraint
           N = mapreduce(nvertices, +, m.geoms) # compile is smart enough to constant propagate this away.
           k = nvertices(eltype(m.geoms))
           ntuple(i -> vertices(m.geoms[ceil(Int, i/k)])[1 + i % k], Val(N))
       end
myvertices (generic function with 1 method)

julia> m
MultiTriangle{2,Float64}
├─ Triangle((0.550249, 0.314319), (0.657976, 0.225926), (0.743882, 0.414126))
├─ Triangle((0.0381363, 0.475113), (0.745496, 0.223022), (0.922711, 0.0582245))
├─ Triangle((0.868872, 0.000274105), (0.217163, 0.322933), (0.354929, 0.543288))
└─ Triangle((0.890692, 0.383539), (0.482963, 0.218332), (0.167305, 0.309235))

julia> @btime myvertices($m);
  120.630 ns (2 allocations: 240 bytes)

julia> @btime vertices($m);
  323.735 ns (4 allocations: 960 bytes)
```
still allocates, but much less. ( I think it can be made completely non allocating, but couldn't find how).
Also, it returns a tuple instead of a vector, that might be a breaking change? But I thought it was fitting since normal `vertices` returns a tuple.

For a general case of completely arbitrary multi geometry I couldn't think of a way to avoid looping over all the geometries. If you know that all are a fixed size, other than knowing the total number of vertices, you also know where to look for each vertex.

juliohm:
Thank you for the suggestions. We will review the return type of all `vertices` methods and will then decide what to do in general.

juliohm:
All `vertices` methods currently return `AbstractVector`, except for the method with `Ngon`, which returns a `NTuple`. One way to make the output more homogeneous is to return `SVector` instead. We can then look into the method with `Multi` to return a lazy vector.

juliohm:
[...]

Issue #1128, opened 2024-11-11
Stack Overflow when instantiating `Multi` 


ghyatzo:
Hey, out of curiosity I was investigating #1127

But I noticed another bug:
```julia
julia> tris  = connect.([(1,5,3), (4,6,2)], Triangle)
2-element Vector{Connectivity{Triangle, 3}}:
 Triangle(1, 5, 3)
 Triangle(4, 6, 2)

julia> collect(tris)
2-element Vector{Connectivity{Triangle, 3}}:
 Triangle(1, 5, 3)
 Triangle(4, 6, 2)

julia> Multi(tris)
ERROR: StackOverflowError:
Stacktrace:
 [1] GenericMemory
   @ .\boot.jl:516 [inlined]
 [2] Array
   @ .\boot.jl:578 [inlined]
 [3] Array
   @ .\boot.jl:591 [inlined]
 [4] Array
   @ .\boot.jl:599 [inlined]
 [5] _collect_indices
   @ .\array.jl:723 [inlined]
 [6] collect
   @ .\array.jl:707 [inlined]
 [7] Multi(geoms::Vector{Connectivity{Triangle, 3}}) (repeats 52260 times)
   @ Meshes C:\Users\camillo.schenone\.julia\packages\Meshes\VQC9c\src\multigeoms.jl:23
```

due to this line most probably: https://github.com/JuliaGeometry/Meshes.jl/blob/42f8543991b0da3442cbe7accebc89a2ac558b0e/src/geometries/multigeom.jl#L23

since `collect(tris) == tris` the call `Multi(geom) = Multi(collect(geom))` will keep calling itself.

I am most likely using this in the wrong way, given the lack of practicality with this package, but still... it would be nice to avoid the recursion, or have some check in place to stop the wrong usage

EDIT:
weirdly enough using straight Triangles doesn't incur in this issue?
```julia
julia> g = [Triangle((0.,0.), (1.,0.), (0.,1.)),Triangle((0.,0.), (1.,0.), (0.,1.))]
2-element Vector{Triangle{2, Float64}}:
 Triangle((0.0, 0.0), (1.0, 0.0), (0.0, 1.0))
 Triangle((0.0, 0.0), (1.0, 0.0), (0.0, 1.0))

julia> Multi(g)
MultiTriangle{2,Float64}
├─ Triangle((0.0, 0.0), (1.0, 0.0), (0.0, 1.0))
[...]

PR #1133, opened 2024-11-12
Introducing iteration on vertices

ghyatzo:
Hi,

following #1127 and #1131 I'm starting this draft PR.
I've added the main iteration logic for single geometry and multi geometry types.

This si incomplete as I would require some input on the following points
1. check any wrong assumption i've made so far
2. relevant types in the package
3. code placement (where in the source tree) and general code organization

in particular, I am not very familiar with all the nomenclature and type system structure used in this package, so I would appreciate an indication of which types we are interested in and/or for which onese we want to be able to call `vertices`.

WRT to point 2, I've gone for stuffing all iteration related code into a single file for now, let me know if this has to change.

cheers!


juliohm:
Thank you @ghyatzo for the contribution. To facilitate the usage of this iterator, we just need to implement a new `eachvertex` function for `Polytope` `Multi`, and `Mesh` subtypes. Can you please refactor the PR to add this function instead?

For the `Multi` case you can use `Iterators.flatten` for example.

ghyatzo:
Hi, yes absolutely can do.

Regarding flatten, I've done some tries, but the `flatten` iterator implementations I could come up with are both slower and worse allocating than the custom one here:
```
julia> @btime begin                                                                                            
           s = 0                                                                                               
           for _ in flat_multiitr                                                                              
               s+=1                                                                                            
           end                                                                                                 
           s                                                                                                   
       end                                                                                                     
  2.472 μs (45 allocations: 3.38 KiB)
9

julia> @btime begin                                                                                            
           s = 0                                                                                               
           for _ in custom_multiitr                                                                            
               s+=1                                                                                            
           end                                                                                                 
           s                                                                                                   
       end                                                                                                     
  1.167 μs (36 allocations: 1.27 KiB)
9

julia> @btime Base.iterate($flat_multiitr)                                                                     
  448.391 ns (5 allocations: 304 bytes)                                                                        
(Point(x: 0.011963944663157644 m, y: 0.13991304491415013 m), (2, VertexItr{Triangle{𝔼{2}, CoordRefSystems.Cartesian2D{CoordRefSystems.NoDatum, Unitful.Quantity{Float64, 𝐋, Unitful.FreeUnits{(m,), 𝐋, nothing}}}}}(Triangle((x: 0.0119639 m, y: 0.139913 m), (x: 0.347587 m, y: 0.958666 m), (x: 0.145339 m, y: 0.109886 m))), 2))          
                                                                                                               
julia> @btime Base.iterate($custom_multiitr)                                                                   
  53.077 ns (1 allocation: 32 bytes)
(Point(x: 0.011963944663157644 m, y: 0.13991304491415013 m), (1, 2))
```

flatten requires an iterator of iterators, which means that we have to build an iterator for each subelement in the multigeometry.
While the custom one can cheat and implement a pseudo iterator since we can manually call and it doesn't have to go through `Base.iterate`.
We could (i think) circumvent this by defining `Base.iterate` directly for the single elements, i.e. `Base.iterate(p::Polytope)`, but that would get in the way of one day maybe be able to iterate over the edges or faces of a geometry instead of the vertices.
I think making a specific `VertexIterator` that acts on geometries is a more solid approach.

juliohm:
@ghyatzo I meant that we simply need to define the function `eachvertex(::Polytope)` without additional constructs. It can be as simple as:

```julia
eachvertex(p::Polytope) = (v for v in vertices(p))
```

The geometry that will benefit the most is the `Multi`, which will avoid the allocations.

ghyatzo:
The reason why I didn't use simply a generator `(v for v in vertices(p))` is to make the `Multi` iterator (which is the one we care about) more performant. if we simply rely on generators it means that we have to instantiate a generator for each geometry.
while with the approach I went with you have a single iterator that returns the correct element at each step.
It is similar to how the `iterator` for an `AbstractArray` is the array itself, here we can use the geometry itself to be the iterator object, without the need to instantiate extra objetcs.

It should still be generic enough since underneat we are calling `vertex(el, idx)`

Please let me know if there is anything I overlooked or wrong with my reasoning.
I'll try to refactor it more to your preference in the mean time.

ghyatzo:
I've improved the flatten approach using a generator of generators and not an array as I was naively doing before.
But I still see much more allocations and time than the custom iterator approach.

```julia-repl
julia> eachvertex_custom(e::Union{Mesh,Polytope,Multi}) = VertexItr(e)
julia> eachvertex_gen(p::Polytope) = (v for v in vertices(p))
julia> eachvertex_flat(e::Multi) = Iterators.flatten(eachvertex_gen(g) for g in e.geoms)                       
eachvertex_flat (generic function with 1 method)
julia> m                                                                                                       
MultiNgon
├─ Triangle((x: 0.0119639 m, y: 0.139913 m), (x: 0.347587 m, y: 0.958666 m), (x: 0.145339 m, y: 0.109886 m))
└─ Hexagon((x: 0.25018 m, y: 0.85677 m), ..., (x: 0.57332 m, y: 0.3497 m))

julia> @btime begin                                                                                            
           s = 0                                                                                               
           for _ in eachvertex_flat($m)                                                                                                                                                                                       
               s+=1                                                                                            
           end                                                                                                 
           s                                                                                                   
       end                                                                                                     
  1.104 μs (39 allocations: 2.95 KiB)
9

julia> @btime begin                                                                                            
           s = 0                                                                                               
           for _ in eachvertex_custom($m)                                                                                                                                                                                     
               s+=1                                                                                            
           end                                                                                                 
           s                                                                                                   
       end                                                                                                     
  497.851 ns (9 allocations: 288 bytes)
9
```
those 9 allocations are from  an `Int64` allocation in the `iterate` function that If ironed out can become completely non allocating.

Anyway, I completely understand prefering the simpler solution.

Also: is it enough to just have the `eachvertex` function just for `Polytopes`, `Multi` and `Mesh`?  let me know if I've missed something here as well.

juliohm:
Isn't ok to define

```julia
eachvertex(m::Multi) = (v for g in m.geoms for v in vertices(g))
```

in the `Multi` geometry case? Why do we need nested generators?

ghyatzo:
```
(v for g in m.geoms for v in vertices(g))
```
is shorthand for
```
Iterators.flatten((v for v in vertices(g)) for g in e.geoms)
```

from lowering the code:
```
julia> eachvertex_flatter(m::Multi) = (v for g in m.geoms for v in vertices(g))

julia> @code_lowered eachvertex_flatter(m)                                                                     
CodeInfo(
1 ─ %1 = Main.:(var"#13#14")
│        #13 = %new(%1)
│   %3 = #13
│   %4 = Base.getproperty(e, :geoms)
│   %5 = Base.Generator(%3, %4)
│   %6 = Base.Flatten(%5) <---- we construct a flatten iterator anyway
└──      return %6
)
```

and has similar performances 
```
@btime begin                                                                                            
           s = 0                                                                                               
           for _ in eachvertex_flatter($m)                                                                                                                                                                                    
               s+=1                                                                                            
           end                                                                                                 
           s                                                                                                   
       end                                                                                                     
  1.092 μs (39 allocations: 2.95 KiB)
9
```

juliohm:
> is it enough to just have the `eachvertex` function just for `Polytopes`, `Multi` and `Mesh`?

Yes, but please restrict to `MultiPolytope` instead of general `Multi` geometries.

> ```
> (v for g in m.geoms for v in vertices(g))
> ```
> 
> 
>     
>   
> 
> is shorthand for
> 
> 
> ```
> Iterators.flatten((v for v in vertices(g)) for g in e.geoms)
> ```

Interesting. Thanks for clarifying this.

src/Meshes.jl
```diff
@@ -37,6 +37,7 @@ import Base: sort
 import Base: ==, !
 import Base: +, -, *
 import Base: <, >, ≤, ≥
+import Base: iterate, length, IteratorSize, eltype
```
juliohm: Please define the methods in-place with `Base.iterate` without importing the names.

juliohm:
Appreciate if you can keep it simple for now without an auxiliary iterator struct. Please define the `eachvertex` function in the respective geometry files, and add new tests with `@allocated`.

We can reconsider the auxiliary iterator struct in the future. I understand it allocates a fixed amount no matter the size of the `MultiPolytope`.

ghyatzo:
Alright, will do. shall I make a separate PR and keep this as draft with the struct approach?

juliohm:
Feel free to update the PR here directly. You can copy the VertexIter struct in the comments to save it for future reference.

ghyatzo:
For future reference:
```julia
struct VertexItr{T}
    el::T
end

_v_iterate(el::Polytope, i) =
    (@inline; (i - 1) % UInt < nvertices(el) % UInt ? (@inbounds vertex(el, i), i + 1) : nothing)

_v_iterate(el::Mesh, i) =
    (@inline; (i - 1) % UInt < nvertices(el) % UInt ? (@inbounds vertex(el, i), i + 1) : nothing)

Base.iterate(itr::VertexItr{<:Mesh}, i=1) = _v_iterate(itr.el, i)

Base.iterate(itr::VertexItr{<:Polytope}, i=1) = _v_iterate(itr.el, i)

Base.iterate(itr::VertexItr{<:MultiPolytope}, state=(1, 1)) = begin
    ig, ivg = state
    ig > length(itr.el.geoms) && return nothing

    is = _v_iterate(itr.el.geoms[ig], ivg)
    is === nothing && return Base.iterate(itr, (ig + 1, 1))

    v, ivg = is
    return (v, (ig, ivg))
end

Base.length(itr::VertexItr{<:Mesh}) = nvertices(T)
Base.length(itr::VertexItr{<:Polytope}) = nvertices(T)
Base.length(itr::VertexItr{<:MultiPolytope}) = sum(nvertices, itr.el.geoms)

Base.IteratorSize(itr::VertexItr) = Base.HasLength()
Base.eltype(::VertexItr) = Point

eachvertex(e::T) where {T} = error("Vertex iterator not implemented for type $T")
eachvertex(e::Union{Mesh,Polytope,MultiPolytope}) = VertexItr(e)
```

ghyatzo:
I would like some pointers WRT the new tests, I'm a bit unsure where best to put them
and what you mean exactly when you meant with `@allocated`.


I've also took the liberty of rewriting some `vertices()` function to instead `collect`ing the iterator instead of repeating the same thing. Let me know if you agree with the change.

src/geometries/multigeom.jl
```diff
@@ -48,12 +48,14 @@ Base.isapprox(m₁::Multi, m₂::Multi; atol=atol(lentype(m₁)), kwargs...) =
 # POLYTOPE
 # ---------
 
-vertex(m::MultiPolytope, ind) = vertices(m)[ind]
+vertex(m::MultiPolytope, ind) = getindex(Base.iterate(Base.Iterators.drop(eachvertex(m), ind - 1)), 1) # nth(itr, n)
```
juliohm:
```suggestion
vertex(m::MultiPolytope, ind) = first(iterate(Iterators.drop(eachvertex(m), ind - 1)))
```

src/domains/meshes.jl
```diff
@@ -32,6 +32,13 @@ Return the number of vertices of the `mesh`.
 """
 nvertices(m::Mesh) = nvertices(topology(m))
 
+"""
+    eachvertex(mesh)
+
+Return an iterator for the vertices in the `mesh`.
```
juliohm:
```suggestion
Return an iterator for the vertices of the `mesh`.
```

src/geometries/multigeom.jl
ghyatzo: if using `first` the `iterate` is not needed!
can just be `first(Iterators.drop(eachvertex(m), ind-1))`

test/polytopes.jl
```diff
@@ -302,6 +302,10 @@ end
   @test vertices(Ngon{3}(pts...)) == verts
   @test vertices(Ngon{3}(tups...)) == verts
 
+  @test collect(eachvertex(Ngon(pts))) == verts
+  @test @allocated eachvertex(Ngon(pts)) == 0
+  @test @allocated Base.iterate(eachvertex(Ngon(pts))) == 0
```
juliohm: Can we replace this test by a test with a for loop?

```julia
@test @allocated begin
  for v in eachvertex(ngon)
    # do nothing
  end
end 0
```

src/geometries/multigeom.jl
juliohm: that is even better!

test/polytopes.jl
ghyatzo: I've quickly checked with grids and multi, but those still allocate, so I didn't put it for those since setting a magic number of allocation felt finnicky.
but for polytopes it should be always non allocating so it should be fine.

juliohm:
> I would like some pointers WRT the new tests, I'm a bit unsure where best to put them

You can add tests in the corresponding test files. For instance, `test/polytopes.jl`, `test/multigeoms.jl` and `test/meshes.jl`.

> and what you mean exactly when you meant with `@allocated`.

I shared a comment above about that. Just a test that uses the `eachvertex` in a for loop, making sure that the for loop doesn't allocate. You could even nest for loops to make sure that everything is working as expected, motivated by the original discussion on Discourse.

> I've also took the liberty of rewriting some `vertices()` function to instead `collect`ing the iterator instead of repeating the same thing.

Does it lead to the same type inference? I know that list comprehensions are very efficient, and that the compiler can do magic with them. I am not sure if this is the case with the `collect` of generator.

ghyatzo:
> Does it lead to the same type inference? I know that list comprehensions are very efficient, and that the compiler can do magic with them. I am not sure if this is the case with the collect of generator.

I'll do some tests, but I expect it does, since I have the sneaking suspicion that doing `[.. for ... in ...]` is just syntax sugar for `collect((... for ... in ...))`

Edit:
Indeed
```
julia> Meta.@lower [x for x in 1:10]
:($(Expr(:thunk, CodeInfo(
    @ none within `top-level scope`
1 ─ %1 = 1:10
│   %2 = Base.Generator(Base.identity, %1)
│   %3 = Base.collect(%2)
└──      return %3
))))
```

ghyatzo:
> I shared a comment above about that. Just a test that uses the eachvertex in a for loop, making sure that the for loop doesn't allocate. You could even nest for loops to make sure that everything is working as expected, motivated by the original discussion on Discourse.

I am not sure how to exclude the allocation of the object being iterated over, like one can do with `@btime`.
for example:
```
julia> rg = RegularGrid(point1, point2, (1, 1))
julia> @btime eachindex($rg)
  2.416 ns (0 allocations: 0 bytes)
julia> @btime iterate(eachindex($rg))
  3.041 ns (0 allocations: 0 bytes)
```
but
```
julia> @allocated for v in eachvertex(rg)
       end
227008
# this does not happen for the simple polytopes
julia> @allocated for v in eachvertex(t()) #<--- t() is a short helper that builds a random Triangle
       end
0
```
another example:
```
julia> @btime for _ in eachvertex_true($rg)
       end
  16.032 ns (0 allocations: 0 bytes)

julia> @btime for _ in eachvertex_true(rg)
       end
  60.000 μs (1478 allocations: 63.55 KiB)
```

juliohm:
The vertices of the `RegularGrid` subtypes are built on the fly upon demand. It is strange to see allocations. All vertices are stored in static memory, for instance.

ghyatzo:
Seems like all the allocations arrive from here:
```julia
function iterate(g::Generator, s...)
    @inline
    y = iterate(g.iter, s...)
    y === nothing && return nothing
    y = y::Tuple{Any, Any} # try to give inference some idea of what to expect about the behavior of the next line
    return (g.f(y[1]), y[2])   # <---  this line here in particular
end
```
which is the code for iterating generators in Base.

in particular, I've seen these allocations using `@check_allock`:
```julia
julia> try ttt() catch err; err.errors[1] end
Allocating runtime call to "jl_f_getfield" in ./Base.jl:49
  | getproperty(x, f::Symbol) = (@inline; getfield(x, f))

Stacktrace:
 [1] getproperty
   @ ./Base.jl:49 [inlined]
 [2] iterate
   @ ./generator.jl:48 [inlined]
 [3] var"##ttt#446"()
   @ Main ./REPL[190]:2


julia> try ttt() catch err; err.errors[2] end
Allocating runtime call to "jl_get_nth_field_checked" in ./tuple.jl:31
  | getindex(@nospecialize(t::Tuple), i::Int) = getfield(t, i, @_boundscheck)

Stacktrace:
 [1] getindex
   @ ./tuple.jl:31 [inlined]
 [2] iterate
   @ ./generator.jl:48 [inlined]
 [3] var"##ttt#446"()
   @ Main ./REPL[190]:2

julia> try ttt() catch err; err.errors[3] end
Dynamic dispatch in ./generator.jl:48
  | return (g.f(y[1]), y[2])

Stacktrace:
 [1] iterate
   @ ./generator.jl:48 [inlined]
 [2] var"##ttt#446"()
   @ Main ./REPL[190]:2


julia> try ttt() catch err; err.errors[4] end
Allocating runtime call to "jl_f_getfield" in ./Base.jl:49
  | getproperty(x, f::Symbol) = (@inline; getfield(x, f))

Stacktrace:
 [1] getproperty
   @ ./Base.jl:49 [inlined]
 [2] iterate
   @ ./generator.jl:45 [inlined]
 [3] var"##ttt#446"()
   @ Main ./REPL[190]:3


julia> try ttt() catch err; err.errors[5] end
Dynamic dispatch to function iterate in ./generator.jl:45
  | y = iterate(g.iter, s...)

Stacktrace:
 [1] iterate
   @ ./generator.jl:45 [inlined]
 [2] var"##ttt#446"()
   @ Main ./REPL[190]:3


julia> try ttt() catch err; err.errors[6] end
Allocating runtime call to "jl_f_getfield" in ./Base.jl:49
  | getproperty(x, f::Symbol) = (@inline; getfield(x, f))

Stacktrace:
 [1] getproperty
   @ ./Base.jl:49 [inlined]
 [2] iterate
   @ ./generator.jl:48 [inlined]
 [3] var"##ttt#446"()
   @ Main ./REPL[190]:3


julia> try ttt() catch err; err.errors[7] end
Allocating runtime call to "jl_get_nth_field_checked" in ./tuple.jl:31
  | getindex(@nospecialize(t::Tuple), i::Int) = getfield(t, i, @_boundscheck)

Stacktrace:
 [1] getindex
   @ ./tuple.jl:31 [inlined]
 [2] iterate
   @ ./generator.jl:48 [inlined]
 [3] var"##ttt#446"()
   @ Main ./REPL[190]:3


julia> try ttt() catch err; err.errors[8] end
Dynamic dispatch in ./generator.jl:48
  | return (g.f(y[1]), y[2])

Stacktrace:
 [1] iterate
   @ ./generator.jl:48 [inlined]
 [2] var"##ttt#446"()
   @ Main ./REPL[190]:3
```

not completely sure what is going on. Seems like the compiler is giving up in trying to infer stuff and fallbacks to runtime allocation. In this case `f` is `vertex(m, ind)`

juliohm:
That is very interesting. The `vertex(grid, ind)` should be fully static. From your investigations I understand that the issue is in the nested generator, and that your initial approach with a custom `VertexIter` eliminates these allocations.

If that is the case, we have a good motive to introduce the custom iterator in a new `iterators.jl` file, use it in the `eachvertex` methods defined in the corresponding geometry files, and include tests for full coverage.

Thank you for the deep analysis of memory usage, it really helps with these design choices. 💯

ghyatzo:
I am a doofus, I was checking the allocations using non const global variables, that is why it was struggling.
I've redid the testing locally using a function enclosure and it successfully doesn't allocate for most stuff now, even by just using generators.
```julia
julia> @allocated begin                                                                                                                                                                                                                 
           for v in eachvertex(T)   # const Triangle                                                                                                                                                                                                    
           end                                                                                                                                                                                                                          
       end                                                                                                                                                                                                                              
0                                                                                                                                                                                                                                       

julia> @allocated begin                                                                                                                                                                                                                 
           for v in eachvertex(H)  # const Hexagon                                                                                                                                                                                                     
           end                                                                                                                                                                                                                          
       end                                                                                                                                                                                                                              
0

julia> @allocated begin                                                                                                                                                                                                                 
           for v in eachvertex(M)  # const Multi([T, H])                                                                                                                                                                                                     
           end                                                                                                                                                                                                                          
       end                                                                                                                                                                                                                              
3312 # <--- this still allocates.  

julia> @allocated begin                                                                                                                                                                                                                 
           for v in eachvertex(RG)   # const RegularGrid                                                                                                                                                                                                   
           end                                                                                                                                                                                                                          
       end                                                                                                                                                                                                                              
0

julia> @allocated begin                                                                                                                                                                                                                 
           for v in eachvertex(SG)  # const StructuredGrid                                                                                                                                                                                                     
           end                                                                                                                                                                                                                          
       end                                                                                                                                                                                                                              
0
```

The issues still seems to be the compiler choking on the nested generator at some point.
I'll setup again the custom iterator to see if we can get rid of it!

juliohm:
But if the nested generators work well with non global variables, we are set, right? It is simpler code with the desired behavior.

ghyatzo:
Yeah, but the code I shared above is with constants, not global variables.
And it still allocates. Also inside a function barrier:
```julia
julia> forloops(g) = for v in eachvertex(g) end
julia> @allocated forloops(m)
3024
```

juliohm:
Got it. Thanks for checking!

Em sex., 15 de nov. de 2024, 12:01, cschen ***@***.***>
escreveu:

ghyatzo:
I have been hacking away at it, but I can't seem to get rid of the type instability in the `Multi` iteration.
for example, using as an example `m = Multi([T, H])` where `T` is a `Triangle` and `H` is a `Hexagon`.

the compiler correctly infers that `m` is of type `Multi{M, C, <:Ngon}`, but when iterating over the geometries it gives up on infering the type of each geometry. It knows that `m.geoms` is of type `Vector{Ngon{N}} where N` but it stops there.

Which incurs in dynamic dispatch calls to find the number of vertices through `nvertices` and therefore allocations (although much less than the generators).

I've tried multiple approaches but I fear that without adding some extra information in the structure itself to help the compiler
having complete inference of all elements inside the vector might be hard.

Couldn't find a way around it yet, let me know if you have any insights.

juliohm:
This should be fine for most practical purposes. The case where Multi
becomes relevant is with PolyArea, a different type of Ngon that is not
aware of the number N of vertices at compile time. If you convert both the
Triangle and the Hexagon to PolyArea you get zero allocations?

Although we support Multi of Ngon, and other types of Multi not common
elsewhere, the main use case is with PolyArea.

Em sáb., 16 de nov. de 2024, 19:23, cschen ***@***.***>
escreveu:

ghyatzo:
Yes, I guess I got stuck on a non issue then 😄.
```julia
julia> @allocated for v in eachvertex(mpa) end     # multi area                                                                                                                                                                         
0

julia> @allocated for v in eachvertex(mring) end   # multi ring                                                                                                                                                                         
0

julia> @allocated for v in eachvertex(mngon) end   # multi ngon                                                                                                                                                                         
288

julia> @allocated for v in eachvertex(rg) end      # regular grid                                                                                                                                                                       
0

julia> @allocated for v in eachvertex(sg) end      # structured grid                                                                                                                                                                    
0
```

I'll push the changes, added a check for #1128 as well.

src/geometries/multigeom.jl
```diff
@@ -20,7 +20,12 @@ struct Multi{M<:Manifold,C<:CRS,G<:Geometry{M,C}} <: Geometry{M,C}
 end
 
 # constructor with iterator of geometries
-Multi(geoms) = Multi(collect(geoms))
+
+Multi(geoms) = begin
+  all(gT <: Geometry{M,C} where {M<:Manifold,C<:CRS} for gT in typeof.(geoms)) ||
+  throw("All elements used must be subtypes of the same Geometry{<:Manifold, <:CRS}")
+  Multi(collect(geoms))
+end
```
juliohm: Please remove this change, it is unrelated to the issue, and we want to support heterogeneous geometry lists.

juliohm (changes requested):
Can you please confirm that the custom iterator type was necessary to avoid allocations in the `Multi` of  `PolyArea` too? If that is not the case, let's keep it simple and use the generator of generator approach.

Please remove unrelated changes such as the `Multi` constructor added.

src/geometries/multigeom.jl
ghyatzo: `gT <: Geometry{M, C}` matches everything that is a subtype of geometry. It's not enforcing the same type for all.
Could have been worded better. This is off topic though. removed as requested, I'll make a separate PR if desired.

ghyatzo:
> Can you please confirm that the custom iterator type was necessary to avoid allocations in the Multi of  PolyArea too? If that is not the case, let's keep it simple and use the generator of generator approach.

```julia
julia> evgen(m::Multi) = (v for g in m.geoms for v in vertices(g))
evgen (generic function with 1 method)

julia> @btime iterate(eachvertex(mpa));
  11.803 ns (0 allocations: 0 bytes)

julia> @btime iterate(eachvertex(mring));
  3.375 ns (0 allocations: 0 bytes)

julia> @btime iterate(eachvertex(mngon));
  52.442 ns (1 allocation: 32 bytes)

julia> @btime iterate(evgen(mpa));
  28.057 ns (1 allocation: 48 bytes)

julia> @btime iterate(evgen(mring));
  10.761 ns (1 allocation: 48 bytes)

julia> @btime iterate(evgen(mngon));
  245.726 ns (7 allocations: 384 bytes)
```
this is for a single iteration.

PS: I am having trouble running tests locally as they break on a method error for `SimpleMeshes`.
```julia
ERROR: LoadError: MethodError: no method matching Meshes.SimpleMesh(::Vector{Meshes.Point}, ::Meshes.HalfEdgeTopology)
The type `Meshes.SimpleMesh` exists, but no method is defined for this combination of argument types when trying to construct it.

Closest candidates are:
  Meshes.SimpleMesh(::V, ::TP) where {M<:Meshes.Manifold, C<:CoordRefSystems.CRS, V<:AbstractArray{Meshes.Point{M, C}, 1}, TP<:Meshes.Topology}
   @ Meshes ~/Projects/+julia/Meshes/src/domains/meshes/simplemesh.jl:35
  Meshes.SimpleMesh(::AbstractVector{<:Tuple}, ::Meshes.Topology)
   @ Meshes ~/Projects/+julia/Meshes/src/domains/meshes/simplemesh.jl:39
  Meshes.SimpleMesh(::Any, ::AbstractVector{<:Meshes.Connectivity}; relations)
   @ Meshes ~/Projects/+julia/Meshes/src/domains/meshes/simplemesh.jl:41

Stacktrace:
  [1] topoconvert(TP::Type{Meshes.HalfEdgeTopology}, m::Meshes.CartesianGrid{Meshes.𝔼{2}, CoordRefSystems.Cartesian2D{CoordRefSystems.NoDatum, Unitful.Quantity{Float64, 𝐋, Unitful.FreeUnits{(m,), 𝐋, nothing}}}, 2, Tuple{Unitful.Quantity{Float64, 𝐋, Unitful.FreeUnits{(m,), 𝐋, nothing}}, Unitful.Quantity{Float64, 𝐋, Unitful.FreeUnits{(m,), 𝐋, nothing}}}})
    @ Meshes ~/Projects/+julia/Meshes/src/domains/meshes.jl:123
  [2] top-level scope
    @ ~/Projects/+julia/Meshes/test/meshes.jl:363
  [3] eval
    @ ./boot.jl:430 [inlined]
  [4] include_string(mapexpr::typeof(identity), mod::Module, code::String, filename::String)
    @ Base ./loading.jl:2643
  [5] include_string(m::Module, txt::String, fname::String)
    @ Base ./loading.jl:2653
  [6] #invokelatest#2
    @ ./essentials.jl:1055 [inlined]
  [7] invokelatest
    @ ./essentials.jl:1052 [inlined]
  [8] #8
    @ ~/.local/julia/packages/TestItemRunner/Qioo0/src/TestItemRunner.jl:118 [inlined]
  [9] withpath(f::TestItemRunner.var"#8#12"{String, String, Module}, path::String)
    @ TestItemRunner ~/.local/julia/packages/TestItemRunner/Qioo0/src/vendored_code.jl:7
 [10] #7
    @ ~/.local/julia/packages/TestItemRunner/Qioo0/src/TestItemRunner.jl:117 [inlined]
 [11] cd(f::TestItemRunner.var"#7#11"{String, String, Module}, dir::String)
    @ Base.Filesystem ./file.jl:112
 [12] run_testitem(filepath::String, use_default_usings::Bool, setups::Vector{Symbol}, package_name::String, original_code::String, line::Int64, column::Int64, test_setup_module_set::TestItemRunner.TestSetupModuleSet, testsetups::Dict{Symbol, Any})
    @ TestItemRunner ~/.local/julia/packages/TestItemRunner/Qioo0/src/TestItemRunner.jl:116
 [13] run_tests(path::String; filter::Nothing, verbose::Bool)
    @ TestItemRunner ~/.local/julia/packages/TestItemRunner/Qioo0/src/TestItemRunner.jl:221
 [14] run_tests(path::String)
    @ TestItemRunner ~/.local/julia/packages/TestItemRunner/Qioo0/src/TestItemRunner.jl:133
 [15] top-level scope
    @ ~/Projects/+julia/Meshes/test/runtests.jl:4
 [16] include(fname::String)
    @ Main ./sysimg.jl:38
 [17] top-level scope
    @ none:6
in expression starting at /Users/cshen/Projects/+julia/Meshes/test/meshes.jl:363
in expression starting at /Users/cshen/Projects/+julia/Meshes/test/runtests.jl:4
ERROR: Package Meshes errored during testing
```

src/geometries/multigeom.jl
```diff
@@ -48,9 +48,9 @@ Base.isapprox(m₁::Multi, m₂::Multi; atol=atol(lentype(m₁)), kwargs...) =
 # POLYTOPE
 # ---------
 
-vertex(m::MultiPolytope, ind) = vertices(m)[ind]
+vertex(m::MultiPolytope, ind) = first(Base.Iterators.drop(eachvertex(m), ind - 1)) # nth(itr, n)
```
juliohm:
```suggestion
vertex(m::MultiPolytope, ind) = first(Base.Iterators.drop(eachvertex(m), ind - 1))
```

src/geometries/multigeom.jl
```diff
@@ -48,9 +48,9 @@ Base.isapprox(m₁::Multi, m₂::Multi; atol=atol(lentype(m₁)), kwargs...) =
 # POLYTOPE
 # ---------
 
-vertex(m::MultiPolytope, ind) = vertices(m)[ind]
+vertex(m::MultiPolytope, ind) = first(Base.Iterators.drop(eachvertex(m), ind - 1))
```
juliohm:
```suggestion
vertex(m::MultiPolytope, ind) = first(Iterators.drop(eachvertex(m), ind - 1))
```

juliohm:
I cleaned the PR as much as possible before we can continue with the work here.

From your investigations, the only problematic case happens when we rely on a naive generator of generators in `Multi` geometries. In this case, we can rely on a custom iterator defined in the same file.

The error you are seeing with `SimpleMesh` is probably a consequence of the changes in the `vertices` implementations. Perhaps the new implementation in terms of `eachvertex` is not inferring the concrete `eltype` somehow? Worth checking it.

Thank you for adjusting the PR and considering the suggestions. I think we are almost there for a final review.

src/geometries/multigeom.jl
```diff
@@ -89,3 +91,29 @@ function Base.show(io::IO, ::MIME"text/plain", m::Multi)
...
+  ig > length(itr.el.geoms) && return nothing
+  is = _v_iterate(itr.el.geoms[ig], ivg)
+  is === nothing && return Base.iterate(itr, (ig + 1, 1))
+  v, ivg = is
+  return (v, (ig, ivg))
+end
+
+Base.length(itr::VertexItr{<:MultiPolytope}) = sum(nvertices, itr.el.geoms)
+Base.IteratorSize(::VertexItr) = Base.HasLength()
+Base.eltype(::VertexItr) = Point
```
juliohm: This is not a concrete type. We need to return the underlying Point type of the Multi

ghyatzo: That is a very good point, excuse my sloppiness but my unfamiliarity with the package is biting me. I'll make the change.

ghyatzo:
I'm trying to run the test locally and the `@allocated` macro is being very annoying with ghost allocations when dealing with external variables and requiring a dry run to exclude compilation before each call.

~~What do you think of introducting `@ballocated` from `BenchmarkTools` instead of the plain `@allocated`?~~
EDIT: I've added a simple function to `testutils.jl` that does all the dance without the need of bringing in extra dependencies.

ghyatzo:
might have found another pain point:
```julia
julia> poly
PolyArea
  outer
  └─ Ring((x: 0.0 m, y: 0.0 m), ..., (x: 0.0 m, y: 1.0 m))
  inner
  ├─ Ring((x: 0.2 m, y: 0.2 m), ..., (x: 0.2 m, y: 0.4 m))
  └─ Ring((x: 0.6 m, y: 0.2 m), ..., (x: 0.6 m, y: 0.4 m))

julia> @btime iterate(eachvertex($poly))
  79.460 ns (4 allocations: 448 bytes)
(Point(x: 0.0 m, y: 0.0 m), 1)
```
it might be another case of generators choking on "collection of geometries" types.

EDIT:
ah no: it is due `vertex(p::Polytope, i)` calling `vertices(p)[i]` which calls `vertices(p::PolyArea) = mapreduce(vertices, vcat, rings)` which allocates new memory on each vcat.

I guess now that we have vertex iteration the flows of the `vertex` and `vertices` functions should be revisited to make use of it

src/geometries/multigeom.jl
```diff
@@ -48,12 +48,14 @@ Base.isapprox(m₁::Multi, m₂::Multi; atol=atol(lentype(m₁)), kwargs...) =
...
 
-vertex(m::MultiPolytope, ind) = vertices(m)[ind]
+vertex(m::MultiPolytope, ind) = first(Iterators.drop(eachvertex(m), ind - 1))
 
-vertices(m::MultiPolytope) = [vertex for geom in m.geoms for vertex in vertices(geom)]
+vertices(m::MultiPolytope) = collect(eachvertex(m))
 
 nvertices(m::MultiPolytope) = sum(nvertices, m.geoms)
 
+eachvertex(m::MultiPolytope) = VertexItr(m)
```
eliascarv:
```suggestion
eachvertex(m::MultiPolytope) = (v for g in m.geoms for v in vertices(g))
```
The `VertexItr` is better for Multi with non-homogeneous geometries, but for homogeneous geometries, `Iterators.flatten` is better:
```julia
julia> itr1(m) = Meshes.VertexItr(m)
itr1 (generic function with 1 method)

julia> itr2(m) = (v for g in m.geoms for v in vertices(g))
itr2 (generic function with 1 method)

julia> function test1(m)
           s = 0
           for _ in itr1(m)
               s += 1
           end
           s
       end
test1 (generic function with 1 method)

julia> function test2(m)
           s = 0
           for _ in itr2(m)
               s += 1
           end
           s
       end
test2 (generic function with 1 method)

julia> r1 = rand(Ring, crs=Cartesian2D);

julia> r2 = rand(Ring, crs=Cartesian2D);

julia> r3 = rand(Ring, crs=Cartesian2D);

julia> r4 = rand(Ring, crs=Cartesian2D);

julia> p1 = PolyArea(r1, r2)
PolyArea
  outer
  └─ Ring((x: 0.679604 m, y: 0.060601 m), ..., (x: 0.0708917 m, y: 0.00982635 m))
  inner
  └─ Ring((x: 0.65877 m, y: 0.799253 m), ..., (x: 0.208459 m, y: 0.231627 m))

julia> p2 = PolyArea(r3, r4)
PolyArea
  outer
  └─ Ring((x: 0.792342 m, y: 0.685956 m), ..., (x: 0.0187149 m, y: 0.527211 m))
  inner
  └─ Ring((x: 0.509887 m, y: 0.190736 m), ..., (x: 0.346163 m, y: 0.540988 m))

julia> mp = Multi([p1, p2])
MultiPolyArea
├─ PolyArea(6-Ring, 16-Ring)
└─ PolyArea(21-Ring, 28-Ring)

julia> mr = Multi([r1, r2, r3, r4])
MultiRing
├─ Ring((x: 0.679604 m, y: 0.060601 m), ..., (x: 0.0708917 m, y: 0.00982635 m))
├─ Ring((x: 0.65877 m, y: 0.799253 m), ..., (x: 0.208459 m, y: 0.231627 m))
├─ Ring((x: 0.792342 m, y: 0.685956 m), ..., (x: 0.0187149 m, y: 0.527211 m))
└─ Ring((x: 0.509887 m, y: 0.190736 m), ..., (x: 0.346163 m, y: 0.540988 m))

julia> @btime test1($mp)
  23.461 μs (142 allocations: 49.86 KiB)
71

julia> @btime test1($mr)
  217.921 ns (0 allocations: 0 bytes)
71

julia> @btime test2($mp)
  578.714 ns (4 allocations: 1.25 KiB)
71

julia> @btime test2($mr)
  22.792 ns (0 allocations: 0 bytes)
```

src/geometries/multigeom.jl
```diff
@@ -89,3 +91,28 @@ function Base.show(io::IO, ::MIME"text/plain", m::Multi)
...
+  is = _v_iterate(itr.el.geoms[ig], ivg)
+  is === nothing && return Base.iterate(itr, (ig + 1, 1))
+  v, ivg = is
+  return (v, (ig, ivg))
+end
+
+Base.IteratorSize(::VertexItr) = Base.HasLength()
+Base.IteratorEltype(::VertexItr) = Base.HasEltype()
+Base.length(itr::VertexItr{<:MultiPolytope}) = sum(nvertices, itr.el.geoms)
+Base.eltype(::VertexItr{<:MultiPolytope{K,M,C}}) where {K,M<:Manifold,C<:CRS} = Point{M,C}
```
eliascarv:
```suggestion
```

eliascarv (changes requested):

juliohm:
@ghyatzo I think Iterators.flatten addresses all the issues where `Multi` is relevant as @eliascarv showed above.

Can we go ahead and apply the changes?

test/testutils.jl
```diff
@@ -176,3 +176,11 @@ function _isapproxtest(g::Geometry, ::Val{3})
...
   @test isapprox(g, Translate(0u"m", 0u"m", τ32)(g32), atol=1.1τ32)
 end
+
+function _iter_alloctest(coll::Union{Geometry,Mesh})
+  iterate(eachvertex(coll)) # precompile run
+
+  bytes = @allocated for _ in eachvertex(coll)
+  end
+  bytes
+end
```
eliascarv:
```suggestion
function eachvertexalloc(g)
  iterate(eachvertex(g)) # precompile run

  @allocated for _ in eachvertex(g)
  end
end
```
To follow the naming pattern.

test/polytopes.jl
```diff
@@ -302,6 +312,11 @@ end
   @test vertices(Ngon{3}(pts...)) == verts
   @test vertices(Ngon{3}(tups...)) == verts
 
+  # vertex iteration
+  const ngon = Ngon(pts)
+  @test collect(eachvertex(ngon)) == verts
+  @test _iter_alloctest(ngon) == 0
```
eliascarv:
```suggestion
  ngon = Ngon(pts)
  @test collect(eachvertex(ngon)) == verts
  @test eachvertexalloc(ngon) == 0
```

test/polytopes.jl
```diff
@@ -257,6 +262,11 @@ end
   @test @elapsed(issimple(r)) < 0.02
   @test @allocated(issimple(r)) < 950000
 
+  # vertex iteration
+  const cring = Ring(Point.([(0, 0), (1, 0), (1, 1), (0, 1)]))
+  @test collect(eachvertex(cring)) == Point.([(0, 0), (1, 0), (1, 1), (0, 1)])
+  @test _iter_alloctest(cring) == 0
```
eliascarv:
```suggestion
  r = Ring(cart.([(0, 0), (1, 0), (1, 1), (0, 1)]))
  @test collect(eachvertex(r)) == cart.([(0, 0), (1, 0), (1, 1), (0, 1)])
  @test eachvertexalloc(r) == 0
```

test/polytopes.jl
```diff
@@ -82,6 +82,11 @@
   s = Segment(latlon(0, 135), latlon(0, 45))
   @test_broken measure(s) ≈ 3C / 4
 
+  # vertex iteration
+  const cseg = Segment(Point(1.0, 1.0, 1.0, 1.0), Point(2.0, 2.0, 2.0, 2.0))
+  @test collect(eachvertex(cseg)) == [Point(1.0, 1.0, 1.0, 1.0), Point(2.0, 2.0, 2.0, 2.0)]
+  @test _iter_alloctest(cseg) == 0
```
eliascarv:
```suggestion
  s = Segment(cart(0, 0), cart(1, 1))
  @test collect(eachvertex(s)) == [cart(0, 0), cart(1, 1)]
  @test eachvertexalloc(s) == 0
```

test/multigeoms.jl
```diff
@@ -10,16 +10,21 @@
...
 
+  const cmulti = Multi([poly, poly])
+  @test _iter_alloctest(cmulti) == 0
+
   poly1 = PolyArea(cart.([(0, 0), (1, 0), (1, 1), (0, 1)]))
   poly2 = PolyArea(cart.([(1, 1), (2, 1), (2, 2), (1, 2)]))
   multi = Multi([poly1, poly2])
   @test vertices(multi) == [vertices(poly1); vertices(poly2)]
   @test nvertices(multi) == nvertices(poly1) + nvertices(poly2)
+  @test collect(eachvertex(multi)) == [vertices(poly1); vertices(poly2)]
```
eliascarv:
```suggestion
  @test collect(eachvertex(multi)) == [vertices(poly1); vertices(poly2)]
  @test eachvertexalloc(multi) == 0
```

test/multigeoms.jl
```diff
@@ -10,16 +10,21 @@
...
   @test vertex(multi, 1) == vertex(poly, 1)
+  @test collect(eachvertex(multi)) == [vertices(poly); vertices(poly)]
   @test vertices(multi) == [vertices(poly); vertices(poly)]
   @test nvertices(multi) == nvertices(poly) + nvertices(poly)
   @test boundary(multi) == merge(boundary(poly), boundary(poly))
   @test rings(multi) == [rings(poly); rings(poly)]
 
+  const cmulti = Multi([poly, poly])
+  @test _iter_alloctest(cmulti) == 0
+
```
eliascarv:
```suggestion
```

test/multigeoms.jl
```diff
@@ -10,16 +10,21 @@
   @test crs(multi) <: Cartesian{NoDatum}
   @test Meshes.lentype(multi) == ℳ
   @test vertex(multi, 1) == vertex(poly, 1)
+  @test collect(eachvertex(multi)) == [vertices(poly); vertices(poly)]
```
eliascarv:
```suggestion
  @test collect(eachvertex(multi)) == [vertices(poly); vertices(poly)]
  @test eachvertexalloc(multi) == 0
```

test/meshes.jl
```diff
@@ -873,13 +882,16 @@ end
...
+  @test collect(eachvertex(mesh)) == vertices(mesh)
+  @test _iter_alloctest(mesh) == 0
   @test collect(elements(topology(mesh))) == connect.([(1, 2, 3), (4, 5, 6)])
 
   # merge operation with 3D geometries
   mesh₁ = SimpleMesh(cart.([(0, 0, 0), (1, 0, 0), (0, 1, 0), (0, 0, 1)]), connect.([(1, 2, 3, 4)], Tetrahedron))
   mesh₂ = SimpleMesh(cart.([(1, 0, 0), (1, 1, 0), (0, 1, 0), (1, 1, 1)]), connect.([(1, 2, 3, 4)], Tetrahedron))
   mesh = merge(mesh₁, mesh₂)
   @test vertices(mesh) == [vertices(mesh₁); vertices(mesh₂)]
+  @test collect(eachvertex(mesh)) == vertices(mesh)
```
eliascarv:
```suggestion
  @test collect(eachvertex(mesh)) == vertices(mesh)
  @test eachvertexalloc(mesh) == 0
```

test/meshes.jl
```diff
@@ -873,13 +882,16 @@ end
   mesh₂ = SimpleMesh(cart.([(1, 0), (1, 1), (0, 1)]), connect.([(1, 2, 3)]))
   mesh = merge(mesh₁, mesh₂)
   @test vertices(mesh) == [vertices(mesh₁); vertices(mesh₂)]
+  @test collect(eachvertex(mesh)) == vertices(mesh)
+  @test _iter_alloctest(mesh) == 0
```
eliascarv:
```suggestion
  @test eachvertexalloc(mesh) == 0
```

test/meshes.jl
```diff
@@ -800,6 +806,9 @@ end
   @test area(mesh) ≈ T(1) * u"m^2"
   @test extrema(mesh) == (cart(0, 0), cart(1, 1))
 
+  const cmesh = SimpleMesh(points, connec)
+  @test _iter_alloctest(cmesh) == 0
+
```
eliascarv:
```suggestion
```

test/meshes.jl
```diff
@@ -789,6 +794,7 @@ end
   @test crs(mesh) <: Cartesian{NoDatum}
   @test Meshes.lentype(mesh) == ℳ
   @test vertices(mesh) == points
+  @test collect(eachvertex(mesh)) == points
```
eliascarv:
```suggestion
  @test collect(eachvertex(mesh)) == points
  @test eachvertexalloc(mesh) == 0
```

test/meshes.jl
```diff
@@ -127,6 +127,11 @@
   @test minimum(sub) == Point(Polar(T(1), T(2)))
   @test maximum(sub) == Point(Polar(T(4), T(7)))
 
+  # vertex iteration
+  const rg = RegularGrid(Point(1, 1), Point(2, 2), (1, 1))
+  @test collect(eachvertex(rg)) == Point.([(1, 1), (2, 1), (1, 2), (2, 2)])
+  @test _iter_alloctest(rg) == 0
```
eliascarv:
```suggestion
  grid = RegularGrid(cart(1, 1), cart(2, 2), (1, 1))
  @test collect(eachvertex(grid)) == cart.([(1, 1), (2, 1), (1, 2), (2, 2)])
  @test eachvertexalloc(grid) == 0
```

eliascarv (changes requested):

ghyatzo:
ok accepted everything, can we close this up now?

test/testutils.jl
```diff
@@ -176,3 +176,10 @@ function _isapproxtest(g::Geometry, ::Val{3})
   @test isapprox(g, Translate(0u"m", τ32, 0u"m")(g32), atol=1.1τ32)
   @test isapprox(g, Translate(0u"m", 0u"m", τ32)(g32), atol=1.1τ32)
 end
+
+function eachvertexalloc(g)
+  iterate(eachvertex(g)) # precompile run
+
+  @allocated for _ in eachvertex(coll)
```
eliascarv:
```suggestion
  @allocated for _ in eachvertex(g)
```
Typo.

eliascarv:
Not yet, I will add more tests before merging.

test/testutils.jl
```diff
@@ -176,3 +176,18 @@ function _isapproxtest(g::Geometry, ::Val{3})
   @test isapprox(g, Translate(0u"m", τ32, 0u"m")(g32), atol=1.1τ32)
   @test isapprox(g, Translate(0u"m", 0u"m", τ32)(g32), atol=1.1τ32)
 end
+
+function eachvertexalloc(g)
+  iterate(eachvertex(g)) # precompile run
+
```
juliohm:
```suggestion
```

eliascarv (approved):

juliohm:
Thank you @ghyatzo for initiating this work, and @eliascarv for finishing it up!

We will work on the final detail now with `MultiPolyArea` in a separate PR to make all cases allocation-free.

PR #1135, opened 2024-11-18
Avoid allocations in `vertex` of `PolyArea`

eliascarv:


src/geometries/polytopes/polyarea.jl
```diff
@@ -31,6 +31,18 @@ PolyArea(outer...) = PolyArea(collect(outer))
...
 
+function vertex(p::PolyArea, ind)
+  offset = 0
+  for r in p.rings
+    offset′ = offset
+    offset += nvertices(r)
+    if ind ≤ offset
+      return vertex(r, ind - offset′)
+    end
+  end
```
juliohm:
```suggestion
  offset = 0
  for r in p.rings
    nverts = nvertices(r)
    if ind ≤ offset + nverts
      return vertex(r, ind - offset)
    end
    offset += nverts
  end
```

juliohm (approved):

PR #1140, opened 2024-11-19
Fix viz of projected grids with invalid elements

eliascarv:


juliohm:
Closing this one as too old. The `viz` implementation changed drastically.

Issue #1126, opened 2024-11-09
Add Bentley-Ottmann algorithm


juliohm:
The [Bentley-Ottmann algorithm](https://en.wikipedia.org/wiki/Bentley%E2%80%93Ottmann_algorithm) is a sweep-line algorithm for efficient pairwise segment-segment intersection in 2D space: https://youtu.be/nNtiZM-j3Pk?si=bwPt-btl2S0Ps8oO

I've started implementing this algorithm, but got stuck due to limitations in the AVLTree of the DataStructures.jl package. Below is the draft for future reference, in case someone finds the time to finish it up:

```julia
using Meshes
using DataStructures

"""
    bentleyottmann(segments)

Compute pairwise intersections between n `segments`
in O(n⋅log(n)) time using Bentley-Ottmann sweep line
algorithm.
"""
function bentleyottmann(segments)
  # adjust vertices of segments
  segs = map(segments) do s
    a, b = extrema(s)
    a > b ? reverse(s) : s
  end

  # retrieve relevant info
  s = first(segs)
  p = minimum(s)
  P = typeof(p)
  S = Tuple{P,P}

  # initialization
  𝒬 = AVLTree{P}()
  𝒯 = AVLTree{S}()
  ℒ = Dict{P,Vector{S}}()
  𝒰 = Dict{P,Vector{S}}()
  𝒞 = Dict{P,Vector{S}}()
  for s in segs
    a, b = extrema(s)
    push!(𝒬, a)
    push!(𝒬, b)
    haskey(ℒ, a) ? push!(ℒ[a], (a, b)) : (ℒ[a] = [(a, b)])
    haskey(𝒰, b) ? push!(𝒰[b], (a, b)) : (𝒰[b] = [(a, b)])
    haskey(ℒ, b) || (ℒ[b] = S[])
    haskey(𝒰, a) || (𝒰[a] = S[])
    haskey(𝒞, a) || (𝒞[a] = S[])
    haskey(𝒞, b) || (𝒞[b] = S[])
  end
  m = Point(-Inf, -Inf)
  M = Point(Inf, Inf)
  push!(𝒯, (m, m))
  push!(𝒯, (M, M))

  # sweep line
  I = Dict{P,Vector{S}}()
  while length(𝒬) > 0
    p = 𝒬[1]
    delete!(𝒬, p)
    handle!(I, p, 𝒬, 𝒯, ℒ, 𝒰, 𝒞)
  end
  I
end

[...]

PR #1143, opened 2024-11-21
Add Bentley-Ottman Algorithm

souma4:
see issue [#1126 ](https://github.com/JuliaGeometry/Meshes.jl/issues/1126) and PR (#1125 )[https://github.com/JuliaGeometry/Meshes.jl/pull/1125]


souma4:
finally getting back to this. latest commit message is my perspective

juliohm:
Welcome back, Jeffrey! =)

Could you please refactor the PR to only include the Bentley-Ottman algorithm?

Regarding the output format, we could go for a dict-like data structure as suggested. This could be

- a `Dict{K,V} where {K<:Tuple{Int,Int},V<:Int}` or
- a sparse matrix where the entry `M[i,j]` has an integer code for the intersection type
 
I believe this dict-like encoding is very efficient and the Julia compiler can optimize it away in downstream algorithms. Just be careful to provide the type of keys and values, the default `Dict` constructor can lead to `Any` types.

souma4:
Ah. messed up the refactor. Sorry everyone. I'll make a new draft and cherry pick what I need to clean up the commit history

juliohm:
Git is always tricky. Thanks for closing and starting a fresh one.

PR #1153, opened 2024-12-17
Add `TranformedDomain`

eliascarv:


src/domains/transformeddomain.jl
```diff
@@ -0,0 +1,34 @@
...
+  end
+end
+
+function TranformedDomain(d::Domain, t::Transform)
+  g = t(first(d))
+  TransformedMesh{manifold(g),crs(g)}(d, t)
+end
+
+# specialize constructor to avoid deep structures
+TransformedMesh(d::TranformedDomain, t::Transform) = TransformedMesh(d.domain, d.transform → t)
```
juliohm: This should probably go in transformedmesh.jl

PR #1155, opened 2024-12-18
Add parametric function for `Chain`

JoshuaLampert:
This adds a parametric function for `Chain`s, such that a `Ring` or a `Rope` can be evaluated at a parameter $t\in [0,1]$. My use case is sampling at the boundary of `Ngon`s and `Box`es. I stumbled upon this by trying:
```julia
julia> box = Box((0.0, 0.0), (1.0, 1.0))
Box
├─ min: Point(x: 0.0 m, y: 0.0 m)
└─ max: Point(x: 1.0 m, y: 1.0 m)

julia> ring = boundary(box)
Ring
├─ Point(x: 0.0 m, y: 0.0 m)
├─ Point(x: 1.0 m, y: 0.0 m)
├─ Point(x: 1.0 m, y: 1.0 m)
└─ Point(x: 0.0 m, y: 1.0 m)

julia> sampler = RegularSampling(10)
RegularSampling{1}((10,))

julia> collect(sample(ring, sampler))
10-element Vector{Point{𝔼{2}, CoordRefSystems.Cartesian2D{CoordRefSystems.NoDatum, Unitful.Quantity{Float64, 𝐋, Unitful.FreeUnits{(m,), 𝐋, nothing}}}}}:
 Point(x: 0.0 m, y: 0.0 m)
 Point(x: 0.4 m, y: 0.0 m)
 Point(x: 0.8 m, y: 0.0 m)
 Point(x: 1.0 m, y: 0.19999999999999996 m)
 Point(x: 1.0 m, y: 0.6000000000000001 m)
 Point(x: 1.0 m, y: 1.0 m)
 Point(x: 0.6000000000000001 m, y: 1.0 m)
 Point(x: 0.20000000000000018 m, y: 1.0 m)
 Point(x: 0.0 m, y: 0.7999999999999998 m)
 Point(x: 0.0 m, y: 0.3999999999999999 m)
```

where the last failed before.

cc @mikeingold: We could think of using this to make the specializations for `Rope` and `Ring` in MeshIntegrals.jl obsolete.


src/geometries/polytopes.jl
```diff
@@ -156,6 +156,16 @@ function angles(c::Chain)
   map(i -> ∠(vs[i - 1], vs[i], vs[i + 1]), i1:i2)
 end
 
+function (c::Chain)(t)
+  if t < 0 || t > 1
+    throw(DomainError(t, "c(t) is not defined for t outside [0, 1]."))
+  end
+  s = collect(segments(c))
```
JoshuaLampert: I'm not sure if this is the best way to do it. If there is an alternative that is not or at least less allocating this would be of course preferable.

JoshuaLampert: Would
```julia
  v = vertices(c)
  n = length(v) - !isclosed(c)
  k = max(1, ceil(Int, n * t))
  for (i, seg) in enumerate(segments(c))
    if i == k
      return seg(n * t - k + 1)
    end
  end
```
be better?

mikeingold:
> cc @mikeingold: We could think of using this to make the specializations for `Rope` and `Ring` in MeshIntegrals.jl obsolete.

Without having tried this out to see how well it would work, I’d actually probably lean towards continuing to decompose these into `Segment`s prior to integration. If we treat the entire `Chain` as a single parametric on a common domain, then `differential` will usually be discontinuous in value at those segment boundaries, and step changes in integrand values tend to be pretty catastrophic to performance when using adaptive integration rules.

(I’m actually planning to tweak these decomposing methods to make use of reusable buffers that I think will significantly reduce the performance penalty for performing multiple sub-integrals.)

JoshuaLampert:
Yes, I think you are right @mikeingold. But I think having a parametric function for those geometries can be helpful nevertheless like in the example I gave above.

mikeingold:
Agreed. I can definitely imagine other uses where having these functions would be useful.

src/geometries/polytopes.jl
juliohm: We can write `s, _ = iterate(segments(c), k)` to retrieve the `k`-th (or `k+1`-th?)  element of the generator.

JoshuaLampert: Ah, that's what I was looking for, but missed. Thanks @juliohm. Implemented in [05a195b](https://github.com/JuliaGeometry/Meshes.jl/pull/1155/commits/05a195be68055a2da3cf8bcd49a970aac5bc1017).

src/geometries/polytopes.jl
```diff
@@ -156,6 +156,17 @@ function angles(c::Chain)
   map(i -> ∠(vs[i - 1], vs[i], vs[i + 1]), i1:i2)
 end
 
+function (c::Chain)(t)
+  if t < 0 || t > 1
+    throw(DomainError(t, "c(t) is not defined for t outside [0, 1]."))
+  end
+  v = vertices(c)
+  n = length(v) - !isclosed(c)
```
juliohm:
```suggestion
  n = nvertices(c) - !isclosed(c)
```

src/geometries/polytopes.jl
```diff
@@ -156,6 +156,16 @@ function angles(c::Chain)
   map(i -> ∠(vs[i - 1], vs[i], vs[i + 1]), i1:i2)
 end
 
+function (c::Chain)(t)
+  if t < 0 || t > 1
+    throw(DomainError(t, "c(t) is not defined for t outside [0, 1]."))
+  end
+  n = nvertices(c) - !isclosed(c)
+  k = max(1, ceil(Int, n * t))
```
juliohm: Shouldn't we utilize the `measure` of each segment to find `k`? Say we have a rope `r` with 2 segments along the horizontal axis, the first with `2m` length and the second with `1m` length. What should be the result of `r(0.5)`? I understand that we should pick `k=1` in this case and the final result should be a point with x coordinate at `1.5m` . Are we on the same page?

JoshuaLampert: Yeah, you are right. That makes much more sense.

JoshuaLampert: I'll implement that tomorrow (it's late in Europe).

JoshuaLampert: I rewrote the function in terms of `measure` in [28d07dc](https://github.com/JuliaGeometry/Meshes.jl/pull/1155/commits/28d07dcd6da823ce7504952aeea023f314a357f6). I think it's still not ideal because we need some quite high tolerances in the tests, but at least it should do the job despite the rounding errors.

test/polytopes.jl
```diff
@@ -264,6 +264,21 @@ end
   r = Ring(merc.([(0, 0), (1, 0), (1, 1), (0, 1)]))
   @test crs(centroid(r)) === crs(r)
 
+  # parameterization
```
juliohm, line 267: We need basic tests with `cart` coordinates. Perhaps an example with the `boundary` of the a `Box(cart(0,0), cart(1,1))` and another example with the `Rope(cart(0,0), cart(2, 0), cart(3,0))`?

JoshuaLampert: Added tests in [44c2270](https://github.com/JuliaGeometry/Meshes.jl/pull/1155/commits/44c227040cb63b1b0c33df0fc9ff57dc8c434d81).

test/polytopes.jl
```diff
@@ -264,6 +264,21 @@ end
...
+  @test ri(T(0.5)) ≈ latlon(60, 90) atol = 1e7 * eps(T) * u"m"
+  @test ri(T(0.75)) ≈ latlon(82.5, 75) atol = 1e7 * eps(T) * u"m"
+  @test ri(T(1)) == latlon(45, 0)
+
+  ro = Rope(latlon(45, 0), latlon(45, 90), latlon(90, 90))
+  @test ro(T(0)) == latlon(45, 0)
+  @test ro(T(0.25)) ≈ latlon(45, 39.375) atol = 1e7 * eps(T) * u"m"
+  @test ro(T(0.5)) ≈ latlon(45, 78.75) atol = 1e7 * eps(T) * u"m"
+  @test ro(T(0.75)) ≈ latlon(63.75, 90) atol = 1e7 * eps(T) * u"m"
+  @test ro(T(1)) == latlon(90, 90)
```
juliohm: Approximate comparison with `LatLon` coordinates can be challenging because it is based on a fallback method that converts to `Cartesian` coordinates over the surface of the Earth. A tiny change in latitude/longitude degrees leads to a large distance in meters.

Our `measure` of geodesic `Segment`, i.e., a curved segment over the Earth surface, is not accurate at the moment. For now, I think we can only test the values t=0 and t=1 safely.

JoshuaLampert: I removed the approximate `LatLon` tests in [bcaf0f2](https://github.com/JuliaGeometry/Meshes.jl/pull/1155/commits/bcaf0f261475c278e91e8602d77ee761b4386766).

juliohm (approved):
Thank you @JoshuaLampert. Let's just wait for the tests before merging it.

JoshuaLampert:
When do you plan to release a new version @juliohm?

juliohm:
Just waiting for a new fix on the new ValidCoords transform. Will release a
patch today or tomorrow at most.

Em qui., 19 de dez. de 2024, 10:57, Joshua Lampert ***@***.***>
escreveu:

juliohm:
@JoshuaLampert released a patch to avoid blocking your work: https://github.com/JuliaRegistries/General/pull/121670

Other possible ValidCoords fixes will come in a future patch.

JoshuaLampert:
Great, thanks!

PR #1157, opened 2024-12-30
reject Point construction with more than 3 dimensions

### LINKED ISSUE #1151: Checking Point dimension numbers on construction
Discussion [here](https://julialang.zulipchat.com/#narrow/channel/276201-geostats.2Ejl/topic/Point.20problems) on Zulip but an isolated issue to fix is the opportunity to create `Point` instances with dimensions higher than the rest of the implementations assume. For example a 4 dimensional point can be created:
```
p = Point(1, 2, 3, 3)
├─ x1: 1.0 m
├─ x2: 2.0 m
├─ x3: 3.0 m
└─ x4: 3.0 m
```
but there is no way to access the 4th dimension/coordinate of this point without first converting it to a `Vec` using `to(p)`. It doesn't make sense for dimensions >3 to be accessible by vector but not otherwise. @juliohm feel free to correct/add to this and assign me.

markmbaum:
This small change throws an error if `Point` primitives are constructed with more than three dimensions/coordinates.

closes #1151

### DIFF
--- a/src/Meshes.jl
+++ b/src/Meshes.jl
@@ -14,7 +14,7 @@ using Random
 
 using Bessels: gamma
 using Unitful: AbstractQuantity, numtype
-using CoordRefSystems: Basic, Geographic, Projected
+using CoordRefSystems: Basic, Geographic, Projected, ncoords
 using Distances: PreMetric, Euclidean, Mahalanobis
 using Distances: Haversine, SphericalAngle
 using Distances: evaluate, result_type
--- a/src/geometries/primitives/point.jl
+++ b/src/geometries/primitives/point.jl
@@ -35,6 +35,12 @@ Point(1m, 2m, 3m) # integer is converted to float by design
 """
 struct Point{M<:Manifold,C<:CRS} <: Primitive{M,C}
   coords::C
+  # here the inner constuctor ensures the point does not have more than 3 dimensions
+  function Point{M,C}(coords::C) where {M<:Manifold,C<:CRS}
+    ndim = ncoords(coords)
+    ndim > 3 && error("Points can only be constructed with up to 3 dimensions, but got $ndim.")
+    new{M,C}(coords)
+  end
 end
 
 Point{M}(coords::C) where {M<:Manifold,C<:CRS} = Point{M,C}(coords)


markmbaum:
It appears there is a (probably stale) test that constructs a 4 dimensional point: https://github.com/JuliaGeometry/Meshes.jl/blob/975b238cc0b0cbb1cc33ffb0e2128ae3ea55e6fc/test/primitives.jl#L107

Ok to remove that @juliohm ?

src/Meshes.jl
```diff
@@ -14,7 +14,7 @@ using Random
 
 using Bessels: gamma
 using Unitful: AbstractQuantity, numtype
-using CoordRefSystems: Basic, Geographic, Projected
+using CoordRefSystems: Basic, Geographic, Projected, ncoords
```
juliohm, line 17: Please use this function without importing the name into package scope

src/geometries/primitives/point.jl
```diff
@@ -35,6 +35,12 @@ Point(1m, 2m, 3m) # integer is converted to float by design
 """
 struct Point{M<:Manifold,C<:CRS} <: Primitive{M,C}
   coords::C
+  # here the inner constuctor ensures the point does not have more than 3 dimensions
```
juliohm, line 38:
```suggestion
```

src/geometries/primitives/point.jl
```diff
@@ -35,6 +35,12 @@ Point(1m, 2m, 3m) # integer is converted to float by design
...
+  function Point{M,C}(coords::C) where {M<:Manifold,C<:CRS}
+    ndim = ncoords(coords)
+    ndim > 3 && error("Points can only be constructed with up to 3 dimensions, but got $ndim.")
+    new{M,C}(coords)
+  end
```
juliohm, line 43:
```suggestion
  function Point{M,C}(coords::C) where {M<:Manifold,C<:CRS}
    if CoordRefSystems.ncoords(coords) > 3
      throw(ArgumentError("points can only be constructed with up to 3 coordinates"))
    end
    new{M,C}(coords)
  end
```

juliohm (changes requested):
Thank you for making this constraint explicit. Review comments attached.

We need to think if this error message should be added here or in CoordRefSystems.jl directly. We are basically constraining the `Cartesian` constructor, given that it is the only CRS with varying number of coordinates.

juliohm:
After thinking about it a bit more, the correct place to fix this is CoordRefSystems.jl in the `Cartesian` constructor.

PR #1159, opened 2025-01-14
Replace reduce(hcat, ...) by stack(...)

juliohm:


eliascarv (approved):

juliohm:
The error in the docs is unrelated, probably caused by a bug in the latest Makie.jl release.

PR #1162, opened 2025-01-20
Implement a parametrization function for `Wedge`

mikeingold:
## Changes
- Implements a parametrization function for `Wedge`, `wedge(u, v, w)`:
  - Use vertices to generate two quadrangle faces with consistent point order.
  - Use `(u, v)` to locate corresponding points on each quadrangle.
  - Use `(w)` to locate a point on the segment between those points.
- Adds two `@test`s on this parametrization function:
  - check whether function produced an expected result.
  - check whether out-of-bounds parametric coordinates throw an error.


juliohm:
Did some minor simplifications assuming that the Quadrangle and Segment parametrizations will take care of the type number type T. Let's see if tests pass.

juliohm (approved):

PR #1163, opened 2025-01-20
Implement a parametrization function for `Pyramid`

mikeingold:
## Changes
- Implements a parametrization function for `Pyramid`, `pyramid(u, v, w)`:
  - Use vertices to generate a `Quadrangle` base and locate the vertex.
  - Use `(u, v)` to locate a point on the base quadrangle.
  - Use `(w)` to locate a point on the segment between this base quadrangle point and the vertex.
- Adds three `@test`s on this parametrization function:
  - check whether function produced an expected result.
  - check whether out-of-bounds parametric coordinates throw an error.


mikeingold:
I just discovered an issue with this parametrization function: it doesn't account for the barycentric coordinate system used by `Triangle`. I've reverted to draft and will let you know when this is fixed.

juliohm (approved):

PR #1166, opened 2025-01-29
Refactor regular grid constructors

juliohm:
I believe this is cleaner and retains the exact same behavior for `CartesianGrid`. It also adds some missing constructors for `RegularGrid`.

### DIFF
--- a/src/domains/meshes/cartesiangrid.jl
+++ b/src/domains/meshes/cartesiangrid.jl
@@ -56,10 +56,10 @@ See also [`RegularGrid`](@ref).
 """
 const CartesianGrid{M<:𝔼,C<:Cartesian} = RegularGrid{M,C}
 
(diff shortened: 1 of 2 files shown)


src/domains/meshes/cartesiangrid.jl
```diff
@@ -56,53 +56,10 @@ See also [`RegularGrid`](@ref).
...
-
-# -----------------
-# HELPER FUNCTIONS
-# -----------------
-
-_cartpoint(p) = Point(convert(Cartesian, coords(p)))
+CartesianGrid(args...; kwargs...) = RegularGrid(cartesian.(args)...; kwargs...)
+
+# enforce Cartesian coordinates for all Point arguments
+cartesian(p::Point) = Point(convert(Cartesian, coords(p)))
```
eliascarv:
```suggestion
_cartesian(p::Point) = Point(convert(Cartesian, coords(p)))
```
Maybe this function could be a helper function. It doesn't look like an API function.

PR #1168, opened 2025-02-11
Add Bentley-Ottman Algorithm 

souma4:
see issue [#1126 ](https://github.com/JuliaGeometry/Meshes.jl/issues/1126) and PR (#1125 )[https://github.com/JuliaGeometry/Meshes.jl/pull/1125]

I needed to refactor and I messed up with the last draft #1143. 

This produces a dictionary of vertices with a vector of segment values. Future needs: a general method for rebuilding polygons with this structure.

### DIFF
--- a/src/utils/sweepline.jl
+++ b/src/utils/sweepline.jl
@@ -1,0 +1,236 @@
+# Implementation of Bentley-Ottmann algorith
+# https://en.wikipedia.org/wiki/Bentley%E2%80%93Ottmann_algorithm
+
+using BinaryTrees
+
+"""
+    bentleyottmann(segments)
+
+Compute pairwise intersections between n `segments`
+in O(n⋅log(n)) time using Bentley-Ottmann sweep line
+algorithm.
+"""
+function bentleyottmann(segments)
+  # adjust vertices of segments
+  segs = map(segments) do s
+    a, b = extrema(s)
+    a > b ? reverse(s) : s
+  end
+
+  # retrieve relevant info
+  s = first(segs)
+  p = minimum(s)
+  P = typeof(p)
+  S = Tuple{P,P}
+
+  # initialization
+  𝒬 = BinaryTrees.AVLTree{P}()
+  𝒯 = BinaryTrees.AVLTree{S}()
+  ℒ = Dict{P,Vector{S}}()
+  𝒰 = Dict{P,Vector{S}}()
+  𝒞 = Dict{P,Vector{S}}()
+  for s in segs
+    a, b = extrema(s)
+    BinaryTrees.insert!(𝒬, a)
+    BinaryTrees.insert!(𝒬, b)
+    haskey(ℒ, a) ? push!(ℒ[a], (a, b)) : (ℒ[a] = [(a, b)])
+    haskey(𝒰, b) ? push!(𝒰[b], (a, b)) : (𝒰[b] = [(a, b)])
+    haskey(ℒ, b) || (ℒ[b] = S[])
+    haskey(𝒰, a) || (𝒰[a] = S[])
+    haskey(𝒞, a) || (𝒞[a] = S[])
+    haskey(𝒞, b) || (𝒞[b] = S[])
+  end
+
+  # sweep line
+  I = Dict{P,Vector{S}}()
+  while !isnothing(BinaryTrees.root(𝒬))
+    p = _key(_leftmost(BinaryTrees.root(𝒬)))
+    BinaryTrees.delete!(𝒬, p)
+    handle!(I, p, 𝒬, 𝒯, ℒ, 𝒰, 𝒞)
+  end
+  I
+end
+
+function handle!(I, p, 𝒬, 𝒯, ℒ, 𝒰, 𝒞)
+  start_segments = get(ℒ, p, S[])
+  end_segments = get(𝒰, p, S[])
+  intersection_segments = get(𝒞, p, S[])
+  _process_start_segments!(start_segments, 𝒬, 𝒯, 𝒞)
+  _process_end_segments!(end_segments, 𝒬, 𝒯, 𝒞)
+  _process_intersection_segments!(intersection_segments, 𝒬, 𝒯, 𝒞)
+
+  if length(start_segments ∪ end_segments ∪ intersection_segments) > 1
+    I[p] = start_segments ∪ end_segments ∪ intersection_segments
+  end
+end
+
+function _process_start_segments!(start_segments, 𝒬, 𝒯, 𝒞)
+  [BinaryTrees.insert!(𝒯, s) for s in start_segments]
+  for s in start_segments
+    above, below = find_above_below(BinaryTrees.root(𝒯), BinaryTrees.search(𝒯, s))
+    s = Segment(s)
+    if above !== nothing && below !== nothing
+      new_geom, new_type = _newevent(Segment(_key(above)), Segment(_key(below)))
+      if new_type == IntersectionType(0)
+        BinaryTrees.insert!(𝒬, new_geom)
+        haskey(𝒞, new_geom) ? push!(𝒞[new_geom], _key(above), _key(below)) : (𝒞[new_geom] = [_key(above), _key(below)])
+      end
+    end
+    if below !== nothing
+      new_geom, new_type = _newevent(Segment(_key(below)), s)
+      if new_type == IntersectionType(0)
+        BinaryTrees.insert!(𝒬, new_geom)
+        haskey(𝒞, new_geom) ? push!(𝒞[new_geom], _key(below), _segdata(s)) : (𝒞[new_geom] = [_key(below), _segdata(s)])
+      end
+    end
+    if above !== nothing
+      new_geom, new_type = _newevent(s, Segment(_key(above)))
+      if new_type == IntersectionType(0)
+        BinaryTrees.insert!(𝒬, new_geom)
+        haskey(𝒞, new_geom) ? push!(𝒞[new_geom], _segdata(s), _key(above)) : (𝒞[new_geom] = [_segdata(s), _key(above)])
+      end
+    end
+  end
+end
+
+function _process_end_segments!(end_segments, 𝒬, 𝒯, 𝒞)
+  for s in end_segments
+    above, below = find_above_below(BinaryTrees.root(𝒯), BinaryTrees.search(𝒯, s))
+    BinaryTrees.delete!(𝒯, s)
+    s = Segment(s)
+    if above !== nothing && below !== nothing
+      new_geom, new_type = _newevent(Segment(_key(above)), Segment(_key(below)))
+      if new_type == IntersectionType(0)
+        BinaryTrees.insert!(𝒬, new_geom)
+        haskey(𝒞, new_geom) ? push!(𝒞[new_geom], _key(above), _key(below)) : (𝒞[new_geom] = [_key(above), _key(below)])
+      end
+    end
+  end
+end
+
+function _process_intersection_segments!(intersection_segments, 𝒬, 𝒯, 𝒞)
+  for s in intersection_segments
+    _, below = find_above_below(BinaryTrees.root(𝒯), BinaryTrees.search(𝒯, s))
+    if below !== nothing
+      # Swap positions of s and t in 𝒯
+      BinaryTrees.delete!(𝒯, s)
+      BinaryTrees.delete!(𝒯, _key(below))
+      BinaryTrees.insert!(𝒯, _key(below))
+      BinaryTrees.insert!(𝒯, s)
+
+      # Find segments r and u
+      _, r = find_above_below(BinaryTrees.root(𝒯), BinaryTrees.search(𝒯, _key(below)))
+      u, _ = find_above_below(BinaryTrees.root(𝒯), BinaryTrees.search(𝒯, s))
+
+      # Remove crossing points rs and tu from event queue
+      if r !== nothing
+        new_geom, new_type = _newevent(Segment(_key(r)), Segment(s))
+        if new_type == IntersectionType(0)
+          BinaryTrees.delete!(𝒬, new_geom)
+        end
+      end
+      if u !== nothing
+        new_geom, new_type = _newevent(Segment(_key(u)), Segment(_key(below)))
+        if new_type == IntersectionType(0)
+          BinaryTrees.delete!(𝒬, new_geom)
+        end
+      end
+
+      # Add crossing points rt and su to event queue
+      if r !== nothing
+        new_geom, new_type = _newevent(Segment(_key(r)), Segment(_key(below)))
+        if new_type == IntersectionType(0)
+          BinaryTrees.insert!(𝒬, new_geom)
+          haskey(𝒞, new_geom) ? push!(𝒞[new_geom], _key(r), _key(below)) : (𝒞[new_geom] = [_key(r), _key(below)])
+        end
+      end
+      if u !== nothing
+        new_geom, new_type = _newevent(Segment(_key(u)), Segment(s))
+        if new_type == IntersectionType(0)
+          BinaryTrees.insert!(𝒬, new_geom)
+          haskey(𝒞, new_geom) ? push!(𝒞[new_geom], _key(u), s) : (𝒞[new_geom] = [_key(u), s])
+        end
+      end
+    end
+  end
+end
+
+_segdata(seg::Segment) = seg.vertices.data
+_key(node::BinaryTrees.AVLNode) = node.key
+_key(node::Nothing) = nothing
+_leftmost(node::BinaryTrees.AVLNode) = node.left === nothing ? node : _leftmost(node.left)
+_geom(intersect::Intersection) = intersect.geom
+_type(intersect::Intersection) = intersect.type
+function _newevent(s₁::Segment, s₂::Segment)
+  new_event = intersection(s₁, s₂)
+  if new_event !== nothing
+    _geom(new_event), _type(new_event)
+  else
+    nothing, nothing
+  end
+end
+
+# Helper: return the leftmost node (minimum) in a subtree.
+function _bst_minimum(node::BinaryTrees.AVLNode)
+  while node.left !== nothing
+    node = node.left
+  end
+  return node
+end
+
+# Helper: return the rightmost node (maximum) in a subtree.
+function _bst_maximum(node::BinaryTrees.AVLNode)
+  while node.right !== nothing
+    node = node.right
+  end
+  return node
+end
+
+"""
+    find_above_below(root, x)
+
+Find the node above and below `x` in the binary search tree rooted at `root`.
+Returns a tuple `(above, below)` where `above` is the node with the smallest key
+greater than `x.key` and `below` is the node with the largest key smaller than `x.key`.
+If `x` is not found, returns the best candidates for `above` and `below`.
+"""
+function find_above_below(root::BinaryTrees.AVLNode, x::BinaryTrees.AVLNode)
+  above = nothing
+  below = nothing
+  current = root
+  # Traverse from the root to the target node, updating candidates.
+  while current !== nothing && current.key != x.key
+    if x.key < current.key
+      # current is a potential above (successor)
+      above = current
+      current = current.left
+    else # x.key > current.key
+      # current is a potential below (predecessor)
+      below = current
+      current = current.right
+    end
+  end
+
+  # If the node wasn't found, return the best candidate values
+  if current === nothing
+    return (above, below)
+  end
+
+  # Found the node with key equal to x.key.
+  # Now, if there is a left subtree, the true below (predecessor) is the maximum in that subtree.
+  if current.left !== nothing
+    below = _bst_maximum(current.left)
+  end
+  # Similarly, if there is a right subtree, the true above (successor) is the minimum in that subtree.
+  if current.right !== nothing
+    above = _bst_minimum(current.right)
+  end
+
+  (above, below)
+end
+find_above_below(root::BinaryTrees.AVLNode, x::Nothing) = (nothing, nothing)
+function Segment(node₁::BinaryTrees.BinaryNode, node₂::BinaryTrees.BinaryNode)
+  node₁ = _key(node₁)
+  node₂ = _key(node₂)
+  Segment((node₁, node₂))
+end
--- a/test/utils.jl
+++ b/test/utils.jl
@@ -54,4 +54,14 @@
   c = (T(30), T(60))
   p = latlon(c) |> Proj(Cartesian)
   @inferred Meshes.withcrs(p, c, LatLon)
+
+  # test bentley-ottmann algorithm
+  S = [
+    Segment(cart(0, 0), cart(1, 1)),
+    Segment(cart(1, 0), cart(0, 1)),
+    Segment(cart(0, 0), cart(0, 1)),
+    Segment(cart(0, 0), cart(1, 0)),
+    Segment(cart(0, 1), cart(1, 1)),
+    Segment(cart(1, 0), cart(1, 1))
+  ]
 end


juliohm:
@souma4 please let me know when the PR is ready for review. I have some review comments already, just waiting for your green light.

souma4:
@juliohm Thanks for reminding me!

src/intersections.jl
```diff
@@ -115,3 +115,4 @@ include("intersections/planes.jl")
 include("intersections/boxes.jl")
 include("intersections/polygons.jl")
 include("intersections/domains.jl")
+include("intersections/sweepline.jl")
```
juliohm: Let's move this to our utils folder. It is a bit different than the other files in the intersections folder.

src/intersections/sweepline.jl
```diff
@@ -0,0 +1,240 @@
+# Implementation of Bentley-Ottmann algorith
+# https://en.wikipedia.org/wiki/Bentley%E2%80%93Ottmann_algorithm
+
+using BinaryTrees
```
juliohm: We collect all our import/using statements in the main Meshes.jl file.

src/intersections/sweepline.jl
```diff
@@ -0,0 +1,240 @@
...
+  while !isnothing(BinaryTrees.root(𝒬))
+    p = _key(_leftmost(BinaryTrees.root(𝒬)))
+    BinaryTrees.delete!(𝒬, p)
+    handle!(I, p, 𝒬, 𝒯, ℒ, 𝒰, 𝒞)
+  end
+  I
+end
+
+function handle!(I, p, 𝒬, 𝒯, ℒ, 𝒰, 𝒞)
+  start_segments = get(ℒ, p, S[])
```
juliohm: Our code style avoids underscores in the middle. Let's find different names.

src/intersections/sweepline.jl
```diff
@@ -0,0 +1,240 @@
...
+        if new_type == IntersectionType(0)
+          BinaryTrees.insert!(𝒬, new_geom)
+          haskey(𝒞, new_geom) ? push!(𝒞[new_geom], _key(u), s) : (𝒞[new_geom] = [_key(u), s])
+        end
+      end
+    end
+  end
+end
+
+_segdata(seg::Segment) = seg.vertices.data
```
juliohm: We don't access fields directly. There is always a public api function `vertices(seg)` to get what we need. If there is no such function, then we probably need to add something before the PR.

src/intersections/sweepline.jl
```diff
@@ -0,0 +1,240 @@
...
+_key(node::BinaryTrees.AVLNode) = node.key
+_key(node::Nothing) = nothing
+_leftmost(node::BinaryTrees.AVLNode) = node.left === nothing ? node : _leftmost(node.left)
```
juliohm: Is this a missing API in BinaryTrees.jl? Should we open a PR there to provide general traversal functionality?

src/intersections/sweepline.jl
```diff
@@ -0,0 +1,240 @@
...
+_geom(intersect::Intersection) = intersect.geom
```
juliohm: This should be `get(intersect)`

src/intersections/sweepline.jl
```diff
@@ -0,0 +1,240 @@
...
+_type(intersect::Intersection) = intersect.type
```
juliohm: This should be `type(intersect)`

src/intersections/sweepline.jl
```diff
@@ -0,0 +1,240 @@
...
+function _newevent(s₁::Segment, s₂::Segment)
+  new_event = intersection(s₁, s₂)
+  if new_event !== nothing
```
juliohm: prefer `isnothing` instead of `x !== nothing`, it is not the same check

src/intersections/sweepline.jl
```diff
@@ -0,0 +1,240 @@
...
+    _geom(new_event), _type(new_event)
+  else
+    nothing, nothing
+  end
+end
+
+# Helper: return the leftmost node (minimum) in a subtree.
+function _bst_minimum(node::BinaryTrees.AVLNode)
```
juliohm: Another missing API function in BinaryTrees.jl?

src/intersections/sweepline.jl
```diff
@@ -0,0 +1,240 @@
...
+  end
+  # Similarly, if there is a right subtree, the true above (successor) is the minimum in that subtree.
+  if current.right !== nothing
+    above = _bst_minimum(current.right)
+  end
+
+  (above, below)
+end
+find_above_below(root::BinaryTrees.AVLNode, x::Nothing) = (nothing, nothing)
+function Segment(node₁::BinaryTrees.BinaryNode, node₂::BinaryTrees.BinaryNode)
```
juliohm: Please avoid introducing a function with the same name of an existing geometry type. That can generate invalidations with the Julia compiler, and can be a source of potential confusion.

test/intersections.jl
```diff
@@ -1187,3 +1187,27 @@ end
...
+    Segment(cart(1, 0), cart(1, 1))
+  ]
+  I = BentleyOttmann(S)
+  S_check = Dict(
+    cart(1, 1) => [(cart(0, 0), cart(1, 1)), (cart(0, 1), cart(1, 1)), (cart(1, 0), cart(1, 1))],
+    cart(0, 1) => [(cart(0, 1), cart(1, 0)), (cart(0, 1), cart(1, 1)), (cart(0, 0), cart(0, 1))],
+    cart(0.5, 0.5) => [(cart(0, 0), cart(1, 1)), (cart(0, 1), cart(1, 0))],
+    cart(1, 0) => [(cart(1, 0), cart(1, 1)), (cart(0, 1), cart(1, 0)), (cart(0, 0), cart(1, 0))],
+    cart(0, 0) => [(cart(0, 0), cart(1, 1)), (cart(0, 0), cart(0, 1)), (cart(0, 0), cart(1, 0))]
+  )
```
juliohm: I think we can simplify the output format, and make it more suitable for use in the downstream algorithms (e.g. Weiler-Atherton clipping). Consider an enumeration of the input segments 1, 2, ..., n. The output could be split into three simple vectors to help the Julia compiler:

```julia
points, types, segs = bentleyottmann(segments)
```

where `points[i]` is the i-th intersection point, `types[i]` is its [intersection type](https://github.com/JuliaGeometry/Meshes.jl/blob/9bc87223ca43f0282467e7e931e90d50576c2b74/src/intersections.jl#L5-L30), and `segs[i]` is a vector of indices pointing to the segments involved in the intersection. For example, if segments 1, 2 and 3 intersect in their middle at a point p, we could output

```julia
points = [p]
types = [Crossing]
segs = [[1,2,3]]
```

We should add tests to make sure that the `eltype` of these vectors is inferred correctly.

juliohm (changes requested):
Thank you @souma4 for working on this. It is really nice to see the amazing progress you've made already ❤️ 

Attached is my first round of review with a couple of suggestions to improve the PR. Let's work on it together slowly, possibly improving the BinaryTrees.jl dependency before coming back here.

Looking forward to the next round. It looks very promising!

src/intersections/sweepline.jl
souma4: Clarifying Q. including function names?

src/intersections/sweepline.jl
souma4: I think there's justification to open a PR for traversal algorithms. I'm thinking _leftmost could change to minimum? that's what it is doing and matches other API's. May need to import StatsBase though. Also should we include the find_above_below in that PR? It's basically an advanced search so maybe we can keep it local to Meshes. Let me know.

src/intersections/sweepline.jl
souma4: These actually could be quite useful. build this to handle each node, then the above _leftmost (maybe renamed to minimum?) function can just recurse _this subtree minimum.

src/intersections/sweepline.jl
```diff
@@ -0,0 +1,240 @@
...
+          BinaryTrees.insert!(𝒬, new_geom)
+          haskey(𝒞, new_geom) ? push!(𝒞[new_geom], _key(u), s) : (𝒞[new_geom] = [_key(u), s])
+        end
+      end
+    end
+  end
+end
+
+_segdata(seg::Segment) = seg.vertices.data
+_key(node::BinaryTrees.AVLNode) = node.key
```
eliascarv: Please, use the `BinaryTrees.key(node)` function.

src/intersections/sweepline.jl
juliohm: Yes. We usually avoid underscores. The only place where we add them is in the beginning of internal functions sometimes.

src/intersections/sweepline.jl
juliohm: If we feel that these operations are general operations with AVLtrees, we can work on a PR there first to provide what is needed for this PR here.

src/intersections/sweepline.jl
juliohm: I am not following the details closely but I trust your gut with these design decisions. We can review details when the PR is opened in the BinaryTrees.jl repo. I can sit down and review the structure carefully.

test/intersections.jl
souma4: In trying to implement this, points tend to have lots of segments they're related to (especially if it's a polygon) and can have a mix of types. In turn this can lead to a lot of redundancy. I'm thinking a way of cleaning this up could be using a Dictionary of points that contains a vector containing a tuple of crossing type and a vector of segment indices. This doesn't enable nice points, types, segs output for the entire list of segments, but can lend itself nicely for iterating through each layer by key. What do you think?

juliohm: Could you share an example where we have a mix of intersection types? Perhaps there is a precedence rule that we could exploit to decide which type is more appropriate?

souma4: I think we can leverage a clever precedence. So the newest test Segment vector has an example.

I extended `Segment(Point(0,0), Point(1,1))` to `Segment(Point(0,0), Point(1.1,1.1))` while all else is the same. the output I draw attention to is `(x: 1.0 m, y: 1.0 m) => [(CornerTouching, [5, 6]), (Crossing, [])]`. The index of 1 is the (0,0) (1.1, 1.1) line, but it's completely missed here. Additionally, technically, this intersection point is edge touching from the incoming segments. This occurs because (1,1) is not added to \scrC because it's not a crossing type, so it's never recorded as a notable point for (0,0)(1.1,1.1).  For this one point we have two different intersection types, EdgeTouching and CornerTouching.

The language gets really confusing here because we have crossing types, and touching types. All crossing types create a new vertex. Some touching types create a new vertex. So I see two ways of handling this. Either we take out some resolution (which is what I have right now) where all we care about are corners (already existing vertices) and "crossings" (new vertices). Alternatively, we can record all intersection types and which segments have that intersection. The simplest way would be pushing a Tuple of crossing type and segment index to the dictionary within the processing steps for every segment. This would result in a lot of CornerTouching repetition. The most space efficient way would be using a dictionary of crossing types with the segments pushed within a dictionary of points, but that's a dictionary of dictionaries, which is probably a type nightmare.

Thank you for pressing me on making this return type better. I really like these changes.

Let me know what you think.

juliohm: @souma4 I will wait for the BinaryTrees.jl PR to get merged, and for the rebase of this PR in terms of the new functionality before sharing a second round of review.

In the meantime, I will think about the issue you raised regarding the output format.

juliohm: I called `viz(S)` with the variable `S` defined in the test set:

image

The top right corner contains the intersection of interest. I can see how a fine-grained classification of intersection types can get tricky, and I wonder if we should adopt a more coarse-grained classification in terms of `Crossing`, `Touching` and possibly `Overlapping` (not sure the Bentley-Ottman algorithm can identify overlapping segments).

If we stick to this coarse-grained classification the intersection of interest would be classified as `Touching` because it involves three segments. If we look at the combination of these 3 segments 2 by 2, we either have `CornerTouching` or `EdgeTouching`. In any case they are all `Touching`. We also need to define what to do when we have a mix of `Crossing` and `Touching` pairs. I believe the precedence of `Crossing` is higher because it is a signal that something needs to be done in downstream algorithms (e.g. clipping).

Do you think this classification scheme is enough for our purposes? If that is case, we can come back to the simpler output format with three outputs (points, types, segs).

souma4: looking at [Greiner-Hormann](https://www.sciencedirect.com/science/article/pii/S259014861930007X) and [Martinez](https://wellquite.org/posts/lets_build/polygons/new_algorithm_paper.pdf) more closely on how they label, I think I see something we can do. Martinez tests for "in/out" during the plane sweep. However there is value in splitting planesweep from martinez because there are good reasons to only want points within an arbitrary vector of segments. To implement that I think all we need to do is test each point whether it's in both polygons, then use `Segments` to build the polygon, maybe with the crossing type to help prioritize testing the next point.

 Meanwhile Greiner-hormann is a lot harder. A part of me wonders if it's even worth implementing at that point, but for that I think we need `Crossing`, `Touching` and `Overlapping.` That can tell us whether the next point is outside of the clipping domain or inside while still recognizing self intersects I think. The main challenge is that due to this factoring (which I think is a good, generalized approach), we don't keep track who is in what, so I think that's what a lot of the clipping algorithm will do. So a `CornerTouching` actually doesn't tell us whether the corner is a clipping polygon or the clipped polygon so it's not valuable from a traversal standpoint. I think if we have a priority of `Crossing` `Overlap` `Touching` we can use that to prioritize tests of future points to determine the next line segment.

I don't quite have The Vision for how this will work, so I'm going to stew on this for awhile.

juliohm: The Bentley-Ottman algorithm is quite useful on its own. If you can rebase the PR to use the new features in BinaryTrees.jl I can give another round of review, and add more tests with different sets of segments.

src/utils.jl
```diff
@@ -11,3 +11,4 @@ include("utils/cmp.jl")
 include("utils/units.jl")
 include("utils/crs.jl")
 include("utils/misc.jl")
+include("utils/sweepline.jl")
```
```suggestion
include("utils/intersection.jl")
```

src/utils/sweepline.jl
```diff
@@ -0,0 +1,180 @@
...
+
+Compute pairwise intersections between n `segments`
+in O(n⋅log(n)) time using Bentley-Ottmann sweep line
+algorithm.
+
+Outputs a Dictionary of {Point, Vector{Tuple{Point, Point}}}
+where the key is each intersection point and the values are all
+pairs of segments that intersect at that point.
+"""
+function BentleyOttmann(segments)
```
juliohm:
```suggestion
function bentleyottmann(segments)
```

src/utils/sweepline.jl
```diff
@@ -0,0 +1,180 @@
...
+    haskey(𝒞, b) || (𝒞[b] = S[])
+    lookup[(a, b)] = i
+  end
+
+  # sweep line
+  I = Dict{P,Vector{Tuple{IntersectionType,Vector{Int}}}}()
+  while !isnothing(BinaryTrees.root(𝒬))
+    p = BinaryTrees.key(BinaryTrees.minnode(𝒬))
+    BinaryTrees.delete!(𝒬, p)
+    Meshes.handle!(I, lookup, p, S, 𝒬, 𝒯, ℒ, 𝒰, 𝒞)
```
juliohm:
```suggestion
    handle!(I, lookup, p, S, 𝒬, 𝒯, ℒ, 𝒰, 𝒞)
```

juliohm (changes requested):
Attaching a few minor comments. Tomorrow I will try to find some time to run tests locally.

The good news is that the output of `@code_warntype` is all green.

juliohm:
Tests are failing because the utility function is not exported. We can either export it or qualify the calls in the test suite with the `Meshes`  prefix.

src/utils/sweepline.jl
```diff
@@ -0,0 +1,180 @@
+# Implementation of Bentley-Ottmann algorith
+# https://en.wikipedia.org/wiki/Bentley%E2%80%93Ottmann_algorithm
```
juliohm: Please move this link to the docstring of the function under the `## References` section like we do with other functions and types.

src/utils/sweepline.jl
```diff
@@ -0,0 +1,180 @@
...
+
```
juliohm: Please copy/paste the LICENSE header from other files.

src/utils/sweepline.jl
```diff
@@ -0,0 +1,180 @@
...
+"""
+    BentleyOttmann(segments)
```
juliohm:
```suggestion
    bentleyottmann(segments)
```

src/utils/sweepline.jl
```diff
@@ -0,0 +1,180 @@
...
+    haskey(𝒰, b) ? push!(𝒰[b], (a, b)) : (𝒰[b] = [(a, b)])
+    haskey(ℒ, b) || (ℒ[b] = S[])
+    haskey(𝒰, a) || (𝒰[a] = S[])
+    haskey(𝒞, a) || (𝒞[a] = S[])
+    haskey(𝒞, b) || (𝒞[b] = S[])
+    lookup[(a, b)] = i
+  end
+
+  # sweep line
+  I = Dict{P,Vector{Tuple{IntersectionType,Vector{Int}}}}()
```
juliohm: I am still a bit concerned with this nested structure, and will sit down next week to look into this output format more carefully.

src/utils/sweepline.jl
```diff
@@ -0,0 +1,180 @@
...
+  end
+  I
+end
+
+function handle!(I, lookup, p, S, 𝒬, 𝒯, ℒ, 𝒰, 𝒞)
+  𝒮ₛ = get(ℒ, p, S[])
+  𝒮ₑ = get(𝒰, p, S[])
+  𝒮ᵢ = get(𝒞, p, S[])
+  _processends!(𝒮ₑ, 𝒬, 𝒯, 𝒞)
+  _processstarts!(𝒮ₛ, 𝒬, 𝒯, 𝒞)
```
juliohm: We could use begin/end or start/finish as terminology.

src/utils/sweepline.jl
```diff
@@ -0,0 +1,180 @@
...
+end
+
+function _processends!(𝒮ₑ, 𝒬, 𝒯, 𝒞)
+  for s in 𝒮ₑ
+    prev, next = BinaryTrees.prevnext(𝒯, s)
+    BinaryTrees.delete!(𝒯, s)
+    s = Segment(s)
+    if !isnothing(prev) && !isnothing(next)
+      newgeom, newtype = _newevent(Segment(BinaryTrees.key(next)), Segment(BinaryTrees.key(prev)))
+      if newtype == IntersectionType(0)
```
juliohm: Please replace `IntersectionType(0)` by the corresponding name (e.g. `Crossing`). We can't rely on the order of names in the enum.

juliohm (changes requested):
More basic fixes before the actual review of the algorithm.

src/utils/sweepline.jl
souma4: yeah I am too tbh. I think the dictionary setup is useful because it can let us quickly and easily create and index a given point and explore the segments that make up the intersection types. However, those vectors can pose an allocation issue, and it may get unruly in pathological cases. It also uses the least memory (As far as I can guesstimate). 

I think `EdgeTouching` should be part of the handling for `Crossing` because it's 1. a new vertex 2. means either an incomplete polygon (so impossible with polygon inputs) or a T crossing indicating we switch to the other polygons segments. 

As I read through the [degenerate expansion of GH ](https://www.sciencedirect.com/science/article/pii/S259014861930007X), I wonder if we're all over complicating this. Maybe all we need is a point and indices of segments that intersect at that point. This can then be used to insert vertices into the polygon, where we can use three adjacent vertices (and the connected segment indices to help find those vertices) to determine a cross/bounce and go from there.

Determining whether to traverse counterclockwise or clockwise just seems tricky in general. 

edit: in my tests today, maybe shifting to sets is needed. Because in the current test case `P(1,1)` gets added to `\scrC` twice. So atleast for those internal start/end/crossing dictionaries we can either use a set implementation or a unique check

src/utils/sweepline.jl
souma4: I'll go with begin and end. The papers tend to use "left endpoint" and "right endpoint" but I don't like how that is directional when we are ordering by x, then by y (I'm imagining like a Left->Rright scanning line that upon reaching a point scans Bottom->Top). Start/finish vibes as more function based rather than property based.

src/utils/sweepline.jl
juliohm: > Maybe all we need is a point and indices of segments that intersect at that point

Agree. We could use Bentley-Ottman as a simple pre-processing step to find the segments that intersect, no matter the type of intersection. Downstream clipping algorithms could then run more elaborate intersection algorithms with these segments to identify the intersection type.

Could you please update the PR to return a vector of vector of indices?

src/utils/sweepline.jl
```diff
@@ -0,0 +1,182 @@
+# ------------------------------------------------------------------
+# Licensed under the MIT License. See LICENSE in the project root.
+# ------------------------------------------------------------------
```
juliohm:
```suggestion
# ------------------------------------------------------------------

```

src/utils/sweepline.jl
```diff
@@ -0,0 +1,182 @@
...
+    bentleyottmann(segments)
+
+Compute pairwise intersections between n `segments`
+in O(n⋅log(n)) time using Bentley-Ottmann sweep line
+algorithm.
+
+Outputs a Dictionary of {Point, Vector{Tuple{Point, Point}}}
+where the key is each intersection point and the values are all
+pairs of segments that intersect at that point.
+
```
juliohm:
```suggestion
```

src/utils/sweepline.jl
```diff
@@ -0,0 +1,182 @@
...
+    haskey(ℒ, b) || (ℒ[b] = S[])
+    haskey(𝒰, a) || (𝒰[a] = S[])
+    haskey(𝒞, a) || (𝒞[a] = S[])
+    haskey(𝒞, b) || (𝒞[b] = S[])
+    lookup[(a, b)] = i
+  end
+
+  # sweep line
+  points = Vector{P}()
+  segs = Vector{Vector{Int}}()
```
juliohm: Notice that the variable `segs` stores two values of different type inside this function. This leads to type instability. Please use a different variable name for the output.

Project.toml
```diff
@@ -5,6 +5,7 @@ version = "0.52.2"
 
 [deps]
 Bessels = "0e736298-9ec6-45e8-9647-e4fc86a2fe38"
+BinaryTrees = "289e92be-c05a-437b-9e67-5b0c799891f8"
```
eliascarv, line 8: Please, add compat.

src/Meshes.jl
```diff
@@ -4,6 +4,7 @@
 
 module Meshes
 
+using BinaryTrees
```
eliascarv: Maybe using `import` here would be better. What do you think @juliohm?

src/utils/sweepline.jl
```diff
@@ -0,0 +1,182 @@
...
+  _processintersects!(ℐ, 𝒬, 𝒯, 𝒞)
+  if !isempty(ℬ ∪ ℰ ∪ ℐ)
+    segments = ℬ ∪ ℰ ∪ ℐ
+    push!(points, p)
+    push!(segs, _pushintersection(lookup, segments))
+  end
+end
+
+function _processbegin!(ℬ, 𝒬, 𝒯, 𝒞)
+  [BinaryTrees.insert!(𝒯, s) for s in ℬ]
```
eliascarv:
```suggestion
  for s in ℬ
    BinaryTrees.insert!(𝒯, s)
  end
```
Array comprehension allocates a new object. Avoid this by using a for loop.

src/utils/sweepline.jl
```diff
@@ -0,0 +1,182 @@
...
+  for s in ℬ
+    prev, next = BinaryTrees.prevnext(𝒯, s)
+    s = Segment(s)
+    if !isnothing(prev) && !isnothing(next)
+      newgeom, newtype = _newevent(Segment(BinaryTrees.key(next)), Segment(BinaryTrees.key(prev)))
+      if _checkintersection(newtype)
+        BinaryTrees.insert!(𝒬, newgeom)
+        haskey(𝒞, newgeom) ? push!(𝒞[newgeom], BinaryTrees.key(next), BinaryTrees.key(prev)) :
+        (𝒞[newgeom] = [BinaryTrees.key(next), BinaryTrees.key(prev)])
```
eliascarv: I think `if else` is better for large expressions than `? :`. What do you think @juliohm?

src/utils/sweepline.jl
```diff
@@ -0,0 +1,182 @@
...
+  if !isnothing(newevent)
+    get(newevent), type(newevent)
+  else
+    nothing, nothing
+  end
+end
+
+function _checkintersection(type)
+  type == Crossing || type == EdgeTouching
+end
```
eliascarv:
```suggestion
_checkintersection(type) = type == Crossing || type == EdgeTouching
```

src/utils/sweepline.jl
```diff
@@ -0,0 +1,182 @@
...
+        if _checkintersection(newtype)
+          BinaryTrees.insert!(𝒬, newgeom)
+          haskey(𝒞, newgeom) ? push!(𝒞[newgeom], BinaryTrees.key(u), s) : (𝒞[newgeom] = [BinaryTrees.key(u), s])
+        end
+      end
+    end
+  end
+end
+
+_pushintersection(lookup, segments) = unique([lookup[segment] for segment in segments])
```
eliascarv:
```suggestion
_pushintersection(lookup, segments) = unique(lookup[segment] for segment in segments)
```
The `unique` functions works with generators, in this case, it returns a `Vector`.

src/utils/sweepline.jl
```diff
@@ -0,0 +1,182 @@
...
+  segs = Vector{Vector{Int}}()
+  while !isnothing(BinaryTrees.root(𝒬))
+    p = BinaryTrees.key(BinaryTrees.minnode(𝒬))
+    BinaryTrees.delete!(𝒬, p)
+    handle!(points, segs, lookup, p, S, 𝒬, 𝒯, ℒ, 𝒰, 𝒞)
+  end
+  points, segs
+end
+
+function handle!(points, segs, lookup, p, S, 𝒬, 𝒯, ℒ, 𝒰, 𝒞)
```
eliascarv: `handle!` looks like a helper function to me, I think `_handle!` would be a better name.

src/utils/sweepline.jl
souma4: For clarification, is the issue that seg's is defined twice with different types in the same scope, or is it specifically with the Vector{Vector{Int}} type construction? I'm interpreting it as the former, but I wouldn't be shocked if the latter is the case.

src/utils/sweepline.jl
souma4: I'm considering refactoring this to a helper and calling BinaryTrees.key() within the function call. Something like `_newintersection!(\scrC, newgeom, BinaryTrees.key(next), BinaryTrees.key(prev))`.

src/utils/sweepline.jl
juliohm: The former issue. You can learn more with the `@code_warntype` macro:

```julia
function foo()
  a = 1 # Int
  a = 1.0 # Float64
end

@code_warntype foo()
MethodInstance for foo()
  from foo() @ Main REPL[2]:1
Arguments
  #self#::Core.Const(Main.foo)
Locals
  a::Union{Float64, Int64}
Body::Float64
1 ─     (a = 1)
│       (a = 1.0)
└──     return 1.0
```

Notice the inferred type of the local variable, it is `Union{Float64,Int64}`. This can compromise performance downstream in cascades of union types. Even though it shouldn't cause a problem in this particular case, it is not good practice. The more we use non-union types, the happier is the Julia compiler and the more optimizations we get.

souma4: awesome thank you. I guess I assumed the reassignment could differentiate between a type promotion/change vs a complete object change. Which, tbh, as I think more on that, that is silly. Types are types, regardless of intent, and the compiler is not a mind reader. Thank you.

src/Meshes.jl
juliohm: I think `using` is the correct pattern as we are not extending any method, but just using the functions. The `import` can also be useful if we don't want to pollute the namespace, but in this case the BinaryTrees.jl api is not exported.

Project.toml
```diff
@@ -22,8 +23,15 @@ TiledIteration = "06e1c1a7-607b-532d-9fad-de7d9aa2abac"
 TransformsBase = "28dd2a49-a57a-4bfb-84ca-1a49db9b96b8"
 Unitful = "1986cc42-f94f-5a68-af5c-568840ba703d"
 
+[weakdeps]
+Makie = "ee78f7c6-11fb-53f2-987a-cfe4a2b5a57a"
+
+[extensions]
+MeshesMakieExt = "Makie"
+
```
juliohm, line 31: @souma4 could you please confirm that these changes were produced by the Pkg manager?

src/utils/sweepline.jl
```diff
@@ -0,0 +1,179 @@
...
+    bentleyottmann(segments)
+
+Compute pairwise intersections between n `segments`
+in O(n⋅log(n)) time using Bentley-Ottmann sweep line
+algorithm.
+
+## References
+
+* Bentley, J. L., & Ottmann, T. 1979. [Algorithms for reporting and counting geometric intersections]
+  (https://www.itseng.org/research/papers/topics/VLSI_Physical_Design_Automation/Physical_Verification/DRC/Geometric_Intersection_Problems/1979-Bentley.pdf)
```
juliohm:
```suggestion
  (https://ieeexplore.ieee.org/document/1675432)
```

src/utils/sweepline.jl
```diff
@@ -0,0 +1,179 @@
...
+
+  # sweep line
+  points = Vector{P}()
+  segmentidxs = Vector{Vector{Int}}()
+  while !isnothing(BinaryTrees.root(𝒬))
+    p = BinaryTrees.key(BinaryTrees.minnode(𝒬))
+    BinaryTrees.delete!(𝒬, p)
+    _handle!(points, segmentidxs, lookup, p, S, 𝒬, 𝒯, ℒ, 𝒰, 𝒞)
+  end
+  points, segmentidxs
```
juliohm:
```suggestion
  seginds = Vector{Vector{Int}}()
  while !isnothing(BinaryTrees.root(𝒬))
    p = BinaryTrees.key(BinaryTrees.minnode(𝒬))
    BinaryTrees.delete!(𝒬, p)
    _handle!(points, seginds, lookup, p, S, 𝒬, 𝒯, ℒ, 𝒰, 𝒞)
  end
  points, seginds
```

src/utils/sweepline.jl
juliohm: I agree with @eliascarv that a simple if-else would do the job in this case.

src/utils/sweepline.jl
```diff
@@ -0,0 +1,179 @@
+# ------------------------------------------------------------------
```
juliohm: This file needs to be renamed to match the `include` statement.

Project.toml
souma4: they were using compat and only editing `BinaryTrees.jl`

src/utils/sweepline.jl
souma4: 🤦 sheesh I am so scatterbrained and forgetting so many easy things... thank you for the patience.

juliohm:
@souma4 the tests are still failing. Could you please take a look before I dive into the implementation?

souma4:
For some reason objects were being coerced into `SVector`, My best guess is `S` and `S[]` was somehow falling back onto `SVector`, so I changed over to `V` and `Vector{V}()`. I chose V for `vertices` akin to the parametric types used elsewhere

src/utils/intersect.jl
```diff
@@ -0,0 +1,179 @@
...
+    haskey(𝒰, a) || (𝒰[a] = S[])
+    haskey(𝒞, a) || (𝒞[a] = S[])
+    haskey(𝒞, b) || (𝒞[b] = S[])
+    lookup[(a, b)] = i
+  end
+
+  # sweep line
+  points = Vector{P}()
+  seginds = Vector{Vector{Int}}()
+  while !isnothing(BinaryTrees.root(𝒬))
```
juliohm: Is this asking for a `isempty(tree)` API?

src/utils/intersect.jl
```diff
@@ -0,0 +1,179 @@
...
+  end
+  points, seginds
+end
+
+function _handle!(points, seginds, lookup, p, S, 𝒬, ℛ, ℒ, 𝒰, 𝒞)
+  ℬ = get(ℒ, p, S[])
+  ℰ = get(𝒰, p, S[])
+  ℐ = get(𝒞, p, S[])
+  _processend!(ℰ, 𝒬, ℛ, 𝒞)
+  _processbegin!(ℬ, 𝒬, ℛ, 𝒞)
```
juliohm: Can we swap the order of these lines to better match the algorithm in the paper?

src/utils/intersect.jl
```diff
@@ -0,0 +1,152 @@
...
+end
+
+function _processbegin!(ℬ, 𝒬, ℛ, 𝒞)
+  for s in ℬ
+    BinaryTrees.insert!(ℛ, s)
+  end
+  for s in ℬ
+    prev, next = BinaryTrees.prevnext(ℛ, s)
+    if !isnothing(prev) && !isnothing(next)
+      _newevent!(𝒬, 𝒞, BinaryTrees.key(next), BinaryTrees.key(prev))
```
juliohm: Can we swap the order of `prev` and `next` arguments here for consistency?

src/utils/intersect.jl
```diff
@@ -0,0 +1,152 @@
...
+    end
```
juliohm: I can't match this if block to the algorithm described in the paper. Could you please confirm you are implementing Algorithm 2.2?

src/utils/intersect.jl
```diff
@@ -0,0 +1,152 @@
...
+    BinaryTrees.delete!(ℛ, s)
+  end
+end
+
+function _processintersects!(ℐ, 𝒬, ℛ, 𝒞)
+  for s in ℐ
+    prev, _ = BinaryTrees.prevnext(ℛ, s)
+    if !isnothing(prev)
+      # find segments r and u
+      r, _ = BinaryTrees.prevnext(ℛ, s)
```
juliohm: `r` is the same as `prev` here, so it is strange to call `prevnext` again. Am I missing some important detail?

souma4: You're correct it's the same, I included this to more closely match algorithm details to "swap s and t". I'll rewrite this to just set r = prev to keep the syntax without recomputing.

src/utils/intersect.jl
souma4: I think I know where this came from. It was a hold over from prior attempts to get crossings correct, but this is redundant now that the rest of the code is correct. removing it.

src/utils/intersect.jl
souma4: I think we can at this point. With prior iterations, I swapped the two because segments that should have been removed were still in `\scrR` but that likely came as a result of other redundant code. Tests with swapping them seem to pan out.

src/utils/intersect.jl
souma4: It is asking for that so, yeah it can.

juliohm:
Thank you @souma4! I am cleaning things up a bit before the reviewing the core algorithm. Perhaps we can simplify it further. I will try to do that during the cleanup process.

src/utils/intersect.jl
```diff
@@ -0,0 +1,149 @@
...
+  end
+end
+
+function _processmid!(ℳₚ, 𝒬, ℛ, ℳ)
+  for s in ℳₚ
+    prev, _ = BinaryTrees.prevnext(ℛ, s)
+    if !isnothing(prev)
+      # find segments r and u
+      r = prev
+      _, u = BinaryTrees.prevnext(ℛ, BinaryTrees.key(prev))
```
juliohm: Isn't the next of the prev the same as the original? Am I missing something?

juliohm:
@souma4 could you please rebase this PR onto the latest remote master branch?

I tried to push my rebased changes following the standard git-rebase methodology, but it didn't work. Perhaps you need to enable edits in your fork? It can also be just me not knowing gitology well.

src/utils/intersect.jl
souma4: I think I'm confusing myself with how the algorithm phrases this step. I've updated the code to actually reflect what is supposed to happen here. I think we just got lucky with our tests. It essentially looks for whats below s and whats above t (which is next of s), then handles points off of those.

juliohm:
@souma4 I've added a test with an infinite loop. Appreciate it if you could take a look.

souma4:
So this works, which is awesome. Problem, doing some performance testing shows its n^2. Most of the time spent (according to profview) is occurring in the intersection checks from the `\` solving for intersect parameters. I do wonder, though, if dictionary checking is really adding up. When looking at allocations, the `inds = [lookup[s] for s in ....] ` does appear to be a significant driver.

juliohm:
Really nice progress. How are you checking the n^2 performance?

Some tests are failing with Float32.

souma4:
I checked time complexity with the REPL off hand. created `numbers = 1:10:10_000`, generated n of numbers random segments over numbers, then iterated through each random generation, sending the output of `@belapsed` to a `time` variable. Then plotted it.  I did fit a line by doing `n .* n .* S` where S is a scaling factor I just messed around with until it fit good enough for me.

Yeah, some tests are failing with Float32, I'm not sure why, and I'll deal with it tomorrow haha.

src/utils/intersect.jl
```diff
@@ -0,0 +1,143 @@
...
+        end
+      end
+    end
+  end
+end
+
+function _newevent!(𝒬, ℳ, s₁, s₂; kwargs...)
+  intersection(Segment(s₁), Segment(s₂)) do I
+    if type(I) == Crossing || type(I) == EdgeTouching
+      p = get(I)
```
juliohm: Here we can simply call `p` the result of `roundcoords`

src/utils/intersect.jl
```diff
@@ -0,0 +1,143 @@
...
+    BinaryTrees.insert!(𝒬, b)
+    haskey(ℬ, a) ? push!(ℬ[a], (a, b)) : (ℬ[a] = [(a, b)])
+    haskey(ℰ, b) ? push!(ℰ[b], (a, b)) : (ℰ[b] = [(a, b)])
+    lookup[(a, b)] = i
+  end
+
+  # sweep line algorithm
+  points = Vector{P}()
+  seginds = Vector{Vector{Int}}()
+  visited = Dict{P,Int}()
```
juliohm: I believe we can eliminate this `visited` auxiliary variable if we start the function with all points rounded. The first block of code in the function could round the coords of points and then orient the segments. Everything is done in terms of rounded coordinates.

souma4: so when the `visited` variable is removed, in the second test we get a duplicate of `Point((10,4))`. This is because the adjacency tree can't process segment 4 crossing 10,4 until many of the other segments are dropped. And the `inds` variable simply pushes to the `points` variable.

juliohm:
@souma4 I will try to review tomorrow or over the weekend. Thank you for looking into it! :)

juliohm:
In the meantime, I have a question regarding the overall structure of the algorithm.

I noticed that the pseudo-code by Bentley-Ottman in their original paper has a different structure:

image

It can be read as

```julia
for p in ...
  if s is a segment with p in the left
    # do something
  elseif s is a segment with p in the right
    # do somethine else
  else # p is some intersection
    # do something else
  end
end
```

In this PR the structure is a bit different:

```julia
for p in ...
  for s in left segments
    # do something
  end
  for s in right segments
    # do something else
  end
  for s in middle segments
    # do something else
  end
end
```

I understand that the latter structure is explicitly "greedy" and loops over all segments related to point p. The former structure in the paper seems incomplete in terms of these details. Just double checking if we are not missing something. Did you have a chance to think about it? Are these two structures equivalent?

souma4:
Yeah I've wracked my head over this a good bit. From my testing and thinking, the structures are equivalent when segments are expected to be completely independent (ie no corners) and there aren't more than two lines intersecting. But when corners and more segments are involved, the former algorithm will "skip" line segments and/or not capture intersections.  This is because the earlier segments for a given point may intersect with later segments, but due to handling, this intersection may not be captured because there are segments "in between" the two focal segments within the status structure.

test/utils.jl
```diff
@@ -62,3 +62,63 @@
   @test Meshes.roundcoords(p₂, digits=10) == p₁
   @inferred Meshes.roundcoords(p₁, digits=10)
 end
+
+@testitem "bentleyottmann" setup = [Setup] begin
+  digits = Int(4) # precision in number of digits
```
eliascarv:
```suggestion
  digits = 4 # precision in number of digits
```
This is not necessary, `Int` is the default type for integer literals in Julia.

juliohm:
I introduced a helper `_digits` function to extract the digits based on the `atol` of the underlying floating point type. Let's see if tests pass. If they still fail in Float32, we can relax the heuristic a bit more.

src/utils/intersect.jl
```diff
@@ -0,0 +1,163 @@
...
+      p = roundcoords(get(I); digits)
+      BinaryTrees.delete!(𝒬, p)
+    end
+    nothing
+  end
+end
+
+function _digits(segments)
+  s = first(segments)
+  ℒ = Meshes.lentype(s)
```
eliascarv:
```suggestion
  ℒ = lentype(s)
```

juliohm:
The heuristic for the number of digits worked well. Assuming a tolerance of `1e-10`, we get `10-1` digits by default. The same for `1e-5`, which becomes `5-1` digits.

I noticed that the output of the algorithm is including the original points besides the intersections. For instance, here is what I get in the second test:

```julia
viz(segs)
viz!(points, color="red")
```
image

Could you please take a look to make sure we are only returning intersection points?

souma4:
Oh if we only want intersections (not all points) then this is an easy change. I'll commit fixes and update tests to reflect. Although I am finding some variance in tests, so I'll need to debug and see if it's real or not.

Outside of that, the facetgrid test returns nothing because the facet generator only generates cornerintersections, which are start/endpoints and not included.

juliohm:
> Outside of that, the facetgrid test returns nothing because the facet generator only generates cornerintersections, which are start/endpoints and not included.

Very good point. I think we need to refactor this test by constructing segments manually in a grid-like arrangement. We could do the following instead:

```julia
horizontal = [Segment((1, i), (n, i)) for i in 1:n]
vertical = [Segment((i, 1), (i, n)) for i in 1:n]
segs = [horizontal; vertical]
```

juliohm:
Thank you @souma4 for looking into it ❤️ Tests with random segments show that the algorithm is not there yet. Perhaps we could try to follow the algorithmic structure in Chapter 2 of *Berg et al. Computational Geometry - Applications and Algorithms* instead. This structure is discussed step-by-step in the video I linked in the issue. More specifically, it is discussed in this second video: https://www.youtube.com/watch?v=qkhUNzCGDt0

(If you need access to the book, I have the PDF and can send it to you directly)

We are almost there!

souma4:
agreed. My plan was to pull up De Berg to see exactly what they do. I'm fairly busy this week, but, it'll keep moving haha. I think we are pretty close though.

src/utils/intersect.jl
```diff
@@ -0,0 +1,286 @@
...
+  while !BinaryTrees.isempty(𝒬)
+    # current point (or event)
+    p = BinaryTrees.key(BinaryTrees.minnode(𝒬))
+
+    # delete point from event queue
+    BinaryTrees.delete!(𝒬, p)
+    # handle event, i.e. update 𝒬, ℛ and ℳ
+    ℬₚ = get(ℬ, p, S[]) # segments with p at the begin
+    ℰₚ = get(ℰ, p, S[]) # segments with p at the end
+    P = typeof(p)
```
juliohm: I believe this is not needed, `P` is already defined in the beginning of the function.

src/utils/intersect.jl
```diff
@@ -0,0 +1,285 @@
...
+
+# -----------------
+# HELPER FUNCTIONS
+# -----------------
+
+function _newevent!(𝒬, p, s₁, s₂, digits)
+  intersection(Segment(s₁), Segment(s₂)) do I
+    if type(I) == Crossing || type(I) == EdgeTouching
+      p′ = coordround(get(I); digits)
+      if p′ ≥ p && isnothing(BinaryTrees.search(𝒬, p′))
```
juliohm: I am a bit worried with this check in the sense that we shouldn't need to perform a `search` in the binary tree before inserting the node, specially now that we use `coordround` to quantize the points into discrete coordinate values.

src/utils/intersect.jl
```diff
@@ -0,0 +1,285 @@
...
+
+  # Ensure the point is not the endpoint (avoids duplicates)
+  skip = (x₂ - TOL ≤ x ≤ x₂ + TOL) && (y₂ - TOL ≤ y ≤ y₂ + TOL)
+  # if collinear and not an endpoint
+  collinear = dy * (x - x₁) - dx * (y - y₁)
+  boundcheck = (x₁ - TOL ≤ x ≤ x₂ + TOL) && (y₁ - TOL ≤ y ≤ y₂ + TOL)
+
+  if abs(collinear) ≤ TOL * ℒ #&& boundcheck
+    skip || push!(ℳₚ, seg)
+  end
```
juliohm: Did you find an issue with our built-in intersection method? It is not clear to me yet why we need to handle coordinates manually within a custom intersection algorithm. From what I recall, the Bentley-Ottmann algorithm assumes an atom function `intersect(s1, s2)` with the usual definition.

juliohm:
@souma4 thank you very much for your resilience! I wish Bentley-Ottmann had shared a precise pseudo-code in their original article. That would have saved us a huge amount of time handing these tricky corner cases.

Could you please remind what was the issue with that original vanilla implementation? Was it purely caused by floating point arithmetic or there was something else?

src/utils/intersect.jl
souma4: turns out this wasn't even needed. \scrM was previous just defined the same way as S (`Tuple{P,P}`)and I forgot to delete the internal P creation.

src/utils/intersect.jl
souma4: I did. The only consistent way of calculating the x intersect (which now that I think about this again, maybe x intersect and all derivates should be y intersect, because `x=# intersect segment` can just be seen as a "y intercept", but my intial logic was "this is the x at the intersect"). Is to either do this cross product solution or to set a `Segment((x,-Inf),(x,Inf))` intersection, but in the intersection script, it uses a method that can't compute the solution due to the infinities. Another potential solution is to use a y extent determined in the initialization and send that to find intersection. Right now I'm finding the current `intersect` methods struggle with the curved lines, so I felt fine doing the linear thing.

src/utils/intersect.jl
souma4: Turns out this search isn't even needed.

souma4:
The issue with the original is it struggled with verticals and multiple intersections. Floating point was mainly an implementation detail that didn't transfer over nicely from the good math. But hey, it works, it seems to have good time complexity, woohoo.

src/utils/intersect.jl
juliohm: > Right now I'm finding the current `intersect` methods struggle with the curved lines, so I felt fine doing the linear thing.

Do you mean our `intersect` atom has bugs that we need to tackle first?

> Is to either do this cross product solution or to set a `Segment((x,-Inf),(x,Inf))` intersection, but in the intersection script, it uses a method that can't compute the solution due to the infinities.

Handing coordinates at infinity is indeed tricky. I wonder what the code would look like with the alternative solution

> Another potential solution is to use a y extent determined in the initialization and send that to find intersection.

I understand it would just call our built-in `intersect` with the original segments and 4 additional segments beyond the `boundingbox`. For example, `segs |> boundingbox |> boundary |> segments |> collect`. We could also add an extra room to the bounding box with `Stretch(factor)`.

juliohm:
> The issue with the original is it struggled with verticals and multiple intersections. Floating point was mainly an implementation detail that didn't transfer over nicely from the good math. But hey, it works, it seems to have good time complexity, woohoo.

Let's investigate these two issues separately: verticals and multiple intersections.

Regarding the verticals, what exactly is the issue? I remember the solution in that video involved a small rotation of the sweep line, which was implemented implicitly in the order relation when we compare "x" and then "y" coordinate in order (lexicographical comparison). Our `isless` operation used in the binary tree should already take care of that?

Regarding the multiple intersections, I believe this is a more serious issue. The algorithm presented in the video (and in the book) seemed ok with multiple intersections. The algorithm implemented in the PR may be doing something slightly different, even though you adapted it from the original paper by Bentley-Ottmann.

souma4:
Right now, the algorithm in this PR follows the version from the video and de Berg et al.—essentially modified Bentley-Ottmann.

### Handling Multiple Intersections
I'll start with handling multiple segments intersecting at a point, since that ties into vertical segment handling later. When several segments meet at a point p, we need to:

1. Find which segments intersect at `p` (`_findintersections`)

2. Remove all those segments from the status.

3. Reinsert them in order from bottom to top based on where they intersect the sweepline (I used `_nudge` to make the order correct)

4. Identify the lowest and highest segments that pass through p.

5. Check for new intersections just below the lowest and above the highest segments.

To do this ordering properly, I compute the xintersection in the `_SweepSegment` struct. That allows comparison using a custom `is.less` that stores the segment based on the y-coordinate where each segment intersects the sweep line. This makes the status structure behave correctly—it’s ordered by y-position at the current sweep x.

I tried other methods (hardcoding order, swapping axes, using midpoints, holding segments as values instead of keys), but none handled edge cases well. Luckily, computing the intersection is cheap. Once calculated, comparisons are simple: go right if you're below the segment, left if you're above.

### Status Structure and Staleness
The status tree is ordered by these computed xintersect values. But I only recompute `_SweepSegments` when they're involved in an intersection. That means the structure can get stale: the relative order is still valid, but the actual intersection points might not be.

So, when checking "is this segment active?" we need to recalculate the intersection to be sure. That’s what `_findintersections` does—it performs a guided search using the same relative order logic as before. It stays O(log n), unlike previous O(n) brute-force checks.

I considered trying to do what I'll call a "Meta Tree" (there might be a real name for it) where all nodes are recomputed upon a change of some "meta state," for example, a change in the sweepline. But I opted against that because this would cause redundant calculations since the relative order of the tree is still unchanged. That's why I do the recompute on the fly strategy.

### Verticals
Vertical segments are a special case. They must always be considered the "top" segment at any point they pass through. But I couldn’t just set their xintersect to their top endpoint because it would miss every intersection between `p` and the endpoint. Instead, I make it relative to the current sweep position at `p.y`, while ensuring it’s still above everything else. This effectively "pushes" the vertical segment to the top of the tree as needed.

juliohm:
Thank you @souma4 for the detailed comments. I understand that your `_SweepSegment` struct is an auxiliary type with custom `isless` for the binary tree. And that this is only needed because of the `+/-Inf` coordinates. Is that correct?

Could we try that other approach you suggested where we add two additional sentinel segments in the beginning of the algorithm using the `boundingbox` of the original set of segments? If you do `box = segs |> boundingbox |> Stretch(factor)`, you get a bounding box that covers the original set of segments with some margin. You can then take the left and right segments of the boundary as the sentinels, which will certainly be positioned in the first and last position of the sweep line. Any iteration of the algorithm will update the active segments intersecting the sweep line without ever going beyond these sentinel segments. Makes sense? Please let me know if I am missing something.

Ideally, we could achieve the desired results with two tricks besides the built-in `intersect` predicate:

1. Quantization of coordinates in the beginning of the function and throughout the intersections (our `digits` keyword argument) to avoid floating point issues.
2. Addition of two sentinel segments at "infinity" using the `boundingbox` of the original set of segments.

souma4:
So the auxiliary type is there because we need to order the status structure by where the segments intersect the sweepline. The `+/-Inf` does not affect that. I'd say the `_SweepSegment` structure is necessary for correct ordering, regardless of whether it uses intersect or not. 

Where the `+/-Inf` comes into play is in the computation of intersections. That would allow us to use the `intersect` predicate. My thinking is to use the ymin and ymax from the bounding box (plus stretch factor) to construct the sweepline and modify x to update the sweepline position.

juliohm:
Thanks for pushing this @souma4 ❤️ I will try to review by the end of the week 🙏🏽

souma4:
I'm honestly glad for the extra time, managed to fix some huge uneeded time sinks. I know there's still optimization there (just have to wrap my head around it) and FP32 still failing is aggravating. I need to research more robust checks. I saw one paper suggesting signed area is "second order" which makes FP64 work but not 32. A solution might be buried in there without requiring a whole new data structure. But you'll probably have some ideas to resolve type instabilities in the intersection calc.

src/utils/intersect.jl
```diff
@@ -0,0 +1,348 @@
...
+"""
+function bentleyottmann(segments; digits=_digits(segments))
+  refₚ = first(segments)
+  T = lentype(refₚ)
+  # get the coordinate reference system and manifold for stable point type
+  M = manifold(refₚ)
+  C = crs(refₚ)
+  P = Point{M,C}
+  S = Tuple{P,P}
+  U = Set{S}
```
juliohm: It seems that all you want to do here is extract the type `P`. You can do `P = typeof(minimum(first(segments)))`.

Appreciate it if you could refactor the code with clear comments to help the review. I understand that `S = Tuple{P,P}` will be the type used to represent segments as pairs of points in the algorithm, but `U = Set{S}` is unclear up to this point.

src/utils/intersect.jl
```diff
@@ -0,0 +1,348 @@
...
+  # orient segments
+  segs = map(segments) do s
+    a, b = coordround.(extrema(s), digits=digits)
+    a > b ? (b, a) : (a, b)
+  end
+
+  pmin, pmax = segs |> boundingbox |> Stretch(1.05) |> extrema
+  _, ymin = CoordRefSystems.values(coords(pmin))
+  _, ymax = CoordRefSystems.values(coords(pmax))
+  ybounds = (ymin, ymax)
```
juliohm: This can go in an auxiliary function `_ybounds`. Notice that `1.05` is `Float64`, but we want to be agnostic to the floating point type. You can wrap the value with `T(1.05)` where `T = numtype(lentype(P))`. Also notice that we try to be consistent in our codebase and use `T` for the float type and `\scrL` for the length type in `Cartesian`, `Projected` and other coordinates.

src/utils/intersect.jl
```diff
@@ -0,0 +1,348 @@
...
+
+  # retrieve types
+  # initialization
```
juliohm: Appreciate if you can review the comments. It looks like we are mixing old with new comments.

src/utils/intersect.jl
```diff
@@ -0,0 +1,348 @@
...
+  𝒬 = BinaryTrees.AVLTree{P,Tuple{U,U,U}}()
```
juliohm: It is not clear what `Tuple{U,U,U}` represents.

src/utils/intersect.jl
```diff
@@ -0,0 +1,348 @@
...
+      BinaryTrees.insert!(𝒬, a, (U([(a, b)]), U(), U()))
+    end
+
+    # add ending point and segment
+    bν = BinaryTrees.search(𝒬, b)
+    if !isnothing(bν)
+      union!(BinaryTrees.value(bν)[2], U([(a, b)]))
+    else
+      BinaryTrees.insert!(𝒬, b, (U(), U([(a, b)]), U()))
+    end
```
juliohm: This initialization used to be much cleaner. I am afraid this might be a sign of over-engineering?

src/utils/intersect.jl
```diff
@@ -0,0 +1,348 @@
...
+    ℬ, ℰ, ℳ = BinaryTrees.value(node)
+
+    # crosses that arent endpoints (including them can lead to duplicates)
+    ℳₚ = setdiff(ℳ, ℰ)
+
+    # handle status line
+    _handlestatus!(ℛ, ℬ, ℳₚ, ℰ, sweepline, p)
+
+    # build sorted bundle of segments
+    bundle = @isdefined(bundle) ? (empty!(bundle); bundle) : Vector{_SweepSegment{P,T}}()
```
juliohm: I believe `@isdefined` is not adequate in package code. It is a helper macro intended for REPL use. It is probably slow too.

src/utils/intersect.jl
souma4: I found that for some reason `P = typeof(minimum(first(segments)))` led to suboptimal type inference, so pulling manifold and crs directly resolved it. I've switched it back to the more straightfoward way because the overhead is relatively small

src/utils/intersect.jl
souma4: Updated comments throughout

src/utils/intersect.jl
souma4: The old initialization used dictionaries for beginning, crossing, and ending. We could go back to that, but this current way essentially eliminates extra lookups. I've edited comments to clarify it's purpose. I've also put these in a wrapper to make the script more clear about purpose.

src/utils/intersect.jl
souma4: I've switched this to predefining `bundle`. The `@isdefined` was only needed for the first run anyway

src/Meshes.jl
```diff
@@ -29,6 +30,7 @@ using DelaunayTriangulation: get_polygon_points
 using StatsBase: AbstractWeights, Weights, quantile
 using TiledIteration: TileIterator
 using ScopedValues: ScopedValue
+using AdaptivePredicates: orient2, orient2fast
```
juliohm: `orient2fast` is not used in the PR

Also, could you please confirm that `Meshes.signarea` was not enough and that we really need `orient2` to make this work?

souma4: Thank you for asking me to confirm because `signarea` actually was sufficient. I added AdaptivePredicates.jl as a way to have maximum robustness and to somewhat get around FP32 lack of precision. Plus, the overhead for it was minimal. But `signarea` is giving the same passing tests. I've just left a breadcrumb comment in case we ever need to revisit this.

test/utils.jl
```diff
@@ -80,6 +80,99 @@ end
   @inferred Meshes.coordround(p₁, digits=10)
 end
 
+@testitem "bentleyottmann" setup = [Setup] begin
+  # simple endpoint case
+  segs = Segment.([(cart(0, 0), cart(2, 2)), (cart(0, 2), cart(2, 0)), (cart(0, 1), cart(0.5, 1))])
+  points, seginds = Meshes.bentleyottmann(segs)
+  @test length(points) == 7
+  @test length(seginds) == 7
```
juliohm: I believe we discussed this issue before. The algorithm should only return the intersection points. Right now it is returning all points, leading to unnecessary allocations.

juliohm:
@souma4 I tried to cleanup the code further, but I am finding it difficult to do so. I believe you are used to other programming languages where types appear in many places. Can you please try to reduce the amount of explicit type handling? Many functions like `_ybounds` didn't need a `T` type argument. You could simply extract the type internally in the function body (if necessary).

Also, try to follow our code style with comments in lowercase, avoid unnecessary underscores (e.g., `_SweepLine`), etc. I know it is hard to catch up with the current code style, but the project is so large at this point that you can try to read other source files to mimic the current style as much as possible.

I still believe that we can simplify this implementation further to a point where we can clearly read it and know what is happening without too much back and forth with low-level information passed around. For example, sometimes you define auxiliary functions in terms of `s = (a, b)` and sometimes you define them in terms of `a` and `b` separate arguments. The more consistent these arguments are, the smaller are the chances we hit bugs. Also, try to name all variables consistently everywhere. If a function call a variable `p`, the other function should use `p` for the same purpose. It is hard to track the algorithm otherwise.

I really think we have something promising here, but we will only reach a point where we feel safe about the implementation when these iterations and reviews start to converge into a cohesive, consistent code. Thanks for the amazing effort you already put into this 🙏🏽 ❤️

src/utils/intersect.jl
```diff
@@ -0,0 +1,357 @@
...
+  end
+end
+
+# compute the number of significant digits based on the segment type
+# this is used to determine the precision of the points
+function _digits(segments)
+  seg = first(segments)
+  ℒ = lentype(seg)
+  τ = ustrip(eps(ℒ))
+  round(Int, 0.8 * (-log10(τ))) # 0.8 is a heuristic to avoid numerical issues
```
juliohm, line 243: This 0.8 was not here before. Can we simply use `tau = atol(numtype(lentype(seg)))`?

souma4: That does work, however it does seem to make one more Float32 test fail (easy to skip it). The 0.8 arose because I had switched to using machine epsilon. If I remember right that was largely to more specifically reflect the precision of the numtype. But this atol switch is a bit more forgiving and seems to work. I'm cool with `tau = atol(numtype(lentype(seg))` rather than being reliant on a tuned parameter from `eps`.

juliohm: I don't remember the code that was there before, but the 0.8 factor is very ad-hoc. We shouldn't have that.

src/utils/intersect.jl
```diff
@@ -0,0 +1,346 @@
...
+
+  # extract y coordinate values
+  map(p -> coords(p).y, extrema(sbox))
+end
+
+# convenience function to get the segment from a our AVLNode{SweepSegment} structure
+_keyseg(ns) = _segment(BinaryTrees.key(ns))
+
+# handles  the degenerate case of trivial (0-length) segments
+_istrivial(seg) = seg[1] == seg[2]
```
juliohm: I am concerned that I am still reviewing dead code. This function isn't used anywhere.

juliohm:
Closing in favor of #1251.

We can come back to the Bentley-Ottman algorithm in the future.

PR #1172, opened 2025-02-17
Add iscurve, issurface, issolid predicates

juliohm:
These predicates can be useful in MeshIntegrals.jl and other downstream projects.

### DIFF
--- a/docs/src/predicates.md
+++ b/docs/src/predicates.md
@@ -26,6 +26,24 @@ isparametrized
 paramdim
 ```
 
+## iscurve
+
+```@docs
+iscurve
+```
+
+## issurface
+
+```@docs
+issurface
+```
+
+## issolid
+
+```@docs
+issolid
+```
+
 ## isperiodic
 
 ```@docs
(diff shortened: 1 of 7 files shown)


mikeingold:
I’m just seeing this PR now and wishing I could have incorporated it into the recent `v0.17.0` release. @JoshuaLampert I tend to be cautious about bumping dependency compat minimums outside breaking releases but we do have some precedence for it; do you have strong feelings on whether this should go in `v0.17.1` or wait for `v0.18.0`?

JoshuaLampert:
I mean, v0.17.0 is not released yet. The registrator is still complaining about the release notes for some reason. So we could still put it in v0.17.0 if you prefer that.

mikeingold:
> I mean, v0.17.0 is not released yet. The registrator is still complaining about the release notes for some reason. So we could still put it in v0.17.0 if you prefer that.

Didn’t realize the release bot was still dissatisfied with the latest string. I’m away from my desktop at to moment but I’ll plan to just cancel the release PR, implement this change, and then push a new Re-Release commit.

JoshuaLampert:
Sounds good to me.

PR #1176, opened 2025-03-17
Add rounding utility for points

souma4:
@juliohm small round utility for points. Potential use in #1168 and other algorithms where floating point error may cause problems.

### DIFF
--- a/src/utils/basic.jl
+++ b/src/utils/basic.jl
@@ -55,3 +55,15 @@ Generate the coordinate arrays `XYZ` from the coordinate vectors `xyz`.
   end
   Expr(:tuple, exprs...)
 end
+
+"""
+    round(point, r=RoundNearest; digits=0, base=10)
+    round(point, r=RoundNearest; sigdigits=0)
+
+rounds the coordinates of a point to specified presicion.
+"""
+function round(a::Point, r::RoundingMode=RoundNearest; kwargs...)
+  A = to(a)
+  rounded = Unitful.round.(eltype(A), A, r; kwargs...)
+  Point(Tuple(rounded))
+end
--- a/src/utils/basic.jl
+++ b/src/utils/basic.jl
@@ -62,8 +62,10 @@ end
 
 rounds the coordinates of a point to specified presicion.
 """
-function round(a::Point, r::RoundingMode=RoundNearest; kwargs...)
-  A = to(a)
-  rounded = Unitful.round.(eltype(A), A, r; kwargs...)
-  Point(Tuple(rounded))
+function Base.round(a::Point, r::RoundingMode=RoundNearest; kwargs...)
+  c = coords(a)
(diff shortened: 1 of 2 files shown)


src/utils/basic.jl
```diff
@@ -55,3 +55,15 @@ Generate the coordinate arrays `XYZ` from the coordinate vectors `xyz`.
...
 end
+
+"""
+    round(point, r=RoundNearest; digits=0, base=10)
+    round(point, r=RoundNearest; sigdigits=0)
+
+rounds the coordinates of a point to specified presicion.
+"""
+function round(a::Point, r::RoundingMode=RoundNearest; kwargs...)
+  A = to(a)
```
juliohm: Instead of extracting the Cartesian vector with `to(point)`, you can `c = coords(point)` to get the correct coordinates. You can then round the `CoordRefSystem.values(c)`, and construct a new coordinate with `cnew = constructor(c)(newcoords...)`.

Given the new coordinates, we can wrap them into a `Point(newcoords)` again.

src/utils/basic.jl
```diff
@@ -55,3 +55,15 @@ Generate the coordinate arrays `XYZ` from the coordinate vectors `xyz`.
...
   Expr(:tuple, exprs...)
 end
+
+"""
+    round(point, r=RoundNearest; digits=0, base=10)
+    round(point, r=RoundNearest; sigdigits=0)
+
+rounds the coordinates of a point to specified presicion.
+"""
+function round(a::Point, r::RoundingMode=RoundNearest; kwargs...)
```
juliohm:
```suggestion
function Base.round(a::Point, r::RoundingMode=RoundNearest; kwargs...)
```

I think we can simply add a new method to the existing `round` function.

src/utils/basic.jl
```diff
@@ -55,3 +55,15 @@ Generate the coordinate arrays `XYZ` from the coordinate vectors `xyz`.
...
+  A = to(a)
+  rounded = Unitful.round.(eltype(A), A, r; kwargs...)
```
juliohm:
```suggestion
  rounded = round.(eltype(A), A, r; kwargs...)
```

Could you please double check if that is possible? I am assuming that Unitful.jl also extends the `Base.round` and therefore we could use the function name without the qualifier. The multiple dispatch will automatically pick the Unitful.jl implementation given the unitful arguments.

juliohm (changes requested):
Thank you @souma4 ❤️ Left a few minor suggestions.

src/utils/basic.jl
souma4: I wondered if there were another way to do it without extracting with `to()`.

juliohm (approved):
I changed the function name to `roundcoords` given that it is an internal utility function. Later on we can brainstorm if it would make sense to export it as part of a public API.

PR #1181, opened 2025-04-02
Add no-op `topology(::Topology)` method

halleysfifthinc:
This allows calling `adjacencymatrix` on a `Topology`, which should be fine since the only other use of the `mesh` argument in `adjacencymatrix` is a call to `nfaces`, which already has a method for `Topology`s.

### DIFF
--- a/src/topologies.jl
+++ b/src/topologies.jl
@@ -15,6 +15,8 @@ A data structure for constructing topological relations in a [`Mesh`](@ref).
 """
 abstract type Topology end
 
+topology(t::Topology) = t
+
 """
     vertex(topology, ind)
 


juliohm:
Hi @halleysfifthinc , the `topology` is only defined for `Domain` subtypes, and perhaps it should only be defined for `Mesh<:Domain`. Given a mesh, we extract the topology in order to derive topological relations as described here: https://juliageometry.github.io/MeshesDocs/stable/domains/meshes/#Relations

juliohm:
Perhaps what you are after is a new method of `adjacentmatrix` for `Topology`? That would be a more promising direction to pursue. Some matrices like `measurematrix` rely on the geometry (vertices) of the mesh, so can only be defined for `Mesh`, but `adjacentmatrix` and `laplacematrix` with uniform weights could also be defined for `Topology`. The question is: why we need this extra complexity if the object we often manipulate is a `Mesh`?

Feel free to reach out over our Zulip channel for instant chat. It is better than GitHub to brainstorm ideas.

PR #1182, opened 2025-04-02
Document partition order for `BisectPointPartition`

### LINKED ISSUE #1180: Document order of `BisectPointPartition` partitions as part of the API
I'm using multiple (>2) iterations of `BisectPointPartition` to partition a mesh. I saw in the source code that the partition order is fixed, where behind/below the bisecting plane (`left`) is first, and ahead/above the plane (`right`) is last.

It would be really helpful if I could depend on this order, but it isn't documented in the docstring. So, I'm not sure if I can depend on the current order behavior to be the same in future Meshes.jl versions. Is committing to maintaining that order something you are open to?

halleysfifthinc:
Fixes #1180.

### DIFF
--- a/src/partitioning/bisectpoint.jl
+++ b/src/partitioning/bisectpoint.jl
@@ -8,9 +8,9 @@
 A method for partitioning objects into two half spaces
 defined by a `normal` direction and a reference `point`.
 
-The first half space contains objects on the "left" of (i.e. below) the bisecting plane,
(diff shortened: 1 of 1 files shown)


juliohm:
Thank you @halleysfifthinc ! Could you please add a test to ensure that this order is not broken in future code changes?

juliohm (approved):
I noticed we already have tests for this behavior 💯

PR #1183, opened 2025-04-07
Refactor `adjsort` into `adjsortperm`

halleysfifthinc:
***Benchmarks have been updated!***

Using BenchmarkTools, I measure the following speed and memory improvements in `adjsortperm` over `adjsort`.

<details><summary>Benchmarking setup</summary>
<p>

Using the two meshes from [this gist](https://gist.github.com/halleysfifthinc/5b0d45a44f4dd451daca097aa35b9846).

```julia
quads = GeoIO.load("quads.obj").geometry;
mixed = GeoIO.load("quads_tris.obj").geometry;
quads_connec = collect(faces(topology(quads), 2))
mixed_connec = collect(faces(topology(mixed), 2))
```

</p>
</details> 

Comparison of `adjsortperm` (PR) vs `adjsort` (master) for a 72 tri, 36 quad torus:

```julia
  time:   -97.80% => improvement (5.00% tolerance)
  memory: -98.38% => improvement (1.00% tolerance)
```

<details><summary>Details</summary>
<p>

<details><summary><code>adjsort</code> master</summary>
<p>

```julia
julia> @benchmark adjsort(faces) setup=(faces=shuffle(mixed_connec)) seconds=15 evals=1 samples=15000
```
```
BenchmarkTools.Trial: 15000 samples with 1 evaluation per sample.
 Range (min … max):  788.378 μs …  21.293 ms  ┊ GC (min … max): 0.00% … 95.07%
 Time  (median):     868.029 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   926.702 μs ± 501.305 μs  ┊ GC (mean ± σ):  4.58% ±  7.58%

   ▁▂▃▄▆▇██▇▆▅▄▃▃▂▂▂▁▁▁▁▁▁▁     ▁▂▁                             ▃
  ▇█████████████████████████████████▆▆▆▆▅▅▆▅▆▇▆▆▆▆▅▆▆▄▄▅▄▄▁▅▁▁▃ █
  788 μs        Histogram: log(frequency) by time       1.38 ms <

 Memory estimate: 451.41 KiB, allocs estimate: 8673.
```

</p>
</details> 

<details><summary><code>adjsortperm</code> PR</summary>
<p>

```julia
julia> @benchmark adjsortperm(faces) setup=(faces=shuffle(mixed_connec)) seconds=15 evals=1 samples=15000
```
```
BenchmarkTools.Trial: 15000 samples with 1 evaluation per sample.
 Range (min … max):  14.086 μs …  9.780 ms  ┊ GC (min … max): 0.00% … 99.40%
 Time  (median):     19.307 μs              ┊ GC (median):    0.00%
 Time  (mean ± σ):   20.389 μs ± 79.752 μs  ┊ GC (mean ± σ):  3.18% ±  0.81%

                ▁▄▄▅▇███▇▆▅▅▄▃▁▁                               
  ▁▁▁▁▁▁▂▂▂▃▃▅▆██████████████████▆▆▅▅▄▄▄▄▃▃▃▂▂▂▂▂▂▂▁▁▁▁▁▁▁▁▁▁ ▄
  14.1 μs         Histogram: frequency by time        27.7 μs <

 Memory estimate: 7.32 KiB, allocs estimate: 25.
```

</p>
</details>

</p>
</details> 

Comparison of `adjsortperm` (PR) vs `adjsort` (master) for a 72 quad torus:

```julia
  time:   -91.12% => improvement (5.00% tolerance)
  memory: -95.41% => improvement (1.00% tolerance)
```

<details><summary>Details</summary>
<p>

<details><summary><code>adjsort</code> master</summary>
<p>

```julia
julia> @benchmark adjsort(faces) setup=(faces=shuffle(quads_connec)) seconds=15 evals=1 samples=15000
```
```
BenchmarkTools.Trial: 15000 samples with 1 evaluation per sample.
 Range (min … max):  117.852 μs …  13.738 ms  ┊ GC (min … max): 0.00% … 97.99%
 Time  (median):     125.267 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   144.715 μs ± 270.531 μs  ┊ GC (mean ± σ):  8.43% ±  4.73%

  ▂▇██▇▆▅▄▄▄▃▃▂▁           ▁▂▂▁▁▁ ▁▂▂▂▂▁▁                       ▃
  ████████████████▇▇▇▇▆▅▇▇███████████████▇▇▇▇▆▅▅▄▅▅▃▁▃▃▁▃▅▆▆▅▆▆ █
  118 μs        Histogram: log(frequency) by time        242 μs <

 Memory estimate: 147.16 KiB, allocs estimate: 2691.
```

</p>
</details> 

<details><summary><code>adjsortperm</code> PR</summary>
<p>

```julia
julia> @benchmark adjsortperm(faces) setup=(faces=shuffle(quads_connec)) seconds=15 evals=1 samples=15000
```
```
BenchmarkTools.Trial: 15000 samples with 1 evaluation per sample.
 Range (min … max):   9.238 μs …  6.570 ms  ┊ GC (min … max): 0.00% … 99.14%
 Time  (median):     12.273 μs              ┊ GC (median):    0.00%
 Time  (mean ± σ):   12.852 μs ± 53.564 μs  ┊ GC (mean ± σ):  3.38% ±  0.81%

                ▁▂▅▆▆▇▆█▆▆▅▅▃▂▁                                
  ▁▁▁▁▁▁▁▂▂▃▃▄▆▇███████████████▇▇▅▅▄▃▂▂▂▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁ ▃
  9.24 μs         Histogram: frequency by time        17.5 μs <

 Memory estimate: 6.76 KiB, allocs estimate: 24.
```

</p>
</details> 

</p>
</details>


src/topologies/halfedge.jl
```diff
@@ -253,7 +251,7 @@ function adjsort(elems::AbstractVector{<:Connectivity})
     end
   end
 
-  connect.(adjs)
+  return map(connect, adjs)
```
juliohm:
```suggestion
  map(connect, adjs)
```

src/topologies/halfedge.jl
```diff
@@ -134,7 +133,7 @@ function HalfEdgeTopology(elems::AbstractVector{<:Connectivity}; sort=true)
 
   # sort elements to make sure that they
   # are traversed in adjacent-first order
-  adjelems = sort ? adjsort(elems) : elems
+  adjelems = sort ? adjsort(elems)::typeof(elems) : elems
```
juliohm:
```suggestion
  adjelems = sort ? adjsort(elems) : elems
```

src/topologies/halfedge.jl
```diff
@@ -144,7 +143,7 @@ function HalfEdgeTopology(elems::AbstractVector{<:Connectivity}; sort=true)
   # initialize with first element
   half4pair = Dict{Tuple{Int,Int},HalfEdge}()
   elem = first(adjelems)
-  inds = collect(indices(elem))
+  inds::Vector{Int} = collect(indices(elem))
```
juliohm:
```suggestion
  inds = collect(indices(elem))
```

src/topologies/halfedge.jl
```diff
@@ -184,8 +183,8 @@ function HalfEdgeTopology(elems::AbstractVector{<:Connectivity}; sort=true)
 
   # add missing pointers
   for (e, elem) in Iterators.enumerate(adjelems)
-    inds = CCW[e] ? indices(elem) : reverse(indices(elem))
-    v = CircularVector(collect(inds))
+    inds = CCW[e] ? collect(indices(elem)) : reverse(collect(indices(elem)))
+    v = CircularVector(inds)
```
juliohm: Does this change the performance? The original code is more readable.

src/topologies/halfedge.jl
```diff
@@ -221,8 +219,9 @@ end
 function adjsort(elems::AbstractVector{<:Connectivity})
   # initialize list of adjacent elements
   # with first element from original list
-  list = indices.(elems)
-  adjs = Tuple[popfirst!(list)]
+  list = map(indices, elems)
+  adjs = similar(list, 0)
+  push!(adjs, popfirst!(list))
```
juliohm: Nice code pattern. First time I see `similar` used for empty initialization.

src/topologies/halfedge.jl
```diff
@@ -232,17 +231,16 @@ function adjsort(elems::AbstractVector{<:Connectivity})
...
-      if !isempty(einds)
-        # lookup all elements that share at
-        # least two vertices (i.e. edge)
-        for j in sort(einds, rev=true)
-          if length(vinds ∩ list[j]) > 1
-            found = true
-            push!(adjs, popat!(list, j))
-          end
-        end
+    for i::Int in vinds
```
juliohm:
```suggestion
    for i in vinds
```

juliohm (changes requested):
Hi @halleysfifthinc ! Thank you for trying to improve the performance of this constructor. That is super welcome.

Could you please comment on the need to enforce type annotations? Do they really contribute to the performance gains you are seeing? If we could avoid type annotations as much as possible, that would be ideal.

I left some review comments in a first round of review. Happy to review the other parts of the code later.

src/topologies/halfedge.jl
halleysfifthinc: The compiler struggles with the return type of `adjsort` when `elems` is a mix of `<:Ngon` connectivities (Cthulhu shows the return type as just `AbstractVector`, but we know that the return type of `adjsort` should be the same as the type of `elems`.

juliohm: What if we replace `connect.(adjs)` by `map(connect, adjs)` in the last line of `adjsort`? Does it help inference?

src/topologies/halfedge.jl
halleysfifthinc: `indices` is type-unstable (since the length is part of the `Tuple` type), so the `indices` return type when `elem` can be different `Ngon` subtypes is fairly wide (`NTuple{N, Int64} where N`) which leads to the `inds` type estimate(?) being `Union{Vector{Union{}}, Vector{Int64}}`. However, we know that N will never be less than 1, so the return type should always be `Vector{Int}`.  Immediately `collect`ing keeps the same type for `inds` throughout the function, and the left-hand-side type annotation is stronger (makes a non `Vector{Int}` assignment to `inds` an error).

(There's no obvious way to annotate the `indices` return type any more specifically due to its inherent type instability.)

src/topologies/halfedge.jl
halleysfifthinc: This is necessary to go along with the earlier `inds` type-annotation, but also improves the type stability here even without the earlier annotation, as explained above.

src/topologies/halfedge.jl
juliohm: Are you saying that the inference issue occurs even with `map`?

halleysfifthinc:
The type-annotations were added based on investigation with Cthulhu.jl. The type-annotations benefit the later code by allowing the compiler to work with more specific types[^1]. Justification for each annotation given in the review comments.

In Blender, I created a test mesh (torus, 72 quads), and the benchmarks vs no type annotations is:

<details><summary>All quads, no type-annots benchmark</summary>
<p>

```julia
BenchmarkTools.Trial: 10000 samples with 1 evaluation per sample.
 Range (min … max):  626.260 μs …  32.464 ms  ┊ GC (min … max): 0.00% … 97.20%
 Time  (median):     674.720 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   762.418 μs ± 731.223 μs  ┊ GC (mean ± σ):  7.88% ±  8.53%

   ▃▅▇██▇▆▅▅▄▄▃▂▂▂▁▁▁      ▁▃▄▃▂▁▁▁                             ▂
  ███████████████████████▇████████████▇▆▄▆▅▅▅▆▅▆▄▆▄▅▃▅▃▅▅▅▃▆▅▆▃ █
  626 μs        Histogram: log(frequency) by time       1.11 ms <

 Memory estimate: 573.52 KiB, allocs estimate: 14665.
```

</p>
</details> 

```julia
  time:   +323.22% => regression (5.00% tolerance)
  memory: +271.06% => regression (1.00% tolerance)
```

I then split half the quads into tris (36 quads, 72 tris) and got:

<details><summary>Quads and tris; no type-annots benchmark</summary>
<p>

```julia
BenchmarkTools.Trial: 7528 samples with 1 evaluation per sample.
 Range (min … max):  3.641 ms …  35.721 ms  ┊ GC (min … max): 0.00% … 87.22%
 Time  (median):     3.837 ms               ┊ GC (median):    0.00%
 Time  (mean ± σ):   3.971 ms ± 970.236 μs  ┊ GC (mean ± σ):  2.81% ±  7.91%

  ▆█▆▂  ▂▁                                                    ▁
  █████▇██▆▅▆▄▅▅▅▄▄▁▄▃▄▅▃▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▆▆▇ █
  3.64 ms      Histogram: log(frequency) by time      10.7 ms <

 Memory estimate: 1.32 MiB, allocs estimate: 39578.
```

</p>
</details> 

```julia
  time:   +348.65% => regression (5.00% tolerance)
  memory: +216.15% => regression (1.00% tolerance)
```

[^1]: It seems that the type annotations alone (without the other changes) might actually make things slower, but when combined, enhances the benefit of the other changes.

src/topologies/halfedge.jl
halleysfifthinc: Just removed this annotation for an earlier/superseding annotation on `adjs`.

src/topologies/halfedge.jl
halleysfifthinc: Yes; the inferred return type of `adjsort` is still poor even with `map`.

juliohm:
That is very nice. Could you please share a MWE so that I can reproduce benchmark results and test the PR locally? :)

halleysfifthinc:
[Quads and mixed faces meshes here](https://gist.github.com/halleysfifthinc/5b0d45a44f4dd451daca097aa35b9846).

```julia
using Meshes, GeoIO, BenchmarkTools

quads = GeoIO.load("quads.obj").geometry
quads_tris = GeoIO.load("quads_tris.obj").geometry

@benchmark topoconvert(HalfEdgeTopology, quads)
@benchmark topoconvert(HalfEdgeTopology, quads_tris)
```

`adjsort` is responsible for ~85% of the runtime and ~50% of allocations (count), but the `HalfEdgeTopology(::AbstractVector{Tuple{HalfEdge,HalfEdge}})` constructor contributes ~80% of total allocated amount (from the `Dict`s). So although [47fa3f4](https://github.com/JuliaGeometry/Meshes.jl/pull/1183/commits/47fa3f4d64001f4c4366dce9bcad90a534cbd2e2) only removes a few individual allocations, they are larger allocations and save a decent amount of memory.

src/topologies/halfedge.jl
```diff
@@ -98,11 +98,23 @@ struct HalfEdgeTopology <: Topology
...
+    (h₁, h₂) = isnothing(h₁.elem) ? (h₂, h₁) : (h₁, h₂)
+
+    j = (i - 1) * 2 + 1
+    halfedges[j] = h₁
+    halfedges[j + 1] = h₂
+
+    # map pair of vertices to an edge (i.e. two halves)
+    u, v = h₁.head, h₂.head
+    edge4pair[minmax(u, v)] = i
+  end
```
juliohm: Nice! This change replaces 3 loops by 1 👏🏽

src/topologies/halfedge.jl
juliohm: Hi @halleysfifthinc , after looking into the code more carefully, I believe we can completely erase the type instabilities if we refactor the `adjsort` function to return a `Vector{Vector{Int}}` instead of a `Vector{<:Connectivity}`. The algorithm that follows only consumes the indices.

In the case where `sort=false`, we also need to process the unsorted `elems` into a `Vector{Vector{Int}}` using a simple `map(collect ∘ indices, elems)` call.

I believe that after that everything will be type stable.

halleysfifthinc: :+1:  I will look into that approach.

halleysfifthinc:
I reimplemented `adjsort` as `adjsortperm`, which now returns a sort permutation (faster, and removes the need for the `indexin` call, which was a non-negligible runtime contributor). Before/after benchmarks have been updated in the top comment (its quite a bit faster).

Along the way, I found a weird mesh/topo that is manifold, however, the old `adjsort` led to one face being missing from the final converted topology. Despite my best efforts, I could not determine what/where/why the problem was. [This gist](https://gist.github.com/halleysfifthinc/27236babe51bd01a204b093a70e33c8d) contains two meshes which are identical save for the triangle orders and orientations (same vertices). I am happy to include the bad mesh as a test if you wish.

halleysfifthinc:
[Failed CI](https://github.com/JuliaGeometry/Meshes.jl/actions/runs/14457935044/job/40544973505?pr=1183#step:6:652) doesn't seem related.

src/topologies/halfedge.jl
```diff
@@ -208,52 +206,60 @@ function HalfEdgeTopology(elems::AbstractVector{<:Connectivity}; sort=true)
...
+      while iter ≤ length(oinds)
+        oelem = elems[oinds[iter]]
+        vinds′ = indices(oelem)
+        if any(==(v), vinds′) && !isdisjoint(v!, vinds′)
+          found = true
+          push!(einds, popat!(oinds, iter))
+        else
+          found = false
+          iter += 1
+          continue
```
juliohm: This `continue` has effect? Given that it is the last statement in the `while` loop, I think we can erase it?

juliohm:
@halleysfifthinc thank you for refactoring the code.

I refactored it a bit more to make sure the new algorithm is easy to read. Do we still need to reintroduce type annotations to improve performance? Appreciate it if you could take another look. If type annotations are necessary, we could start with annotating the return type of `adjsortperm`, by typing `einds::Vector{Vector{Int}}` at the last line of the function.

If we could encapsulate these type annotations in the `adjsortperm` function, that would decrease our chances to break type instability in the future.

juliohm:
> Along the way, I found a weird mesh/topo that is manifold, however, the old `adjsort` led to one face being missing from the final converted topology.

Do you mean that the old implementation had a bug? Is it present in the new implementation in this PR?

src/topologies/halfedge.jl
```diff
@@ -220,49 +216,50 @@ function HalfEdgeTopology(elems::AbstractVector{<:Connectivity}; sort=true)
...
-  listi = collect(eachindex(elems)[2:end])
   # initialize list of adjacent elements
   # with first element from original list
-  adjs = Int[1]
-  sizehint!(adjs, length(elems))
+  einds = Int[1]
+  sizehint!(einds, length(elems))
+
+  # remaining list of elements to process
+  oinds = collect(2:length(elems))
```
halleysfifthinc: This is less safe than `collect(eachindex(elems)[2:end])`, because the `elems::AbstractVector` type parameter allows a vector with a `firstindex !== 1`.

```suggestion
  oinds = collect(eachindex(elems)[2:end])
```

juliohm: Isn't the algorithm written in terms of a local enumeration already? We initialize `einds = Int[1]` to mean the first element in the vector of elements, and then enumerate the remaining elements as `2:N`? Later on, we index `oinds[iter]` where `iter` starts as `1` too.

halleysfifthinc: Ah good catch, properly `einds` should be initialized as `einds = Int[firstindex(elems)]` along with the above. All the other indexing is into `oinds` which is a regular `Vector` (i.e. starts at `1`) or indexing into `elems` using (ultimately) values from `oinds` which, with the above, is filled with guaranteed inbounds values produced by `eachindex`.

juliohm: Good point. Feel free to adjust 👍🏽

halleysfifthinc:
Just pushed yet another refactor, this version is much faster still  (and as an unintentional bonus, now actually properly separates connected components, which as previously discussed, is something I need).

I converted to a draft, as I am still unsatisfied with some behaviors, but can't continue spending time on this for the moment.

I will update the benchmarks when I get a chance.

juliohm:
That is really amazing @halleysfifthinc. The new connected components function is quite useful in other contexts. We should probably convert it to a utility function after this PR is merged. Looking forward to it!

halleysfifthinc:
Benchmarks updated (now specifically for `adjsort`/`adjsortperm`)!

Re: `connected_components` utility, I agree. I'd like to see it exported, or `public` at least, so I can use it and avoid the Graphs.jl dependency.

> Do you mean that the old implementation had a bug? Is it present in the new implementation in this PR?

Yes, the old `adjsort` was leading to one face not being added/recognized in the final `HalfEdgeTopology`. That specific problem is no longer present with the new `adjsortperm` implementation.

However, the old `adjsort` was only part of the problem, I believe the main `HalfEdgeTopology(::Vector{<:Connectivity}` function is incorrect/unreliable for certain mixes of face orientations. I have a mesh[^1] which has inconsistencies after being converted to a `HalfEdgeTopology` (`next` or `prev` edges not matching the current `elem`). I have been able to determine that the bug is related to inconsistent orientations among the faces, but I am struggling to find any logical errors in the existing orientation handling code here and above https://github.com/JuliaGeometry/Meshes.jl/blob/28c75552ef2bf3f8da9050a853f0a7025bed9217/src/topologies/halfedge.jl#L177-L182

[^1]:  I can't share the mesh, and I don't understand the cause of the bug well enough to simplify the mesh like I did for the meshes attached to [my earlier comment](https://github.com/JuliaGeometry/Meshes.jl/pull/1183#issuecomment-2803353936))

juliohm:
> Benchmarks updated (now specifically for adjsort/adjsortperm)!

🚀 

Please let us know when the PR is ready for review @halleysfifthinc 🙂

halleysfifthinc:
**tl;dr**: The face ordering with the new `connected_components`/`adjsortperm` isn't adjacently sorted under the strictest definition (subsequent faces share at least one already existing edge). Obviously, it is your call whether this is acceptable, but the updated `HalfEdgeTopology(::Vector{<:Connectivity})` function handles the current (and previous) behavior. (Although nearly all of the performance gains are from the refactored adjacency sort.)

I determined that the bug I was hunting is because adjacency is now functionally defined as any face that has >2 previously observed vertices. This allows a face (triangle 4 in the photo) to be added, even if all of its edges are new/previously unobserved. The orientation of this triangle cannot be checked against existing faces, and we can end up with irreparable inconsistencies when the "hole" (triangle 5 in the photo) is filled (if the orientation of triangle 4 is ultimately inconsistent with the correct orientation of triangle 5).


That bug is not present in the original code on `master`.

juliohm:
Thank you for the update @halleysfifthinc. Before I spend time reviewing the minor details in the diff, could you please confirm the nature of the changes in this PR?

Is it correct to say that there are

1. Changes in the `HalfEdgeTopology` constructor with vector of `Connectivity` to improve allocations and type stability
2. Changes in the `adjsortperm` to improve runtime in terms of a new `connected_components` function

and nothing else?

What about the behavior change you mentioned? Where is it coming from? You mean that the new `adjsortperm` implementation is not equivalent to the current implementation in `master`? What is the new definition of adjacency if not >2 vertices in common?

I wonder if we could split this PR into smaller PRs that are easier to review and test. The `HalfEdgeTopology` is used in various topological relations and algorithms, and we should do our best to avoid introducing new bugs.

halleysfifthinc:
> Is it correct to say that there are

Yes that is a correct summary. I'm happy to split this up for easier review if you would like.

> You mean that the new `adjsortperm` implementation is not equivalent to the current implementation in `master`?

Yes, the new `adjsortperm` does not produce identical sorts to the current implementation in `master`. This is partially by design (the current `adjsort` is not stable due to the reverse iteration order, which means that sorting a sorted `Vector{<:Connectivity}` still changes the face order).

> What is the new definition of adjacency if not >2 vertices in common?

The old definition was theoretically defined by >2 previously observed vertices, but due to the way `adjsort` was implemented, the actual functioning definition was >1 previously observed *edge* (defined by a pair of previously observed vertices). The distinction is subtle (and therefore took me a while to realize), but you can have a face that shares 2 previously observed vertices without those vertices having been previously observed as an edge/pair (e.g. triangle 4 in the above photo diagram).

I won't have the time to investigate the new test failures for a few days.

juliohm:
> I'm happy to split this up for easier review if you would like.

That would be super appreciated. Smaller PRs with all tests passing can be quickly merged and shared with all users of the package via patch releases. If you can identify subsets of changes that just improve performance, without behavior change, we can quickly review and approve.

juliohm:
@halleysfifthinc did you have a chance to split this PR into smaller ones? 🙂

juliohm:
#1193 merged 💯

halleysfifthinc:
Before I split this anymore, I want to run it past you: the only remaining logical split I see in these changes is to separate the changes to `adjsort`/`adjsortperm` and the changes to `HalfEdgeTopology(::Vector{<:Connectivity})`, which can mostly be independently separated from the changes to `adjsort`. Is this PR still larger than you'd like to review/merge (once tests are passing of course), or would you like me to split as described above?

juliohm:
The more self-contained the changes, the better. If it is not too much to ask, of course. With the split we can also add more tests for the bug you identified while working on it.

halleysfifthinc:
Once #1194 is merged, I have a rebased version of this ready to push.

Besides the architectural change of returning a sort permutation instead of a sorted copy, there are now 2 algorithmic differences which cause differences in sorted order between the PR and `master` algorithms:

1. Iterative search for the next adjacent element is done in a forward direction in this PR, which results in a more stable sort:
    - For example, each of the latter 3 elements here `connect.([(1, 2, 3), (3, 4, 2), (2, 1, 5), (6, 3, 1)])` are adjacent to the first element. So, any permutation of the latter 3 elements after the first is a valid adjacent sort, including the current order. The reverse iteration order by `adjsort` on `master` will always result in these elements being reordered (e.g. as `connect.([(1, 2, 3), (6, 3, 1), (2, 1, 5), (3, 4, 2)])` in the above example), despite already having a valid adjacent order.
2. "Adjacent" is now (technically more loosely) defined as "an element which contains more than 1 previously observed index". This differs from the stricter (but more computationally intense) definition of "an element which shares an edge (2 indices) with the current element".

(These two changes mean that there are some expected edge indices in the tests which I will need to update.)

In combination with #1194, which enables the `HalfEdgeTopology` construction to tolerate unsorted connectivities, and since `adjsort` is not public and was only used in `HalfEdgeTopology` construction, I think this slightly looser adjacency definition should be acceptable.

juliohm:
Now that #1194 is merged, I wonder if we still need `adjsort`. Would it be correct to assume that the new algorithm to convert vector of connectivity into vector of half-edges is robust to any order of elements?

Originally, I introduced `adjsort` to workaround incosistent orientation in "most cases", but now that the algorithm is refactored, `adjsort` might not be needed. If we can have an idea of average runtime and correctness with and without `adjsort`, that would be awesome.

halleysfifthinc:
> Would it be correct to assume that the new algorithm to convert vector of connectivity into vector of half-edges is robust to any order of elements?

As far as I can logically determine, that assumption is correct.

> If we can have an idea of average runtime and correctness with and without `adjsort`, that would be awesome.

Definitely. There is still a benefit to pre-sorting the elements. I compared a pre-shuffled connectivity vector to a pre-sorted connectivity vector (using `sort=false` for both cases), and I found:

```julia
# all quad torus
  time:   -18.08% => improvement (5.00% tolerance)
  memory: -6.09% => improvement (1.00% tolerance)

# quads and tris torus
  time:   -40.78% => improvement (5.00% tolerance)
  memory: -36.61% => improvement (1.00% tolerance)
```

As far as "correctness", the only differences between sorted or not will be the order of the edges. I don't have a strong opinion, but I think the possibility of multiple equally valid sort permutations (e.g. for `connect.([(1, 2, 3), (3, 4, 2), (2, 1, 5), (6, 3, 1)])`) makes a reasonable case against making a strong (or any) commitment for edge index stability.

src/topologies/halfedge.jl
```diff
@@ -332,45 +331,84 @@ Base.convert(::Type{HalfEdgeTopology}, t::Topology) = HalfEdgeTopology(collect(e
...
-        end
+# permutation of elements in adjacent-first order
+function adjsortperm(elems::AbstractVector{<:Connectivity})
+  reduce(vcat, conneccomps(elems))
+end
+
+# connected components from list of elements
+function conneccomps(elems::AbstractVector{<:Connectivity})
+  # initialize list of connected components
+  comps = [Int[firstindex(elems)]]
```
juliohm: @halleysfifthinc could we drop `Int` here or Cthulu.jl is telling you that this is important?

src/topologies/halfedge.jl
```diff
@@ -332,45 +331,84 @@ Base.convert(::Type{HalfEdgeTopology}, t::Topology) = HalfEdgeTopology(collect(e
...
+
+  # initialize list of seen vertices
+  seen = Set(indices(first(elems)))
```
juliohm: I tried to simplify the code in a few places without looking into Cthulhu.jl outputs. Please feel free to comment if you see opportunities to improve performance further.

src/topologies/halfedge.jl
```diff
@@ -332,45 +331,84 @@ Base.convert(::Type{HalfEdgeTopology}, t::Topology) = HalfEdgeTopology(collect(e
...
-            push!(adjs, popat!(list, j))
-          end
-        end
+# permutation of elements in adjacent-first order
+function adjsortperm(elems::AbstractVector{<:Connectivity})
+  reduce(vcat, conneccomps(elems))
+end
+
+# connected components from list of elements
+function conneccomps(elems::AbstractVector{<:Connectivity})
```
juliohm, line 340: I wonder if this new function for connected components is too specific for 2D manifold meshes or if it could be generalized to 3D meshes of any kind given the definition of adjacency here (>=2 shared vertices). If we believe that it should work in general, we could promote it in a separate PR with a dedicated docstring.

src/topologies/halfedge.jl
halleysfifthinc: The `Int` there could be removed if you want. I think the code is clearer/more readable with the explicit type there, instead of needing to think through the code to know what the inner vector will be.

src/topologies/halfedge.jl
halleysfifthinc: :man_shrugging: I'm not sure; I'm not yet familiar with 3D meshes.

src/topologies/halfedge.jl
```diff
@@ -331,91 +331,84 @@ Base.convert(::Type{HalfEdgeTopology}, t::Topology) = HalfEdgeTopology(collect(e
...
-  if count(in_seen, inds) > 1
-    for v in inds
-      push!(seen, v)
-    end
-    return true
-  end
-  return false
-end
+# connected components from list of elements
+function conneccomps(elems::AbstractVector{<:Connectivity})
```
halleysfifthinc, line 340: IMO, this function is not used often enough to make an abbreviated name worth the cost of being less clear.

src/topologies/halfedge.jl
```diff
@@ -331,91 +331,84 @@ Base.convert(::Type{HalfEdgeTopology}, t::Topology) = HalfEdgeTopology(collect(e
...
+      # manually union-split two most common polytopes
+      # for type stability and maximum performance
+      isadjacent = if pltype(elem) <: Triangle
+        adjelem!(seen, inds)
+      elseif pltype(elem) <: Quadrangle
+        adjelem!(seen, inds)
       else
-        _update_adjacency!(seen, indices(lelem), in_seen)
+        adjelem!(seen, inds)
       end
```
halleysfifthinc: The change here in 5dacc16f negates the manual union-splitting. `elem` is not type stable for a vector of mixed connectivity types, so any functions called on it will need runtime dispatch. The "union-splitting" makes the contents of the first two branches type-stable (and therefore no runtime dispatch), because Julia knows that e.g. in the first branch `elem isa Connectivity{Triangle,3}`, so within that branch, any use of `elem` can now be inferred, so the `indices` and `adjelem!` calls are also type stable.

Calling `pltype` on the still unstable `elem` will be type unstable. Additionally, `inds` is now unstable, since it is defined outside the union-split branches.

Without union-splitting (for the mixed quads & tris torus):
```
BenchmarkTools.Trial: 10000 samples with 1 evaluation per sample.
 Range (min … max):  194.245 μs …   7.735 ms  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     294.418 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   299.659 μs ± 107.799 μs  ┊ GC (mean ± σ):  0.39% ± 1.59%

                 ▁▂▄▅▅▆▇▇███▇█▆▅▄▃▂▁                             
  ▁▁▁▁▁▁▂▂▂▃▄▅▅▆▇████████████████████▇▆▅▅▄▃▃▃▂▂▂▂▂▂▁▁▁▁▁▁▁▁▁▁▁▁ ▄
  194 μs           Histogram: frequency by time          437 μs <

 Memory estimate: 21.63 KiB, allocs estimate: 456.
```

With (functioning) union-splitting:
```
BenchmarkTools.Trial: 10000 samples with 1 evaluation per sample.
 Range (min … max):  19.136 μs …  10.355 ms  ┊ GC (min … max): 0.00% … 99.16%
 Time  (median):     26.008 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   28.500 μs ± 110.980 μs  ┊ GC (mean ± σ):  5.00% ±  1.40%

             ▂▃▆▆██▇▇▆▄▃▂▁                                      
  ▁▁▁▂▂▂▃▄▅▇███████████████▇▆▆▅▅▄▄▅▄▄▄▄▄▃▃▃▃▃▃▂▂▂▂▂▂▂▂▁▂▂▁▁▁▁▁ ▄
  19.1 μs         Histogram: frequency by time         40.6 μs <

 Memory estimate: 14.02 KiB, allocs estimate: 212.

  time:   -90.50% => improvement (5.00% tolerance)
  memory: -35.45% => improvement (1.00% tolerance)
```

src/topologies/halfedge.jl
juliohm: We try to omit types as much as possible as part of our code style. In this context the type is pretty clear too.

juliohm:
```suggestion
  comps = [[firstindex(elems)]]
```

src/topologies/halfedge.jl
juliohm: No worries. Let's keep this utility function in halfedge.jl and then look into it again in the future if a 3D applications shows up.

juliohm:
All test changes are correct ✅ 

I will benchmark the PR against master in a few minutes.

src/topologies/halfedge.jl
halleysfifthinc: The previous manual `seen` initialization is faster vs `seen = Set(indices(first(elems)))`:
```
  time:   -17.02% => improvement (5.00% tolerance)
  memory: -0.21% => invariant (1.00% tolerance)
```

src/topologies/halfedge.jl
juliohm: It is just our code style. We avoid underscores as much as possible, even if it is an internal function. I added a comment right above to make it clear what the function does.

src/topologies/halfedge.jl
juliohm: That is super interesting. So we can't use `pltype` nor `indices` outside the branches because `elem` itself is not a concrete type. I will refactor the code once again, thanks for sharing.

Learning a lot about low-level compiler limitations in these last PRs.

halleysfifthinc: (I already pushed fixes for this.)

src/topologies/halfedge.jl
```diff
@@ -332,45 +331,87 @@ Base.convert(::Type{HalfEdgeTopology}, t::Topology) = HalfEdgeTopology(collect(e
...
+# connected components from list of elements
+function conneccomps(elems::AbstractVector{<:Connectivity})
+  # initialize list of connected components
+  comps = [[firstindex(elems)]]
+
+  # initialize list of seen vertices
+  seen = Set{Int}()
+  for v in indices(first(elems))
+    push!(seen, v)
+  end
```
juliohm, line 348: Is this explicit loop needed or the `Set(indices(first(elems))` would be equivalent compiler-wise? (I didn't have time to dive into Cthulhu.jl docs yet).

src/topologies/halfedge.jl
```diff
@@ -332,45 +331,87 @@ Base.convert(::Type{HalfEdgeTopology}, t::Topology) = HalfEdgeTopology(collect(e
...
+  while !isempty(remaining)
+    iter = 1
+    while iter ≤ length(remaining)
+      elem = elems[remaining[iter]]
+
+      # manually union-split two most common polytopes
+      # for type stability and maximum performance
+      isadjacent = if elem isa Connectivity{Triangle,3}
+        adjelem!(seen, elem)
+      elseif elem isa Connectivity{Quadrangle,4}
```
juliohm, line 363:
```suggestion
      isadjacent = if elem isa Connectivity{Triangle}
        adjelem!(seen, elem)
      elseif elem isa Connectivity{Quadrangle}
```

Can we omit the second type parameter?

src/topologies/halfedge.jl
halleysfifthinc: That has nothing to do with Cthulhu. The explicit loop is 17% faster; c.f. the benchmark I shared in the comment.

src/topologies/halfedge.jl
halleysfifthinc: Benchmarks indicate they are very much necessary. I expect it is related to the fact that `Connectivity{Triangle}` is a `UnionAll` type (incompletely specified).

src/topologies/halfedge.jl
juliohm: Got it. It is a bit unexpected to see that the constructor is 17% slower in this simple setup.

juliohm (approved):
Another amazing speedup 💯 🚀

juliohm:
Patch released: https://github.com/JuliaRegistries/General/pull/131456

PR #1194, opened 2025-05-13
Improve `HalfEdgeTopology(::Vector{<:Connectivity})` correctness

halleysfifthinc:
This PR reveals/fixes a couple of bugs in the `HalfEdgeTopology` constructor.

1. `Dict`'s are unordered, so constructing the `halves::Vector{Tuple{HalfEdge,HalfEdge}}` by iterating through `half4pair` pairs produces a roughly random order (by element) for halves. https://github.com/JuliaGeometry/Meshes.jl/blob/3a3d0e3a41e4b8b224061fad232ce464b07bc4e3/src/topologies/halfedge.jl#L220 This has downstream affects on e.g. edge indices and the starting vertex index from `loop`. (Cause of most/all? test failures.)
2. Incorrect orientation consistency handling in the presence of incorrectly- or unsorted elements. A couple of things to note:
    - Treating incorrectly- or unsorted elements as a "no warranty" situation is perfectly fine, but the current/`master` state of the function can lead to non-closed edge loops, such that the `loop` function never terminates and allocates memory until Julia is OOM killed[^1].
      - ```julia
        # without this redefinition, `loop` will eat all your memory
        function Meshes.loop(e::HalfEdge)
        n = e.next
        v = [e.head]
        cnt = 1
        while n != e
            if cnt < 3
                push!(v, n.head)
                n = n.next
                cnt += 1
            else
                throw(error("too many edges!"))
            end
        end
        v
        end

        e = connect.([(1,2,3), (1,3,4), (2,5,3), (5,4,6), (3,5,4)], Triangle)
        t = HalfEdgeTopology(e; sort=false)
        Meshes.loop(half4elem(t, 1)) # works
        Meshes.loop(half4elem(t, 2)) # errors
        ```
    - Minor nit: The orientation consistency variable name (`CCW`) was misleading, as actual element orientation isn't checked (and cannot be without actual vertices afaict?). Also, when there are multiple unconnected components (which I believe is still valid under a 2-manifold?) the orientation consistency checks don't/can't cross the disconnection, leading to further potential mismatches between the name `CCW` and reality. I renamed it to make this clearer.

The cause of most of the current test failures are due to changing to the `halves` construction to be element ordered. I believe this is more correct, and that any stable/consistent edge indices with the previous version are because the hashing algorithm in Base has not been changed between Julia versions (that will no longer be true after 1.13 with JuliaLang/julia#57509).

Another cause of test failures is a change to the reverse index iteration, which now ensures that the first index returned by `loop` (which affects several downstream dependent functions) is either:
  1. The original first index of the element (regardless of whether the original element had an inconsistent orientation)
  2. The (head) index of the first observed edge of the element

With regard to sorting and orientation consistency, the PR is correct (no infinite memory use) even for incorrectly- or unsorted elements when called with `sort=false`, but the edge indices will/may change between `sort=false` and `sort=true`.

If you agree with this PR's general direction. I will update the tests and share benchmarks.

[^1]: In my code, I am redefining `loop` similar to the given example, but using a ScopedValue so that the maximum number of edges can be changed from outside the function if needed. I'm happy to share/submit that if you are interested, but I realize that such a solution might be too limiting generally to be added to the library.


test/topologies.jl
```diff
@@ -372,8 +372,17 @@ end
...
+      while i ≤ length(inds)
+        @test he.elem == e
+        @test he.head ∈ inds
+        @test he.next.elem == e
+        @test he.prev.elem == e
+        @test he.next.prev == he
+        @test he.prev.next == he
+        he = he.next
+        i += 1
+      end
```
juliohm:
```suggestion
      for _ in inds
        @test he.elem == e
        @test he.head ∈ inds
        @test he.next.elem == e
        @test he.prev.elem == e
        @test he.next.prev == he
        @test he.prev.next == he
        he = he.next
      end
```

test/topologies.jl
```diff
@@ -492,9 +502,14 @@ end
   # more challenging case with inconsistent orientation
   e = connect.([(4, 1, 5), (2, 6, 4), (3, 5, 6), (4, 5, 6)])
   t = HalfEdgeTopology(e)
+  test_halfedge(e, t)
   n = collect(elements(t))
   @test n == connect.([(5, 4, 1), (6, 2, 4), (6, 5, 3), (4, 5, 6)])
 
+  e = connect.([(1, 2, 3), (1, 3, 4), (2, 5, 3), (5, 4, 6), (3, 5, 4)], Triangle)
```
juliohm:
```suggestion
  e = connect.([(1, 2, 3), (1, 3, 4), (2, 5, 3), (5, 4, 6), (3, 5, 4)])
```

src/topologies/halfedge.jl
```diff
@@ -139,89 +139,138 @@ function HalfEdgeTopology(halves::AbstractVector{Tuple{HalfEdge,HalfEdge}}; nele
...
-  eleminds = sort ? indexin(adjelems, elems) : 1:length(elems)
+  _elems = sort ? adjsort(elems) : elems
+  eleminds = sort ? indexin(_elems, elems) : 1:length(elems)
+  adjelems::Vector{Vector{Int}} = map(collect ∘ indices, _elems)
 
   # start assuming that all elements are
-  # oriented consistently as CCW
-  CCW = trues(length(adjelems))
+  # oriented consistently
+  REV_DIR = falses(length(adjelems))
```
juliohm: Do you mean `REV_DIR` as in the verb "reverse direction"? Let's rename to `isreversed`, I think it is clearer.

src/topologies/halfedge.jl
```diff
@@ -139,89 +139,138 @@ function HalfEdgeTopology(halves::AbstractVector{Tuple{HalfEdge,HalfEdge}}; nele
...
+  for i in eachindex(inds)
+    uv = (inds[i], inds[mod1(i + 1, n)])
+    if haskey(half4pair, uv)
+      return true
+    end
+  end
+  return false
+end
+
+const NULL_EDGE = HalfEdge(0, nothing)
```
juliohm: Please avoid _ in variable and function names.

juliohm: Since this null edge is only used inside the helper function, it can be defined therein.

src/topologies/halfedge.jl
```diff
@@ -139,89 +139,138 @@ function HalfEdgeTopology(halves::AbstractVector{Tuple{HalfEdge,HalfEdge}}; nele
...
+function any_claimed_edges_exist(inds, half4pair)
+  n = length(inds)
+  for i in eachindex(inds)
+    uv = (inds[i], inds[mod1(i + 1, n)])
+    if !isnothing(get(half4pair, uv, NULL_EDGE).elem)
+      return true
```
juliohm: Let's avoid this auxiliary NULL_EDGE and make the logic more explicit. We can check the key in the `half4pair` and then do something else if it is not present.

src/topologies/halfedge.jl
```diff
@@ -139,89 +139,138 @@ function HalfEdgeTopology(halves::AbstractVector{Tuple{HalfEdge,HalfEdge}}; nele
...
 function HalfEdgeTopology(elems::AbstractVector{<:Connectivity}; sort=true)
   assertion(all(e -> paramdim(e) == 2, elems), "invalid element for half-edge topology")
 
   # sort elements to make sure that they
   # are traversed in adjacent-first order
-  adjelems = sort ? adjsort(elems) : elems
-  eleminds = sort ? indexin(adjelems, elems) : 1:length(elems)
+  _elems = sort ? adjsort(elems) : elems
+  eleminds = sort ? indexin(_elems, elems) : 1:length(elems)
+  adjelems::Vector{Vector{Int}} = map(collect ∘ indices, _elems)
```
juliohm:
```suggestion
  elemsort = sort ? adjsort(elems) : elems
  eleminds = sort ? indexin(elemsort, elems) : 1:length(elems)
  adjelems::Vector{Vector{Int}} = map(collect ∘ indices, elemsort)
```

src/topologies/halfedge.jl
halleysfifthinc: This way (using `get` with a default of `NULL_EDGE`) is faster because it avoids double indexing the Dict (compared to e.g. `haskey(half4pair, uv) && !isnothing(half4pair[uv].elem)`). The const externally defined `NULL_EDGE` is helpful because `HalfEdge` is mutable, which means defining a `HalfEdge` allocates. Defining `NULL_EDGE` in the function would incur that definition/allocation cost every time the function is called.

test/toporelations.jl
```diff
@@ -445,30 +445,30 @@ end
   elems = connect.([(1, 2, 3), (4, 3, 2)])
   t = HalfEdgeTopology(elems)
   ∂ = Boundary{2,0}(t)
-  @test ∂(1) == (2, 3, 1)
+  @test ∂(1) == (1, 2, 3)
```
halleysfifthinc, line 448: Element-ordered `halves` results in the head of the first edge matching the original first index of the element.

test/toporelations.jl
```diff
@@ -445,30 +445,30 @@ end
...
   ∂ = Boundary{1,0}(t)
-  @test ∂(1) == (3, 2)
-  @test ∂(2) == (1, 2)
+  @test ∂(1) == (1, 2)
+  @test ∂(2) == (2, 3)
   @test ∂(3) == (3, 1)
-  @test ∂(4) == (2, 4)
-  @test ∂(5) == (4, 3)
+  @test ∂(4) == (4, 3)
+  @test ∂(5) == (2, 4)
```
halleysfifthinc, line 458: Edge indices changed due to element-ordered `halves`

test/toporelations.jl
```diff
@@ -445,30 +445,30 @@ end
...
   𝒞 = Coboundary{0,1}(t)
-  @test 𝒞(1) == (2, 3)
-  @test 𝒞(2) == (4, 1, 2)
-  @test 𝒞(3) == (3, 1, 5)
-  @test 𝒞(4) == (5, 4)
+  @test 𝒞(1) == (1, 3)
+  @test 𝒞(2) == (5, 2, 1)
+  @test 𝒞(3) == (3, 2, 4)
+  @test 𝒞(4) == (4, 5)
```
halleysfifthinc, line 463: Due to different edge indices

test/toporelations.jl
```diff
@@ -445,30 +445,30 @@ end
...
   𝒞 = Coboundary{0,2}(t)
   @test 𝒞(1) == (1,)
   @test 𝒞(2) == (2, 1)
   @test 𝒞(3) == (1, 2)
   @test 𝒞(4) == (2,)
   𝒞 = Coboundary{1,2}(t)
-  @test 𝒞(1) == (2, 1)
-  @test 𝒞(2) == (1,)
+  @test 𝒞(1) == (1,)
+  @test 𝒞(2) == (1,2)
```
halleysfifthinc: Due to different edge indices

test/toporelations.jl
```diff
@@ -539,17 +539,17 @@ end
...
   # 4 quadrangles in a grid
   elems = connect.([(1, 2, 5, 4), (2, 3, 6, 5), (4, 5, 8, 7), (5, 6, 9, 8)])
   t = HalfEdgeTopology(elems)
   𝒜 = Adjacency{2}(t)
-  @test 𝒜(1) == (3, 2)
+  @test 𝒜(1) == (2, 3)
   @test 𝒜(2) == (1, 4)
   @test 𝒜(3) == (1, 4)
-  @test 𝒜(4) == (3, 2)
+  @test 𝒜(4) == (2, 3)
```
halleysfifthinc, line 552: Due to element ordered `halves`

test/toporelations.jl
```diff
@@ -539,17 +539,17 @@ end
   𝒜 = Adjacency{2}(t)
   @test 𝒜(1) == (2, 3, 4)
   @test 𝒜(2) == (1, 3)
-  @test 𝒜(3) == (2, 4, 1)
+  @test 𝒜(3) == (1, 2, 4)
```
halleysfifthinc, line 542: Due to element ordered `halves`

test/topologies.jl
```diff
@@ -492,14 +500,19 @@ end
...
+  test_halfedge(e, t)
 
   # indexable api
   g = GridTopology(10, 10)
   t = convert(HalfEdgeTopology, g)
-  @test t[begin] == connect((13, 12, 1, 2), Quadrangle)
-  @test t[end] == connect((110, 121, 120, 109), Quadrangle)
+  @test t[begin] == g[begin]
+  @test t[end] == connect((120, 109, 110, 121), Quadrangle)
   @test t[10] == connect((22, 21, 10, 11), Quadrangle)
```
halleysfifthinc, line 516: The only difference between the GridTopology and HalfEdgeTopology is the initial index, I'm not familiar with GridTopology to know if/what could be done to unify that, but the initial indices didn't match previously (at least for `end` and `10`).

halleysfifthinc:
Benchmarking procedure same as previously described, using `sort=false` to limit extraneous testing time/allocations:

Quads only:
```
  time:   -66.93% => improvement (5.00% tolerance)
  memory: -46.14% => improvement (1.00% tolerance)
```

Quads and tris:
```
  time:   -66.81% => improvement (5.00% tolerance)
  memory: -43.88% => improvement (1.00% tolerance)
```

juliohm:
Thank you for the updates @halleysfifthinc. I am planning to review a couple of PRs in CoordRefSystems.jl, Meshes.jl and GeoIO.jl this week.

juliohm:
@halleysfifthinc the tests were failing locally for me, falling in the `assertion` you introduced. I did some cleanup to facilitate the review moving forward.

src/topologies/halfedge.jl
```diff
@@ -144,126 +144,116 @@ function HalfEdgeTopology(elems::AbstractVector{<:Connectivity}; sort=true)
...
-    half4pair[(v[i], v[i + 1])] = HalfEdge(v[i], eleminds[1])
+  elem = first(eleminds)
+  inds = first(adjelems)
+  n = length(inds)
+  for i in eachindex(inds)
+    u = inds[i]
+    v = inds[mod1(i + 1, n)]
+    # insert halfedges u -> v and v -> u
+    he = get!(() -> HalfEdge(u, elem), half4pair, (u, v))
+    half = get!(() -> HalfEdge(v, nothing), half4pair, (v, u))
```
juliohm, line 165: Is there a reason why you prefer `get!` with this anonymous function instead of simply using `half4pair[(u,v)] = HalfEdge(u, elem)` for example?

src/topologies/halfedge.jl
```diff
@@ -144,126 +144,116 @@ function HalfEdgeTopology(elems::AbstractVector{<:Connectivity}; sort=true)
...
+        # at least one half-edge has been found, so we can assess
+        # the orientation w.r.t. previously added elements/edges
+        added = true
+        disconnected = false
+        deleteat!(remaining, iter)
       else
-        # insert pair in consistent orientation
-        half4pair[(v[i], v[i + 1])] = HalfEdge(v[i], eleminds[e])
+        iter += 1
+        continue
```
juliohm, line 189: Could you please confirm this `continue` is only skipping the rest of the inner `while` loop?

src/topologies/halfedge.jl
halleysfifthinc: Directly `setindex`-ing will silently overwrite previously defined keys; I believe some of the issues I was encountering were due to that. The alternative to avoid overwriting is an additional `haskey` check, which does incur a computational cost (albeit, somewhat minor).

An different use of `get!` is like so `get!(half4pair, (u, v), HalfEdge(u, elem))`, which looks slightly cleaner, but also has a performance penalty, because the default element is created even if not used, unlike with the anonymous function, which is only called if needed.

src/topologies/halfedge.jl
halleysfifthinc: That's correct. `break` and `continue` only affect the whichever loop is most local to the `break`/`continue` call.

src/topologies/halfedge.jl
```diff
@@ -338,3 +328,68 @@ nfacets(t::HalfEdgeTopology) = length(t.halfedges) ÷ 2
...
+    uv = (inds[i], inds[mod1(i + 1, n)])
+    if haskey(half4pair, uv)
+      return true
+    end
+  end
+  return false
+end
+
+function anyhalfclaimed(inds, half4pair)
+  ∅ = HalfEdge(0, nothing)
```
halleysfifthinc: This can't be defined internally; c.f. the bug I discovered in base Julia >= 1.11, discussed here https://discourse.julialang.org/t/unmutated-mutable-type-being-corrupted/129161 and at JuliaLang/julia#58470

juliohm: Wow! I am starting to feel concerned about Julia >= v1.11. Can we rewrite this function to avoid the `∅` variable? Let's be more explicit with an additional `if` statement to avoid the bug in present and future versions of the language.

halleysfifthinc:
> the tests were failing locally for me

Before or after the readability changes in [dab47eb](https://github.com/JuliaGeometry/Meshes.jl/pull/1194/commits/dab47eb6ccd5ffd506b7ebff826e10081596ff0e)? Tests are consistently passing for me before that commit, and also consistently pass after I sidestep the base Julia bug. :crossed_fingers:  nothing turns up in this CI run

juliohm:
> Before or after the readability changes in [dab47eb](https://github.com/JuliaGeometry/Meshes.jl/pull/1194/commits/dab47eb6ccd5ffd506b7ebff826e10081596ff0e)?

Almost sure it was before the changes. I did run the tests locally, then made the changes.

src/topologies/halfedge.jl
halleysfifthinc: The change I pushed in [7f0dab5](https://github.com/JuliaGeometry/Meshes.jl/pull/1194/commits/7f0dab5bce4a1c65ccfe7178e358484725193537) successfully avoids the issue from what I can tell/test.

halleysfifthinc:
Hmmm. Yesterday, I did see a test failure once before the readability commit, but I'm not able to replicate that today. I wonder if it was user error on my part (e.g. I wasn't actually on the right commit, etc).

I have been running one of the previously failing tests in an infinite loop for >20mins now and have not triggered any more errors. (I also have been running with these patches [with the `const NULLEDGE` version] for several days without experiencing any issues either.)

juliohm:
I confirm that all tests pass locally 💯

Will review the test changes one by one now, before assessing the performance improvements 👍🏽

src/topologies/halfedge.jl
```diff
@@ -144,126 +144,119 @@ function HalfEdgeTopology(elems::AbstractVector{<:Connectivity}; sort=true)
...
+        for i in eachindex(inds)
+          u = inds[i]
+          v = inds[mod1(i + 1, n)]
+          he = get!(() -> HalfEdge(u, elem), half4pair, (u, v))
+          he.elem = elem # this may be a pre-allocated half-edge with a nothing `elem`
+          half = get!(() -> HalfEdge(v, nothing), half4pair, (v, u))
+          he.half = half
+          half.half = he
+        end
       end
```
juliohm: In the meantime, I wonder if we could collapse these two branches into a single one where `inds` are looped in reverse order when `reversed[other]` is true. I see that the code is pretty much the same, except that the first branch has an additional `isnothing` check and an `assertion`. If you feel that we can't simplify this further, that is fine too. Just trying to keep things as simple as possible for future maintenance.

test/topologies.jl
```diff
@@ -492,14 +500,19 @@ end
...
+  e = connect.([(1, 2, 3), (1, 3, 4), (2, 5, 3), (5, 4, 6), (3, 5, 4)])
+  t = HalfEdgeTopology(e)
+  test_halfedge(e, t)
 
   # indexable api
   g = GridTopology(10, 10)
   t = convert(HalfEdgeTopology, g)
-  @test t[begin] == connect((13, 12, 1, 2), Quadrangle)
-  @test t[end] == connect((110, 121, 120, 109), Quadrangle)
+  @test t[begin] == g[begin]
```
juliohm:
```suggestion
  @test t[begin] == connect((1, 2, 13, 12), Quadrangle)
```

halleysfifthinc: My thought was that directly comparing to `g` would be more in line with testing that the edges indices in GridTopology match that of the converted HalfEdgeTopology. Currently the first element is the only one with matching edges/indices. The other two (10 and the last element) do not match the initial index of the corresponding elements in `g`. I don't have an opinion or preference, but you had mentioned that matching GridTopology was a goal.

juliohm:
All test changes are perfect 💯 I will benchmark the PR against the master branch now...

test/topologies.jl
juliohm: I noticed that even before the PR we didn't have guarantees to match the order with GridTopology. Since the resulting order is arbitrary, I think it makes more sense to hard-code it in the tests. That way we can quickly sketch the grid on paper and debug based on the test values.

juliohm:
I confirm a speedup of approximately 2x when `sort=false` 🚀

juliohm (approved):

juliohm:
@JoshuaLampert do you have any additional comment before we merge this? I know it is a very tricky algorithm to follow, so don't feel any pressure to review in detail. Any bird's-eye review would be nice given the sensitivity of these changes 🙂

src/topologies/halfedge.jl
```diff
@@ -144,126 +144,115 @@ function HalfEdgeTopology(elems::AbstractVector{<:Connectivity}; sort=true)
...
-function adjsort(elems::AbstractVector{<:Connectivity})
-  # initialize list of adjacent elements
-  # with first element from original list
-  list = indices.(elems)
-  adjs = Tuple[popfirst!(list)]
+  # add missing pointers and save halfedges in a vector of pairs
+  halves = Vector{Tuple{HalfEdge,HalfEdge}}()
+  visited = Set{Tuple{Int,Int}}()
+  for (e, inds) in enumerate(adjelems)
+    inds = isreversed[e] ? circshift!(reverse!(inds), 1) : inds
```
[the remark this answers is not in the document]
juliohm: @halleysfifthinc worth checking this comment by Copilot. I believe there is no issue in this case because the `adjelems` are looped in a single pass from beginning to the end.

juliohm:
The test that is failing consistently in MacOS is unrelated to the changes in this PR. We need to increase the threshold in a separate PR later.

src/topologies/halfedge.jl
```diff
@@ -144,126 +144,115 @@ function HalfEdgeTopology(elems::AbstractVector{<:Connectivity}; sort=true)
 
   # sort elements to make sure that they
   # are traversed in adjacent-first order
-  adjelems = sort ? adjsort(elems) : elems
-  eleminds = sort ? indexin(adjelems, elems) : 1:length(elems)
+  elemsort = sort ? adjsort(elems) : elems
+  eleminds = sort ? indexin(elemsort, elems) : 1:length(elems)
+  adjelems::Vector{Vector{Int}} = map(collect ∘ indices, elemsort)
```
JoshuaLampert, line 149: This type annotation improves performance? Usually I would like to avoid these type annotations if not necessary. I read in of the recent PRs here that this improved performance, but I just want to double-check.

src/topologies/halfedge.jl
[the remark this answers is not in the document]
halleysfifthinc: Correct. The mutation was intentional to avoid the copy and doesn't affect anything because that is the last time we use `inds`.

juliohm:
> I was wondering whether this counts as breaking change because, IIUC, it changes the ordering. Or was the ordering nothing to rely on before anyway and therefore this change can be done in a non-breaking way?

The ordering of vertices in `HalfEdgeTopology` is arbitrary and hidden from end users through our topological relations `Adjacency`, `Boundary` and `CoBoundary`. Our idea with the current design is that users can write algorithms in terms of topological relations, which are agnostic to the topological data structure underneath.

Originally I thought of releasing this as a patch release to upstream the performance improvements throughout the ecosystem, but if you guys feel that this has the potential to break code, we can go with a minor release.

src/topologies/halfedge.jl
juliohm: I share this feeling. From my local tests the type annotation leads to a 3x speedup, even though the output of `@code_warntype` is the same. @halleysfifthinc is using Cthulu.jl to optimize the code, and I think I will start using it too instead of simple performance macros from Base.

juliohm:
Thank you all for the valuable contributions! Merging another important fix and speedup 🚀 👏🏽

src/topologies/halfedge.jl
halleysfifthinc: Yes, this type annotation is narrower than what the compiler/type-system can currently determine and allows the rest of the function to specialize more appropriately.

JoshuaLampert: If Cthulhu.jl reports type inference issues without the annotation, I agree it makes sense to include it. Thanks for the clarification.👍

halleysfifthinc:
> if you guys feel that this has the potential to break code, we can go with a minor release

I would be consider this a non-breaking bugfix, as the previous indices were indisputably (IMO) incorrect (albeit, consistently so). Any code relying on the previous orderings is potentially also incorrect, and therefore implicitly already broken.

juliohm:
Very good point. This PR also fixes a bug so it would be a good idea to
propagate the fix.

Em ter., 20 de mai. de 2025, 17:57, Allen Hill ***@***.***>
escreveu:

JoshuaLampert:
It's fine with me to not consider it as breaking.

juliohm:
Patch released: https://github.com/JuliaRegistries/General/pull/131425

PR #1198, opened 2025-05-28
Improve `intersects(::Ray, ::Triangle)` type-stability and performance

halleysfifthinc:
- The existing `intersects(a, b) = !isnothing(a ∩ b)` methods needlessly introduce
    type-instability in the intermediate intersection value, when a simpler (and still
    valid, afaict) approach is to check if the (type-stable) enum value in the
    `Intersection` is not `NotIntersecting`
- A closure was capturing and `Box`ing a variable, leading to type-instability and a bunch
    of allocations
- `ScopedValue` accesses are fast, but they do take time (~7 ns on my machine, measured by
    benchmarking `atol(1u"m")`). Saving one `ScopedValue` access per `intersect(::Ray,
    ::Triangle)` call adds up when testing for intersections across a (large) mesh.

### DIFF
--- a/src/predicates/intersects.jl
+++ b/src/predicates/intersects.jl
@@ -33,7 +33,7 @@ intersects(r::Ray, b::Box) = someintersection(r, b)
 
 intersects(b::Box, r::Ray) = intersects(r, b)
 
-intersects(r::Ray, t::Triangle) = type(intersection(r, t)) !== NotIntersecting
+intersects(r::Ray, t::Triangle) = someintersection(r, t)
 
 intersects(t::Triangle, r::Ray) = intersects(r, t)
 


halleysfifthinc:
It seems like there should be a way to hoist the `ScopedValue` accesses to once in the function, because they (the `ATOL(32|64)`) values can't be changed (externally) during the function call.

src/intersections/rays.jl
```diff
@@ -153,14 +153,15 @@ function intersection(f, ray::Ray, tri::Triangle)
   det = e₁ ⋅ p
 
   # keep det > 0, modify T accordingly
-  if det > atol(det)
+  det_atol = atol(det)
```
juliohm:
```suggestion
  atoldet = atol(det)
```
We don't use underscore.

juliohm (changes requested):
Small changes with great potential 💯 

Do you have benchmark results with the current version of the PR?

juliohm:
I can take a second look early morning tomorrow (BRT time zone).

halleysfifthinc:
Benchmarks! Iterating through every triangle (but 1) on a ~750k element mesh:

Benchmark call: `@benchmark count(Base.Fix1(intersects, Ray(centroid(tri), normal(tri))), view(msh, filter(!=(j), 1:nfaces(msh, 2)))) setup=(j=rand(1:nfaces(msh, 2)); tri=msh[j])` (Somewhat indirect, but demonstrates gains more clearly/dramatically than single `intersect` benchmarks). 

On `master`:
```julia
BenchmarkTools.Trial: 32 samples with 1 evaluation per sample.
 Range (min … max):  146.081 ms … 283.795 ms  ┊ GC (min … max): 4.31% … 46.45%
 Time  (median):     151.798 ms               ┊ GC (median):    5.43%
 Time  (mean ± σ):   160.771 ms ±  25.347 ms  ┊ GC (mean ± σ):  7.71% ±  7.91%

  █▁▃                                                            
  ███▇▆▁▄▆▁▄▁▁▆▁▁▄▁▄▁▄▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▄ ▁
  146 ms           Histogram: frequency by time          284 ms <

 Memory estimate: 71.62 MiB, allocs estimate: 3912436.
```

PR:
```julia
BenchmarkTools.Trial: 111 samples with 1 evaluation per sample.
 Range (min … max):  42.019 ms … 51.613 ms  ┊ GC (min … max): 0.00% … 14.51%
 Time  (median):     44.086 ms              ┊ GC (median):    0.00%
 Time  (mean ± σ):   45.113 ms ±  2.207 ms  ┊ GC (mean ± σ):  1.47% ±  2.18%

             █▂       ▂                                        
  ▃▃▄▁▁▃▁▁▅▅▆██▃▃▁▆▁▁▆█▄▃▃▃▁▁▃▁▁▁▁▁▁▁▁▃▄▄▃▃▁▃▁▁▁▁▃▄▃▆▁▁▃▁▁▁▁▃ ▃
  42 ms           Histogram: frequency by time        50.6 ms <

 Memory estimate: 11.92 MiB, allocs estimate: 23.
```

juliohm:
I've introduced the `someintersection` helper function to improve readability and future maintenance.

Couldn't reproduce the benchmark results because I don't have access to the `msh` used.

PR #1199, opened 2025-06-03
Make `elem` loop local to avoid Box

halleysfifthinc:
Because `elem` is previously defined in the main function body, redefining it in the `while`
loops and using it from a capture causes it to be `Box`ed with the infamous
performance/allocation cost to `Box` encounters.

  time:   -8.34% => improvement (5.00% tolerance)
  memory: -23.19% => improvement (1.00% tolerance)


juliohm:
@halleysfifthinc I made the changes to avoid the same name inside the `while` loop. Could you please confirm they address the issue?

halleysfifthinc:
Confirmed.

juliohm:
Patch released: https://github.com/JuliaRegistries/General/pull/132306

Issue #1205, opened 2025-06-10
Implement `isconvex` for `Hexahedron`


juliohm:
The method is missing and can be useful in some algorithms.

raman-maker:
Yes, I need this feature in this post. It can help me get Hexahedron Volume plot.  https://discourse.julialang.org/t/hexahedron-volume-plot/131489/8

PR #1206, opened 2025-06-10
RegularDiscretization of Triangle and Tetrahedron

juliohm:
This PR adds methods for `RegularDiscretization` of `Triangle` and `Tetrahedron`. The parametrization of simplices of 2 or more dimensions is not orthogonal so we can't rely on `RegularSampling`.

I didn't add a test for `Tetrahedron` due to a missing `isconvex` method reported in #1205.

### DIFF
--- a/src/discretization/regular.jl
+++ b/src/discretization/regular.jl
@@ -27,8 +27,10 @@ function discretize(geometry::Geometry, method::RegularDiscretization)
   end
 end
 
+# triangle is parametrized with barycentric coordinates, so we can't rely on regular sampling
 discretize(triangle::Triangle, method::RegularDiscretization) = discretizewithinbox(triangle, method)
 
+# tetrahedron is parametrized with barycentric coordinates, so we can't rely on regular sampling
 discretize(tetrahedron::Tetrahedron, method::RegularDiscretization) = discretizewithinbox(tetrahedron, method)
 
 function discretizewithinbox(geometry::Geometry, method::RegularDiscretization)


src/discretization/regular.jl
```diff
@@ -23,12 +23,20 @@ function discretize(geometry::Geometry, method::RegularDiscretization)
...
     SimpleMesh(collect(verts), tmesh)
   else
-    box = boundingbox(geometry)
-    grid = discretize(box, method)
-    view(grid, geometry)
+    discretizewithinbox(geometry, method)
   end
 end
 
+discretize(triangle::Triangle, method::RegularDiscretization) = discretizewithinbox(triangle, method)
```
JoshuaLampert, line 31: Maybe add a comment why we need to overload this method for `Triangle` and `Tetrahedron`? IIUC, it's because the parametric functions for them are defined for barycentric coordinates in contrast to $[0, 1]^n$. I needed a bit to realize that.

JoshuaLampert (commented):
LGTM. I only left one minor suggestion below.

src/discretization/regular.jl
juliohm: Thank you for the suggestion. I will add this comment 👍🏽

JoshuaLampert (approved):
Thanks!
Just FYI (probably you are already aware of it): Something similar probably has to be done for `RegularSampling`. I just tried on master

```julia
julia> using Meshes

julia> tri = Triangle(Point(0, 0), Point(1, 0), Point(0, 1))
Triangle
├─ Point(x: 0.0 m, y: 0.0 m)
├─ Point(x: 1.0 m, y: 0.0 m)
└─ Point(x: 0.0 m, y: 1.0 m)

julia> p = sample(tri, RegularSampling(10))
Base.Iterators.Flatten{Base.Generator{Tuple{Base.Generator{Base.Iterators.ProductIterator{Tuple{StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}}, Meshes.var"#459#463"{Triangle{𝔼{2}, CoordRefSystems.Cartesian2D{CoordRefSystems.NoDatum, Unitful.Quantity{Float64, 𝐋, Unitful.FreeUnits{(m,), 𝐋, nothing}}}}}}, Base.Generator{Tuple{}, typeof(identity)}}, typeof(identity)}}(Base.Generator{Tuple{Base.Generator{Base.Iterators.ProductIterator{Tuple{StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}}, Meshes.var"#459#463"{Triangle{𝔼{2}, CoordRefSystems.Cartesian2D{CoordRefSystems.NoDatum, Unitful.Quantity{Float64, 𝐋, Unitful.FreeUnits{(m,), 𝐋, nothing}}}}}}, Base.Generator{Tuple{}, typeof(identity)}}, typeof(identity)}(identity, (Base.Generator{Base.Iterators.ProductIterator{Tuple{StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}}, Meshes.var"#459#463"{Triangle{𝔼{2}, CoordRefSystems.Cartesian2D{CoordRefSystems.NoDatum, Unitful.Quantity{Float64, 𝐋, Unitful.FreeUnits{(m,), 𝐋, nothing}}}}}}(Meshes.var"#459#463"{Triangle{𝔼{2}, CoordRefSystems.Cartesian2D{CoordRefSystems.NoDatum, Unitful.Quantity{Float64, 𝐋, Unitful.FreeUnits{(m,), 𝐋, nothing}}}}}(Triangle((x: 0.0 m, y: 0.0 m), (x: 1.0 m, y: 0.0 m), (x: 0.0 m, y: 1.0 m))), Base.Iterators.ProductIterator{Tuple{StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}}((0.0:0.1111111111111111:1.0, 0.0:0.1111111111111111:1.0))), Base.Generator{Tuple{}, typeof(identity)}(identity, ()))))

julia> collect(p)
ERROR: DomainError with (1.0, 0.1111111111111111):
invalid barycentric coordinates for triangle.
```

So, this looks like a similar issue, which would be good to resolve.

juliohm:
I think the `RegularSampling` method is not well-defined with simplices that have more than 1 parametric dimension due to the non-orthogonality of the barycentric coordinates. If we sample the first coordinate in a triangle, the second is defined by the sum-to-one constraint. The points wouldn't be distributed regularly inside the triangle, at least with our current definition of `RegularSampling`.

JoshuaLampert:
Ah, I would have expected that `RegularSampling` would simply return (an iterator of) the points of `RegularSampling` in the bounding box that lie in the triangle.

juliohm:
That is a good point. Maybe we should consider this more general definition!

JoshuaLampert:
Sounds good.

juliohm:
If we allow `RegularDiscretization` then it is indeed strange to not allow `RegularSampling`. I will open an issue to track this.

JoshuaLampert:
Alternatively, one could think of providing a translation from $[0,1]^2$ to the barycentric coordinates like we do in MeshIntegrals.jl: https://github.com/JuliaGeometry/MeshIntegrals.jl/blob/main/src/specializations/Triangle.jl#L30. But I don't know how much sense this makes for sampling to be honest. But it would also make things easier if every parametrized geometry would be parametrized by $[0,1]^n$ and not just almost every. See also https://github.com/JuliaGeometry/Meshes.jl/pull/1093#issuecomment-2408452647.

juliohm:
Yes, that inconsistency bothered me too in the past, but most algorithms I know that use triangles and tetrahedra end up requiring barycentric coordinates. The mapping of the box coordinates to the barycentric coordinates (and vice-versa) could be useful in other contexts besides numerical integration. The other examples with unbounded parametric coordinates like `Line` and `Plane` would also require some thought.

PR #1214, opened 2025-06-14
Improve convex hull implementations

### LINKED ISSUE #1211: Bug in Graham scan
The [attached gist](https://gist.github.com/juliohm/d5dfce90642efdb4d512362a0d12018f) contains a set of points for which the `GrahamScan` method fails:

```julia
using Meshes
using CSV

csv = CSV.File("grahamscan.csv")

pts = Point.(csv.x, csv.y)

chull = hull(pts, GrahamScan())

viz(chull)
```

Image

juliohm:
Fix #1211

Besides fixing the bug, this PR also removes the usage of `to` in the hull algorithms, generalizing them to other CRS besides Cartesian2D. I also believe the performance of Graham scan will be improved due to the replacement of `\angle` by `signarea` in the main loop.


src/hulls/graham.jl
```diff
@@ -34,34 +34,32 @@ function hull(points, ::GrahamScan)
...
+  for B in Iterators.drop(q, 1)
+    Δ = signarea(r[end - 1], r[end], B)
+    while isnegative(Δ) && length(r) > 2
       pop!(r)
+      Δ = signarea(r[end - 1], r[end], B)
     end
-    if !iscollinear(r[end - 1], r[end], B)
+    if ispositive(Δ)
+      push!(r, B)
+    elseif evaluate(Euclidean(), r[end - 1], r[end]) < evaluate(Euclidean(), r[end - 1], B)
```
juliohm, line 56: In this case we have three points in the plane that are collinear, but we don't know if the order of these points is A -- B -- C or A -- C -- B. I am using the Euclidean distance to find it out. Do you know a more efficient method?

JoshuaLampert (approved):
Generally, this looks good to me. However, I am not familiar with the algorithm and am not very familiar with these kind of methods. So I can't check the correctness or suggest algorithmic improvements.

juliohm:
No problem. Any review already helps 👍🏽

PR #1215, opened 2025-06-16
Add RegularSampling methods for Triangle and Tetrahedron

### LINKED ISSUE #1208: Implement `RegularSampling` for `Triangle` and `Tetrahedron`
Based on our recent discussion in #1206 , we could consider a more general definition of `RegularSampling` that simply returns an iterator of points inside the simplices using a bounding box and a `point in simplex` check.

juliohm:
Fix #1208

I believe this is the behavior that matches what we currently have in `RegularDiscretization`. We simply sample the bounding box and discard points outside of the simplex.


JoshuaLampert:
Nice! I will review this PR tomorrow morning (CET).

src/sampling/regular.jl
```diff
@@ -87,6 +87,16 @@ extrapoints(c::FrustumSurface, sz) = (bottom(c)(0, 0), top(c)(0, 0))
 # SPECIAL CASES
 # --------------
 
+function sample(rng::AbstractRNG, triangle::Triangle, method::RegularSampling)
```
JoshuaLampert, line 90: It's a matter of taste, but since these two implementations are the same for `Triangle` and `Tetrahedron`, we could also combine them as one dispatching on `Union{Triangle, Tetrahedron}`.

test/sampling.jl
```diff
@@ -303,9 +303,17 @@ end
     @test p ≈ t
   end
 
-  grid = cartgrid(10, 10)
-  points = sample(grid, RegularSampling(100, 200))
-  @test length(collect(points)) == 20000
+  t = Triangle(cart(0, 0), cart(1, 0), cart(0, 1))
+  ps = sample(t, RegularSampling(10))
+  @test all(∈(t), ps)
```
JoshuaLampert, line 308:
```suggestion
  @test all(∈(t), ps)
  @test length(collect(ps)) == 55
```
Maybe also test the number of sampled points to ensure we don't miss any point (`N * (N + 1) / 2 == 55` for `N = 10`)?

test/sampling.jl
```diff
@@ -303,9 +303,17 @@ end
...
+
+  t = Tetrahedron(cart(0, 0, 0), cart(1, 0, 0), cart(0, 1, 0), cart(0, 0, 1))
+  ps = sample(t, RegularSampling(10))
+  @test all(∈(t), ps)
```
JoshuaLampert, line 313:
```suggestion
  @test all(∈(t), ps)
  @test length(collect(ps)) == 220
```
Same as above (`N * (N + 1) * (N + 2) / 6 == 220` for `N = 10`).

JoshuaLampert (changes requested):
Looks good to me. I left some minor suggestions below.

src/sampling/regular.jl
juliohm: We try to avoid union types in dispatch. It can quickly become a nightmare. For example, Distances.jl relies on some union types for dispatch and because of that, people sometimes can't specialize their own metrics.

juliohm:
Accepted the suggestions. Waiting for tests 👍🏽

JoshuaLampert:
That's interesting. Tests are failing because we miss quite some points. Is that because points lie on the boundary of the triangle/tetrahedron and `∈` rejects some of them?

juliohm:
The new tests failed. Could you please double check the expected number of points inside the triangle and tetrahedron? We could also remove the tests.

JoshuaLampert:
I included the boundary points in the expected number of points, but some of them (actually depending on floating point type and platform) are not caught by `∈`. I think it would be a bit confusing if some of the boundary points are included and some not. Seems like the default tolerances are too strict here.

JoshuaLampert:
This is what I get with a `Float32` triangle:
discretized_triangle
A `Float64` includes all the boundary points on the diagonal on my machine.

juliohm:
Floating point issues are always difficult to handle in general. I suggest
that we remove the tests that count points. There will always be some
triangle configuration for which points will be left out.

Em ter., 17 de jun. de 2025, 07:08, Joshua Lampert ***@***.***>
escreveu:

JoshuaLampert:
Ok, would be fine with me.

JoshuaLampert:
Or we could use `RegularSampling(9)`. Then all point locations can be represented exactly in `Float32` and (I think) there will be floating point issues.

juliohm:
Please feel free to explore these changes. I'm currently off my workstation.

Em ter., 17 de jun. de 2025, 07:23, Joshua Lampert ***@***.***>
escreveu:

juliohm:
Thank you @JoshuaLampert for reviewing it. Merging and releasing a patch.

PR #1220, opened 2025-06-26
Refactor Sutherland-Hodgeman clipping

### LINKED ISSUE #1218: Bug in `SutherlandHodgmanClipping` method
The [attached gist](https://gist.github.com/juliohm/2910b19447b84c02970ed105c54c0225) can be used to reproduce the bug:

```julia
using Meshes
using DelimitedFiles

data1 = readdlm("poly1.csv", ',', skipstart=1)
data2 = readdlm("poly2.csv", ',', skipstart=1)

poly1 = PolyArea(Point.(data1[:,1], data1[:,2]))
poly2 = PolyArea(Point.(data2[:,1], data2[:,2]))

poly1 ∩ poly2
```
```julia
ERROR: MethodError: Cannot `convert` an object of type Nothing to an object of type Point{𝔼{2}, Cartesian2D{NoDatum, Quantity{Float64, 𝐋, Unitful.FreeUnits{(m,), 𝐋, nothing}}}}
```

Probably caused by floating point issues.

juliohm:
Fix #1218 

This PR improves the `intersectparameters` utility and fixes bugs downstream whenever the coordinate values are too large in absolute value. It also refactors the Sutherland-Hodgeman clipping method to retain the same circular vector type in all iteration steps.

### DIFF
--- a/test/clipping.jl
+++ b/test/clipping.jl
@@ -77,8 +77,8 @@
   # https://github.com/JuliaGeometry/Meshes.jl/issues/1218
   data1 = readdlm(joinpath(datadir, "issue1218-1.dat"), ',')
   data2 = readdlm(joinpath(datadir, "issue1218-2.dat"), ',')
-  ring1 = Ring(Point.(T.(data1[:, 1]), T.(data1[:, 2])))
-  ring2 = Ring(Point.(T.(data2[:, 1]), T.(data2[:, 2])))
+  ring1 = Ring(cart.(data1[:, 1], data1[:, 2]))
+  ring2 = Ring(cart.(data2[:, 1], data2[:, 2]))
   cring = clip(ring1, ring2, SutherlandHodgmanClipping())
   perim = length(cring)
   if T === Float32
--- a/test/hulls.jl
+++ b/test/hulls.jl
@@ -141,7 +141,7 @@
 
       # https://github.com/JuliaGeometry/Meshes.jl/issues/1211
       data = readdlm(joinpath(datadir, "issue1211.dat"))
-      points = Point.(T.(data[:, 1]), T.(data[:, 2]))
+      points = cart.(data[:, 1], data[:, 2])
       chull = hull(points, method)
       @test area(chull) ≈ T(0.0015160200648848573)u"m^2"
     end


juliohm:
@souma4 that might be relevant to your Bentley-Ottman PR.

src/utils/misc.jl
```diff
@@ -72,31 +72,34 @@ and the rank `rₐ` of the augmented matrix `[A y⃗]` are
...
 
-  # calculate the rank of the rectangular matrix
-  r = sum(>(τ), sum(abs, view(R, :, 1:2), dims=2))
+  # check if a vector is non zero
+  isnonzero(v) = !isapprox(v, zero(v), atol=τ)
+
+  # calculate ranks by checking the zero rows of
+  # the factor R in the QR matrix factorization
+  _, R = qr(A)
+  r = sum(isnonzero, eachrow(R[:, SVector(1, 2)]))
```
JoshuaLampert, line 97:
```suggestion
  r = sum(isnonzero, eachrow(view(R, :, SVector(1, 2))))
```
This slicing allocates needlessly, doesn't it?

src/utils/misc.jl
```diff
@@ -72,31 +72,34 @@ and the rank `rₐ` of the augmented matrix `[A y⃗]` are
...
+  rₐ = sum(isnonzero, eachrow(R))
 
-  # calculate parameters of intersection or closest point
+  # calculate parameters of intersection
   if r ≥ 2
-    λ = A \ y
+    λ = A[:, SVector(1, 2)] \ A[:, 3]
```
JoshuaLampert, line 102:
```suggestion
    λ = view(A, :, SVector(1, 2)) \ SVector(A, :, 3)
```
Similar to above.

JoshuaLampert (changes requested):
LGTM. I only left two related comments below regarding allocations.

juliohm:
In this case the matrices are static, so there are no allocations involved even when we use `getindex`. The `view` has the downside that the type inference no longer works. In order to avoid explicit type annotations, I used `getindex`.

juliohm:
Did some local benchmarks with simple rings:

```
  time:   -60.27% => improvement (5.00% tolerance)
  memory: -39.07% => improvement (1.00% tolerance)
```

juliohm:
@JoshuaLampert I will go ahead with the merge. I believe we have nice speedups besides the bug fix :)

PR #1222, opened 2025-06-27
Make simplexify(::Mesh) type stable

juliohm:
Testing our new benchmark infrastructure 🚀 

Implemented the same union-split trick proposed by @halleysfifthinc in `HalfEdgeTopology`.


juliohm:
The manual union-split is indeed a bit strange. Hopefully the Julia compiler will get smarter over time.

PR #1223, opened 2025-06-27
Multithreaded winding(point, mesh)

### LINKED ISSUE #1197: Use multiple threads in generalized `winding` numbers
The algorithm is embarrassingly parallel, and should be trivial to run in parallel with multiple threads:

https://github.com/JuliaGeometry/Meshes.jl/blob/49654506e8871f237adcf6a403bae6535200930f/src/winding.jl#L44-L65

I am opening the issue in case someone wants to give it a try.

juliohm:
Fix #1197 

I am seeing 2x speedups locally with 8 threads.

Implemented the same trick that @disberd used in `sideof`.

### DIFF
--- a/benchmark/benchmarks.jl
+++ b/benchmark/benchmarks.jl
@@ -1,6 +1,17 @@
 using BenchmarkTools
 using Meshes
 
+# auxiliary variables for tests
+pₒ = Point(0, 0)
+ps = rand(Point, 100)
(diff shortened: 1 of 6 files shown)


juliohm:
@kylebeggs this speedup might be relevant to you.

kylebeggs:
@juliohm thanks for tagging me! Good to know

Issue #810, opened 2024-03-25
`Box`-`Segment` intersection in 3D


cserteGT3:
I have a box, and a segment, that clearly doesn't intersect, but `intersects` returns true:

```julia
julia> testbox = Box(Point3(-130, -18.78, 46.25), Point3(-110.75, -5.05, 60.0))
Box{3,Float64}
├─ min: Point(-130.0, -18.78, 46.25)
└─ max: Point(-110.75, -5.05, 60.0)

julia> testline = Segment(Point(-70,-50,0), Point(-130, 0, 50))
Segment{3,Float64}
├─ Point(-70.0, -50.0, 0.0)
└─ Point(-130.0, 0.0, 50.0)

julia> intersects(testbox, testline)
true
```

Using `Ray` returns false, as expected:

```julia
julia> testbox = Box(Point3(-130, -18.78, 46.25), Point3(-110.75, -5.05, 60.0))
Box{3,Float64}
├─ min: Point(-130.0, -18.78, 46.25)
└─ max: Point(-110.75, -5.05, 60.0)

julia> testray = Ray(Point(-70,-50,0), Point(-130, 0, 50)-Point(-70,-50,0))
Ray{3,Float64}
├─ p: Point(-70.0, -50.0, 0.0)
└─ v: Vec(-60.0, 50.0, 50.0)

julia> intersects(testbox, testray)
false
```

juliohm:
Probably related to #809 

cc: @MachSilva

MachSilva:
As of today, I have not been able to reproduce this error. I'll include it in the tests.

cserteGT3:
With the latest Meshes release, me neither. I could run a git bisect, to check when was this issue resolved, but I think it's unnecessary at this point (will do though if anyone is interested).

Thank you for looking at this and my other intersection issues!

juliohm:
> I'll include it in the tests.

That would be great 👍🏽

PR #1227, opened 2025-07-05
Fix GJK implementation for 3D case

### LINKED ISSUE #809: `Box`-`Triangle` intersection in 3D
I discovered the following:

```julia
t = Triangle((Point(-60.0, -35.0, 55.0), Point(-68.75, -33.43333, 54.90143), Point(-68.75, -35.0, 55.0)))
b = Box((-77.5, -35.0, 54.6073),(-68.75, -31.8914, 55.0))
intersects(b, t)
```
Running this code causes an infinite loop here:
https://github.com/JuliaGeometry/Meshes.jl/blob/4f1a5d86e23d80465e10c45a183bbd7249e6c7fc/src/predicates/intersects.jl#L96-L104

In this particular case changing the argument order solves the issue (running `intersects(t,b)`), but in general it does not solve the issue.
I tried normalizing my input, but it doesn't help.

```julia
s = maximum(sides(boundingbox([t, b])))
tn = Scale(1 / s)(t)
bn = Scale(1 / s)(b)
intersects(bn, tn)
```

I am not sure why is this the case, on other triangle-box pairs it works fine. Maybe because they have one vertice that is common?

MachSilva:
Fixes #809 

The GJK implementation for the 3D case assumes a position for each point in the point array that represents the tetrahedron. The point array is [D, C, B, A] with faces ABC, ADB, BDC and ACD. Each face is a triangle with points in the counterclockwise ordering whose normal vector points outwards. At the end of every iteration, the implementation should remove a point and leave the points that will form the triangle BDC of the next iteration.

When the point C was removed in the second case, the remaining points were not in the correct order. Thus, swapping the points fixed the issue.

Also, add test case for issue #810

### DIFF
--- a/src/predicates/intersects.jl
+++ b/src/predicates/intersects.jl
@@ -196,6 +196,7 @@ function _gjk!(::Val{3}, O, points)
       d = ABCᵀ
     elseif ispositive(ADBᵀ ⋅ AO)
       popat!(points, 2) # pop C
+      points[1], points[2] = points[2], points[1]
       d = ADBᵀ
     elseif ispositive(ACDᵀ ⋅ AO)
       popat!(points, 3) # pop B
--- a/test/predicates.jl
+++ b/test/predicates.jl
@@ -416,6 +416,15 @@ end
   @test intersects(t, b)
   @test intersects(b, t)
 
+  # https://github.com/JuliaGeometry/Meshes.jl/issues/810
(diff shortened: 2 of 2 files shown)


juliohm:
@MachSilva ! 🫶 

That solves a long-standing issue. Thank you!

juliohm (approved):

cserteGT3:
Thank you very much @MachSilva for fixing this issue! 🫶

PR #1229, opened 2025-07-10
Refactor Ray-Triangle intersection

juliohm:
I refactored the code to follow the exact same steps of the non-culling branch described in the Möller, T. & Trumbore, B. 1997 paper. I understand that this branch will compute the correct intersection even when triangles are back-faces.

The new implementation avoids unnecessary intermediate computations like the variable `T` that is only computed after the determinant check. It also follows the notation of the paper more closely to facilitate future maintenance.


juliohm:
@halleysfifthinc given that you touched this code recently to improve type stability, I appreciate if you could take a look to make sure we are not missing something. Please let me know if you are busy and I can try to ping other contributors for review 🙂

halleysfifthinc (approved):
LGTM

halleysfifthinc:
Is this intersection benchmarked by CI yet? Probably a good idea to add now given that the intent of this PR is at least partially to improve performance, afaict?

juliohm:
> Is this intersection benchmarked by CI yet? Probably a good idea to add now given that the intent of this PR is at least partially to improve performance, afaict?

That is a good idea. I will add a benchmark to make sure we don't introduce regressions.

juliohm:
Benchmarks show great speedups when the ray and the triangle do not intersect. Given that this is the most likely configuration, this can make a real difference in practice.

PR #1231, opened 2025-07-11
Set default tolerances to match CoordRefSystems.jl values

juliohm:
The default tolerances in CoordRefSystems.jl were carefully set to make sure that coordinates of objects in the real world are equal up to a physical difference of

```julia
julia> CoordRefSystems.atol(1.0u"m") |> u"mm"
0.011601763617363758 m
```

This physical difference is computed based on a tolerance for unitless floating point numbers defined as

```julia
julia> eps(Float64) ^ (3//4)
1.8189894035458565e-12
```

and the real size of the Earth (semi-major axis).

The tolerance for unitless floating point numbers defined therein is slightly smaller than the current tolerance in Meshes.jl (1e-10). This PR fixes this inconsistency and adjusts tests accordingly. It can be considered a breaking change.

Unlike CoordRefSystems.jl, the tolerances for unitful quantities in Meshes.jl are not based on the size of the Earth. This is because users of Meshes.jl are sometimes interested in `Cartesian{NoDatum}` coordinates and on smaller distances between points represented in micro or nano scale. In that case, we simply copy the tolerance for length, area and volume units using exponentiation as before, without additional factors.

### DIFF
--- a/docs/src/tolerances.md
+++ b/docs/src/tolerances.md
@@ -2,6 +2,7 @@
 
 The absolute tolerance used for floating point comparisons is hard-coded in
 the project to `eps(T) ^ (3 // 4)` where `T` is either `Float64` or `Float32`.
+This formula preserves roughly 3/4 of the significant digits of the type `T`.
 You can use [ScopedValues.jl](https://github.com/vchuravy/ScopedValues.jl) to
 customize these tolerance values in specific computations:
 


juliohm:
This is an important sensitive change. Appreciate your review @JoshuaLampert @halleysfifthinc

JoshuaLampert (approved):
In general, this is fine with me. However, I would indeed consider this a breaking change since the stricter tolerances could lead to different results (as we can see in the tests). Is there a justification for this specific tolerance, i.e., for the exponent of `3//4`?

juliohm:
> Is there a justification for this specific tolerance, i.e., for the exponent of `3//4`?

Nothing super justifiable, just a value that closely matches the previous hard-coded value and leads to a sub-millimeter tolerance in CoordRefSystems.jl.

JoshuaLampert:
I remember I had some issues with a `point in segment` check with the tolerances from before (that is why we added the `ScopedValues` solution back then; IIRC I sampled points on a segment and a test if the points are on the segment failed for some of the points). If we use stricter tolerances, I would expect more of those checks to fail, which is why I would consider this breaking.

juliohm:
> Is there a principled reason (e.g. performance?) for having this different than eps?

Asking for `eps(Float64) ≈ 1e-16` precision is too much for most algorithms. The error accumulates quickly on basic algebraic operations. `eps(Float64) ^ (1//2) ≈ 1e-8` preserves "half" of the digits, and `eps(Float64) ^ (3//4) ≈ 1e-12` preserves "3/4" which is a bit more conservative.

I agree with @JoshuaLampert that this is too sensitive to be released as a patch release. We need a minor release to signal that things can potentially break downstream.

juliohm:
Added an extra sentence to the docs explaining the formula. Thank you for the input!

PR #1234, opened 2025-07-19
Add AdaptiveSampling and AdaptiveDiscretization for one-dimensional parametric geometries

sschuldenzucker:
This uses the algorithm from here: https://github.com/sschuldenzucker/parametricadaptivesampling.jl/

Tested with `ParametricCurve` and `BezierCurve`.


sschuldenzucker:
Huh the 32 bit tests seem to fail for subtle numerical deviations

souma4:
So I'll jump in with an unofficial review on the simple-to-fix/broad stuff before Julio and Josh since I have the time. The code conventions are that variable names should be short, descriptive, and not use underscores unless it is an internal, private function, in which case the underscore starts the function name (i.e `_foo()`). I also wonder, with how many SOMEDAYS there are, if this should be more of a running PR to add in those further features.

Other's can chime in on this, but we are planning on adding [BinaryTrees.jl](https://github.com/JuliaCollections/BinaryTrees.jl) to this package, so that might be a way of creating your sorted sequence (currently BinaryHeap) without adding an extra datastructure dependency in the long term. 

As for the FP32, you concrete type `Float64` in your BinaryHeap which is promoting results to `Float64`, but then testing it against `Float32`, I think swapping those `Float64`'s into `T` (the numtype) should resolve the FP32 test failures.

juliohm:
💯% agree with @souma4 's review.

Try to adjust the code for a second round of review @sschuldenzucker. There are code style adjustments and dependency adjustments to make. If all you need is a BinaryTree, then BinaryTrees.jl is a better (more lightweight) dependency. DataStructures.jl has too much in it.

juliohm:
Adaptive sampling/discretization will be super useful for visualization and other purposes. Depending on the performance of the implementation compared with `RegularSampling` and `RegularDiscretization` we could even consider it as the default discretization method for visualization of curved 1D geometries (e.g., BezierCurve).

souma4:
At some point I'll write up an update to the contribution document to give a push for what the style of this code base is, but scouring old messages and stuff I notice I get these. I don't see all of them from my quick review, but good to have them in one place.
1. short descriptive variable names without underscores (unless it's a private function, which is an underscore up front)
2. maintain variable naming patterns throughout function barriers. (i.e. if you create points `p...`, the functions that use those should have the argument `p`).
3. Try to keep variable names consistent with the code base, too. So T is for numtype. `\scrL` is for lentype. etc.
4. Structs, even private ones, don't have preceding underscores.
5. minimize parameterization of structs. (I think you do very well with this, but I tend to over parameterize as a way to reason about code, so good to mention just in case).
6. Concrete types should be kept to a minimum. Parametric types tend to handle it better.
7. When in doubt, look at the GeoStats.jl stack to see what patterns you see and try to shoot for those.

juliohm:
Thanks for putting this list together @souma4 ! 🫶 

I would also add that we try to follow the notation from scientific publications when that is possible. That way new contributors can read the paper and the implementation side by side to propose improvements more easily.

sschuldenzucker:
Hi!

1. Removed underscores from variable names and hopefully made them a bit more readable.
2. Did my best, though sometimes this hits a generality bound (e.g., called function accepts any callable `f` and is called with a geometry `geom` for `f`)
3. Did my best.
4. Already ok.
5. Already ok, though I'm wondering if this leads to type instability.
6. Done, replaces this one mention of Float64 with the error type `E` and it seems fine now.

On `DataStructures.jl` vs `BinaryTrees.jl`: This only implements binary search (AVL) trees but I need heaps. Note that we already depend on `DataStructures` indirectly (`Manifest.toml` did not change) and I generally don't see any problem with pulling in some very standard data structures, either.

As for the SOMEDAYs: The ones that are not in the docstring are mostly about code quality; happy for any pointers.

For the ones in the docstring, these seem to sit at various levels of difficulty. Here's an annotated list. Would love to do more but there seem to be some tricky parts:

1. Only geometries with one-dimensional parameter domains are supported. It would be cool to implement higher dimensions. We then probably want a discretization, not just sampling. - **Somewhat hard. The main ingredient would be a way some machienary to split simplices into sub-simplices (or polygons into sub-polygons) and that does the required accounting for me. Do you have that?**
2. Infinite or open t ranges where we would adaptively sample larger/smaller (towards infinity or the edges) t values. This may be an instance of a more general hook into refinement. - **AFAIU we don't have the geometries for this right now? `ParametrizedCurve` only supports closed t ranges right?**
3. Some way of detecting and avoiding polar points. Basically a way to say "it's not useful to sample this part, better to cut it out". Not clear how we'd do this in a robust way without affecting some use cases negatively. - **There seems to be a fundamental trade-off between detecting polar points and not excluding extreme non-polar regions. Seems tricky to get right generically, though there certainly are some methods.**
4. Optional penalty for recursion depth (or something like this) to avoid excessive concentration at polar points if `maxPoints` gets exhausted. This would have to be configured by the user with knowledge of the function, can be detrimental otherwise. Unclear if we want this. - **We may just not need this; if we do, a simple feature to add. Would implement the rest first and see if we need this then.**
5. Option to split into more than two parts per recursion step, or to split not into halves but by some other proportions. This could help when details are missed by the 2-split recursion (see Caveats). - **I wouldn't split into more than two parts but we can choose the split point dynamically based on the derivative (see below).**
6. Point density target in value space. Could help in situations where the midway point is interpolated well linearly, but there is additional variation at some other point. - **Should be reasonably easy to add but not sure if needed / actually desired.**
7. Option to drop points that are ultimately not needed if the range expands during refinement (may reduce number of points, might be useful for some later computation steps). - **Easy to add, may not be needed.**
8. A way to understand that the range may actually be the wrong thing, e.g., when we use
9. `aspect_ratio=1` in the plot and the plot therefore creates additional space, or some xlim or ylim. Hard to do generically. Maybe make an option to pass the plot range explicitly. - **This sounds like a PITA to do; we could have an easy `aspect_ratio` option if it's really desired for the specific use case above.**
10. Is there some smartness we can do if the (second) derivative of f is available? Choosing the splitting point better than at the interval midpoint seems attractive. - **This is probably either easy OR not useful ;should be easy to figure out. How would we pipe the derivative through the geometry infra though? Maybe the right abstract concept is "curvature" here?**

juliohm:
> Note that we already depend on `DataStructures` indirectly

That is really unfortunate. DataStructures.jl is one of those packages that provides various incomplete implementations without a single active maintainer. I've encountered so many issues in the past that I realized it would be safer to just not use it at all.

It seems that the dependency is coming from another poorly maintained package, which is doubly unfortunate:

```jl
(jl_YsCiCv) pkg> why DataStructures
  Meshes → StatsBase → DataStructures
  Meshes → StatsBase → SortingAlgorithms → DataStructures
```

Our ultimate goal is to remove StatsBase.jl from the list of dependencies. It dates back to the origins of Julia when the Statistics stdlib was part of the Julia distribution and people were trying to add more stats functionality independently of the Julia release process.

docs/src/algorithms/discretization.md
```diff
@@ -157,6 +157,20 @@ mesh = discretize(sphere, RegularDiscretization(10,10))
...
+```@docs
+AdaptiveDiscretization
+```
+
+```@example discretization
+crv = ParametrizedCurve((t -> Point((sin(t * 2π) * t, cos(t * 2π) * t))), (0, 1))
+method = AdaptiveDiscretization(; min_points=10.0, max_points=100.0)
+mesh = discretize(crv, method)
+
+viz(mesh, showsegments = true)
```
juliohm, line 171: Please rename `crv` to `curve`. Also, the keyword arguments have underscores, which is not adherent to our code style.

docs/src/algorithms/sampling.md
```diff
@@ -129,3 +129,19 @@ points  = sample(sphere, sampler) |> collect
...
+AdaptiveSampling
+```
+
+```@example sampling
+# Sample between 10 and 100 points
+sampler = AdaptiveSampling(; min_points=10.0, max_points=100.0)
+crv = ParametrizedCurve((t -> Point((sin(t * 2π) * t, cos(t * 2π) * t))), (0, 1))
+points  = sample(crv, sampler) |> collect
+
+viz(points)
```
juliohm, line 145: Same suggestions I shared in the other discretization example.

src/sampling/adaptive.jl
```diff
@@ -0,0 +1,204 @@
+# ------------------------------------------------------------------
+# Licensed under the MIT License. See LICENSE in the project root.
+# ------------------------------------------------------------------
+
+using DataStructures: BinaryHeap
+using LinearAlgebra: norm
+import IterTools
```
juliohm, line 7: The only place in the source code where we import modules is the main Meshes.jl file. That way we don't loose track of current imports and avoid naming conflicts.

src/sampling/adaptive.jl
```diff
@@ -0,0 +1,204 @@
...
+
+"""
+    AdaptiveSampling(; [options])
+
+Incrementally sample points such that the error from linear interpolation between these points is small. This is useful for plotting parametric functions or Bezier curves where curvature is very non-uniform.
+
+This is *only* implemented for geometries with a one-dimensional parameter domain (i.e., realistically, [`ParametrizedCurve`](@ref) or [`BezierCurve`](@ref))!
+
```
juliohm, line 15:
```suggestion
```

No need to make this information available in the docstring. At some point in the future, we hope to fill in the gaps.

src/sampling/adaptive.jl
```diff
@@ -0,0 +1,204 @@
...
+## Options
+
+* `tol` - Tolerance for the linear interpolation error. The meaning depends on `errfun`.
+* `errfun` - Function to compute the interpolation error. The default is [`errRelativeBBox`](@ref) = the l-infinity distance component-wise relative to (an estimate of) the range of the bounding box; this is a good default for plotting. If provided, it must satisfy `errfun(::ParametrizedSegment, ::MutableBox)::E`
+* `minPoints` - Number of points sampled uniformly initially.
+* `maxPoints` - Maximum number of points to sample. We do our best to, if this is hit, get rid of the worst approximation errors first.
```
juliohm, line 21: Please revise the options as they follow a camel case notation, not supported by our code style. Also, it is worth thinking which of these options will be changed in practice. Do you expect end-users to change the `errfun`?

src/sampling/adaptive.jl
```diff
@@ -0,0 +1,204 @@
...
+
+### Operation
+
+This function works as follows: first, it performs an initial presampling step. Then, it considers the segment where the error of linear interpolation across the segment vs the midpoint value (midpoint in t space) is worst and splits it into two halves. Then repeat until all errors are below the tolerance or we run out of points. The estimate of the value range is updated as we add new points.
+
```
juliohm, line 26: This description should appear in the beginning of the docstring as it defines the method.

For example, you could combine the first sentence of the docstring with the text in this paragraph to obtain something like

```
Incrementally sample points such that the error from linear interpolation between these points is small. First, the method performs an initial presampling step. Then, it considers the segment where the error of linear interpolation across the segment vs the midpoint value (midpoint in t space) is worst and splits it into two halves. Then repeat until all errors are below the tolerance or we run out of points. The estimate of the value range is updated as we add new points.
```

src/sampling/adaptive.jl
```diff
@@ -0,0 +1,204 @@
...
+### Caveats
+
+If `minPoints` is low, `range` in `errfun` may be catastrophically small, at least for initial refinement. You probably don't want that. More generally, if the initial samples do not represent the value range reasonably well, in some cases, refinement may add excessive points, and may even reach `maxPoints` without any meaningful progress.
+
+Recursion is by splitting into halves in t space, so if your function has details that are missed by this recursive grid, they won't show up. In this case, increasing `minPoints` may help.
```
juliohm, line 31: These should appear in a `### Notes` section as we do with other implementations in the code base.

src/sampling/adaptive.jl
```diff
@@ -0,0 +1,204 @@
...
+- Option to split into more than two parts per recursion step, or to split not into halves but by some other proportions. This could help when details are missed by the 2-split recursion (see Caveats).
+- Point density target in value space. Could help in situations where the midway point is interpolated well linearly, but there is additional variation at some other point.
+- Option to drop points that are ultimately not needed if the range expands during refinement (may reduce number of points, might be useful for some later computation steps).
+- A way to understand that the range may actually be the wrong thing, e.g., when we use
+  `aspect_ratio=1` in the plot and the plot therefore creates additional space, or some xlim or
+  ylim. Hard to do generically. Maybe make an option to pass the plot range explicitly.
+- Is there some smartness we can do if the (second) derivative of f is available? Choosing the splitting point better than at the interval midpoint seems attractive.
```
juliohm, line 47: Not needed. If we start adding missing features to all available methods in their docstrings, then users will never stop reading :)

src/sampling/adaptive.jl
```diff
@@ -0,0 +1,204 @@
...
+"""
+struct AdaptiveSampling{E} <: ContinuousSamplingMethod
+    tol::E
+    errfun::Function
+    minPoints::Int
+    maxPoints::Int
+    function AdaptiveSampling(tol::E, errfun, minPoints, maxPoints) where {E}
+        # assertion(minPoints >= 2, "Need at least two initial points")
+        new{E}(tol, errfun, minPoints, maxPoints)
+    end
```
juliohm, line 57: No need for inner constructor with empty body.

src/sampling/adaptive.jl
```diff
@@ -0,0 +1,204 @@
...
+- Option to split into more than two parts per recursion step, or to split not into halves but by some other proportions. This could help when details are missed by the 2-split recursion (see Caveats).
+- Point density target in value space. Could help in situations where the midway point is interpolated well linearly, but there is additional variation at some other point.
+- Option to drop points that are ultimately not needed if the range expands during refinement (may reduce number of points, might be useful for some later computation steps).
+- A way to understand that the range may actually be the wrong thing, e.g., when we use
+  `aspect_ratio=1` in the plot and the plot therefore creates additional space, or some xlim or
+  ylim. Hard to do generically. Maybe make an option to pass the plot range explicitly.
+- Is there some smartness we can do if the (second) derivative of f is available? Choosing the splitting point better than at the interval midpoint seems attractive.
+"""
+struct AdaptiveSampling{E} <: ContinuousSamplingMethod
+    tol::E
```
juliohm, line 50: Are you assuming that the tolerance is expressed as a unitless floating point number or as a tolerance with length units? We usually adopt the latter in our methods because they are more useful in applications. A user might want to sample the curve until the deviation/error is smaller than 1mm for example.

src/sampling/adaptive.jl
```diff
@@ -0,0 +1,204 @@
...
+    errfun::Function
+    minPoints::Int
+    maxPoints::Int
```
juliohm, line 53: Not compliant with our code style.

src/sampling/adaptive.jl
```diff
@@ -0,0 +1,204 @@
...
+"A segment of three `(t, v)` values where `t` is the parameter value and `v` is the point. Require `t1 < tMid < t2` but not checked."
+# SOMEDAY is there a better struct for this somewhere already?
+struct ParametrizedSegment{T,V}
+    t1::T
+    v1::V
+    tMid::T
+    vMid::V
+    t2::T
+    v2::V
+end
```
juliohm, line 85: Why not use the `Segment` type? It is parametrized already.

src/sampling/adaptive.jl
```diff
@@ -0,0 +1,204 @@
...
+function ParametrizedSegment(f, t1, t2)
+    ParametrizedSegment(f, t1, f(t1), t2, f(t2))
+end
+
+"Like `Box` but mutable."
+mutable struct MutableBox{V}
+    # Implicit: All lengths are equal
+    mins::Vector{V}
+    maxs::Vector{V}
+end
```
juliohm, line 102: I don't think mutability is enough justification to create another box type.

test/discretization.jl
```diff
@@ -545,6 +545,18 @@ end
   @test nvertices.(mesh) ⊆ [4]
 end
 
+@testitem "AdaptiveDiscretization" setup=[Setup] begin
+  # This is a bit of a dummy test because AdaptiveDiscretization just wraps AdaptiveSampling.
+  min_points = 5
+  max_points = 10
```
juliohm, line 551: Not compliant with code style.

juliohm (changes requested):
Thank you @sschuldenzucker for attempting your first contribution! 🎉 

We still have work to do in order to improve the quality of the code and adhere to the project standards. Please feel free to ask any question as you are working on the review comments.

juliohm:
ping @sschuldenzucker

PR #1236, opened 2025-07-29
Add `intersects(::Intersection)` method

halleysfifthinc:
This is a convenient method to have when you want to also grab the intersection, if the
geometries actually intersect.

Example use case/pattern:

```julia
i = intersection(r::Ray, t::Triangle)
if intersects(i)
    # stuff...
end
```

which reads better/more directly than

```julia
i = intersection(r::Ray, t::Triangle)
if type(i) !== NotIntersecting
    # stuff...
end
```

### DIFF
--- a/src/predicates/intersects.jl
+++ b/src/predicates/intersects.jl
@@ -21,19 +21,21 @@ a well-defined [`supportfun`](@ref).
 """
 function intersects end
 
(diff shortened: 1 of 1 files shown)


juliohm:
I don't think we need this method. The `intersects` predicate has a well-defined meaning.

The internal `someintersection` function that you introduced in a past PR does the job if you really need a separate function name.

Issue #1237, opened 2025-07-29
Organize CI as dependent sequence


halleysfifthinc:
It might be helpful to set the different CI jobs to wait to run after logical predecessors pass. This would reduce our CI use (@juliohm has mentioned he thinks the jobs are throttled sometimes due to the number of them and their compute load?)

E.g. Documentation and benchmarks both wait to run until CI has run and passed. (Optionally, don't even run the tests workflow until the style check passes?)

Should be doable using https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idneeds and https://docs.github.com/en/actions/reference/workflows-and-actions/reusable-workflows. I have such a setup for my `Peaks.jl` package, where the documentation build waits until the tests pass.

juliohm:
If you know how to set this up in the GitHub Action, please submit a PR and we can review. It could definitely save resources!

juliohm:
For some reason the issue was not closed  by the merged PR. Closing manually.

PR #1239, opened 2025-08-01
Reorganize CI into conditional hierarchy

halleysfifthinc:
The general CI structure should be as follows:

- Anything can be triggered manually (`workflow_dispatch`)
- Style/format suggestions always runs
- Tests run for pushes to `master` and tags, and only run in PRs if there are modifications to `Project.toml` or within `src`, `ext`, or `test`
- Benchmarks always runs following successful tests (and transitively by anything that triggers tests)
- Documentation runs for pushes to `master` and tags, and runs in PRs if:
  - Tests workflow successfully ran (catches changed docstrings)
  - Anything was modified in `docs` (catches anything else that could affect docs builds)

I don't have a great idea on how to test some of these without just merging this and then submitting some test PRs to trigger specific behaviors.

### DIFF
--- a/.github/workflows/Benchmarks.yml
+++ b/.github/workflows/Benchmarks.yml
@@ -1,13 +1,21 @@
 name: Benchmarks
 on:
+  workflow_run:
+    workflows: ["CI"]
+    types:
+      - completed
   pull_request_target:
     branches: [ master ]
+  workflow_dispatch:
+  
 permissions:
   pull-requests: write
 
 jobs:
   bench:
     runs-on: ubuntu-latest
+    # Only run if CI workflow succeeded or if manually triggered
+    if: ${{ github.event.workflow_run.conclusion == 'success' || github.event_name == 'pull_request_target' }}
     strategy:
       matrix:
         version:
@@ -17,4 +25,4 @@ jobs:
       - uses: MilesCranmer/AirspeedVelocity.jl@action-v1
         with:
           julia-version: ${{ matrix.version }}
-          bench-on: ${{ github.event.pull_request.head.sha }}
+          bench-on: ${{ github.event.pull_request.head.sha || github.event.workflow_run.head_sha }}
--- a/.github/workflows/Benchmarks.yml
+++ b/.github/workflows/Benchmarks.yml
@@ -3,12 +3,49 @@ on:
   pull_request_target:
     types: [labeled]
     branches: [ master ]
+  workflow_run:
+    workflows: ["CI"]
+    types: [completed]
+    branches: [ master ]
   workflow_dispatch:
   
 permissions:
   pull-requests: write
 
 jobs:
+  add-benchmark-label:
+    name: Add benchmark label
+    runs-on: ubuntu-latest
+    if: ${{ github.event_name == 'workflow_run' && github.event.workflow_run.conclusion == 'success' }}
+    steps:
+      - name: Add benchmarkable label
+        uses: actions/github-script@v7
+        with:
+          script: |
+            // Get the PR associated with this workflow run
+            const { data: prs } = await github.rest.pulls.list({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              base: 'master',
+              head: `${context.payload.workflow_run.head_repository.owner.login}:${context.payload.workflow_run.head_branch}`,
+              state: 'open'
+            });
+
+            if (prs.length === 0) {
+              console.log('No open PR found for this workflow run');
+              return;
+            }
+
+            const pr = prs[0];
+            console.log(`Adding benchmarkable label to PR #${pr.number}`);
+
+            await github.rest.issues.addLabels({
+              issue_number: pr.number,
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              labels: ['benchmarkable']
+            });
+
   bench:
     runs-on: ubuntu-latest
     # Only run if benchmarkable label was added or if manually triggered
--- a/.github/workflows/Benchmarks.yml
+++ b/.github/workflows/Benchmarks.yml
@@ -1,10 +1,7 @@
 name: Benchmarks
 on:
-  workflow_run:
-    workflows: ["CI"]
-    types:
-      - completed
   pull_request_target:
+    types: [labeled]
     branches: [ master ]
   workflow_dispatch:
   
@@ -14,8 +11,8 @@ permissions:
 jobs:
   bench:
     runs-on: ubuntu-latest
-    # Only run if CI workflow succeeded or if manually triggered
-    if: ${{ github.event.workflow_run.conclusion == 'success' || github.event_name == 'pull_request_target' }}
+    # Only run if benchmarkable label was added or if manually triggered
+    if: ${{ contains(github.event.label.name, 'benchmarkable') || github.event_name == 'workflow_dispatch' }}
     strategy:
       matrix:
         version:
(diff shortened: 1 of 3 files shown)


.github/workflows/CI.yml
```diff
@@ -3,17 +3,25 @@ on:
   pull_request:
     branches:
       - master
+    paths:
+      - 'src/'
+      - 'test/'
+      - 'ext/'
```
juliohm:
```suggestion
      - 'ext/'
      - 'test/'
```

juliohm:
Requested a review from @JoshuaLampert as he has much more experience with these settings.

juliohm:
Fix #1237

halleysfifthinc:
> Could you please elaborate on this item? What exactly `workflow_dispatch` does?

It's a weird name, but it just allows us to trigger the workflow manually from the workflow's page under the "Actions" tab without opening a PR or pushing a new commit. e.g. You could go to the repository "Actions" tab, click on the "CI" workflow, and you would see this banner:
image
That can be useful for example to re-run tests for a new `nightly`.

.github/workflows/Benchmarks.yml
```diff
@@ -1,13 +1,21 @@
 name: Benchmarks
 on:
+  workflow_run:
+    workflows: ["CI"]
+    types:
+      - completed
   pull_request_target:
     branches: [ master ]
```
halleysfifthinc: The `pull_request_target` is triggering the benchmarks workflow immediately, which is not desired. I'm exploring alternatives, but they might prevent the PR comments style of results vs in the checks tab. How important/strong a preference is the benchmark results in comments vs in the checks tab?

JoshuaLampert: I prefer having the comments. As you say, changing this to `pull_request` has advantages, but also means we do not get the comments if the PR is from a fork. There is some discussion about this starting at https://github.com/MilesCranmer/AirspeedVelocity.jl/issues/97#issuecomment-2936248772. If we want to switch to `pull_request` instead of `pull_request_target` we would set `job-summary: 'true' `, but then the tables are a bit hidden.

.github/workflows/Benchmarks.yml
```diff
@@ -1,13 +1,21 @@
 name: Benchmarks
 on:
+  workflow_run:
+    workflows: ["CI"]
+    types:
+      - completed
```
JoshuaLampert: Do we really want to run the Benchmarks only after CI is finished? This would mean more time to wait for the actions before a PR can be merged. Currently, the benchmarks run for ~3 minutes, but this might increase in the future and then could become annoying to wait.

.github/workflows/Documentation.yml
```diff
@@ -0,0 +1,50 @@
+name: Documentation
+on:
+  workflow_run:
+    workflows: ["CI"]
+    types:
+      - completed
```
JoshuaLampert, line 6: Similar question as above. Documentation needs ~11 minutes to build, i.e. waiting time for a PR until everything is green, would increase.

juliohm:
These are all very good points. It boils down to saving energy vs. saving
time.

If the tests fail, then there is no much point in running benchmarks as we
will not be willing to merge anyways.

On there other hand, it is nice to have the full set of actions finish as
soon as possible.

I am leaning towards saving energy as proposes in the PR. But no strong
opinions.

In any case it would be great to foresee the possible issues of a more
complex hierarchy. We want to spend time reviewing code, and avoid possible
flaws with the hierarchical settings.

Looking forward to further comments.

Em sex., 1 de ago. de 2025, 17:57, Joshua Lampert ***@***.***>
escreveu:

juliohm:
Regarding benchmark results, I also think it is great to read them directly
in the comments.

Em sex., 1 de ago. de 2025, 18:16, Júlio Hoffimann <
***@***.***> escreveu:

.github/workflows/Benchmarks.yml
halleysfifthinc: My thought is that this saves a potentially pointless Benchmarks run if failing tests would cause failures in the benchmark.

This becomes more important in a future with a more substantial benchmark suite: Generally the only benchmarks that matter for a given PR are benchmarks that test new or newly modified code; if the test suite fails, then the (specifically relevant) benchmarks won't finish/succeed either.

IMO that's a reasonable tradeoff for a slightly delayed merge timeframe.

.github/workflows/Documentation.yml
halleysfifthinc: See above response.

halleysfifthinc:
> Do you have any evidence for that?

Nothing direct, I have observed some slow/delayed job scheduling in the Meshes CI a couple of times. My overall goal/thought here is to use the free resources more efficiently and conservatively.

.github/workflows/Benchmarks.yml
JoshuaLampert: I understand this point. I guess it's just a matter of how you set your priorities. I just wanted to state my priority, but if you agree that saving energy has higher priority, I'm fine with merging this.

juliohm:
Let's wait a bit more to brainstorm possible impacts of this before merging.

Em sex., 1 de ago. de 2025, 18:30, Joshua Lampert ***@***.***>
escreveu:

halleysfifthinc:
> Let's wait a bit more to brainstorm possible impacts of this before merging.

Agreed; and I still need to see if I can conditionally trigger the Benchmarks after tests pass.

One thing that might reduce any pain from slightly longer total Actions/checks process is to enable a [merge queue](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue) for Meshes.jl. Then, approval can be given and the PR will be merge automatically iff all the checks complete successfully (e.g. approve the content of a docs change which will be merged if the docs builds successfully, etc).

juliohm:
To summarize some of the points raised above:

1. It would be nice to preserve benchmarks as comments.
2. We could try to run benchmarks and docs *after* CI tests pass.

If (2) requires too much CI engineering work, then it is probably better to stick to our greedy approach.

@halleysfifthinc LLM bots might help here. If you ask ChatGPT or similar to produce the desired behavior with GitHub Actions, it should be able to produce something reasonable.

halleysfifthinc:
Ok, with some assistance from Claude, I believe I have the benchmarks set to trigger only after a successful CI workflow, but because the benchmarks runs using the workflow from the base branch (and not the PR branch) we can't test this without merging (and then reverting/updating as needed).

A possible alternative is that I can push this branch to the Meshes repo, tweak it so that all the workflows trigger on PRs to this branch, and then confirm that everything works as intended.

> We could try to run benchmarks and docs after CI tests pass.

If I'm understanding right, @juliohm is suggesting we try this PR/CI arrangement, see how it works out in practice, and revert if its too inconvenient? That is fine with me.

juliohm:
Thank you for looking into this, @halleysfifthinc. The proposed changes go beyond my CI maintenance skills, and if I had to decide between keeping it simple (the current CI config) versus adding sophistication (this PR) I would choose the former alternative.

@JoshuaLampert what is your take on this?

JoshuaLampert:
Interesting approach to add a label once CI was successful. I am not 100% sure, but this might work. As I said above, I don't know if this is worth it, but simply out of curiosity if this works, we can merge this and test it at least for a while. If it turns out to be annoying and/or difficult to maintain, we can always go back to the original version.

juliohm:
Thank you all for the discussion. Let's see how this works out in practice 👍🏽

PR #1240, opened 2025-08-11
Use linked-list workflows when constructing `HalfEdgeTopology`s

halleysfifthinc:
The HalfEdgeTopology construction process (topological adjacency sort, and then creating the oriented edges) heavily uses a code pattern of removing elements from an array that contains items we haven't dealt with yet (seen or processed, etc). Previously, this looked like `popat!` calls to an array, but the `popat!` runtime is dependent on the location of removed element, e.g. removing the last
element is fastest, and removing the first (or second, depending on implementation) element is slowest. This translates to order dependent runtime, where the initial order of connectivity elements can dramatically affect the runtime of the HalfEdgeTopology construction.

This PR switches to a linked-list, which has O(1) element removal, using a simplified, vendored[^1] copy of `MutableLinkedList` from DataStructures.jl. The benchmarks should show up to a ~1.8x speedup (depending on the order of the connectivity array), based on what I'm seeing on my machine.

[^1]: Vendored because the MutableLinkedList in DataStructures has apparently been needing
    love for quite some time and an old PR (which I used for the vendored copy) which
    addresses longstanding issues hasn't been accepted yet.


halleysfifthinc:
(Converted to a draft pending fixes to CI triggers by #1241 )

halleysfifthinc:
@juliohm would you prefer the style in the `"mutable_list.jl"` file to be updated for Meshes.jl style, or keep the original style to maintain a stronger resemblance to the "upstream" DataStructures.jl code (in case there are ever any relevant update)?

halleysfifthinc:
**tl;dr**: I don't think adding a dependency for this single use is the right call at the moment, but something that should be re-evaluated in the future if we discover a broader beneficial use of the linked-list.

1. This code already exists in a registered package (and in a PR to that package). If you want this to be an external package, the right approach would be for us to push for a review/acceptance of that PR, and add DataStructures.jl as a dependency.
2. This vendored linked-list isn't intended to be used more broadly in Meshes.jl beyond this specific use, and I stripped all other unused functionality from the original implementation specifically to discourage other use. (Linked-lists are rarely the correct choice.) The other benefit of removing unused functionality is that it simplifies the maintenance burden, which should be effectively nil anyways.

> think about the interface more carefully, including function names and possible overlaps with Base names.

The only new function/interface I added compared to the upstream DataStructures implementation is the `popprev!` function. It is unlikely that Base Julia would add a `popprev!` function, but we could add an underscore prefix if desired to avoid that possibility.

juliohm:
> **tl;dr**: I don't think adding a dependency for this single use is the right call at the moment, but something that should be re-evaluated in the future if we discover a broader beneficial use of the linked-list.

I understand where you are coming from. You are trying to minimize the amount of work necessary to get things done in your specific use case. Unfortunately, Meshes.jl is not the place to "push code and fix later". We need to make sure that added code is well-tested, specially if it is an entirely new data structure.

> 1. This code already exists in a registered package (and in a PR to that package). If you want this to be an external package, the right approach would be for us to push for a review/acceptance of that PR, and add DataStructures.jl as a dependency.

DataStructures.jl suffers from issues related to social inertia: it is a bag of data structures developed by different people and maintained by no one. As a result, no single maintainer feels confident to merge PRs. A dead end.

We introduced BinaryTrees.jl after careful review, considering a general interface that accounted for all possible use cases of binary trees. LinkedLists.jl would provide the same encapsulation for linked lists (mutable or not), and would be maintained by a reduced (and known) set of people (me, you, ...).

> 2\. This vendored linked-list isn't intended to be used more broadly in Meshes.jl beyond this specific use, and I stripped all other unused functionality from the original implementation specifically to discourage other use

This is a valid development philosophy in other projects, but we favor the opposite approach in Meshes.jl: general functionality that might be useful to other people are developed in well-tested packages and only used in Meshes.jl.

These pushbacks might feel annoying, I know, but they are necessary for project/software health in the long-term.

juliohm:
@halleysfifthinc out of curiosity, are you trying to beat some benchmark somewhere that involves the `HalfEdgeTopology`? Or is this demand coming from a need to construct many extremely large surface meshes?

halleysfifthinc:
The latter. We are working with ~100-120k face meshes and performing multiple partition => independent (vs views) mesh creation from the partitions. HalfEdgeTopology creation is a significant proportion of our software's runtime, and we need the orientability and topological information of the HalfEdgeTopology.

halleysfifthinc:
> Unfortunately, Meshes.jl is not the place to "push code and fix later".

That's a disingenuous characterization of what I'm proposing, regardless of whether you agree with me.

> As a result, no single maintainer feels confident to merge PRs. A dead end.

I agree with you there; that was a part of my motivation in vendoring the linked-list implementation.

> general functionality that might be useful to other people are developed in well-tested packages and only used in Meshes.jl.

That's a nice ideal, but its not practical/scalable. Maintainers/contributors have finite time and energy, and creating a separate package for everything that potentially has wide/generic use can easily devolve into a bunch of unmaintained packages.

juliohm:
> That's a disingenuous characterization of what I'm proposing, regardless of whether you agree with me.

Sorry, I was just trying to clarify that we need some extra work to get changes approved into the source code. I've noticed this pattern in some of your previous contributions and wanted to point it out again that we can't rush into a design just to meet external deadlines.

We are happy to brainstorm improvements given the necessary efforts to achieve a "good enough" solution. There are many open PRs that we didn't merge for similar reasons. We do our best to contain the "beast" that is open source software.

PR #1241, opened 2025-08-11
Fix CI path filters

halleysfifthinc:
https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#patterns-to-match-file-paths


juliohm (approved):

JoshuaLampert (approved):
Let's hope this is enough to trigger CI in the other PR.

PR #1242, opened 2025-08-11
Fixup CI triggers, take #2

halleysfifthinc:
Two issues:

1. The add "benchmarks" label step wasn't working; I think/hope the new script in the (also
   new) Labeling workflow will work as originally intended.
2. Documentation wasn't building the PR commit as intended (for `workflow_run` the
   SHA/commit defaults to master, not the PR). I switched the CI-dependent documentation
   build to the same label triggered setup so that it triggers from a `pull_request` context to have the right SHA/commit.
3. I moved the labeling job to its own workflow file to simplify some checks in the
   benchmarks and documentation workflows. The trigger `if` checks are slightly simpler when
   they don't have to deal with the `workflow_run` event trigger.

Sorry for the noise; hopefully this will be worth it in the end.

### DIFF
--- a/.github/workflows/Benchmarks.yml
+++ b/.github/workflows/Benchmarks.yml
@@ -23,13 +23,7 @@ jobs:
         with:
           script: |
             // Get the PR associated with this workflow run
-            const { data: prs } = await github.rest.pulls.list({
-              owner: context.repo.owner,
-              repo: context.repo.repo,
-              base: 'master',
-              head: `${context.payload.workflow_run.head_repository.owner.login}:${context.payload.workflow_run.head_branch}`,
-              state: 'open'
-            });
+            const prs = context.payload.workflow_run.pull_requests;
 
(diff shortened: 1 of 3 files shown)


halleysfifthinc:
This (correctly) didn't trigger CI since there are no Julia or test workflow (i.e. CI.yml) changes.

juliohm (commented):
I will leave it to @JoshuaLampert , but I am starting to feel the complexity growth of this approach. If we can't make it work after this iteration, we should probably revert to a vanilla CI config as before.

JoshuaLampert (approved):
This now also exceeds my CI skills. I can't really judge if this will be working or not. I agree with @juliohm that if this fix is not working as we wish, it is probably easier to go back to the previous version (or maybe keep some smaller changes like CI is only triggered if a certain set of files is touched by the PR or allow `workflow_dispatch`). But let's try this nevertheless.

PR #1243, opened 2025-08-11
Revert overly complex staged/hierarchical CI

halleysfifthinc:
The last PR didn't work out for Documentation triggers, and I agree that this is becoming
too much to deal with/maintain. Apologies again for the noise/waste of time.


juliohm:
Can you please confirm that we also need to remove the extra Documentation.yml?

juliohm:
Just removed it. I believe we are in the exact same config as before now.

PR #1246, opened 2025-08-18
Tweaks to Ball parametrizations

mikeingold:
Very minor suggestion here for something that I noticed while debugging.

# Changes

- Update `b.center` to use `center(b)` API.
- Update `b.radius` to use `radius(b)` API.
- Move these single-use terms to their point of usage for clarity.

### DIFF
--- a/src/geometries/primitives/ball.jl
+++ b/src/geometries/primitives/ball.jl
@@ -41,12 +41,10 @@ function (b::Ball{𝔼{2}})(ρ, φ)
     throw(DomainError((ρ, φ), "b(ρ, φ) is not defined for ρ, φ outside [0, 1]²."))
   end
   T = numtype(lentype(b))
-  c = b.center
-  r = b.radius
-  ρ′ = T(ρ) * r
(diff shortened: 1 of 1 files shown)


juliohm:
Hi Michael! Is there a benefit in terms of performance or code generation?

We tried to follow the same pattern in other parametrizations of
geometries. I should be able to review by tomorrow.

Em seg., 18 de ago. de 2025, 18:58, codecov[bot] ***@***.***>
escreveu:

mikeingold:
> Hi Michael! Is there a benefit in terms of performance or code generation? We tried to follow the same pattern in other parametrizations of geometries. I should be able to review by tomorrow.

I wouldn't expect any significant performance change as the API methods are definitely simple enough for Julia's compiler inline the same code anyway. The primary change was switching from directly accessing struct fields to using the defined API methods for consistency with the other primitives source.

juliohm:
We need to make sure that these tweaks can be applied consistently across types.

PR #1247, opened 2025-08-18
Tweaks to ParaboloidSurface parametrization and methods

mikeingold:
Very minor suggestion here for something that I noticed while debugging.

# Changes

- Update `p.apex` to use `apex(p)` API.
- Update `p.focallength` to use `focallength(p)` API.
- Update `p.radius` to use `radius(p)` API.
- Move single-use terms to their point of usage for clarity.


juliohm:
@mikeingold I think we need a more comprehensive PR that makes these tweaks to all primitives and polytopes. That way we avoid introducing inconsistencies.

PR #1251, opened 2025-09-30
Add pairwise intersection between segments with sweep-line algorithm

souma4:
Upon reviewing the performance of comparative implementations, I decided to move from the very complicated and numerically unstable implementation in [#1168 ](https://github.com/JuliaGeometry/Meshes.jl/pull/1168) to a simpler interval sweepline (ie, we iterate over a sorted sequence by the furthest left x and check intersections if two segments have overlapping intervals).

This is equivalent to the Bentley-Ottmann algorithm in sparse cases and is O(N^2) in dense cases. However, in dense cases, Bentley-Ottmann computes many more intersections (to traverse the Binary search tree). These roughly balance out. In unofficial tests against a dense vector of segments, this implementation was the second fastest, only slower than crate::geo. 

Profiling shows most time is spent exclusively in runtime and GC dispatch within the `Intersection(...) do` block, so improving efficiency there would be valuable.

### DIFF
--- a/src/utils/intersect.jl
+++ b/src/utils/intersect.jl
@@ -52,7 +52,7 @@ function pairwiseintersect(segments; digits=_digits(segments))
 
   # sweepline algorithm
   n = length(segs)
-  P = eltype(first(segs))
+  P = eltype(vertices(first(segs)))
   D = Dict{P,Vector{Int}}()
   for i in 1:n
     for j in (i + 1):n


src/utils/intersect.jl
```diff
@@ -0,0 +1,111 @@
...
+# Licensed under the MIT License. See LICENSE in the project root.
+# ------------------------------------------------------------------
+
+"""
+    intervalsweep(segments; [digits])
+
+Compute pairwise intersections between n `segments`
+with `digits` precision in O(n⋅log(n+k)) time using
+an x-interval sweep line algorithm. Similar to an optimal
+Bentley-Ottmann algorithm in sparse systems, and closer to O(n²) in dense systems.
```
juliohm:
```suggestion
an x-interval sweep line algorithm. Similar to an
optimal Bentley-Ottmann algorithm in sparse systems,
and closer to O(n²) in dense systems.
```

src/utils/intersect.jl
```diff
@@ -0,0 +1,111 @@
+# ------------------------------------------------------------------
+# Licensed under the MIT License. See LICENSE in the project root.
+# ------------------------------------------------------------------
+
+"""
+    intervalsweep(segments; [digits])
```
juliohm:
```suggestion
    pairwiseintersect(segments; [digits])
```

src/utils/intersect.jl
```diff
@@ -0,0 +1,111 @@
...
+"""
+function intervalsweep(segments; digits=_digits(segments))
+  # orient segments and round coordinates
+  segs = map(segments) do seg
+    a, b = coordround.(extrema(seg), digits=digits)
+    a > b ? (b, a) : (a, b)
+  end
+  𝒮 = SweepLineQueue(segs)
+
+  points, seginds = handle!(𝒮; digits=digits)
```
juliohm:
```suggestion
  handle!(SweepLineQueue(segs); digits=digits)
```

src/utils/intersect.jl
```diff
@@ -0,0 +1,111 @@
...
+end
+
+# # compute the number of significant digits based on the segment type
+# # this is used to determine the precision of the points
```
juliohm:
```suggestion
# compute the number of significant digits based on the segment type
# this is used to determine the precision of the points
```

src/utils/intersect.jl
```diff
@@ -0,0 +1,111 @@
...
+    x₁, _ = CoordRefSystems.values(coords(seg[1]))
+    x₂, _ = CoordRefSystems.values(coords(seg[2]))
+    push!(intervals, SweepLineInterval(min(x₁, x₂), max(x₁, x₂), seg, i))
+  end
+  SweepLineQueue(sort!(intervals, by=s -> s.start))
+end
+
+function overlaps(i₁::SweepLineInterval, i₂::SweepLineInterval)
+  i₁.start ≤ i₂.start && i₂.stop ≥ i₁.start
+end
```
juliohm: This function belongs to the `SweepLineInterval` interface. Please move it up to near the corresponding struct.

src/utils/intersect.jl
```diff
@@ -0,0 +1,111 @@
...
+
+function SweepLineQueue(segs::Vector{<:Tuple{Point,Point}})
+  intervals = Vector{SweepLineInterval}()
+  for (i, seg) in enumerate(segs)
+    x₁, _ = CoordRefSystems.values(coords(seg[1]))
+    x₂, _ = CoordRefSystems.values(coords(seg[2]))
+    push!(intervals, SweepLineInterval(min(x₁, x₂), max(x₁, x₂), seg, i))
+  end
+  SweepLineQueue(sort!(intervals, by=s -> s.start))
+end
```
juliohm: It appears to me that `SweepLineQueue` is a simple "list" of intervals. Why not use a simple `Vector`? I don't see special behavior in the interface that justifies a new type.

src/utils/intersect.jl
```diff
@@ -0,0 +1,111 @@
...
+# ----------------
+
+"""
+  handle!(𝒮::SweepLineQueue)
+
+Iterate through the sweep line queue and compute all intersection points
+between overlapping intervals. Returns a tuple of intersection points and
+the sets of segment indices that intersect at each point.
+"""
+function handle!(𝒮::SweepLineQueue; digits=10)
```
juliohm: This function could have a more specific name and could operate directly on a `Vector` of intervals as explained above. We only introduce types when new behavior is needed. From what I can tell you are only using the `Vector` interface in `SweepLineQueue`.

test/utils.jl
```diff
@@ -80,6 +80,100 @@ end
   @inferred Meshes.coordround(p₁, digits=10)
 end
 
+@testitem "intervalsweep" setup = [Setup] begin
```
```suggestion
@testitem "pairwiseintersect" setup = [Setup] begin
```

juliohm (changes requested):
Thank you @souma4! First round of review attached. This one should be much easier to get ready for a merge.

src/utils/intersect.jl
souma4: That's fair. My main thinking was in case we want to reuse the `handle!` function. But based on the below feedback, there isn't a need to.

src/utils/intersect.jl
```diff
@@ -0,0 +1,104 @@
+# ------------------------------------------------------------------
+# Licensed under the MIT License. See LICENSE in the project root.
+# ------------------------------------------------------------------
+
+"""
+    pairwiseintersect(segments; [digits])
+
+Compute pairwise intersections between n `segments`
+with `digits` precision in O(n⋅log(n+k)) time using
+an x-interval sweep line algorithm. Similar to an optimal
```
juliohm:
```suggestion
a sweep line algorithm. Similar to an optimal
```

src/utils/intersect.jl
```diff
@@ -0,0 +1,104 @@
...
+struct SweepLineInterval{T<:Number}
+  start::T
+  stop::T
+  segment::Any
+  index::Int
+end
+
+function overlaps(i₁::SweepLineInterval, i₂::SweepLineInterval)
+  i₁.start ≤ i₂.start && i₁.stop ≥ i₂.start
+end
```
juliohm: Do we really need this data structure?

The new algorithm seems very simple:

For each segment, do
1. Compute start and save in a vector `starts`
2. Compute stop and save in a vector `stops`

Sort vectors `segs`, `starts` and `stops` in terms of `sortperm(starts)`

For each pair of segments, do
1. Fast check with `starts` and `stops`
2. If check is true, call expensive intersection
3. Otherwise, advance the loop

If we simply implement this algorithm with raw vectors `segs`, `starts` and `stops`, it will be probably be more readable and type stable.

Also, the function `sweep1d` can probably be merged into the main `pairwiseintersect` function after that. The only part of the code that deserves encapsulation is the kernel of the loop that uses `starts`, `stops` for a fast check and only then computes intersections with `segs`.

juliohm: Alternatively, if you feel that the readability with a single sort is superior, then use a simple vector of tuples `preproc = [(start, stop, seg), ...]` and `sort(preproc, by=first)`.

When you introduce a data structure like `SweepLineInterval` and use its fields in a separate function (e.g., `by=x->x.start`), you are accessing internals.

souma4: We really don't. The main idea was to have everything tied to the same object, but this way works well. It's actually significantly faster because it is effectively a struct of arrays rather than array of structs. It's now the fastest test I've run thus far.

src/utils/intersect.jl
```diff
@@ -0,0 +1,90 @@
...
+  segs = map(segments) do seg
+    a, b = coordround.(extrema(seg), digits=digits)
+    a > b ? (b, a) : (a, b)
+  end
+
+  starts, stops = map(segs) do seg
+    x₁, _ = CoordRefSystems.values(coords(seg[1]))
+    x₂, _ = CoordRefSystems.values(coords(seg[2]))
+    (x₁, x₂)
+  end |> x -> (first.(x), last.(x))
```
juliohm: Please split this into two simple list comprehensions or map calls to avoid intermediate allocations.

src/utils/intersect.jl
```diff
@@ -0,0 +1,90 @@
...
+  # sort indices based on start x coordinates
+  sortedindices = sortperm(starts)
+  # reorder everything based on sorted indices
+  starts, stops, segs = getindex.((starts, stops, segs), Ref(sortedindices))
```
juliohm: Here you can simply call `getindex` 3 times for clarity:

```julia
starts = starts[inds]
stops = stops[inds]
segs = segs[inds]
```

src/utils/intersect.jl
```diff
@@ -0,0 +1,90 @@
...
+  # keep track of original indices
+  n = length(segs)
+  oldindices = collect(1:n)[sortedindices]
```
juliohm: No need to collect the range. Issues with performance?

src/utils/intersect.jl
```diff
@@ -0,0 +1,90 @@
...
+
+  # primary sweepline algorithm
```
juliohm:
```suggestion
  # sweepline algorithm
```

src/utils/intersect.jl
```diff
@@ -0,0 +1,90 @@
...
+  𝐺 = Dict{Point,Vector{Int}}()
```
juliohm: This type is not concrete. Please use `typeof` if necessary. Is `Dict` the best data structure for this loop?

src/utils/intersect.jl
```diff
@@ -0,0 +1,90 @@
...
+        continue
+      else
+        _addintersection!(𝐺, I, oldindices[i], oldindices[k]; digits=digits)
+      end
+    end
+  end
+  (collect(keys(𝐺)), collect(values(𝐺)))
+end
+
+overlaps(startᵢ, stopᵢ, startₖ) = (startᵢ ≤ startₖ ≤ stopᵢ)
```
juliohm:
```suggestion
_overlaps(startᵢ, stopᵢ, startₖ) = (startᵢ ≤ startₖ ≤ stopᵢ)
```

src/utils/intersect.jl
souma4: doh, you're right, `Point` isn't concrete.  Easy fix. that simple fix improved the speed 2x 😮 

I argue that Dict is the best data structure largely because what we want is a set of unique points and a set of corresponding segment indices for each point. In the case of more than two lines intersecting at a point, the dictionary allows us to easily check if that point already exists, then we can just append new indices to the indices.

`Vector` isfeasible, but they are a bit messier to handle with little to no speed improvement. That's largely because if we want unique keys in a `Vector`, we still have to find which indices are duplicated (a large batch membership equality check) then append segment indices as needed. 

`Set` wouldn't work because the order of the set may be different from the segment indices, thus losing which indices belong to which intersection point.

Now, a true "graph" would probably be optimal, but adds the complexity of implementing said system.

src/utils/intersect.jl
souma4: my brain weirdly thought one could not index a `UnitRange`.

src/utils/intersect.jl
souma4: I did a check on if we could derive a graph from this output easily, and yes we can. We can build adjacency and segment-points (reverse of current point-segments) graphs in 50% of the time as creating the line segment intersections (in a dense case).

src/utils/intersect.jl
```diff
@@ -0,0 +1,90 @@
...
+By default, set `digits` based on the absolute
+tolerance of the length type of the segments.
+
+## References
+
+* Bentley & Ottmann 1979. [Algorithms for reporting and counting
+  geometric intersections](https://ieeexplore.ieee.org/document/1675432)
+"""
+function pairwiseintersect(segments; digits=_digits(segments))
+  P = typeof(vertices(first(segments))[1])
```
juliohm: Please define `P` as a function of `segs` as it is the actual variable used in the algorithm below.

src/utils/intersect.jl
```diff
@@ -0,0 +1,90 @@
...
+_overlaps(startᵢ, stopᵢ, startₖ) = (startᵢ ≤ startₖ ≤ stopᵢ)
+
+function _checkintersection(i, k, starts, stops, segs)
+  overlap = _overlaps(starts[i], stops[i], starts[k])
+  overlap || return :break
+  intersection(Segment(segs[i]), Segment(segs[k])) do 𝑖
+    t = type(𝑖)
+    (t === Crossing || t === EdgeTouching) ? get(𝑖) : :continue
+  end
+end
```
juliohm: Can we move this logic to inside the main loop? I think the code is simple enough to be pasted there without the `:break` and `:continue` symbols as return type.

src/utils/intersect.jl
```diff
@@ -0,0 +1,85 @@
...
+  stops = [CoordRefSystems.values(coords(seg[2]))[1] for seg in segs]
+  # sort indices based on start x coordinates
+  inds = sortperm(starts)
+  # reorder everything based on sorted indices
+  starts = starts[inds]
+  stops = stops[inds]
+  segs = segs[inds]
+  # keep track of original indices
+  n = length(segs)
+  oldindices = (1:n)[inds]
```
juliohm: Isn't this equal to `inds`? How could we sort `1:n` according to  `inds` and not get `inds` as the result?

src/utils/intersect.jl
```diff
@@ -0,0 +1,85 @@
...
+        t = type(𝑖)
+        (t === Crossing || t === EdgeTouching) ? get(𝑖) : nothing
+      end
+
+      # continue if no intersection, add intersection otherwise
+      if isnothing(I)
+        continue
+      else
+        _addintersection!(𝐺, I, oldindices[i], oldindices[j]; digits=digits)
+      end
```
juliohm: This is overly complicated...

Can you please replace all these variable assignments with the natural implementation:

```julia
for i in eachindex(segs)
  for j in (i+1):length(segs)
    _overlaps(...) || break
    
    intersection(...) do I
      if type(I) == Crossing || type(I) == EdgeTouching
        # add intersection
      end
    end
  end
end
```

src/utils/intersect.jl
souma4: It actually is equivalent 🤦 . I think I had it in my head that sortperm wasn't sortperm, the indices for the sorted order, which is what is used for old indices. thanks as always for catching these.

src/utils/intersect.jl
```diff
@@ -0,0 +1,80 @@
...
+  ℒ = lentype(s)
+  τ = ustrip(eps(ℒ))
+  round(Int, 0.8 * (-log10(τ))) # 0.8 is a heuristic to avoid numerical issues
+end
+
+# add an intersection point to the dictionary with segment indices
+function _addintersection!(D, p::Point, i::Int, j::Int)
+  if haskey(D, p)
+    append!(D[p], (i, j))
+    unique!(D[p])
```
juliohm: Do we really need `unique!` here? Doesn't the loop `i in 1:n && j in (i+1):n` eliminate double visits to pairs `(i, j)`?

juliohm: Nevermind, the values of the dictionary are vectors of integers, not vectors of pairs of integers.

juliohm: If you need uniqueness, the most natural choice is a `Set{Int}`. Could you please benchmark the current version against a version where `D = Dict{P,Set{Int}}()`? The version with sets doesn't need the `unique!` call.

If the output needs to be ordered, then `Vector{Int}` is the best option.

juliohm: Here is a sketch implementation with `Set`:

```julia
          if haskey(D, p)
            push!(D[p], inds[i])
            push!(D[p], inds[j])
          else
            D[p] = Set([inds[i], inds[j]])
          end
```

souma4: yeah prior commits had `Set{Int}` and I was finding a ton of GC occuring within the instantiation of Sets (taking up about a quarter of `@profview`). I think they use dictionaries under the hood which defaults the `sizehint!` (preallocated memory, basically) to a number much larger than the number of expected set objects (it was using a hint of 15).

I dug into this a bit more and found that for small datasets where segments are being added to existing points (our FP64 test), the `Set` implementation is faster. But for a large dataset where every intersection is n=2 segments, the vector approach was notably faster. This tells me that Set's will work best for cases where we need to `union!` existing intersection entries. But if there is a lot of new entries that aren't revisited, then `Vector` works better.

So, this is a tradeoff we can pick. I lean vector because I can't think of any particular geospatial datasets right now where we'd expect many segments for each intersection. More CAD centered problems, though, probably would have more of those.

juliohm: Thank you for double checking it. I will refactor the code once more to clean things up. We could then move the `unique!` call to the end after the loop.

src/utils/intersect.jl
```diff
@@ -0,0 +1,90 @@
+# ------------------------------------------------------------------
+# Licensed under the MIT License. See LICENSE in the project root.
+# ------------------------------------------------------------------
+
+"""
+    pairwiseintersect(segments; [digits])
+
+Compute pairwise intersections between n `segments`
+with `digits` precision in O(n⋅log(n+k)) time using
```
juliohm: Need to define `k` in the docstring

test/utils.jl
```diff
@@ -80,6 +80,100 @@ end
...
+    Segment(cart(0, 0), cart(2, 2)),
+    Segment(cart(3, 1), cart(3, 3)),
+    Segment(cart(3, 3), cart(3, 3))
+  ]
+  points, seginds = sortedintersection(segs)
+  @test length(points) == 1
+
+  # inference test
+  segs = facets(cartgrid(10, 10))
+  @inferred Nothing (Meshes.pairwiseintersect(segs))
```
juliohm: I think we should return `Nothing` from this inferred test, right?

src/utils/intersect.jl
```diff
@@ -0,0 +1,91 @@
...
+
+  # sort segments based on first coordinates
+  inds = sortperm(xₛ)
+  segs = segs[inds]
+  xₛ = xₛ[inds]
+  xₑ = xₑ[inds]
+
+  # sweepline algorithm
+  n = length(segs)
+  P = eltype(first(segs))
```
souma4: Are we sure this should be `eltype`? I thought `eltype` was always `Any` for segments. I also realize on reviewing this again, we're doing this operation on the first Segment (L38), which is made up of `Point` types, and we want to add `Point` types to the dictionary. So maybe we need to do something like `P = eltype(pointify(first(segs)))` or any other permutation that splits down to points.

juliohm: Sorry, I thought we had `eltype` implemented for `Geometry`. It is only implemented for `Domain`.

Will adjust the line. Thank you for catching this!

src/utils/intersect.jl
```diff
@@ -0,0 +1,91 @@
...
+      end
+    end
+  end
+
+  # remove duplicate indices
+  for p in keys(D)
+    unique!(D[p])
+  end
+
+  collect(keys(D)), collect(values(D))
```
juliohm, line 81: Trying to think if the output format is the most convenient for users of this utility function. I believe it is.

Say we are implementing a new clipping algorithm. We could call this function in the beginning of the algorithm and then access the relevant indices of segments to connect the dots, right?

souma4: That was my thinking. Because from these we know which points associate with which segments, which is a significant piece of what is needed. Plus it's very easy to build the segment-to-intersections graph from this if needed. The only thing I can think of that might be useful is if we build this, but also the segment-to-intersections graph at the same time, but that leads to so many outputs with significant overlap, plus it would assuredly be slower because it would be duplicating the "is this segment in the dictionary, are these points all unique?" checks.

juliohm (approved):
There we go! 🎉 

It is a good exercise to compare the first commit of this PR with the final version to gain experience with idiomatic Julia/Meshes.jl code. The simpler, the better 🥇

PR #1256, opened 2025-10-09
Widen vector type in GeometrySet

juliohm:
Fix https://github.com/JuliaEarth/GeoStats.jl/issues/551

This is a delicate change. Appreciate if folks could share their thoughts.


juliohm:
I was just wondering if type inference or performance could be affected by this change. Our benchmarks don't cover it well, but I think we can go ahead and fix the bug.

PR #1259, opened 2025-10-12
Implementing isconvex(Hexahedron)

souma4:
Quick `isconvex(Hexahedron)` implementation ([ #1205](https://github.com/JuliaGeometry/Meshes.jl/issues/1205) ). I believe with the current implementation of `Hexahedron` it's impossible to get a globally concave polygon with only local convexity, thus I only check convexity of the faces rather than also computing the centroid and checking if any vertices are on the other side of that centroid (global convexity). If we find a case that requires global convexity, I have the code to add in easily.

### DIFF
--- a/test/predicates.jl
+++ b/test/predicates.jl
@@ -247,7 +247,7 @@ end
   @test !isconvex(poly2)
   poly = PolyArea(cart.([(0, 0), (1, 0), (1, 1), (0.5, 0.5), (0, 1)]))
   @test !isconvex(poly)
-  hexa1 = Hexahedron(
+  h = Hexahedron(
     cart(0, 0, 0),
     cart(1, 0, 0),
     cart(1, 1, 0),
@@ -257,7 +257,8 @@ end
     cart(1, 1, 1),
     cart(0, 1, 1)
   )
-  hexa2 = Hexahedron(
+  @test isconvex(h)
+  h = Hexahedron(
     cart(0, 0, 0),
     cart(1, 0, 0),
     cart(0.9, 0.9, 0),
@@ -267,7 +268,8 @@ end
     cart(1, 1, 1),
     cart(0, 1, 1)
   )
-  hexa3 = Hexahedron(
+  @test !isconvex(h)
+  h = Hexahedron(
     cart(0, 0, 0),
     cart(1, 0, 0),
     cart(1, 1, 0),
@@ -277,7 +279,8 @@ end
     cart(0.75, 0.26, 0.2),
     cart(0.25, 0.26, 0.2)
   )
-  hexa4 = Hexahedron(
+  @test isconvex(h)
+  h = Hexahedron(
     cart(0, 0, 0),
     cart(1, 0, 0),
     cart(1, 1, 0),
@@ -287,7 +290,8 @@ end
     cart(0.75, 0.25, 0.21),
     cart(0.25, 0.25, 0.21)
   )
-  hexa5 = Hexahedron(
+  @test !isconvex(h)
+  h = Hexahedron(
     cart(0, 0, 0.1),
     cart(1, 0, 0.1),
     cart(1, 1, 0.1),
@@ -297,7 +301,8 @@ end
     cart(1, 1, 0.3),
     cart(0, 1, 0.2)
   )
-  hexa6 = Hexahedron(
+  @test !isconvex(h)
+  h = Hexahedron(
     cart(0, 0, 0),
     cart(1, 0, 0),
     cart(1, 1, 0),
@@ -307,15 +312,22 @@ end
     cart(1, 1, -1),
     cart(0, 1, -1)
   )
-  @test isconvex(hexa1)
-  @test !isconvex(hexa2)
-  @test isconvex(hexa3) # symmetric deformation
-  @test !isconvex(hexa4) # tiny non-symmetric deformation
-  @test !isconvex(hexa5)
-  @test !isconvex(hexa6)
-  for transform in [Affine(AngleAxis(0.75, 1.0, 0.5, 0.25), [-5, 2, 2]), Stretch(1.0, 2.0, 5.0)]
-    @test isconvex(hexa3 |> transform)
-  end
+  @test !isconvex(h)
+  h0 = Hexahedron(
+    cart(0, 0, 0),
+    cart(1, 0, 0),
+    cart(1, 1, 0),
+    cart(0, 1, 0),
+    cart(0.25, 0.25, 0.2),
+    cart(0.75, 0.25, 0.2),
+    cart(0.75, 0.26, 0.2),
+    cart(0.25, 0.26, 0.2)
+  )
+  h1 = h0 |> Affine(AngleAxis(T(0.75), T(1.0), T(0.5), T(0.25)), T[-5, 2, 2])
+  h2 = h0 |> Stretch(T(1.0), T(2.0), T(5.0))
+  @test isconvex(h0)
+  @test isconvex(h1)
+  @test isconvex(h2)
 end
 
 @testitem "issubset" setup = [Setup] begin


src/predicates/isconvex.jl
```diff
@@ -76,3 +76,16 @@ function isconvex(q::Quadrangle)
...
+
+  # Check all faces are convex and consistently oriented
+  for face in msh
+    if !isconvex(face)
+      return false
+    end
+  end
+
+  true
+end
```
juliohm:
```suggestion
isconvex(h::Hexahedron) = all(isconvex, boundary(h))
```

souma4: I thought I was missing something trivial 🤦 `all` duh. Thank you for the second check

souma4:
@juliohm So the two conditions are that 1. each face is convex (coplanar, each vertex lies on the convex hull) 2. that any two points within the polyhedra connect without passing through a surface of the polyhedron. 

If any of the vertices of a regular `Hexahedron` (cube) are swapped, a face becomes concave. Thus the only change would be a regular and single axis symmetric change. The below hexahedron should have all coplanar faces because I just translate the (1,0,1)-(1,1,1) edge to beyond the midpoint of the cube, thus the (1,0,0)-(1,1,0)  vertices should pass through that surface when connected to (0,0,1)-(0,1,1). The current implementation checks it as not convex though, so it is capturing that global property.

```julia
  hexa1 = Hexahedron(
    cart(0, 0, 0),
    cart(1, 0, 0),
    cart(1, 1, 0),
    cart(0, 1, 0),
    cart(0, 0, 1),
    cart(0.3, 0, 0.3),
    cart(0.3, 1, 0.3),
    cart(0, 1, 1)
  )
```

juliohm:
Thank you for clarifying @souma4. I think we should append the necessary condition (2) after the condition (1):

```julia
isconvex(hex) = all(isconvex, boundary(hex)) && condition2(hex)
```

The && operator is a short-circuit, meaning that `condition2` will be skipped when condition (1) is true. So no impact in performance.

src/predicates/isconvex.jl
```diff
@@ -76,3 +76,18 @@ function isconvex(q::Quadrangle)
...
+  vs = vertices(h)
+  for i in eachindex(vs), j in (i + 1):length(vs)
+    p₁, p₂ = vs[i], vs[j]
+    d = Segment(p₁, p₂)
+    # if any segment between two vertices intersects a face other
+    # than the two faces containing the vertices, it's not convex
+    any(intersects(d, e) && !(p₁ ∈ e || p₂ ∈ e) for e in boundary(h)) && return false
+  end
+  true
+end
```
juliohm: What is the minimum (necessary) number of intersection checks? Are you implementing this condition based on some reference paper?

Please give this auxiliary function a more descriptive name. If you are using the `boundary` inside this second condition, then you probably want to refactor the main function to only compute the boundary once, and use it in two internal checks.

souma4: I don't believe there is a minimum number because all it takes is one messed up polyhedron to make it non-convex. The reference is ["Checking the convexity of polytopes and the planarity of subdivisions"](https://www.sciencedirect.com/science/article/pii/S092577219800039X) (I have pdf if needed). There's a cleaner description at [this location](https://liam.flookes.com/cs/geo/index.html).  Their way is for generic polyhedrons, so I've simplified my implementation because of how our hexahedrons are structured. It wouldn't be a heavy lift to implement the generic method but that's for later

Edit: I actually whipped up what I believe is a correct implementation of their algorithm using topology and its 100x faster than this brute force approach.

raman-maker:
I think you should discuss it with some Mathematician on geometry by sending email to him about convexity conditions for different possibilities.

src/predicates/isconvex.jl
```diff
@@ -76,3 +78,33 @@ function isconvex(q::Quadrangle)
...
+
+function _isgloballyconvex(b)
+  t = convert(HalfEdgeTopology, topology(b))
+  # map segments to their adjacent faces
+  𝒞₁₂ = Coboundary{1,2}(t)
+  r = 0
+  for edgeᵢ in 1:nfacets(t)
+    adjfaces = 𝒞₁₂(edgeᵢ)
+    length(adjfaces) == 2 || return false
```
juliohm: This condition is always true right? Every edge in a hexahedron shares exactly 2 quadrangles.

src/predicates/isconvex.jl
```diff
@@ -76,3 +78,33 @@ function isconvex(q::Quadrangle)
...
+    # normal faces
+    n₁, n₂ = map(adjfaces) do faceᵢ
+      face = element(b, faceᵢ)
+      verts = vertices(face)
+      n = length(verts)
+      n ≥ 3 ? normal(Plane(verts[1], verts[2], verts[3])) : throw(ArgumentError("Face must have at least 3 vertices"))
```
juliohm: Every quadrangle face has exactly 4 vertices. Why the n >= 3 check?

juliohm: If you are writing a general implementation for arbitrary Polyhedron, not just Hexahedron, then we better split the methods. The method for Hexahedron should be straightforward because the faces are always quadrangles.

My suggestion is to start with the simplest, most efficient implementation for Hexahedron, and then consider a separate PR for the general Polyhedron case.

souma4: This is what I've been converging towards. `isconvex` for a hexahedron, based on how we build them, I'm 99% sure can be done with the single `all(isconvex, boundary(h))` check. The second these hexahedrons are non-symmetrically deformed, or an indent is created, the face becomes curved. And if there's a self intersection, the face polygons self intersect or curve, so it gets those as well. When I've tried to account for a curving face by adjusting other vertices, I get a tetrahedra. I've updated tests to reflect the main ways nonconvexity can be achieved.

This other checking step (using edges and faces to determine global convexity) would be better suited for `Polyhedron` generics.

src/predicates/isconvex.jl
```diff
@@ -64,6 +64,8 @@ _isconvex(p::Polygon, ::Val{2}) = Set(eachvertex(convexhull(p))) == Set(eachvert
 
 _isconvex(p::Polygon, ::Val{3}) = isconvex(proj2D(p))
 
+isconvex(h::Hexahedron) = all(isconvex, boundary(h))
```
juliohm, line 67: Consider a box where the top face is flipped. Isn't this a counter-example where all faces are convex, but the hexahedron is non-convex? If there is a theorem for general Polyhedron, it should apply to Hexahedron. The second condition you erased should just be adjusted to the simpler case?

My comment was about adding checks about number of vertices and faces per edge. These are known and fixed in the case of Hexahedron.

souma4: So that only has convex faces if it is triangulated, but since we use quadrangles, what we get is the second image below where the (in the image) top left vertex connects to bottom right, which is a self intersection and therefore not convex.
image
image

juliohm: You are right. Forgot about the walls of the cube.

juliohm:
Fix #1205

juliohm:
@JoshuaLampert our benchmark suite is failing on this PR for some reason. Could it be related to some recent update on Julia v1.12 or AirspeedVelocity.jl?

JoshuaLampert:
> @JoshuaLampert our benchmark suite is failing on this PR for some reason. Could it be related to some recent update on Julia v1.12 or AirspeedVelocity.jl?

Good question. I don't know. I haven't seen this issue yet, but it could be a julia v1.12 thing. But I also don't know why the lts benchmark action gets canceled seemingly without an error. Maybe open a dummy PR to change https://github.com/JuliaGeometry/Meshes.jl/blob/e72f2fd9c75e6a63357c4517cee8e0211c84a97a/.github/workflows/Benchmarks.yml#L15 into using julia v1.11 to verify it's v1.12?

MilesCranmer:
Quick note on the CI failure (sorry) - it seems like it's due to this change to Pkg.jl: https://github.com/JuliaLang/Pkg.jl/issues/4422.

PR #1261, opened 2025-10-28
Replace LScene by Axis3

juliohm:
Attached is the impact of switching from `LScene` to `Axis3`. The former has better defaults and the latter has more features (e.g., set title, labels, interactive).


juliohm:
Closing in favor of #1357

PR #1262, opened 2025-10-28
add point to segment and segment to segment evaluate

souma4:
Implementation for `evaluate(d, point, segment)` and `evaluate(d, segment, segment)`. previously we were using infinite `Line`, this simply clamps the projection matrix to [0-1].  There's a good bit of repeated code, so maybe there's a more clever way to refactor this.

### DIFF
--- a/src/distances.jl
+++ b/src/distances.jl
@@ -41,11 +41,11 @@ function evaluate(d::Euclidean, l₁::Line, l₂::Line)
   λ₁, λ₂, r, rₐ = intersectparameters(l₁(0), l₁(1), l₂(0), l₂(1))
 
   if (r == rₐ == 2) || (r == rₐ == 1)  # lines intersect or are colinear
-    return zero(result_type(d, lentype(l₁), lentype(l₂)))
+    zero(result_type(d, lentype(l₁), lentype(l₂)))
   elseif (r == 1) && (rₐ == 2)  # lines are parallel
-    return evaluate(d, l₁(0), l₂)
+    evaluate(d, l₁(0), l₂)
   else  # get distance between closest points on each line
-    return evaluate(d, l₁(λ₁), l₂(λ₂))
+    evaluate(d, l₁(λ₁), l₂(λ₂))
   end
 end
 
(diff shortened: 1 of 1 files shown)


src/distances.jl
```diff
@@ -35,6 +49,25 @@ function evaluate(d::Euclidean, l₁::Line, l₂::Line)
...
+
+Evaluate the minimum Euclidean `distance` between `segment₁` and `segment₂`.
+"""
+function evaluate(d::Euclidean, s₁::Segment, s₂::Segment)
+  λ₁, λ₂, r, rₐ = intersectparameters(s₁(0), s₁(1), s₂(0), s₂(1))
+  λ₁ = clamp(λ₁, 0, 1)
+  λ₂ = clamp(λ₂, 0, 1)
+
+  if (r == rₐ == 1)  # lines intersect or are colinear
+    return zero(result_type(d, lentype(s₁), lentype(s₂)))
```
juliohm: Two collinear segments be shifted along the line leading to non-zero distance:

a-----b   c-----d

The distance in the above example is norm(c-b).

src/distances.jl
```diff
@@ -35,6 +49,25 @@ function evaluate(d::Euclidean, l₁::Line, l₂::Line)
...
+  elseif (r == 1) && (rₐ == 2)  # lines are parallel
+    return evaluate(d, s₁(0), s₂)
```
juliohm: In this case we could take the minimum between two pairs of points: (s1(0), s2(0)) and (s1(0), s2(1)). Would that be more efficient than computing the distance between s1(0) and s2?

src/distances.jl
souma4: ah you're right. I forgot about shifts of those types.

src/distances.jl
souma4: So I've done some testing here and in order to get correct results we'd also need to do `(s1(1), s2(0)), (s1(1), s2(1))`.  Timing wise for the `min...` option and the `(s1(0), s2), (s1(1), s2))` option, these end up in the same ~230 ns realm.

juliohm (approved):
Awesome stuff! 💯

PR #1266, opened 2025-11-03
implement isconvex for Pyramid

### LINKED ISSUE #1265: Missing `isconvex` method for `Pyramid`
I was looking at some pyramids and tried to test if they intersect with other geometries. Currently it throws error with missing method for `insconvex`:

```julia
julia> pir = Pyramid((0,0,0),(1,0,0),(1,1,0),(0,1,0),(0,0,1));

julia> b2d = Box((0.1, 0.1, 0.2), (0.2,0.2,0.2));

julia> intersects(b2d, pir)
ERROR: MethodError: no method matching isconvex(::Pyramid{CoordRefSystems.Cartesian3D{CoordRefSystems.NoDatum, Unitful.Quantity{…}}, 𝔼{3}})
The function `isconvex` exists, but no method is defined for this combination of argument types.

Closest candidates are:
  isconvex(::Triangle)
   @ Meshes C:\Users\csert\.julia\dev\Meshes\src\predicates\isconvex.jl:57
  isconvex(::Quadrangle)
   @ Meshes C:\Users\csert\.julia\dev\Meshes\src\predicates\isconvex.jl:75
  isconvex(::Box)
   @ Meshes C:\Users\csert\.julia\dev\Meshes\src\predicates\isconvex.jl:35
  ...

Stacktrace:
 [1] intersects(g₁::Box{𝔼{3}, CoordRefSystems.Cartesian3D{CoordRefSystems.NoDatum, Unitful.Quantity{Float64, 𝐋, Unitful.FreeUnits{(m,), 𝐋, nothing}}}}, g₂::Pyramid{CoordRefSystems.Cartesian3D{CoordRefSystems.NoDatum, Unitful.Quantity{Float64, 𝐋, Unitful.FreeUnits{(m,), 𝐋, nothing}}}, 𝔼{3}})
   @ Meshes C:\Users\csert\.julia\dev\Meshes\src\predicates\intersects.jl:79
 [2] top-level scope
   @ REPL[6]:1
```

The easiest solution for me seems to be copying the implementation from `Hexahedron`:

```julia
isconvex(p::Pyramid) = all(isconvex, boundary(p))
```

With this change, it works:

```julia
julia> intersects(b2d, pir)
true
```

I can construct a concave pyramid, which I guess doesn't make sense, but the convexity is correct:
```julia
julia> pir2 = Pyramid((0,0,0),(1,0,0),(0.25,0.25,0),(0,1,0),(0,0,1))
Pyramid
├─ Point(x: 0.0 m, y: 0.0 m, z: 0.0 m)
├─ Point(x: 1.0 m, y: 0.0 m, z: 0.0 m)
├─ Point(x: 0.25 m, y: 0.25 m, z: 0.0 m)
├─ Point(x: 0.0 m, y: 1.0 m, z: 0.0 m)
└─ Point(x: 0.0 m, y: 0.0 m, z: 1.0 m)

julia> isconvex(pir2)
false
```

Would that be a correct implementation? I don't understand this topic well enough to see this right away.

cserteGT3:
This PR implements a specialized version for `isconvex` for `Pyramid` (as discussed in #1265). It assumes that the first four vertices construct the base.

Fix #1265

### DIFF
--- a/test/predicates.jl
+++ b/test/predicates.jl
@@ -328,21 +328,9 @@ end
   @test isconvex(h0)
   @test isconvex(h1)
   @test isconvex(h2)
-  p = Pyramid(
-    cart(0, 0, 0),
-    cart(1, 0, 0),
-    cart(1, 1, 0),
-    cart(0, 1, 0),
-    cart(0, 0, 1)
-  )
+  p = Pyramid(cart(0, 0, 0), cart(1, 0, 0), cart(1, 1, 0), cart(0, 1, 0), cart(0, 0, 1))
   @test isconvex(p)
-  p = Pyramid(
-    cart(0, 0, 0),
-    cart(1, 0, 0),
-    cart(0.25, 0.25, 0),
-    cart(0, 1, 0),
-    cart(0, 0, 1)
-  )
+  p = Pyramid(cart(0, 0, 0), cart(1, 0, 0), cart(0.25, 0.25, 0), cart(0, 1, 0), cart(0, 0, 1))
   @test !isconvex(p)
 end
 


src/geometries/polytopes/pyramid.jl
```diff
@@ -11,6 +11,11 @@ A pyramid with points `p1`, `p2`, `p3`, `p4`, `p5`.
 
 nvertices(::Type{<:Pyramid}) = 5
 
+function base(p::Pyramid)
+  a, b, c, d, o = vertices(p)
```
juliohm:
```suggestion
  a, b, c, d, o = p.vertices
```

Given that this is a function/accessor implemented in the same file of the geometry, we tend to use internal field access. There is no clear rule, but I am trying to follow the pattern in other polytope implementations.

cserteGT3: I've updated the file. This line, and another function as well.

cserteGT3:
> Thank you @cserteGT3! Should we export the new `base` function?

It's already exported. I chose the `base` name as it is used by `Cone`.

juliohm:
Could you please add some tests for the new `base` of `Pyramid`? That should complete the PR 💯

src/geometries/polytopes/pyramid.jl
```diff
@@ -20,7 +25,7 @@ function (pyramid::Pyramid)(u, v, w)
   if (u < 0 || u > 1) || (v < 0 || v > 1) || (w < 0 || w > 1)
     throw(DomainError((u, v, w), "pyramid(u, v, w) is not defined for u, v, w outside [0, 1]³."))
   end
-  a, b, c, d, o = vertices(pyramid)
+  a, b, c, d, o = p.vertices
   q = Quadrangle(a, b, c, d)
```
JoshuaLampert:
```suggestion
  q = base(p)
```
Wouldn't it be cleaner to reuse `base` here?

src/geometries/polytopes/pyramid.jl
JoshuaLampert, line 8:
```suggestion
A pyramid with points `p1`, `p2`, `p3`, `p4`, `p5`. The first four points correspond
to the vertices of the base and the last is the top of the pyramid.
```
Now that we assume this, I would like to clarify this also in the docstring.

src/Meshes.jl
```diff
@@ -226,6 +226,7 @@ export
   angles,
   innerangles,
   normal,
+  base,
```
JoshuaLampert, line 229: Now `base` is `export`ed twice, right? In line 190 and here. Is that intended?

juliohm: Thank you for checking @JoshuaLampert. Yes, we try to export all interface functions on each section for consistency to avoid erasing `base` from `Cone` one day and affecting `Pyramid` unintentionally.

There are multiple exports of the same name in a few places.

src/geometries/polytopes/pyramid.jl
juliohm: It has always been the case. I agree it deserves a better docstring.

src/Meshes.jl
JoshuaLampert: Ah ok, good to know.

juliohm:
Tests are failing due to some unrelated issue with DataStructures.jl 🍇

cserteGT3:
I've did similar updates, just added a comment on convexity and fixed a bug I introduced (using variable `p` instead of `pyramid`).

src/geometries/polytopes/pyramid.jl
```diff
@@ -6,11 +6,18 @@
     Pyramid(p1, p2, p3, p4, p5)
 
 A pyramid with points `p1`, `p2`, `p3`, `p4`, `p5`.
+The first four points correspond to the vertices of the base and the last is the top of the pyramid.
+Please note, that the current implementation doesn't handle non-convex bases correctly.
```
juliohm:
```suggestion

The first four points form the base and the last is the is the apex.
```

src/geometries/polytopes/pyramid.jl
```diff
@@ -6,11 +6,18 @@
     Pyramid(p1, p2, p3, p4, p5)
 
 A pyramid with points `p1`, `p2`, `p3`, `p4`, `p5`.
+The first four points correspond to the vertices of the base and the last is the top of the pyramid.
```
JoshuaLampert:
```suggestion
The first four points correspond to the vertices of the base and the last point is the apex..
```
I think Júlio's formulation using `apex` was more accurate than my suggestion using `top` 😄

JoshuaLampert: Ok, you beat me on it :sweat_smile:

src/geometries/polytopes/pyramid.jl
```diff
@@ -6,11 +6,18 @@
     Pyramid(p1, p2, p3, p4, p5)
 
 A pyramid with points `p1`, `p2`, `p3`, `p4`, `p5`.
+
+The first four points form the base and the last is the is the apex.
```
JoshuaLampert:
```suggestion
The first four points form the base and the last is the apex.
```
We are getting there... 😆

src/geometries/polytopes/pyramid.jl
cserteGT3: I agree, but this commit removed the non-convex note (which is fine if you don't agree with it).

juliohm: Notes that describe temporary issues shouldn't reach end-users in the docstring. Our docstrings would be too long otherwise :)

juliohm:
Added a new `apex` method for consistency. Now `Pyramid` has `base` and `apex` and both functions are used in the pyramid(u,v,w) implementation.

src/geometries/polytopes/pyramid.jl
```diff
@@ -6,22 +6,31 @@
...
+
 ==(p₁::Pyramid, p₂::Pyramid) = p₁.vertices == p₂.vertices
 
 Base.isapprox(p₁::Pyramid, p₂::Pyramid; atol=atol(lentype(p₁)), kwargs...) =
   all(isapprox(v₁, v₂; atol, kwargs...) for (v₁, v₂) in zip(p₁.vertices, p₂.vertices))
 
-function (pyramid::Pyramid)(u, v, w)
+function (p::Pyramid)(u, v, w)
   if (u < 0 || u > 1) || (v < 0 || v > 1) || (w < 0 || w > 1)
     throw(DomainError((u, v, w), "pyramid(u, v, w) is not defined for u, v, w outside [0, 1]³."))
```
juliohm:
```suggestion
    throw(DomainError((u, v, w), "p(u, v, w) is not defined for u, v, w outside [0, 1]³."))
```

juliohm:
@cserteGT3 could you please add some tests for the `base` and `apex`? Other than that, I think the PR is ready ✅

test/polytopes.jl
```diff
@@ -954,6 +954,7 @@ end
   equaltest(p)
   isapproxtest(p)
   vertextest(p)
+  @test base(p) == Quadrangle(cart(0, 0, 0), cart(1, 0, 0), cart(1, 1, 0), cart(0, 1, 0))
```
juliohm, line 957:
```suggestion
  @test base(p) == Quadrangle(cart(0, 0, 0), cart(1, 0, 0), cart(1, 1, 0), cart(0, 1, 0))
  @test apex(p) == cart(0, 0, 1)
```

JoshuaLampert:
> @cserteGT3 could you please add some tests for the `base` and `apex`? Other than that, I think the PR is ready ✅

Apart from running the formatter.

juliohm:
Just added the test on GitHub directly 👍🏽

src/geometries/polytopes/pyramid.jl
cserteGT3: Fair enough :)

juliohm:
> Apart from running the formatter.

Good point! @cserteGT3 you can press Ctrl+Shift+I to get the formatting right.

cserteGT3:
> > Apart from running the formatter.
> 
> Good point! @cserteGT3 you can press Ctrl+Shift+I to get the formatting right.

Thank you both for shaping this PR! Ran the tests locally, all passed.

Regarding the formatting: I don't see that it would be different, than other files. Running "Format document" in VSCode doesn't change anything for me in `pyramid.jl`, though I think it is not configured properly, so I have to look into it...

JoshuaLampert:
> > > Apart from running the formatter.
> > 
> > 
> > Good point! @cserteGT3 you can press Ctrl+Shift+I to get the formatting right.
> 
> Thank you both for shaping this PR! Ran the tests locally, all passed.
> 
> Regarding the formatting: I don't see that it would be different, than other files. Running "Format document" in VSCode doesn't change anything for me in `pyramid.jl`, though I think it is not configured properly, so I have to look into it...

test/predicates.jl is not formatted correctly, see https://github.com/JuliaGeometry/Meshes.jl/actions/runs/19064950241/job/54453184291?pr=1266#step:2:572.

cserteGT3:
> > > > Apart from running the formatter.
> > > 
> > > 
> > > Good point! @cserteGT3 you can press Ctrl+Shift+I to get the formatting right.
> > 
> > 
> > Thank you both for shaping this PR! Ran the tests locally, all passed.
> > Regarding the formatting: I don't see that it would be different, than other files. Running "Format document" in VSCode doesn't change anything for me in `pyramid.jl`, though I think it is not configured properly, so I have to look into it...
> 
> test/predicates.jl is not formatted correctly, see https://github.com/JuliaGeometry/Meshes.jl/actions/runs/19064950241/job/54453184291?pr=1266#step:2:572.

Thanks for the pointer! Figured out my VSCode setting and pushed the change.

juliohm:
Opened an issue in StatsBase.jl, which is calling DataStructures.jl: https://github.com/JuliaStats/StatsBase.jl/issues/974 I believe our only use of StatsBase.jl is for weighted sampling. We should seek alternatives to drop this dependency once and for all.

juliohm:
Thank you @cserteGT3 ! Another one for the list 💯 🙂

cserteGT3:
Thank you for the quick merge and the collaborative efforts!

Issue #534, opened 2023-05-21
Visualization of `Line`


Antoinemarteau:
Small bug : seems the infinite long line is not possible to discretize ?
```julia
using Meshes, MeshViz
using CairoMakie: activate!; activate!()
ln = Line(Point(1,1,1), Point(0,0,0))
viz(ln)
```
yields StackOverflowError

Stacktrace:
     [1] simplexify(geometry::Line{3, Float64})
       @ Meshes ~/.julia/packages/Meshes/1Iszy/src/discretization.jl:135
     [2] discretize(geometry::Line{3, Float64})
       @ Meshes ~/.julia/packages/Meshes/1Iszy/src/discretization.jl:103
--- the last 2 lines are repeated 39990 more times ---
 [79983] simplexify(geometry::Line{3, Float64})
       @ Meshes ~/.julia/packages/Meshes/1Iszy/src/discretization.jl:135

Got the same with GLMakie backend.

juliohm:
We do not know how to plot infinite geometries like lines yet. Any help is appreciated on a good approach. If Makie.jl has a trick to represent infinite lines and planes, please let us know and we can add the recipe here.

Antoinemarteau:
I have no idea of what is possible in Makie (I just wanted to report the bug), would you like that I try to do a PR to catch the `viz` calls on `Line` and throw a `TypeError` ?

juliohm:
No need for a PR to catch the error. Thanks!

Em ter., 23 de mai. de 2023 17:20, Antoine Marteau ***@***.***>
escreveu:

juliohm:
Relevant discussion: https://discourse.julialang.org/t/how-to-visualize-infinite-lines-with-makie-jl/100325

juliohm:
It is now possible to visualize infinite lines on the plane. The 3D case is still not solved.

PR #1269, opened 2025-11-07
Improvements to viz of 1D geometries and vectors

juliohm:
Uses `Makie.ablines` for diagonal lines and `Makie.vlines` for vertical lines embedded in 2D space.

See https://github.com/JuliaGeometry/Meshes.jl/issues/534 for the related issue.


juliohm:
We do our best to unify the names of keywords in Makie.jl, and use a generic `segmentsize` option to refer to the size of geometries with parametric dimension 1. The name of this option could be improved in the future though.

juliohm:
I refactored the code once more to convert the recipe of `Vec` to that of `Ray` to avoid code duplication.

Will wait tests and merge 👍🏽

PR #1270, opened 2025-11-20
Improving isconvex robustness

souma4:
The current version of `isconvex` uses the convex hull to determine convexity. This works for all polygons except for pathological self-intersecting cases (e.g. a square where the top two vertices are flipped, leading to an hourglass). The proposed fix uses the primary definition of convexity, which is that all interior angles are <= 180 degrees and that a chord bisecting the polygon will only intersect 2 sides (e.g. no holes). This new implementation ends up being significantly faster too, as a plus.


juliohm:
Very nice! ❤️

PR #1273, opened 2025-11-27
Always propagate transform to boundary

juliohm:
Part of a series of improvements related to `TransformedGeometry`.

### DIFF
--- a/src/sideof.jl
+++ b/src/sideof.jl
@@ -56,10 +56,7 @@ Possible results are `IN`, `OUT` or `ON` the `ring`.
   Differential Coding Boolean Operations on Polygons]
   (https://www.mdpi.com/2073-8994/10/10/477)
 """
-sideof(point::Point, ring::Ring) = _sideof(point, ring)
-sideof(point::Point, ring::TransformedRing) = _sideof(point, ring)
-
-function _sideof(point::Point, ring)
+function sideof(point::Point, ring::Ring)
   assertion(CoordRefSystems.ncoords(crs(point)) == 2, "points must have 2 coordinates")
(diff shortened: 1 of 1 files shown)


juliohm:
@eliascarv we introduced `TransformedGeometry` to delay the materialization of some transforms. It seems that in this case we could avoid the extra branch with `isdistorted` and simply delegate the behavior to the transform itself, which will dispatch on the boundary type.

The other two places where we are  still using `isdistorted` is the `discretize` and `pointify`, which I will look next. Please let me know if you remember other possible issues that might occur with these suggested changes.

eliascarv:
This change probably won't have a major impact. I think always returning `TransformedGeometry` in `boundary` is more correct.

eliascarv (approved):

juliohm:
> I think always returning `TransformedGeometry` in `boundary` is more correct.

We could return `TransformedGeometry(b, t)` always to enforce the delay, even when the transform `t` materializes the argument by design. I thought about it and it seems too restrictive. The fact that `t(b)` can be customized leads to a more flexible design where the author of the transform can choose what to do depending on the argument. We are still propagating the transform, but the author decides when it should be materialized, makes sense?

juliohm:
> This change probably won't have a major impact.

I believe it will. The new behavior already broke one test where we were testing against `TransformedGeometry` explicitly. I am working on a series of changes next, and we will probably need a minor release.

eliascarv:
> We are still propagating the transform, but the author decides when it should be materialized, makes sense?

I completely agree. I actually expressed myself poorly, I think the way it is in PR seems more practical.

PR #1277, opened 2025-11-29
Better discretization of non-Euclidean geometries

juliohm:
Fixes some fundamental issues with the current `discretize` methods to handle non-Euclidean geometries (e.g. curved geometries over the globe).


src/boundary.jl
```diff
@@ -213,10 +213,22 @@ end
...
   b = boundary(parent(g))
-  isnothing(b) ? nothing : transform(g)(b)
+  t = transform(g)
+  if isnothing(b)
+    nothing
+  elseif b isa Geometry
+    TransformedGeometry(b, t)
+  elseif b isa Mesh
+    TransformedMesh(b, t)
+  end
```
juliohm, line 223: @eliascarv it turns out we needed a more general method to delay the discretization, which enforces the `TransformedGeometry` and `TransformedMesh` 👍🏽

PR #1285, opened 2025-12-01
Refactor GeometrySet recipes

juliohm:
This refactor will facilitate future additions.

### DIFF
--- a/src/discretization.jl
+++ b/src/discretization.jl
@@ -58,7 +58,7 @@ _discretize(geometry::Geometry) = _simplexify(geometry)
 
 _discretize(box::Box) = discretize(box, RegularDiscretization(1))
 
-_discretize(ball::Ball{𝔼{2}}) = discretize(ball, RegularDiscretization(50))
+_discretize(ball::Ball{𝔼{2}}) = discretize(ball, RegularDiscretization(2, 50))
 
 _discretize(disk::Disk) = discretize(disk, RegularDiscretization(50))
 
--- a/test/discretization.jl
+++ b/test/discretization.jl
@@ -559,7 +559,7 @@ end
   mesh = discretize(ball)
   @test !(eltype(mesh) <: Triangle)
   @test !(eltype(mesh) <: Quadrangle)
-  @test nelements(mesh) == 2550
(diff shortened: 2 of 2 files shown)


juliohm:
@JoshuaLampert so we could in theory add tests to the new repository and restart the action here to see it in action?

JoshuaLampert:
Yes, exactly. The tests have to be in main of MeshVizTests.jl.

juliohm:
@JoshuaLampert I wonder if we could just point to the existing GeoStatsVizTests.jl since the action accepts any downstream package in any organization. The tests are all there already.

JoshuaLampert:
Sure that's possible. You should only need to update this line
https://github.com/JuliaGeometry/Meshes.jl/blob/d8fb284af44687583500e08e0d7d44b508e91583/.github/workflows/Downstream.yml#L36
and this line:
https://github.com/JuliaGeometry/Meshes.jl/blob/d8fb284af44687583500e08e0d7d44b508e91583/.github/workflows/Downstream.yml#L52
in the action. I thought you wanted to have a separate repo for Meshes.jl specific tests, while GeoStatsVizTests.jl is for the whole GeoStats.jl ecosystem. But if we only had one repo for viz tests, it would of course reduce some maintenance burden.

juliohm:
I'm still trying to figure out what to do in this case. I wonder if
downstream compat entries can compromise the visual tests of Meshes.jl. The
maintenance burden shouldn't be very high to maintain both repos given the
number of repos we already maintain.

Will try to migrate the viz specific recipes to this new MeshVizTests.jl
repo.

Em seg., 1 de dez. de 2025, 13:12, Joshua Lampert ***@***.***>
escreveu:

JoshuaLampert:
I would be fine either way.

juliohm:
@JoshuaLampert I will proceed with GeoStatsVizTests.jl for now, but we can gradually move the Meshes.jl' specific tests to the new MeshVizTests.jl repository if the action works well.

This PR is finally ready after a whole day messing with the recipe system. We finally have something working! I will merge and adjust the action for the next PR tomorrow.

Thank you again for the help!

PR #1286, opened 2025-12-02
Move visual tests to external repository

juliohm:
Let's see if the downstream visual tests are triggered correctly.

### DIFF
--- a/docs/src/geometries/polytopes.md
+++ b/docs/src/geometries/polytopes.md
@@ -23,7 +23,7 @@ Segment
 ```
 
 ```@example polytopes
-Segment((0., 0.), (1.,1.)) |> viz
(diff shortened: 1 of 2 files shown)


juliohm:
@JoshuaLampert I believe everything is working as expected. I moved the few visual tests to the external repository to reduce the burden of maintaining them here.

juliohm:
I thought about it too. I think what we really need in these cases is to write actual non-visual tests for the refinement algorithms. The visual tests are now "covered" in the external repo, and I just finished porting it to TestItems.jl to facilitate testing specific visual tests.

juliohm:
Will merge this one to get things going. Thanks for the amazing help as usual! 💯

PR #1287, opened 2025-12-03
Optimize indices for invertible TransformedGrid

marcosdanieldasilva:
This PR adds a major performance optimization for finding indices on a `TransformedGrid`.

Previously, querying a transformed grid would often fall back to the generic `findall` method. This approach is extremely slow for large grids, as it requires performing an intersection check for every single cell in the mesh one by one.

This new specialized method implements a **Fast Path** using the inverse transformation logic:

1.  **Check Invertibility:** It first checks if the grid's transformation is mathematically invertible (using `isinvertible`).
2.  **CRS Consistency:** It ensures the input geometry is projected to the same Coordinate Reference System as the grid.
3.  **Inverse Mapping:** It applies the inverse transform to the geometry, mapping it back to the parent domain's coordinate space.
4.  **Delegate:** It calls the existing, highly optimized `indices` function on the underlying parent grid (which allows for instant lookups).

If the transform is **not** invertible (e.g., contains `Morphological` operations), the method safely falls back to the generic iteration strategy.

### DIFF
--- a/src/indices.jl
+++ b/src/indices.jl
@@ -44,21 +44,20 @@ indicesfallback(domain, geometry) = findall(intersects(geometry), domain)
 # ----------------
 
 function indices(grid::OrthoRegularGrid, point::Point)
-  point ∉ grid && return Int[]
-
-  # grid properties
+  # point coordinates in grid
   orig = minimum(grid)
   spac = spacing(grid)
   dims = size(grid)
+  xyz = (point - orig) ./ spac
 
-  # integer coordinates
-  ijk = ceil.(Int, (point - orig) ./ spac)
-
-  # fix coordinates that are on the grid border
-  ijk = clamp.(ijk, 1, dims)
-
-  # convert to linear index
-  [LinearIndices(dims)[ijk...]]
+  # check if point is in/out grid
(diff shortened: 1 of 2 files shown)


juliohm:
Excellent @marcosdanieldasilva ! This will help a lot of people very soon 🚀

test/indices.jl
```diff
@@ -202,4 +202,12 @@
...
   @test linds[10, 10] == only(indices(grid, p5))
   @test isempty(indices(grid, p6))
+
+  # fast method for transformed grids with invertible transforms
+  tri = Triangle(cart(5, 5), cart(10, 10), cart(15, 5))
+  grid = cartgrid(20, 20)
+  trans = Translate(2.0, 3.0)
+  ttri = TransformedGeometry(tri, trans)
+  tgrid = TransformedGrid(grid, trans)
+  @test issetequal(indices(tgrid, transtri), indices(grid, tri))
```
juliohm: Can we be more strict and use `==` in this case? My intuition says that the translation shouldn't change the order of indices.

juliohm (commented):
Could you please add another test with the transform pipeline below?

```julia
Affine([1.0 0.0; 0.0 1.0], [1.0, 2.0]) → ReinterpretCoords(Cartesian, LatLon)
```

PR #1298, opened 2025-12-08
More optimizations and fixes related to discretization of quadrangles

juliohm:
This should speedup the discretization of quadrangles by a 2x factor.


juliohm:
image

Simple change, huge impact 🚀

juliohm:
The downstream visual tests are failing as expected. Will update after merging this.

PR #1307, opened 2025-12-10
Remove bbox optimization in vizgrid

juliohm:
It doesn't help performance-wise, and makes the call trace unnecessarily complex.


juliohm:
The visual test failed as expected. The new result is the correct one. Merging and fixing the test data downstream.

PR #1318, opened 2026-01-10
Feature addition, generalized polygon boolean operations

souma4:
This began as an implementation of the Martinez–Rueda (MR) clipping algorithm (#578) to support clipping with concave polygons. Quickly, it became clear the MR framework naturally generalizes to full polygon–polygon boolean operations with relatively little additional code. As a result, this PR introduces a generalized polygon boolean operation rather than a narrowly scoped clipping implementation.

This is substantial. I explored breaking the work into smaller PRs, but each stage feeds directly into the next, and isolating them would have resulted in either duplicated logic or incomplete intermediate abstractions. Given that constraint, a single PR felt like the most coherent approach. I have included a diagram below that outlines the major functions and how they interact.

Primary new features

* polygonbooleanop: a general function supporting polygon–polygon boolean operations.
* Extensions of existing set-like methods (union, setdiff, intersect, symdiff, xor) implemented via polygonbooleanop.
* A fallback for clip that uses the MR algorithm, plus an explicit MartinezRuedaClipping() method.
* Introduction of a dependency on [BinaryTrees.jl](https://github.com/JuliaCollections/BinaryTrees.jl)
.

Potential future extensions

* Polygon–segment clipping (splitting polygons by segments).
* Segment–polygon clipping (restricting segments to those inside a polygon).
* An erase interface for computing polygon differences using plain language.

# Potential concerns and open questions

## Performance

Profiling indicates that approximately 50% of runtime is spent in pairwiseintersect and ~30% in _annotatefill, largely due to BinaryTree insertions and deletions. I have made a concerted effort to minimize allocations and GC pressure throughout, but pairwiseintersect is likely the highest-leverage target for future optimization.
Compared to SutherlandHodgmanClipping (SH), this implementation is expectedly slower—about 4× on a representative real-world test case (~20 ms vs ~80 ms).

## Consistency with existing clip behavior

There are currently discrepancies between MR and SH outputs. Some are benign (e.g., differing starting vertices), but others are more structural:

In certain cases, SH produces polygons with holes where MR produces a single polygon.

In tests such as “PolyArea with box” and “PolyArea with one inner ring inside other and another inner ring outside other”, SH places a hole on the polygon boundary, creating a “ghost segment” (a segment corresponding only to a hole). MR correctly omits this and returns a single polygon.

SH can also produce what should be Multi geometries by representing secondary polygons as holes.

These inconsistencies will likely need to be addressed to ensure predictable downstream behavior.

## Code complexity

I invested significant effort in simplifying the implementation, but the logic remains dense. One notable pain point is that segments for subject and clipping polygons are constructed twice; I was unable to reconcile the `rings` and `segments` interfaces in a way that avoided this duplication without introducing additional complexity. More broadly, there is a lot happening in this PR, and I am open to suggestions on further refactoring or clarification.

Architecture overview

The following diagram shows the major functions introduced in this PR and where they live:
PR_polybool

* P.S. I know this PR looks a lot like an AI one. I just knew this PR was large and needed as much clarity as possible. Ultimately, AI was only used for basic auto completion, generating the dependency figure, and bouncing ideas off of for clarity, simplicity etc. Actual structure and coding was all my writing, because AI is horrible at these more advanced tasks.


juliohm:
Thank you @souma4 for insisting on this.

It is definitely hard to review these massive PRs with lots of pieces that probably contain bugs and untested corner cases.

Please break this PR into smaller PRs that can be safely reviewed.

In these smaller PRs do your best to write idiomatic Julia code. I see the same old patterns from previous PRs where you introduce types just to hold constants (no behavior) and types that could be easily replaced by built-in data structures (e.g. vectors).

juliohm:
Closing in favor of smaller PRs, starting with #1319

PR #1319, opened 2026-01-11
Feature Addition: Polygon boolean (#1)

souma4:
In line with feedback in #1318 here is an initial small PR that adds a shell for the broader algorithm and the first necessary step for this algorithm - which is to compute and insert intersections into polygons.

I'm also adding a utility that's used through the algorithm to take a 4 bit number marking whether a segment has a subject or clipping polygon above or below it. 

These are small, simple internals that should be easy to review.

The next PR will be taking these modified polygons and performing a sweep line to fill in those segment region fills.

Edit: current failing test is `pairwiseintersect`, which hasn't been changed.

### DIFF
--- a/src/polygonboolean.jl
+++ b/src/polygonboolean.jl
@@ -10,16 +10,16 @@ include("polygonboolean/utils.jl")
 # include("polygonboolean/reconstruction.jl")
 
 """
-    polygonbooleanop(geometry, other, operation)
+    polygonbooleanop(poly, other, operation)
 
-Performs boolean `operation` of `geometry` with `other`.
+Performs polygon boolean `operation` of `poly` with `other`.
 
 ## Operands
 
 - `operation` — the operation to perform; one of:
   - `intersect` — regions inside both geometries (∩)
   - `union` — regions inside at least one geometry (∪)
-  - `setdiff` — regions inside `geometry` but not `other`
+  - `setdiff` — regions inside `poly` but not `other`
   - `symdiff` — regions inside exactly one geometry (xor/⊻)
 
 ## Notes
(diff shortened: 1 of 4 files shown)


src/polygonboolean/utils.jl
```diff
@@ -0,0 +1,108 @@
+# ------------------------------------------------------------------
+# Licensed under the MIT License. See LICENSE in the project root.
+# ------------------------------------------------------------------
+
+# insert intersection points into rings
+function _insertintersections!(intersections, seginds, allrings)
```
juliohm: So is this function reimplementing `Base.splice!`? If you take the `vertices(ring)` as a `CircularVector`, you should be able to splice multiple points efficiently.

src/polygonboolean.jl
```diff
@@ -0,0 +1,115 @@
...
+include("polygonboolean/utils.jl")
+# include("polygonboolean/fillsegments.jl")
+# include("polygonboolean/sweep.jl")
+# include("polygonboolean/selection.jl")
+# include("polygonboolean/reconstruction.jl")
+
+"""
+    polygonbooleanop(geometry, other, operation)
+
+Performs boolean `operation` of `geometry` with `other`.
```
juliohm: Can the algorithm be used with any kind of geometry or just polygons (and their rings)?

src/polygonboolean.jl
```diff
@@ -0,0 +1,115 @@
...
+  g₁ ⊻ g₂
+
+Compute the symmetric difference (exclusive or) of geometries `g₁` and `g₂`.
+"""
+Base.xor
+
+Base.union(a::Geometry, b::Geometry) = polygonbooleanop(a, b, union)
+Base.setdiff(a::Geometry, b::Geometry) = polygonbooleanop(a, b, setdiff)
+Base.symdiff(a::Geometry, b::Geometry) = polygonbooleanop(a, b, symdiff)
+Base.xor(a::Geometry, b::Geometry) = polygonbooleanop(a, b, symdiff)
```
juliohm: I believe the dispatch should be done with `Polygon` instead of general `Geometry` given the name of the function used in the implementation.

Also, please move these definitions to the source file where the `Polygon` is defined.

juliohm: And the docstrings should be attached to these specific methods. You are currently adding docstrings to the function name, not the method.

src/polygonboolean.jl
souma4: as of right now, polygons and rings, but in the future polygon-segment and segment-polygon should be added. I'll knock these down to polygon/ring for the time being.

src/polygonboolean/utils.jl
souma4: `CircularVector` does not have a `splice!` method (weirdly). `insert!` also doesn't update the size of `CircularVector` until after insertion, which means it places points at the beginning of the vector. If we add multiple points to the end (e.g., the final segment of a poly has multiple intersections), their order gets reversed. Append is much cleaner.
The other key part here is making sure the right points go into the right rings in the right orientation. This means the checking of `ps` and `pe`, the sorting, etc, are necessary to keep proper geometry.

juliohm: Should we open a PR in CircularArrays.jl to add splice! support?

souma4:
> It is not very clear what this PR is trying to accomplish. I can review it again after you consider the first round of comments.

The primary goal is build a rough shell of the polygonboolean operation and implement the first major feature, which is splitting segments at intersections and inserting those into the polygon.

src/polygonboolean/utils.jl
souma4: I've written an issue there because I'm too busy to make the PR.

souma4: So I haven't received any response on a `splice!` update. Let me know how you'd like to proceed

juliohm: The maintainer of CircularArrays.jl mainly reviews PRs. We should start a PR over there and refactor this one here afterwards.

souma4:
So `splice!` is now in CircularArrays and I've updated this PR. Now I look at it and realize I have all this segment fill logic (for saying if a given polygon is filled above or below a segment). I lean I can cut that right now and we can just hone in on this `insertintersection!` logic.

juliohm:
Thank you for the update @souma4 ! Please let me know when the PR is ready for review and I can reserve some time to look into it over the week or next week.

souma4:
Quick Q just to save us some time for review, would you like me to omit the segment fill logic so we can split this implementation into a lot of very little pieces? Or is this initial moderate PR acceptable?

juliohm:
If you feel that this PR could be split into smaller self-contained PRs, then that is probably a good idea 🙏🏽

souma4:
I'm going to split this into smaller PR's reviews with branches successively built upon each other. I want to make sure we are all on the same page with this implementation.

Issue #497, opened 2023-07-06
Add concave hull methods


juliohm:
There are various methods in the literature to compute the "concave" hull of a 2D point set:

https://www.researchgate.net/publication/312373158_A_Concave_Hull_Based_Algorithm_for_Object_Shape_Reconstruction

We can use ConcaveHull.jl as a test bed, it implements the method that is based on the modification of Jarvis march that we already have implemented with k-nearest neighbors.

juliohm:
We now have `MoreiraMarch` thanks to @souma4 🙂

PR #1328, opened 2026-03-01
Extending `JarvisMarch` to perform concave hulls

souma4:
Implementing a k nearest neighbors approach for Concavehulls #497. The paper referenced in the issue is primarily for combining offset points and is quite a bit more involved than [this paper](http://repositorium.sdum.uminho.pt/bitstream/1822/6429/1/ConcaveHull_ACM_MYS.pdf). This looks very similar to Jarvis March, just with an added check to prevent self intersections. I refactored some of the convex hull tests to make sure we are properly testing hulls in general, but not holding concave hulls to the same standard as convex hulls.

I am curious, though, The tests starting at L43 and L75 give different polygons between concave and convex hulls. The convex hull hugs tightly to the points, but the concave hull smooths them. To me it seems like a valid solution (plus, concave hulls don't have a singular correct solution), but it seems sort of weird that the area of the convex hull is less than the area of the concave hull. Let me know what you think.

edit: Ah, I didn't catch it's just defaulting to convex hull because it fails to get some points inside the polygon. More work to be done. Will mark for review when it gets solved


juliohm:
@souma4 what is the status of this one? Is it ready for review?

src/hulls/concave.jl
```diff
@@ -0,0 +1,108 @@
...
+The algorithm has empirical complexity `O(n)` where `n` is the number of points
+
+## References
+
+* Moreira, A. & Santos, M. Y. 2007. "CONCAVE HULL: A K-NEAREST NEIGHBOURS APPROACH FOR THE COMPUTATION OF THE REGION OCCUPIED BY A SET OF POINTS". In: *Proceedings of the Second International Conference on Computer Graphics Theory and Applications*. SciTePress, pp. 61–68.
+
+"""
+struct Concave <: HullMethod end
+
+function hull(points, ::Concave; k=3)
```
juliohm: Notice that the design we follow is one in which every parameter of the method is stored in the method object itself. The `k` parameter should be a field of the `Concave` struct. Also, try to be more specific with the method name, perhaps `KnnJarvisMarch`?

Can't we adjust our current `JarvisMarch` to include a `k` parameter instead of repeating a similar implementation?

souma4: As far as your first suggestion, that's easy to swap in.

For the second, Jarvis March works by performing a global ordering of points by "right hand turns." What do you think of making a more generic `JarvisHull` that can be used to dispatch the k-th jarvis march algorithm, but `KnnJarvisMarch` is a type that defaults to k=3, and `JarvisMarch` is just a method dispatch for hull where points are used to set k? 

Something to the effect of

```julia
abstract type JarvisHull <: HullMethod end # I can see removing this and Making Knn JarvisMarch a subtype of HullMethod. That would avoid introducing a new abstract type. Just might make reduce extensibility.
struct JarvisMarch <: JarvisHull end
struct KnnJarvisMarch <: JarvisHull
  k::Int
  KnnJarvisMarch() = new(3)
  KnnJarvisMarch(k::Int) = new(k)
end

function hull(points, method::J) where J <: JarvisHull #if we remove the abstract type, then just replace J with the concrete type
 # k-nearest implementation that I propose
 # ...
end

hull(points, ::JarvisMarch) = hull(points, KnnJarvisMarch(length(points))

```

Let me know what you think

souma4:
@juliohm I still want to simplify the internal logic. The check for self intersection is a bit messy right now so I don't think it's ready. And with your other feedback, it might be worth slotting this Knn approach into `jarvis.jl` and deleting the `concave.jl` file.

src/hulls/concave.jl
juliohm: I would keep it as simple as possible. From what I can tell we only need to generalize the implementation of `JarvisMarch` to handle `k`. By default we can set `k=nothing` and update it inside the `hull` call to match the current behavior. If the generalization to handle `k` implies worst performance when `k=npoints`, then we can consider alternative designs.

souma4:
@juliohm This should be good for review now. I refactored the main jarvis algorithm to dispatch on different `JarvisMarch{k}` methods. if on `Nothing` it's same as original implementation (which `JarvisMarch()` creates). If on an integer, then it dispatches the knearest approach. If a valid hull cannot be made at a given `k` it iterates through `k`'s until a valid one is found. According to the profiler, most time is spent checking if a proposed "next segment" intersects the current hull segments. Without that, it's very easy to get inappropriate early termination.

juliohm:
Thank you @souma4. I will try to find some time to review the PR over the week.

src/hulls/jarvis.jl
```diff
@@ -34,16 +43,29 @@ function hull(points, ::JarvisMarch)
...
   # find bottom-left point
   i = argmin(p)
+  # find next point with smallest angle
+  O = p[i]
+  A = O + Vec(zero(ℒ), -oneunit(ℒ))
+
+  # perform primary Jarvis march loop
+  ℐ = _jarvisloop(method, points, p, n, i, O, A)
+  # if no hull is found, return convex hull as fallback
+  isnothing(ℐ) && return convexhull(p)
```
juliohm: We can't call `convexhull`, which is a higher-level abstraction inside `hull`, which is the low-level implementation. I also don't understand why is it possible that the indices are `nothing`. Whenever we call `hull` with a set of points, the result of the Jarvis loop should be valid, right?

src/hulls/jarvis.jl
```diff
@@ -67,6 +89,92 @@ function hull(points, ::JarvisMarch)
...
   end
 
-  # return polygonal area
-  PolyArea(p[ℐ[begin:(end - 1)]])
+  # return indices of hull vertices
+  ℐ
+end
+
+# concave hull
+function _jarvisloop(method::JarvisMarch{I}, points, p, n, i, O, A) where {I<:Integer}
```
juliohm: From the paper I recall that the only difference between JarvisMarch{Nothing} and JarvisMarch{K} the the set of candidate indices. In the former, we consider all remaining candidates, whereas in the latter we consider the K-nearest candidates.

Couldn't we simply preserve the original implementation, with minimal changes, by just introducing a helper function to specialize the set of candidates?

```julia
jarviscandidates(::JarvisMarch{Nothing}, ...) = # all remaining indices
jarviscandidates(::JarvisMarch{<:Integer}, ...) = # K-nearest indices
```

The less we modify the original implementation, the less likely we introduce bugs.

If these functions need an extra state (e.g., KNN searcher), then this state could be instantiated out of the loop, and passed to the function:

```julia
searcher = nothing
searcher = KNearestSearch(...)

jarviscandidates(searcher::Nothing, ...)
jarviscandidates(searcher::KNearest, ...)
```

Alternatively, you could select a function before the Jarvis loop based on `K` to avoid extra complications:

```julia
# define candidates function
candidates = if k isa Integer
  # instantiate searcher once
  searcher = KNearestSearch(...)
  # return function that retrieves k-nearest candidates
  i -> search(points[i], searcher)
else # k is nothing
  # return function that retrieves remaining candidates
  i -> setdiff(1:n, [i])
end

# now candidates can be used regardless of K
js = candidates(i)
```

src/hulls/jarvis.jl
```diff
@@ -67,6 +89,92 @@ function hull(points, ::JarvisMarch)
...
+  m = I(n - 2)
+  k = min(max(method.k, 3), m)
+  assertion(ispositive(k), "k must be a positive integer")
+  k > m && return _jarvisloop(JarvisMarch{Nothing}(nothing), points, p, n, i, O, A) # fallback to convex hull
+
+  # initial state for retries
+  i₀, O₀, A₀ = i, O, A
+
+  # try increasing k until valid hull found
```
juliohm: I am concerned that you are implementing more than just Jarvis March with fixed K. This meta algorithm for increasing k is necessary or is it another layer of abstraction to enforce some kind of expected result?

If for some `k` and some specific points the algorithm is not valid (e.g., infinite loop), we should just return an error. Any additional logic to attempt multiple `k` could be made available in a separate method like `JarvisMarchManyKAttempts` or something.

juliohm (changes requested):
Hi @souma4 , attached is a first round of review.

I believe we still have ways to simplify the implementation and make it more straightforward.

src/hulls/jarvis.jl
souma4: This is so that people don't need to tweak input k if their k is too low to find a valid result. we could easily do a wrap around the function for many k attempts. I'll roll with fixed k

src/hulls/jarvis.jl
souma4: Possibly. The biggest issue would be the self intersection check and reinserting the starting point to the `KNearestSearch` mask, but there is probably a solution. I'll work on that.

src/hulls/jarvis.jl
```diff
@@ -4,22 +4,63 @@
...
+"""
+struct JarvisMarch{K} <: HullMethod
+  k::K
+end
+JarvisMarch() = JarvisMarch{Nothing}(nothing)
+JarvisMarch(k::I) where {I<:Integer} = JarvisMarch{I}(k)
+
 """
-struct JarvisMarch <: HullMethod end
+    AdaptiveJarvisMarch(k)
```
souma4: I've named this iteratively increasing k approach "AdaptiveJarvisMarch." It doesnt change k on the fly as needed, it tries to form the hull at k = m, if that fails then it increments k by 1 and tries again. If you have name suggestions, that'll be awesome.

src/hulls/jarvis.jl
```diff
@@ -34,39 +75,96 @@ function hull(points, ::JarvisMarch)
   # corner cases
   n == 1 && return p[1]
   n == 2 && return Segment(p[1], p[2])
+  n == 3 && return PolyArea(p)
 
   # find bottom-left point
   i = argmin(p)
+  start = i
+  # initialize hull with i
+  ℐ = [i]
```
souma4, line 67: I init `\scrI` for the new `jarviscandidates` function. This is because `j` is requires a search.

src/hulls/jarvis.jl
```diff
@@ -34,39 +75,96 @@ function hull(points, ::JarvisMarch)
...
+function jarvisnext(::KNearestSearch, candidates, p, ℐ, A, O, k)
+  sort!(candidates, by=l -> ∠(A, O, p[l]))
+  # check candidates in order of angle until we find one that doesn't intersect the existing hull
+  for nᵢ in candidates
+    cpoint = p[nᵢ]
+    cseg = Segment(p[ℐ[end]], cpoint)
+    cbox = boundingbox(cseg)
+    offset = cpoint == p[ℐ[begin]] ? 1 : 0
+    limit = length(ℐ) - 1 - offset
+    ok = limit < 2 || !any(2:limit) do indⱼ
```
souma4: This preselection using bounding boxes cut the time required to fun by 50% and it's fairly simple. Helps reduce issues with many expensive intersection checks

test/hulls.jl
```diff
@@ -1,5 +1,5 @@
 @testitem "Hulls" setup = [Setup] begin
-  for method in [GrahamScan(), JarvisMarch()]
+  for method in [GrahamScan(), JarvisMarch(), AdaptiveJarvisMarch()]
```
souma4: I'm testing on AJM only because `JarvisMarch(k)` is a pain to tune for every test. My thinking is that if AJM works as intended, JM must. Given the paper I reference uses AJM as their reference point, feel comfortable with this.

souma4:
Sorry for the commit wall. I screwed up a rebase to the newest version of Meshes. Nice thing is this no longer contains commits from the integral patches.

This is ready for review and I've left some comments to help in the review.

src/hulls/jarvis.jl
souma4: Right now the function returns `nothing` if the choice of k leads to an issue. This was to make the `AdaptiveJarvisMarch` wrapper performant (try-catches are quite slow). 

We could maybe integrate this info into the doc.

juliohm:
Tests are failing, so it is not ready for review yet.

souma4:
huh... I really thought it would pass all tests, but it's some weird issue where infinite loops are occuring when the inputs are random. I'll figure it out.

souma4:
Okay, now we're good for a review

src/hulls/jarvis.jl
```diff
@@ -4,22 +4,63 @@
...
+  pointsꜝ = unique(points)
+  for kᵢ in k:length(pointsꜝ)
+    chul = hull(points, JarvisMarch(kᵢ))
+    # validate we found a hull that contains all points, if so return it, otherwise increase k and try again
+    isnothing(chul) && continue
+    all(points .∈ Ref(chul)) && return chul
+  end
+  # otherwise, fallback to convex hull
+  hull(points, JarvisMarch())
+end
```
juliohm: This algorithm is too simple to deserve a struct. Users will likely need different strategies for picking `k` and interrupting the loop. Please remove `AdaptiveJarvisMarch` from the PR. We can think about it later if really necessary.

src/hulls/jarvis.jl
```diff
@@ -34,39 +75,97 @@ function hull(points, ::JarvisMarch)
...
   # find bottom-left point
   i = argmin(p)
+  start = i
+  # initialize hull with i
+  ℐ = [i]
+
+  # initialize candidate set and searcher for k-nearest neighbors if needed
+  k, searcher, 𝒞 = _jarvissetup(method.k, n, p)
+  # if no candidates exist (k too large), fallback to convex hull
+  isnothing(𝒞) && return hull(points, JarvisMarch())
```
juliohm: This condition should happen in the beginning, right after the corner cases with n==1, n==2, n==3. You can simply check if k >= n and call the convex hull case.

src/hulls/jarvis.jl
```diff
@@ -34,39 +75,97 @@ function hull(points, ::JarvisMarch)
...
+end
+
+# outputs k, searcher, and a mask of candidate point indices
+_jarvissetup(k, n, p) = nothing, nothing, 1:n
+
+function _jarvissetup(k::T, n, p) where {T<:Integer}
+  k = min(max(k, 3), n)
+  assertion(ispositive(k), "k must be a positive integer")
+  k, KNearestSearch(p, k), falses(n)
+end
```
juliohm: Please rename this function to something more descriptive. I understand that it is building the `searcher` when `k` is provided. A more descriptive name could be `jarvissearcher`.

Please avoid overwriting variable names (`k` in this case), specially if they are passed as function arguments. Overwriting variables in Julia can lead to type instabilities (variables changing type) and behavior that is difficult to predict/parallelize.

src/hulls/jarvis.jl
```diff
@@ -34,39 +75,97 @@ function hull(points, ::JarvisMarch)
   # corner cases
   n == 1 && return p[1]
   n == 2 && return Segment(p[1], p[2])
+  n == 3 && return PolyArea(p)
```
juliohm: Isn't this case already covered by the general algorithm? Is this extra line added for performance?

src/hulls/jarvis.jl
```diff
@@ -34,39 +75,97 @@ function hull(points, ::JarvisMarch)
...
+  ℐ = [i]
+
+  # initialize candidate set and searcher for k-nearest neighbors if needed
+  k, searcher, 𝒞 = _jarvissetup(method.k, n, p)
+  # if no candidates exist (k too large), fallback to convex hull
+  isnothing(𝒞) && return hull(points, JarvisMarch())
 
-  # candidates for next point
-  𝒞 = [1:(i - 1); (i + 1):n]
+  candidates = jarviscandidates(searcher, 𝒞, n, p, i, i, start)
```
juliohm: It is much easier to understand the proposed changes if the variable names are not changed. Before this PR, we called `𝒞` the pool of candidate indices. Please preserve the name.

juliohm (changes requested):
Thank you @souma4 for the updated PR. It is looking much cleaner now.

Try to define/document the 2 or 3 main operations that change in the presence or absence of `k`. With these operations well-defined, implement the corresponding methods.

My current understanding is that we only need to define the `jarvissearcher` operation, which creates a searcher or a dummy (`nothing`) object, the `jarviscandidates` operation, that takes the searcher and returns indices, and some sort of `jarvisupdate` operation.

src/hulls/jarvis.jl
souma4: the earlier k > n change basically turns L78 moot, so now it's removed.

src/hulls/jarvis.jl
souma4: So, I see what you mean. The problem is that we seem to be using `convexhull` as a generic hull creator with sensible defaults. However, concave hulls may not be possible with certain k's. I can see this being expected behavior for the lower level `hull` interface, but for `concavehull` there probably should be a sensible default that performs this k iteration strategy. This is so someone can just ask for a concave hull and get a valid result rather than get somewhat unexpected nothing outputs.

What do you think of making the `concavehull` interface perform this iteration approach to get a valid result, but `hull(JarvisMarch{Int})` performs the algorithm with a given k?

src/hulls.jl
```diff
@@ -64,10 +79,28 @@ convexhull(g::Grid) = Box(extrema(g)...)
...
 # IMPLEMENTATIONS
 # ----------------
 
 _gconvexhull(geoms) = _pconvexhull(p for g in geoms for p in boundarypoints(g))
 
 _pconvexhull(points) = hull(points, GrahamScan())
+
+_gconcavehull(geoms) = _pconcavehull(p for g in geoms for p in boundarypoints(g))
+
+_pconcavehull(points) = hull(points, AdaptiveJarvisMarch()) # default k = 3 for concave hulls
```
juliohm: I think this internal `_pconcavehull` function could hold the heuristic you called `AdaptiveJarvisMarch`. The implementation should live here though.

test/hulls.jl
```diff
@@ -1,5 +1,8 @@
 @testitem "Hulls" setup = [Setup] begin
-  for method in [GrahamScan(), JarvisMarch()]
+  # can AdaptiveJarvisMarch(Integer) be constructed with an integer k and work
+  @test hull(cart.([(0, 0), (1, 0), (1, 1), (0, 1), (0.5, -1)]), AdaptiveJarvisMarch(3)) isa PolyArea
+
+  for method in [GrahamScan(), JarvisMarch(), AdaptiveJarvisMarch()]
```
juliohm: Please remove the `AdaptiveJarvisMarch` from this testset and any related change. We need a dedicated test set for concave hulls.

juliohm:
@souma4 when do you think you will have some time to continue the work on this PR? :)

souma4:
This week I'll have the updated version pushed.

souma4:
I think this is good for a review.

src/hulls/jarvis.jl
```diff
@@ -4,22 +4,35 @@
...
 
+If `k` is provided, the algorithm will attempt to compute a concave hull using k
+nearest neighbors. However, the algorithm is not guaranteed to succeed for any particular `k`.
+
 The algorithm has complexity `O(n*h)` where `n` is the number of points
 and `h` is the number of points in the hull.
 
+see `concavehull` for a version that iteratively increases `k` until all
+points are in the hull, which is useful when an effective `k` is not known prior.
+This gurantees correctness.
```
juliohm: The `concavehull` is a higher-level function, and shouldn't be mentioned in the docstring of the implementation. Please remove this paragraph.

src/hulls/jarvis.jl
```diff
@@ -4,22 +4,35 @@
...
 """
     JarvisMarch()
+    JarvisMarch(k)
 
 Compute the convex hull of a set of points or geometries using the
 Jarvis's march algorithm. See [https://en.wikipedia.org/wiki/Gift_wrapping_algorithm]
 (https://en.wikipedia.org/wiki/Gift_wrapping_algorithm).
 
+If `k` is provided, the algorithm will attempt to compute a concave hull using k
+nearest neighbors. However, the algorithm is not guaranteed to succeed for any particular `k`.
```
juliohm: What happens when the algorithm does not succeed? That would be useful information to add to the docstring.

src/hulls/jarvis.jl
```diff
@@ -4,22 +4,35 @@
...
 * Jarvis 1973. [On the identification of the convex hull of a finite set of
   points in the plane](https://www.sciencedirect.com/science/article/abs/pii/0020019073900203)
+* Moreira, A. & Santos, M. Y. 2007. "concave hull: a k-nearest neighbours approach for the computation of the region occupied by a set of points". In: *Proceedings of the Second International Conference on Computer Graphics Theory and Applications*. SciTePress, pp. 61–68.
 """
-struct JarvisMarch <: HullMethod end
+struct JarvisMarch{K} <: HullMethod
+  k::K
+end
+JarvisMarch() = JarvisMarch{Nothing}(nothing)
+JarvisMarch(k::I) where {I<:Integer} = JarvisMarch{I}(k)
```
juliohm:
```suggestion
struct JarvisMarch{K} <: HullMethod
  k::K
end

JarvisMarch() = JarvisMarch{Nothing}(nothing)

JarvisMarch(k::I) where {I<:Integer} = JarvisMarch{I}(k)
```

src/hulls/jarvis.jl
```diff
@@ -34,39 +47,97 @@ function hull(points, ::JarvisMarch)
   # corner cases
   n == 1 && return p[1]
   n == 2 && return Segment(p[1], p[2])
+  kₒ = method.k
+  isnothing(kₒ) || kₒ ≥ n && return hull(points, JarvisMarch())
```
juliohm: We need proper error handling when k > n. If the user specifies `k` manually, he/she should know that it can't be larger than `n`.

src/hulls/jarvis.jl
```diff
@@ -34,39 +47,97 @@ function hull(points, ::JarvisMarch)
...
+  mask[start] = true
+  mask[inds] = false
+  search(p[current], searcher; mask=mask)
+end
+
+# outputs k, searcher, and a mask of candidate point indices
+jarvissearcher(k, n, p) = nothing, nothing, 1:n
+
+function jarvissearcher(kₒ::T, n, p) where {T<:Integer}
+  k = min(max(kₒ, T(3)), n)
```
juliohm: I feel that you constantly mix side effects from different auxiliary functions with side effects that are part of the main function. Why the `jarvissearcher` auxiliary function needs to correct the input `k`? Isn't that the job of the main function? If you are asking for a searcher with invalid `k`, and correcting the `k` inside the auxiliary function, it is really hard to predict what is going  to happen later on in the main function.

Try to improve encapsulation. In this case, you should probably handle the corner cases with k=3 and k > n in the main function before calling `jarvissearcher`.

src/hulls.jl
```diff
@@ -34,6 +34,13 @@ Convex hull of `object`.
 """
 function convexhull end
 
+"""
+    concavehull(object)
+
+Concave hull of `object`.
+"""
+# function concavehull end
```
juliohm: This line is commented out.

src/hulls.jl
```diff
@@ -64,10 +79,39 @@ convexhull(g::Grid) = Box(extrema(g)...)
...
+  n = length(pointsꜝ) - 1
+  for k in 3:n
+    chul = hull(pointsꜝ, JarvisMarch(k))
+    # validate we found a hull that contains all points, if so return it, otherwise increase k and try again
+    isnothing(chul) && continue
+    all(pointsꜝ .∈ Ref(chul)) && return chul
+  end
+  # otherwise, fallback to convex hull
+  hull(pointsꜝ, JarvisMarch())
+end
```
juliohm: Please describe the heuristic that was implemented in the docstring. You're increasing `k` from `3` to `n` until all points are inside of the resulting polygonal area. Is there another heuristic to consider in practical applications? For example, select `k` such that the area is minimized? Select `k` such that all vertices of the resulting polygon are at least `d` meters apart? Should we give users the freedom to pick their own heuristic? If yes, I suggest that we add this `concavehull` utility in a separate PR.

souma4: If we want the opportunity for alternative heuristics, then we could probably make this PR just the `hull(JarvisHull{Integer}) ` implementation. Then in a separate PR we can add `concavehull` that accepts a heuristic argument and we can have the default be minimum k. Further heuristics can be added on top of that. Let me know what you think of that. Personally, from an extensibility standpoint, we should pursue this

src/hulls/jarvis.jl
souma4: That's fair. I'll make a pass through checking for these cases.

juliohm:
@souma4 please let me know when the PR is ready for another review.

souma4:
@juliohm This is ready for a review. The formatter action seems to not like the broadcasting with `cart.([(1,2),(3,4)....)`. I don't see a reason why that wouldn't be the default method of building those objects, though, so let me know if it's fine to keep or if I should just swap out for `[cart(1,2),cart(2,3)...]`. Also, these private functions are small and simple, and I'm uncertain on where we should preprend with `_` because some unexported/private functions in `Meshes.jl` don't have it, but other's do have it. That's a small simple fix if needed, just let me know.

juliohm:
Thank you Jeffrey. Trying to find the time to review it. Hopefully
today/tomorrow.

Em qui., 16 de jul. de 2026, 12:37, Jeffrey Chandler <
***@***.***> escreveu:

src/hulls/jarvis.jl
```diff
@@ -4,22 +4,35 @@
...
 The algorithm has complexity `O(n*h)` where `n` is the number of points
-and `h` is the number of points in the hull.
+and `h` is the number of points in the hull. The concave variant adds a
+k-nearest-neighbor search and hull-intersection checks per hull vertex.
 
 ## References
 
 * Jarvis 1973. [On the identification of the convex hull of a finite set of
   points in the plane](https://www.sciencedirect.com/science/article/abs/pii/0020019073900203)
+* Moreira, A. & Santos, M. Y. 2007. "concave hull: a k-nearest neighbours approach for the computation of the region occupied by a set of points". In: *Proceedings of the Second International Conference on Computer Graphics Theory and Applications*. SciTePress, pp. 61–68.
```
juliohm:
```suggestion
* Moreira, A. & Santos, M. Y. 2007. [concave hull: a k-nearest neighbours approach for the computation of the region occupied by a set of points]()
```
Needs a URL

src/hulls/jarvis.jl
```diff
@@ -4,22 +4,35 @@
...
 """
-struct JarvisMarch <: HullMethod end
+struct JarvisMarch{K} <: HullMethod
+  k::K
+end
+
+JarvisMarch() = JarvisMarch{Nothing}(nothing)
+
+JarvisMarch(k::I) where {I<:Integer} = JarvisMarch{I}(k)
```
juliohm:
```suggestion
JarvisMarch() = JarvisMarch(nothing)
```

src/hulls/jarvis.jl
```diff
@@ -4,22 +4,35 @@
...
 
 ## References
 
 * Jarvis 1973. [On the identification of the convex hull of a finite set of
   points in the plane](https://www.sciencedirect.com/science/article/abs/pii/0020019073900203)
+* Moreira, A. & Santos, M. Y. 2007. "concave hull: a k-nearest neighbours approach for the computation of the region occupied by a set of points". In: *Proceedings of the Second International Conference on Computer Graphics Theory and Applications*. SciTePress, pp. 61–68.
 """
-struct JarvisMarch <: HullMethod end
+struct JarvisMarch{K} <: HullMethod
+  k::K
```
juliohm:
```suggestion
struct JarvisMarch <: HullMethod
  k::Union{Int,Nothing}
```

src/hulls/jarvis.jl
```diff
@@ -34,39 +47,101 @@ function hull(points, ::JarvisMarch)
   # corner cases
   n == 1 && return p[1]
   n == 2 && return Segment(p[1], p[2])
+  k = method.k
```
juliohm:
```suggestion

  k = method.k
```

test/hulls.jl
```diff
@@ -1,15 +1,20 @@
 @testitem "Hulls" setup = [Setup] begin
-  for method in [GrahamScan(), JarvisMarch()]
+  for method in [GrahamScan(), JarvisMarch(), JarvisMarch(5)]
```
juliohm: Let's create a separate testset for JarvisMarch(k) so that we don't need to change these general tests for convex hulls. This "Hulls" test set could be seen as a generic test set for various convex hull methods, we need a separate one with more specific tests for JarvisMarch(k).

juliohm (changes requested):
Hi @souma4 , I couldn't review on the weekend, but found some time today to look into it. Appreciate it if you could isolate the new tests for JarvisMarch(k). After that I will pull the branch to test locally.

src/hulls/jarvis.jl
```diff
@@ -9,19 +9,38 @@ Compute the convex hull of a set of points or geometries using the
...
 * Jarvis 1973. [On the identification of the convex hull of a finite set of
   points in the plane](https://www.sciencedirect.com/science/article/abs/pii/0020019073900203)
+* Moreira, A. & Santos, M. Y. 2007. [concave hull: a k-nearest neighbours approach for the computation of the region occupied by a set of points](https://www.semanticscholar.org/paper/Concave-hull:-A-k-nearest-neighbours-approach-for-a-Moreira-Santos/319a3450f9909043d46eb7ceb4299efceb984d4f)
 """
-struct JarvisMarch <: HullMethod end
+struct JarvisMarch <: HullMethod
+  k::Union{Nothing,Int}
+
+  function JarvisMarch(k)
+    isok = isnothing(k) || (isinteger(k) && k > 2)
```
juliohm, line 32: @souma4 could you please confirm the valid interval for `k`? Is it the closed interval [2, n] or the open interval (2, n)? In other words, do we need strict inequalities?

src/hulls/jarvis.jl
```diff
@@ -31,42 +50,102 @@ function hull(points, ::JarvisMarch)
   p = unique(points)
   n = length(p)
 
+  # sanity check
+  isnothing(k) || assertion(k < n, "k must be less than the number of unique points")
```
juliohm, line 54: Same here, should this be strict inequality?

src/hulls/jarvis.jl
```diff
@@ -31,42 +50,102 @@ function hull(points, ::JarvisMarch)
...
 
     # update ring of indices
     push!(ℐ, j)
+    jarvisupdate!(searcher, visited, j)
   end
 
+  poly = PolyArea(p[ℐ[begin:(end - 1)]])
+
+  # invalid hull, should only happen if k is too small
+  validatehull(k, poly, p) || throw(ArgumentError("could not find concave hull with k = $k, try a larger k"))
```
juliohm: Can you please comment on this validation? It is not clear to me why we need to validate the result inside the function.

test/hulls.jl
```diff
@@ -171,3 +170,79 @@ end
...
+  @test pt[1] == hull(pt, JarvisMarch(3))
+  line = cart.([(0, 0), (1, 0)])
+  @test Segment(line...) == hull(line, JarvisMarch(3))
+  triangle = cart.([(0, 0), (1, 0), (0, 1)])
+  @test_throws AssertionError hull(triangle, JarvisMarch(2))
+  @test_throws AssertionError hull(triangle, JarvisMarch(3))
+  pts = cart.([(0, 0), (1, 0), (1, 1), (0, 1), (0.5, -1)])
+  @test_throws AssertionError hull(pts, JarvisMarch(2))
+  @test_throws AssertionError hull(pts, JarvisMarch(5))
+  @test_throws AssertionError hull(pts, JarvisMarch(6))
```
juliohm, line 186: Some of these tests might need updates.

juliohm (commented):
@souma4 I left a few more comments attached. I can continue the review later after some rest.

src/hulls/jarvis.jl
souma4: open interval.

src/hulls/jarvis.jl
souma4: It's because the k-nearest approach might create a valid polygon, but might not create a valid hull (all points within polygon). The best example is if you have two clusters of points. If k is too small, you'll only circle one of the clusters. Larger k allows jumping the gap. IMO, if someone wants to create hulls around clusters, that should be done prior to calling the hull. For the `Nothing` method, the algorithm guarantees validity. Want me to write a short part into the comment, probably near the method for `validatehull`?

src/hulls/jarvis.jl
souma4: should be `k < n`,

src/hulls/jarvis.jl
juliohm: This comment needs attention.

souma4: Ah I realize it was a pending comment. I'll elaborate on the interval more here. It's an open interval thus strict inequalities are appropriate `2 < x < n`. If `k==2` you just get self intersections.

src/hulls/jarvis.jl
juliohm: I believe this is another instance where we are doing more than what was requested by the user. The algorithm is well-defined for any valid `k>2`. Even though the result might not include all points, that might be desirable in some exotic application? If the same user wants to perform checks later on to ensure that all points are included, then we leave it to him/her. Throwing an error inside the function with a particular condition seems unnecessary. I will remove the specific validity check and will share a final round of review.

src/hulls/jarvis.jl
```diff
@@ -31,42 +50,90 @@ function hull(points, ::JarvisMarch)
...
 end
+
+# helper to create searcher and mask of visited points
+jarvissearcher(p, k::Nothing) = nothing, nothing
+jarvissearcher(p, k::Integer) = KNearestSearch(p, k), falses(length(p))
+
+# helper to get candidate indices for next point
+jarviscandidates!(searcher::Nothing, visited, p, ℐ) = setdiff(1:length(p), last(ℐ, 2))
+function jarviscandidates!(searcher::KNearestSearch, visited, p, ℐ)
+  mask = .!visited
```
juliohm: Notice how we are allocating a new array at every step of the algorithm. Consider replacing the variable `visited` by the complement `unseen`. That should avoid allocation, and the update will just be a local update `unseen[last(ℐ, 2)] .= false`.

src/hulls/jarvis.jl
```diff
@@ -31,42 +50,90 @@ function hull(points, ::JarvisMarch)
...
+    tₒ = nᵢ == ℐ[begin] ? 2 : 1
+    valid = !any(tₒ:(length(ℐ) - 2)) do t
+      eseg = Segment(p[ℐ[t]], p[ℐ[t + 1]])
+      # quick check to see if segments could intersect before doing more expensive segment intersection check
+      intersects(cbox, boundingbox(eseg)) && intersects(cseg, eseg)
+    end
+    valid && return nᵢ
+  end
+  nothing
+end
```
juliohm, line 138: I confess that I don't understand why this function is so complex. My original understanding of the Jarvis's march algorithm with integer `k` was that it was exactly the same algorithm with the only difference being the set `𝒞` of candidates. Why do we need a custom implementation when `𝒞` already has the valid k-nearest candidates? Are you trying to do something more with extra validity checks here too?

juliohm (changes requested):
Thank you for updating the PR and comments @souma4.

Attached a final round of review.

src/hulls/jarvis.jl
souma4: So one issue with the limited set of candidates is that new segments may intersect existing segments, thus creating an invalid polygon. That's basically all this function does. It checks the bounding box intersection, then the proper intersection from a proposed segment to all existing segments.

juliohm: I will try to review this part by tomorrow, assuming that the authors of the paper discuss this issue of self-intersection, etc. I thought that reducing the set of candidates was enough given that they are still ordered by angle.

souma4: In all fairness, it is annoying this issue comes up. A lot of the tests don't fail when the intersection check is removed. I've added the paper's example into the tests. Page 64 has the two edge cases (self-intersections and clusters). We probably should add back in the `validatehull` so that people don't make a concave hull that excludes points. Users would expect all points to be in the hull.

juliohm: IMO this is a matter of definition. You are constantly switching back and forth between the raw JarvisMarch(k) algorithm and the higher-level `concavehull` (to be added in the future) that computes the concave hull with additional requirements (e.g., all points within ring). This PR is about the core algorithm, which leads to results that might not full-fill the requirements of the the `concavehull` utility. The utility will require arbitrary heuristics to enforce behavior, but that shouldn't be the job of the underlying JarvisMarch(k) engine.

Alternatively, we could consider a completely decoupled implementation with a different struct name, implemented with some code repetition plus additional validity checks. I just want to double check the necessity of these validity checks before proceeding with further suggestions. I will use the afternoon to read the cited paper to see if I can understand the challenges more deeply.

souma4: It is a matter of definition, that's for sure, however I'm defining a "hull" as "the smallest shape that encloses a given set of points in a plane or higher dimensions." Because this is a `hull` operation, my thinking is those should be validated for things like "all points included." I do agree that having a higher level `concavehull` for different heuristics is valuable. But that, IMO, is for things like "do we draw around clusters? do we iterate k until a success? etc." I do see a slight argument against validating all points are within the hull because maybe, for whatever reason, someone wants to draw a hull that excludes some points. However, I think that's moving away from a `hull`.

I guess, all this to say, I am worried someone may expect `hull` to contain all points, and without that final validation check, they may not get that result they wanted.

juliohm: The problem is that the performed validation checks are not correcting the issue, they are just throwing an error. Ideally, we would use a modified algorithm to circumvent the issue and restore a valid hull.

I've read the paper again and their proposed algorithm differs from Jarvis's march considerably:

Image

In this PR we attempted to reuse the `JarvisMarch` code assuming that the algorithms were similar, but they are not. Should we close this PR in favor of a completely independent `MoreiraSantosMarch` struct that implements their algorithm line-by-line?

souma4:
closed for `MoreiraSantosMarch` implementation.

Issue #1350, opened 2026-04-09
Review h-adaptive integration defaults


juliohm:
The following MWE takes too long and shows warnings about maximum subdivision:

```julia
fname = "zone.gpkg"

zone = GeoIO.load(fname)

sum(area, zone.geometry)
```
```julia
┌ Warning: maximum number of subdivisions reached `maxsubdiv=32768`, try increasing the keyword argument `maxsubdiv`.
└ @ HAdaptiveIntegration ~/.julia/packages/HAdaptiveIntegration/f6CYW/src/integrate.jl:114
┌ Warning: maximum number of subdivisions reached `maxsubdiv=32768`, try increasing the keyword argument `maxsubdiv`.
└ @ HAdaptiveIntegration ~/.julia/packages/HAdaptiveIntegration/f6CYW/src/integrate.jl:114
2.900380628497628e12 m^2
```

The file `zone.gpkg` can be downloaded [here](https://github.com/JuliaEarth/GeoStats.jl/blob/267dfbcb76932a222cf882fe1c239033839f3b16/docs/src/data/zone.gpkg).

@zmoitier do you have suggestions for better defaults? We are already setting `rtol` and `atol`, but the subdivision is too aggressive:

https://github.com/JuliaGeometry/Meshes.jl/blob/f282c2ede0903519151ed04d3a9b1dcf1a5af893/src/integration.jl#L5-L9

Our `rtol` and `atol` values are

```julia
julia> Meshes.rtol(Float64), Meshes.atol(Float64)
(6.055454452393343e-6, 1.8189894035458565e-12)
```

juliohm:
To give more context, the geometries are zones in the Amazon forest:

<img width="672" height="548" alt="Image" src="https://github.com/user-attachments/assets/d634cc96-3c22-4a57-918d-5f008124e471" />

Our `area` function is defined as `area(g) = integral(_ -> 1, g)`. So we are looking for good defaults to integrate the constant function `1` over these kinds of polygons.

juliohm:
Reduced the MWE to the 3rd geometry in the list:

```julia
julia> area(zone.geometry[3])
┌ Warning: maximum number of subdivisions reached `maxsubdiv=32768`, try increasing the keyword argument `maxsubdiv`.
└ @ HAdaptiveIntegration ~/.julia/packages/HAdaptiveIntegration/f6CYW/src/integrate.jl:114
1.1102012597756082e12 m^2
```

[...]

PR #1352, opened 2026-04-09
Refactor discretization in terms of refinemaxlen

juliohm:
The refinement is explicit now to avoid over-discretization. Now that we have efficient `measure` for non-Euclidean geometries, we don't need to refine the meshes to get accurate results.


juliohm:
@JoshuaLampert with this PR we get an additional 1.3x speedup with the area calculation in #1350.

juliohm:
I will go ahead and merge this so that it can go in a patch release.

PR #1355, opened 2026-04-09
Refactor vizgset for performance

juliohm:
And to reproduce previous behavior in v0.56


juliohm:
Failure on julia-lts-windows-latest-Float64 is random and unrelated. Merging it.

PR #1357, opened 2026-04-11
Replace LScene by Axis3 (attempt 2)

juliohm:
Based on https://github.com/MakieOrg/Makie.jl/pull/5375

Works as expected, but it would be nice if we could define the `preferred_axis_attributes` for the object being plotted like we already do with `args_preferred_axis`. If we can specialize the method for the object, we can extract additional information like the units along each axis to place in the labels.


juliohm:
Finally! 🎉

PR #1360, opened 2026-04-13
Use `Base.setindex` instead of `ntuple` in `derivative`

### LINKED ISSUE #1353: Enzyme support?
Hey folks,

Cool package!

I was wondering if Enzyme support was on the horizon (I vageuly recall it being supported/tested in MeshIntergrals, but seemingly not here?)

JoshuaLampert:
This fixes #1353. @juliohm, should we add some tests with other differentiation backends? Apparently, their support depends on the code in Meshes.jl.


JoshuaLampert:
The `Base.setindex` version is even much faster.
Old version:
```julia
julia> @benchmark jacobian(seg, (0.1,))
BenchmarkTools.Trial: 10000 samples with 99 evaluations per sample.
 Range (min … max):  787.343 ns … 387.779 μs  ┊ GC (min … max): 0.00% … 99.35%
 Time  (median):     868.333 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   928.611 ns ±   3.870 μs  ┊ GC (mean ± σ):  4.15% ±  0.99%

    ▆█▂                                                          
  ▁▅███▅▄▃▃▃▂▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▂▃▄▄▆▆▇▇▇▆▅▄▃▃▂▂▂▁▁▁▁▁▁▁▁▁▁ ▂
  787 ns           Histogram: frequency by time         1.05 μs <

 Memory estimate: 680 bytes, allocs estimate: 22.

julia> @benchmark jacobian(seg, (0.1,), dbackend = DI.AutoEnzyme(mode = Enzyme.Forward))
BenchmarkTools.Trial: 10000 samples with 550 evaluations per sample.
 Range (min … max):  211.669 ns …  59.655 μs  ┊ GC (min … max): 0.00% … 99.48%
 Time  (median):     217.299 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   260.750 ns ± 873.710 ns  ┊ GC (mean ± σ):  7.71% ±  2.43%

  ▅█▇▅▃▂▂▂▁▁▁                 ▂▄▅▅▄▃▂▁▁▂▃▃▂▁                    ▂
  ███████████▇▇▇▆▆▅▅▅▄▃▄▄▃▃▂▂▇███████████████▇▇▆▆▅▅▅▅▅▄▄▄▃▅▄▄▄▄ █
  212 ns        Histogram: log(frequency) by time        335 ns <

 Memory estimate: 224 bytes, allocs estimate: 8.
```
This PR:
```julia
julia> @benchmark jacobian(seg, (0.1,))
BenchmarkTools.Trial: 10000 samples with 998 evaluations per sample.
 Range (min … max):  15.765 ns … 267.128 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     23.042 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   21.256 ns ±   8.595 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

     ▃█▇▁                                ▂▆▅                    
  ▁▂▄████▅▃▂▂▂▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▂▄▄▄▅████▅▃▃▄▆▅▄▂▂▁▁▁▁▁▁▁▁ ▂
  15.8 ns         Histogram: frequency by time         27.3 ns <

 Memory estimate: 32 bytes, allocs estimate: 1.

julia> @benchmark jacobian(seg, (0.1,), dbackend = DI.AutoEnzyme(mode = Enzyme.Forward))
BenchmarkTools.Trial: 10000 samples with 998 evaluations per sample.
 Range (min … max):  17.212 ns …  22.299 μs  ┊ GC (min … max): 0.00% … 99.80%
 Time  (median):     19.752 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   24.162 ns ± 222.886 ns  ┊ GC (mean ± σ):  9.21% ±  1.00%

      ▆█▄                                     ▁                 
  ▂▂▄▇███▇▄▃▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▁▂▂▂▂▂▂▃▄▄▄▅▇██▆▅▄▃▄▄▅▄▃▃▂▂▂▂ ▃
  17.2 ns         Histogram: frequency by time         27.7 ns <

 Memory estimate: 32 bytes, allocs estimate: 1.

julia> @benchmark jacobian(seg, (0.1,), dbackend = DI.AutoEnzyme(mode = Enzyme.Reverse))
BenchmarkTools.Trial: 10000 samples with 983 evaluations per sample.
 Range (min … max):  58.260 ns …  37.359 μs  ┊ GC (min … max):  0.00% … 99.42%
 Time  (median):     64.199 ns               ┊ GC (median):     0.00%
 Time  (mean ± σ):   88.754 ns ± 571.898 ns  ┊ GC (mean ± σ):  16.96% ±  2.81%

   ▅▇█▇▇▆▅▄▃▃▂▁▂▁          ▁      ▁▅▆▆▅▅▄▃▁                    ▂
  ▇████████████████▇▇▇▇▇▇▅████▇▆▆▇████████████▇▆▅▅▄▂▆▅▅▅▅▄▄▄▅▅ █
  58.3 ns       Histogram: log(frequency) by time       119 ns <

 Memory estimate: 160 bytes, allocs estimate: 5.
```

juliohm:
Thank you Joshua! That is nice!

I have mixed feelings about the sensitivity of AD to basic Julia
constructs...

Adding tests for Enzyme would be the most reasonable thing to do. My main
concern is if some of these AD backends get outdated and/or unmaintained.
We don't have control over that though.

Should we add a test set for any other backend besides Enzyme? I think we
can simply eliminate the test set in the future if the package is
deprecated?

Em seg., 13 de abr. de 2026, 11:28, Joshua Lampert ***@***.***>
escreveu:

JoshuaLampert:
> Adding tests for Enzyme would be the most reasonable thing to do. My main
concern is if some of these AD backends get outdated and/or unmaintained.
We don't have control over that though.

Yes, in the worst case we can always remove the test again, but for now I think it's good to have a minimal test (probably not all the tests for each geometry, but only one geometry) for Enzyme.jl. I main concern is a possibly increased CI time when we add Enzyme.jl to the test environment because it takes some time to compile. But I can add a testset and see how much CI time is increased.

> Should we add a test set for any other backend besides Enzyme? I think we
can simply eliminate the test set in the future if the package is
deprecated?

I am also not sure. Having tests for Mooncake.jl could be nice, but maybe something for another PR if we want that.

JoshuaLampert:
I added a basic test for Enzyme forward and reverse mode. The additional CI time doesn't seem to be so bad :+1:
I also added a few items for differentiation and integration to the benchmarks. We can see a really nice speedup and a lot fewer allocations here :rocket:

test/differentiation.jl
```diff
@@ -10,6 +13,11 @@
   @test Jr[1] ≈ b - a
   @test Jm[1] ≈ b - a
 
+  Jm = jacobian(seg, (T(0.5),); dbackend=DI.AutoEnzyme(mode=Enzyme.Forward))
+  @test Jm[1] ≈ b - a
+  Jm = jacobian(seg, (T(0.5),); dbackend=DI.AutoEnzyme(mode=Enzyme.Reverse))
+  @test Jm[1] ≈ b - a
+
```
juliohm: Let's create a separate `testitem` for these backend tests so that we can easily remove in the future:

```julia
@testitem "dbackend" setup = [Setup] begin
  # tests go here
end
```

test/differentiation.jl
```diff
@@ -1,4 +1,7 @@
 @testitem "Jacobian" setup = [Setup] begin
+  import DifferentiationInterface as DI
+  import Enzyme
```
juliohm: Please move all imports to runtests.jl

juliohm (changes requested):
Thank you @JoshuaLampert !

One thing I noticed is that `Base.setindex`'s docstring has a warning:

```julia
help?> Base.setindex
  │ Warning
  │
  │  The following bindings may be internal; they may change or be removed
  │  in future versions:
  │
  │    •  Base.setindex
```

Pasting the warning here for future reference.

juliohm:
I am amazed by the huge speedups! 🚀 and a perplexed at the same time by how such a little change in the code matters for AD.

test/differentiation.jl
JoshuaLampert: Done in [d1045a6](https://github.com/JuliaGeometry/Meshes.jl/pull/1360/commits/d1045a6012ff570051a72c46a49cd86adc16cc83).

test/differentiation.jl
JoshuaLampert: Done in [d1045a6](https://github.com/JuliaGeometry/Meshes.jl/pull/1360/commits/d1045a6012ff570051a72c46a49cd86adc16cc83).

juliohm: I believe you forgot to remove these lines that were moved to a dedicated test item?

JoshuaLampert: :man_facepalming: You are right. Sorry Should be fixed now.

test/differentiation.jl
```diff
@@ -57,6 +62,17 @@
   @test Jc[2] ≈ ((c₁₀ - c₀₀) + (c₁₁ - c₀₁)) / 2
 end
 
+@testitem "Jacobian with Enzyme.jl" setup = [Setup] begin
```
juliohm:
```suggestion
@testitem "Tests with Enzyme.jl" setup = [Setup] begin
```

Please move this testitem to the end of the file. We can append more backends next to it in the future.

JoshuaLampert: I moved the tests to the end of the file and also added tests for `differential`. Isn't "Tests" too generic in the Name of the `@testitem`?

juliohm: > Isn't "Tests" too generic in the Name of the `@testitem`?

I thought of this testitem as "everything related to dbackend=Enzyme.jl". Jacobian, differential, etc. It currently includes differential tests too.

JoshuaLampert: Yeah, I added the `differential` tests because I thought that's your reasoning :sweat_smile:. Renamed in [68dcbdc](https://github.com/JuliaGeometry/Meshes.jl/pull/1360/commits/68dcbdccff7a72988a98f41305a9d9a272aeb985).

juliohm:
Test failing on Windows for unrelated random issue on GitHub CI.

juliohm (approved):

PR #1363, opened 2026-05-16
Add ComponentsPartition

### LINKED ISSUE #1196: Add new `PartitionMethod` for connected components, `ConnectedComponentsPartition`
Not an immediate priority for me, but opening to register the idea.

This function returns an vector of index vectors, where each vector is a connected component  https://github.com/JuliaGeometry/Meshes.jl/blob/49654506e8871f237adcf6a403bae6535200930f/src/topologies/halfedge.jl#L340

The `Partition` takes exactly such a `Vector{Vector{Int}}`, so adding a new partition method should be straightforward
https://github.com/JuliaGeometry/Meshes.jl/blob/49654506e8871f237adcf6a403bae6535200930f/src/partitions.jl#L11-L15

juliohm:
Fix #1196

### DIFF
--- a/docs/src/algorithms/partitioning.md
+++ b/docs/src/algorithms/partitioning.md
@@ -23,10 +23,10 @@ FractionPartition
 BlockPartition
 ```
 
-## Connected
+## Components
 
 ```@docs
-ConnectedPartition
+ComponentsPartition
 ```
 
 ## Bisect-Point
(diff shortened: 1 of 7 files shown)


juliohm:
The failure on Windows-Julia-LTS-Float64 is the same random failure that we saw in other PRs.

Merging it.

PR #1366, opened 2026-05-30
Use Makie's ComputeGraph instead of @lift

### LINKED ISSUE #1364: Use Makie's `ComputeGraph` to avoid race conditions in `viz` recipes
Consider the following MWE:

```julia
using Meshes

import GLMakie as Mke

boxes = Mke.Observable(rand(Box, 2))

viz(boxes)

boxes[] = rand(Box, 3)
```
```julia
ERROR: Failed to verify position attribute:
Faces address 72 vertex attributes but only 24 are present.
Stacktrace:
  [1] error(s::String)
    @ Base ./error.jl:44
  [2] GeometryBasics.Mesh(vertex_attributes::@NamedTuple{…}, fs::Vector{…}, views::Vector{…})
    @ GeometryBasics ~/.julia/packages/GeometryBasics/yB1f1/src/basic_types.jl:550
  [3] #Mesh#148
    @ ~/.julia/packages/GeometryBasics/yB1f1/src/basic_types.jl:703 [inlined]
  [4] Mesh
    @ ~/.julia/packages/GeometryBasics/yB1f1/src/basic_types.jl:699 [inlined]
  [5] (::MeshesMakieExt.var"#vizmesh!##56#vizmesh!##57")(arg1#305::Vector{…}, arg2#306::Vector{…})
    @ MeshesMakieExt ./none:-1
  [6] (::Observables.MapCallback)(value::Any)
    @ Observables ~/.julia/packages/Observables/YdEbO/src/Observables.jl:436
  [7] notify
    @ ~/.julia/packages/Observables/YdEbO/src/Observables.jl:206 [inlined]
  [8] setindex!(observable::Observables.Observable, val::Any)
    @ Observables ~/.julia/packages/Observables/YdEbO/src/Observables.jl:123
  [9] (::Observables.MapCallback)(value::Any)
    @ Observables ~/.julia/packages/Observables/YdEbO/src/Observables.jl:436
 [10] notify
    @ ~/.julia/packages/Observables/YdEbO/src/Observables.jl:206 [inlined]
 [11] setindex!(observable::Observables.Observable, val::Any)
    @ Observables ~/.julia/packages/Observables/YdEbO/src/Observables.jl:123
 [12] (::Observables.MapCallback)(value::Any)
    @ Observables ~/.julia/packages/Observables/YdEbO/src/Observables.jl:436
 [13] notify
    @ ~/.julia/packages/Observables/YdEbO/src/Observables.jl:206 [inlined]
 [14] (::ComputePipeline.var"#ComputeGraph##0#ComputeGraph##1"{ComputePipeline.ComputeGraph})(changeset::Set{Symbol})
    @ ComputePipeline ~/.julia/packages/ComputePipeline/30b0T/src/ComputePipeline.jl:351
 [15] notify
    @ ~/.julia/packages/Observables/YdEbO/src/Observables.jl:206 [inlined]
 [16] foreach(f::typeof(notify), itr::Vector{Observables.Observable})
    @ Base ./abstractarray.jl:3188
 [17] #34
    @ ~/.julia/packages/ComputePipeline/30b0T/src/ComputePipeline.jl:501 [inlined]
 [18] lock(f::ComputePipeline.var"#34#35"{ComputePipeline.ComputeGraph, Symbol, SimpleMesh{…}}, l::ReentrantLock)
    @ Base ./lock.jl:335
 [19] setproperty!
    @ ~/.julia/packages/ComputePipeline/30b0T/src/ComputePipeline.jl:499 [inlined]
 [20] (::ComputePipeline.var"#add_input!##0#add_input!##1"{…})(new_val::SimpleMesh{…})
    @ ComputePipeline ~/.julia/packages/ComputePipeline/30b0T/src/ComputePipeline.jl:807
 [21] notify
    @ ~/.julia/packages/Observables/YdEbO/src/Observables.jl:206 [inlined]
 [22] setindex!(observable::Observables.Observable, val::Any)
    @ Observables ~/.julia/packages/Observables/YdEbO/src/Observables.jl:123
 [23] (::Observables.MapCallback)(value::Any)
    @ Observables ~/.julia/packages/Observables/YdEbO/src/Observables.jl:436
 [24] notify
    @ ~/.julia/packages/Observables/YdEbO/src/Observables.jl:206 [inlined]
 [25] setindex!(observable::Observables.Observable, val::Any)
    @ Observables ~/.julia/packages/Observables/YdEbO/src/Observables.jl:123
 [26] (::Observables.MapCallback)(value::Any)
    @ Observables ~/.julia/packages/Observables/YdEbO/src/Observables.jl:436
 [27] notify
    @ ~/.julia/packages/Observables/YdEbO/src/Observables.jl:206 [inlined]
 [28] setindex!(observable::Observables.Observable, val::Any)
    @ Observables ~/.julia/packages/Observables/YdEbO/src/Observables.jl:123
 [29] (::Observables.MapCallback)(value::Any)
    @ Observables ~/.julia/packages/Observables/YdEbO/src/Observables.jl:436
 [30] notify
    @ ~/.julia/packages/Observables/YdEbO/src/Observables.jl:206 [inlined]
 [31] setindex!(observable::Observables.Observable, val::Any)
    @ Observables ~/.julia/packages/Observables/YdEbO/src/Observables.jl:123
 [32] (::Observables.MapCallback)(value::Any)
    @ Observables ~/.julia/packages/Observables/YdEbO/src/Observables.jl:436
 [33] notify
    @ ~/.julia/packages/Observables/YdEbO/src/Observables.jl:206 [inlined]
 [34] setindex!(observable::Observables.Observable, val::Any)
    @ Observables ~/.julia/packages/Observables/YdEbO/src/Observables.jl:123
 [35] (::Observables.MapCallback)(value::Any)
    @ Observables ~/.julia/packages/Observables/YdEbO/src/Observables.jl:436
 [36] notify
    @ ~/.julia/packages/Observables/YdEbO/src/Observables.jl:206 [inlined]
 [37] (::ComputePipeline.var"#ComputeGraph##0#ComputeGraph##1"{ComputePipeline.ComputeGraph})(changeset::Set{Symbol})
    @ ComputePipeline ~/.julia/packages/ComputePipeline/30b0T/src/ComputePipeline.jl:351
 [38] notify
    @ ~/.julia/packages/Observables/YdEbO/src/Observables.jl:206 [inlined]
 [39] foreach(f::typeof(notify), itr::Vector{Observables.Observable})
    @ Base ./abstractarray.jl:3188
 [40] #34
    @ ~/.julia/packages/ComputePipeline/30b0T/src/ComputePipeline.jl:501 [inlined]
 [41] lock(f::ComputePipeline.var"#34#35"{ComputePipeline.ComputeGraph, Symbol, Vector{Box{…}}}, l::ReentrantLock)
    @ Base ./lock.jl:335
 [42] setproperty!
    @ ~/.julia/packages/ComputePipeline/30b0T/src/ComputePipeline.jl:499 [inlined]
 [43] (::ComputePipeline.var"#add_input!##0#add_input!##1"{ComputePipeline.ComputeGraph, Symbol})(new_val::Vector{Box{…}})
    @ ComputePipeline ~/.julia/packages/ComputePipeline/30b0T/src/ComputePipeline.jl:807
 [44] notify
    @ ~/.julia/packages/Observables/YdEbO/src/Observables.jl:206 [inlined]
 [45] setindex!(observable::Observables.Observable, val::Any)
    @ Observables ~/.julia/packages/Observables/YdEbO/src/Observables.jl:123
 [46] top-level scope
    @ REPL[14]:1
```

When we update the vector of boxes from length 2 to length 3, the `Observable` takes care of automatically updating the internal mesh. However, due to some race condition, the corresponding vector of colors is not updated in time. As a result, the function below is called with an updated `mesh` and an outdated `colorant`:

https://github.com/JuliaGeometry/Meshes.jl/blob/ff3eff1a45fce991dca573211ec6d5e5a651d537/ext/mesh.jl#L59

Makie solved this race condition issue in recent releases with the introduction of `ComputeGraph`s:

https://docs.makie.org/stable/explanations/compute-pipeline#ComputeGraph

We need to rewrite our recipes to replace `Makie.@lift` calls by `map!` calls with a compute graph when possible.

cserteGT3:
I've asked Gemini to update the code (replace `@lift`s with `map!`) but I am still hesitant about this commit/PR. On one hand the issue is solved: geometries (and their color) can be updated without issues. On the other hand, I did not review the changes (other than scrolling through the files), because I don't understand the internals of the visualization code.

To be clear, I'm not asking you to review the LLM-generated code, that I generated, but want to start this discussion. I don't think I would be able to update this code manually, so I would rely on the LLM generated code. I was thinking on writing tests for visualizing all kind of geometries so we could verify that this works as expected, but I would ask for some input for that as well.

### DIFF
--- a/.github/workflows/Downstream.yml
+++ b/.github/workflows/Downstream.yml
@@ -43,15 +43,13 @@ jobs:
       - uses: julia-actions/cache@v3
       - uses: julia-actions/julia-buildpkg@v1
       - name: Retrieve downstream package
-        # Note: we retrieve the current `main` branch of the downstream package (MeshesVizTests.jl) to ensure
+        # Note: we retrieve the current `main` branch of the downstream package to ensure
         # that compatibility errors we make in Meshes.jl are detected already here
         # See also https://github.com/trixi-framework/Trixi.jl/pull/1707#discussion_r1382938895
         uses: actions/checkout@v6
         with:
           repository: JuliaGeometry/${{ matrix.package }}
           path: downstream
-        # Note: this script loads the upstream (to-be-tested) version of Meshes.jl
-        # See also https://github.com/JuliaGeometry/Meshes.jl/pull/1366#issuecomment-4605948374
       - name: Load upstream package into downstream environment
         shell: julia --color=yes --project=downstream {0}
         run: |
--- a/.github/workflows/Downstream.yml
+++ b/.github/workflows/Downstream.yml
@@ -43,13 +43,15 @@ jobs:
       - uses: julia-actions/cache@v3
       - uses: julia-actions/julia-buildpkg@v1
       - name: Retrieve downstream package
-        # Note: we retrieve the current `main` branch of the downstream package to ensure
+        # Note: we retrieve the current `main` branch of the downstream package (MeshesVizTests.jl) to ensure
         # that compatibility errors we make in Meshes.jl are detected already here
         # See also https://github.com/trixi-framework/Trixi.jl/pull/1707#discussion_r1382938895
         uses: actions/checkout@v6
         with:
           repository: JuliaGeometry/${{ matrix.package }}
           path: downstream
+        # Note: this script loads the upstream (to-be-tested) version of Meshes.jl
+        # See also https://github.com/JuliaGeometry/Meshes.jl/pull/1366#issuecomment-4605948374
       - name: Load upstream package into downstream environment
         shell: julia --color=yes --project=downstream {0}
         run: |
--- a/ext/geomset.jl
+++ b/ext/geomset.jl
@@ -2,20 +2,16 @@
 # Licensed under the MIT License. See LICENSE in the project root.
 # ------------------------------------------------------------------
 
-Makie.plot!(plot::Viz{<:Tuple{GeometrySet}}) = vizgset!(plot)
+function Makie.plot!(plot::Viz{<:Tuple{GeometrySet}})
+  Makie.map!(process, plot, [:color, :colormap, :colorrange, :alpha], :colorant)
+  vizgset!(plot)
+end
 
 # split heterogeneous geometry sets into homogeneous vectors
 # of geometries and send these vectors to specialized recipes
-function vizgset!(plot; facets=false)
-  gset = plot.object[]
-
-  # process color spec into colorant
-  if !facets
-    Makie.map!(process, plot, [:color, :colormap, :colorrange, :alpha], :colorant)
-  end
-
+function vizgset!(plot)
   # get geometries
-  geoms = parent(gset)
+  geoms = parent(plot.object[])
 
   # get geometry types
   types = unique(map(typeof, geoms))
@@ -23,29 +19,20 @@ function vizgset!(plot; facets=false)
   for (i, G) in enumerate(types)
     inds_sym = Symbol(:inds_, i)
     gvec_sym = Symbol(:gvec_, i)
+    cvec_sym = Symbol(:cvec_, i)
 
-    Makie.map!(plot, [:object], inds_sym) do g
-      findall(x -> x isa G, parent(g))
-    end
-
-    Makie.map!(plot, [:object, inds_sym], gvec_sym) do g, inds
-      collect(G, parent(g)[inds])
+    Makie.map!(plot, [:object, :colorant], [inds_sym, gvec_sym, cvec_sym]) do g, colorant
+      inds = findall(x -> x isa G, parent(g))
+      gvec = collect(G, parent(g)[inds])
+      cvec = colorant isa AbstractVector ? colorant[inds] : fill(colorant, length(inds))
+      inds, gvec, cvec
     end
 
     gvec = collect(G, geoms[findall(x -> x isa G, geoms)])
     M = manifold(first(gvec))
     pdim = paramdim(first(gvec))
     edim = embeddim(first(gvec))
-
-    if facets
-      vizgsetfacets!(plot, M, Val(pdim), Val(edim), G, gvec_sym)
-    else
-      cvec_sym = Symbol(:cvec_, i)
-      Makie.map!(plot, [:colorant, inds_sym], cvec_sym) do colorant, inds
-        colorant isa AbstractVector ? colorant[inds] : fill(colorant, length(inds))
-      end
-      vizgset!(plot, M, Val(pdim), Val(edim), G, gvec_sym, cvec_sym)
-    end
+    vizgset!(plot, M, Val(pdim), Val(edim), G, gvec_sym, cvec_sym)
   end
 end
 
@@ -269,7 +256,7 @@ end
 # FACETS
 # -------
 
-vizfacets!(plot::Viz{<:Tuple{GeometrySet}}) = vizgset!(plot, facets=false)
+vizfacets!(plot::Viz{<:Tuple{GeometrySet}}) = vizfacets!(plot)
 
 function vizfacets!(plot::Viz{<:Tuple{GeometrySet}}, G::Type, geoms::Symbol)
   gvec = plot[geoms][]
--- a/ext/geomset.jl
+++ b/ext/geomset.jl
@@ -10,27 +10,27 @@ end
 # split heterogeneous geometry sets into homogeneous vectors
 # of geometries and send these vectors to specialized recipes
 function vizgset!(plot)
-  # get geometries
+  # retrieve geometries and their types
   geoms = parent(plot.object[])
-
-  # get geometry types
   types = unique(map(typeof, geoms))
 
   for (i, G) in enumerate(types)
-    gvec_sym = Symbol(:gvec_, i)
-    cvec_sym = Symbol(:cvec_, i)
-
-    Makie.map!(plot, [:object, :colorant], [gvec_sym, cvec_sym]) do g, colorant
-      inds = findall(x -> x isa G, parent(g))
-      gvec = collect(G, parent(g)[inds])
+    # add nodes to compute graph for this type of geometry
+    gvecid = Symbol(:geoms, i)
+    cvecid = Symbol(:colors, i)
+    Makie.map!(plot, [:object, :colorant], [gvecid, cvecid]) do gset, colorant
+      geos = parent(gset)
+      inds = findall(x -> x isa G, geos)
+      gvec = collect(G, geos[inds])
       cvec = colorant isa AbstractVector ? colorant[inds] : fill(colorant, length(inds))
       gvec, cvec
     end
-    gvec = plot[gvec_sym][]
-    M = manifold(first(gvec))
-    pdim = paramdim(first(gvec))
-    edim = embeddim(first(gvec))
-    vizgset!(plot, M, Val(pdim), Val(edim), G, gvec_sym, cvec_sym)
+
+    # dispatch specialized recipe for this type of geometry
+    M = manifold(G)
+    pdim = paramdim(G)
+    edim = embeddim(G)
+    vizgset!(plot, M, Val(pdim), Val(edim), G, gvecid, cvecid)
   end
 end
 
--- a/ext/geomset.jl
+++ b/ext/geomset.jl
@@ -7,7 +7,7 @@ Makie.plot!(plot::Viz{<:Tuple{GeometrySet}}) = vizgset!(plot)
 # split heterogeneous geometry sets into homogeneous vectors
 # of geometries and send these vectors to specialized recipes
 function vizgset!(plot; facets=false)
-  gset = plot[:object][]
+  gset = plot.object[]
 
   # process color spec into colorant
   if !facets
@@ -55,8 +55,8 @@ end
 
 # fallback to visualization of discretized geometries
 function vizgset!(plot, M::Type, pdim::Val, ::Val, G::Type{<:Geometry}, geoms::Symbol, colors::Symbol)
-  showsegments = plot[:showsegments]
-  showpoints = plot[:showpoints]
+  showsegments = plot.showsegments
+  showpoints = plot.showpoints
 
   # refine meshes over the 🌐 manifold until
   # they satisfy the maximum length criterion
@@ -115,8 +115,8 @@ function vizgset!(plot, M::Type, pdim::Val, edim::Val, G::Type{<:Multi}, geoms::
 end
 
 function vizgset!(plot, ::Type, ::Val, ::Val, G::Type{<:Point}, geoms::Symbol, colors::Symbol)
-  pointmarker = plot[:pointmarker]
-  pointsize = plot[:pointsize]
+  pointmarker = plot.pointmarker
+  pointsize = plot.pointsize
 
   coords_sym = Symbol(geoms, :_coords)
 
@@ -130,8 +130,7 @@ function vizgset!(plot, ::Type, ::Val, ::Val, G::Type{<:Point}, geoms::Symbol, c
 end
 
 function vizgset!(plot, ::Type, ::Val, edim::Val, G::Type{<:Ray}, geoms::Symbol, colors::Symbol)
-  segmentsize = plot[:segmentsize]
-  showpoints = plot[:showpoints]
+  showpoints = plot.showpoints
 
   orig_sym = Symbol(geoms, :_orig)
   dirs_sym = Symbol(geoms, :_dirs)
@@ -188,7 +187,7 @@ function vizgset!(plot, ::Type, ::Val, edim::Val, G::Type{<:Ray}, geoms::Symbol,
 end
 
 function vizgset!(plot, ::Type{<:𝔼}, ::Val, ::Val{2}, G::Type{<:Line}, geoms::Symbol, colors::Symbol)
-  segmentsize = plot[:segmentsize]
+  segmentsize = plot.segmentsize
 
   inter_sym = Symbol(geoms, :_inter)
   vinds_sym = Symbol(geoms, :_vinds)
@@ -248,9 +247,9 @@ function vizgset!(plot, ::Type{<:𝔼}, ::Val, ::Val{2}, G::Type{<:Line}, geoms:
 end
 
 function vizgset!(plot, ::Type{<:𝔼}, ::Val{2}, ::Val{2}, G::Type{<:Polygon}, geoms::Symbol, colors::Symbol)
-  showsegments = plot[:showsegments]
-  segmentcolor = plot[:segmentcolor]
-  segmentsize = plot[:segmentsize]
+  showsegments = plot.showsegments
+  segmentcolor = plot.segmentcolor
+  segmentsize = plot.segmentsize
 
   polys_sym = Symbol(geoms, :_polys)
 
@@ -281,9 +280,9 @@ function vizfacets!(plot::Viz{<:Tuple{GeometrySet}}, G::Type, geoms::Symbol)
 end
 
 function vizgsetfacets!(plot, ::Type, ::Val{1}, ::Val, G::Type, geoms::Symbol)
-  pointmarker = plot[:pointmarker]
-  pointcolor = plot[:pointcolor]
-  pointsize = plot[:pointsize]
+  pointmarker = plot.pointmarker
+  pointcolor = plot.pointcolor
+  pointsize = plot.pointsize
 
   pset_sym = Symbol(geoms, :_pset)
 
@@ -297,8 +296,8 @@ function vizgsetfacets!(plot, ::Type, ::Val{1}, ::Val, G::Type, geoms::Symbol)
 end
 
 function vizgsetfacets!(plot, ::Type, ::Val{2}, ::Val, G::Type, geoms::Symbol)
-  segmentcolor = plot[:segmentcolor]
-  segmentsize = plot[:segmentsize]
+  segmentcolor = plot.segmentcolor
+  segmentsize = plot.segmentsize
 
   bset_sym = Symbol(geoms, :_bset)
 
--- a/ext/geomset.jl
+++ b/ext/geomset.jl
@@ -57,7 +57,7 @@ function vizgset!(plot, M::Type, pdim::Val, ::Val, G::Type{<:Geometry}, geoms::S
     end
     vizmany!(plot, plot[meshes_sym], plot[colors])
     if plot.showpoints[]
-      vizfacets!(plot, G, geoms)
+      vizfacets!(plot, geoms)
     end
   elseif pdim === Val(2)
     Makie.map!(plot, [geoms], meshes_sym) do gvec
@@ -65,7 +65,7 @@ function vizgset!(plot, M::Type, pdim::Val, ::Val, G::Type{<:Geometry}, geoms::S
     end
     vizmany!(plot, plot[meshes_sym], plot[colors])
     if plot.showsegments[]
-      vizfacets!(plot, G, geoms)
+      vizfacets!(plot, geoms)
     end
   elseif pdim === Val(3)
     Makie.map!(plot, [geoms], meshes_sym) do gvec
@@ -160,7 +160,7 @@ function vizgset!(plot, ::Type, ::Val, edim::Val, G::Type{<:Ray}, geoms::Symbol,
   end
 
   if plot.showpoints[]
-    vizfacets!(plot, G, geoms)
+    vizfacets!(plot, geoms)
   end
 end
 
@@ -243,17 +243,15 @@ end
 # FACETS
 # -------
 
-vizfacets!(plot::Viz{<:Tuple{GeometrySet}}) = vizfacets!(plot)
-
-function vizfacets!(plot::Viz{<:Tuple{GeometrySet}}, G::Type, geoms::Symbol)
+function vizfacets!(plot::Viz{<:Tuple{GeometrySet}}, geoms::Symbol)
   gvec = plot[geoms][]
   M = manifold(first(gvec))
   pdim = paramdim(first(gvec))
   edim = embeddim(first(gvec))
-  vizgsetfacets!(plot, M, Val(pdim), Val(edim), G, geoms)
+  vizgsetfacets!(plot, M, Val(pdim), Val(edim), geoms)
 end
 
-function vizgsetfacets!(plot, ::Type, ::Val{1}, ::Val, G::Type, geoms::Symbol)
+function vizgsetfacets!(plot, ::Type, ::Val{1}, ::Val, geoms::Symbol)
   pset_sym = Symbol(geoms, :_pset)
 
   # all boundaries are points or multipoints
@@ -265,7 +263,7 @@ function vizgsetfacets!(plot, ::Type, ::Val{1}, ::Val, G::Type, geoms::Symbol)
   viz!(plot, plot[pset_sym], color=plot.pointcolor, pointmarker=plot.pointmarker, pointsize=plot.pointsize)
 end
 
-function vizgsetfacets!(plot, ::Type, ::Val{2}, ::Val, G::Type, geoms::Symbol)
+function vizgsetfacets!(plot, ::Type, ::Val{2}, ::Val, geoms::Symbol)
   bset_sym = Symbol(geoms, :_bset)
 
   # all boundaries are 1D geometries
--- a/ext/geomset.jl
+++ b/ext/geomset.jl
@@ -197,14 +197,11 @@ function vizgset!(plot, ::Type{<:𝔼}, ::Val, ::Val{2}, G::Type{<:Line}, geoms:
   dcolor_sym = Symbol(colors, :_dcolor)
 
   # split vertical and non-vertical lines
-  Makie.map!(plot, [geoms], inter_sym) do gvec
-    [line ∩ Line((0, 0), (0, 1)) for line in gvec]
-  end
-  Makie.map!(plot, [inter_sym], vinds_sym) do inter
-    findall(g -> isnothing(g) || g isa Line, inter)
-  end
-  Makie.map!(plot, [geoms, vinds_sym], dinds_sym) do gvec, vinds
-    setdiff(1:length(gvec), vinds)
+  Makie.map!(plot, [geoms], [inter_sym, vinds_sym, dinds_sym]) do gvec
+    inter = [line ∩ Line((0, 0), (0, 1)) for line in gvec]
+    vinds = findall(g -> isnothing(g) || g isa Line, inter)
+    dinds = setdiff(1:length(gvec), vinds)
+    inter, vinds, dinds
   end
 
   # split colors accordingly
--- a/ext/geomset.jl
+++ b/ext/geomset.jl
@@ -167,7 +167,7 @@ function vizgset!(plot, ::Type{<:𝔼}, ::Val, ::Val{2}, ::Type{<:Line}, gvecid:
   # split colors accordingly
   vcolorid = Symbol(cvecid, :_vcolor)
   dcolorid = Symbol(cvecid, :_dcolor)
-  Makie.map!(plot, [cvecid, vindsid, dindsid], vcolorid) do cvec, vinds, dinds
+  Makie.map!(plot, [cvecid, vindsid, dindsid], [vcolorid, dcolorid]) do cvec, vinds, dinds
     cvec[vinds], cvec[dinds]
   end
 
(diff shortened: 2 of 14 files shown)


juliohm:
Thank you @cserteGT3 ! Could you please rebase the PR? If you could fix the build failure and non-viz tests, I can try to help with additional review.

cserteGT3:
Did the rebase, tests pass now (except building the docs).  Though I have a question about MeshesVizTests.jl: if I saw the logs correctly, it uses the master branch, which means that not the code in this PR is tested. I guess that is intended, though I don't understand why?

JoshuaLampert:
> Though I have a question about MeshesVizTests.jl: if I saw the logs correctly, it uses the master branch, which means that not the code in this PR is tested. I guess that is intended, though I don't understand why?

Indeed. That looks unintended. It's a bit confusing because the template I used for the GitHub Action works as expected, see, e.g., [here](https://github.com/trixi-framework/Trixi.jl/actions/runs/26823926205/job/79085869092#step:9:15) and the workflow files look almost identical. My best guess is that it's because of the `[sources]` section in MeshesVizTests.jl pointing to the master branch of Meshes.jl: https://github.com/JuliaGeometry/MeshesVizTests.jl/blob/dbb4178b63b284afba1c9c73edc0492600fc746a/test/Project.toml#L16.

juliohm:
Do you have suggestions to modify the action or Project.toml to always run
MeshesVizTests.jl with the PR branch? I think that would be the desired
behavior, right?

Em ter., 2 de jun. de 2026, 15:15, Joshua Lampert ***@***.***>
escreveu:

JoshuaLampert:
We could just remove the `[sources]` section in test/Project.toml of MeshesVizTests.jl, I guess? To verify my hypothesis we could first do a trick and run the downstream action here on julia v1.10 instead of v1(.12) because v1.10 should ignore the `[sources]` section.

cserteGT3:
Sorry, I think I got confused by the comments and the actual script. I think it works as we expect it and tests the PR version. If you look at the [Downstream job](https://github.com/JuliaGeometry/Meshes.jl/actions/runs/26840379513/job/79145551952?pr=1368#step:8:11) of #1368:
```
Resolving package versions...
    Updating `~/work/Meshes.jl/Meshes.jl/downstream/Project.toml`
  [eacbb407] + Meshes v0.57.12-DEV `..`
    Updating `~/work/Meshes.jl/Meshes.jl/downstream/Manifest.toml`

```

I changed the package version to `v0.57.12-DEV` in that PR to see if it loads the PR version and indeed it loads that.
Sorry for the noise. I'll update the comment in the workflow file.

JoshuaLampert:
Yes, there is uses the correct version, but not [during the tests](https://github.com/JuliaGeometry/Meshes.jl/actions/runs/26840379513/job/79145551952?pr=1368#step:9:16).

cserteGT3:
> Yes, there is uses the correct version, but not [during the tests](https://github.com/JuliaGeometry/Meshes.jl/actions/runs/26840379513/job/79145551952?pr=1368#step:9:16).

Ah yes, I was skipping between different codes and runs and everything too fast... I'll run the visualization tests on v1.10 as Joshua suggested.

cserteGT3:
Running the downstream workflow on v1.10 seems correct [during the tests](https://github.com/JuliaGeometry/Meshes.jl/actions/runs/26841181984/job/79148440621?pr=1368#step:9:14) too

JoshuaLampert:
Yes, then I think it is because of the `[sources]` section in the test/Project.toml of MeshesVizTests.jl. I would just remove that part. It's not necessary is it, @juliohm?

juliohm:
I don't think it is necessary. We probably added sources in order to update
the test data with the latest master locally. Please feel free to remove
the sources section if it works without it.

Em ter., 2 de jun. de 2026, 16:01, Joshua Lampert ***@***.***>
escreveu:

JoshuaLampert:
After merging https://github.com/JuliaGeometry/MeshesVizTests.jl/pull/4 I restarted the downstream test here and now it uses the correct branch from the PR: https://github.com/JuliaGeometry/Meshes.jl/actions/runs/26840914585/job/79162906305?pr=1366#step:9:118. Thanks for noticing that, @cserteGT3!

cserteGT3:
Thank you @JoshuaLampert too, for the quick replies and the quick merge! :)

juliohm:
Thank you all! I will try to share a review tomorrow. Excited to fix these
race conditions with observables.

Em ter., 2 de jun. de 2026, 17:21, cserteGT3 ***@***.***>
escreveu:

.github/workflows/Downstream.yml
```diff
@@ -43,13 +43,15 @@ jobs:
...
-        # Note: we retrieve the current `main` branch of the downstream package to ensure
+        # Note: we retrieve the current `main` branch of the downstream package (MeshesVizTests.jl) to ensure
         # that compatibility errors we make in Meshes.jl are detected already here
         # See also https://github.com/trixi-framework/Trixi.jl/pull/1707#discussion_r1382938895
         uses: actions/checkout@v6
         with:
           repository: JuliaGeometry/${{ matrix.package }}
           path: downstream
+        # Note: this script loads the upstream (to-be-tested) version of Meshes.jl
+        # See also https://github.com/JuliaGeometry/Meshes.jl/pull/1366#issuecomment-4605948374
```
juliohm: Please undo these changes as they are unrelated to the PR objectives.

ext/grid/cartesian.jl
```diff
@@ -3,39 +3,44 @@
...
     (xₛ, xₑ), (yₛ, yₑ)
   end
-  x = Makie.@lift $xy[1]
-  y = Makie.@lift $xy[2]
+  Makie.map!(plot, [:xy], :x) do xy
+    xy[1]
+  end
+  Makie.map!(plot, [:xy], :y) do xy
+    xy[2]
+  end
```
juliohm: Please merge these three blocks into a single one that takes `:object` as input and produces two nodes `[:x, :y]`.

ext/grid/cartesian.jl
```diff
@@ -3,39 +3,44 @@
...
 
-  if nc[] == nv[]
+  if plot[:nc][] == plot[:nv][]
```
juliohm:
```suggestion
  if plot.nc[] == plot.nv[]
```

ext/grid/cartesian.jl
```diff
@@ -3,39 +3,44 @@
...
     # visualize as built-in image with interpolation
-    C = Makie.@lift reshape($colorant, $sz .+ 1)
-    Makie.image!(plot, x, y, C, interpolate=true)
+    Makie.map!(plot, [:colorant, :sz], :C) do colorant, sz
+      reshape(colorant, sz .+ 1)
+    end
+    Makie.image!(plot, plot[:x], plot[:y], plot[:C], interpolate=true)
```
juliohm:
```suggestion
    Makie.image!(plot, plot.x, plot.y, plot.C, interpolate=true)
```

Let's try to use this getfield syntax of compute graphs as it is cleaner.

ext/grid/cartesian.jl
```diff
@@ -3,39 +3,44 @@
...
   else
     # visualize as built-in image without interpolation
-    C = Makie.@lift $nc == 1 ? fill($colorant, $sz) : reshape($colorant, $sz)
-    Makie.image!(plot, x, y, C, interpolate=false)
+    Makie.map!(plot, [:nc, :colorant, :sz], :C) do nc, colorant, sz
+      nc == 1 ? fill(colorant, sz) : reshape(colorant, sz)
+    end
+    Makie.image!(plot, plot[:x], plot[:y], plot[:C], interpolate=false)
```
juliohm:
```suggestion
    Makie.image!(plot, plot.x, plot.y, plot.C, interpolate=false)
```

ext/grid/cartesian.jl
```diff
@@ -45,54 +50,69 @@ end
...
-  xyz = Makie.@lift map(x -> ustrip.(x), Meshes.xyz($grid))
+  Makie.map!(plot, [:object], :sp) do grid
+    ustrip.(spacing(grid))
+  end
+  Makie.map!(plot, [:object], :xyz) do grid
+    map(x -> ustrip.(x), Meshes.xyz(grid))
+  end
 
-  if nc[] == nv[]
+  if plot[:nc][] == plot[:nv][]
```
juliohm:
```suggestion
  if plot.nc[] == plot.nv[]
```

ext/grid/cartesian.jl
```diff
@@ -45,54 +50,69 @@ end
...
+            GB.QuadFace(i3, i7, i6, i2),
+            GB.QuadFace(i4, i8, i7, i3),
+            GB.QuadFace(i1, i5, i8, i4),
+            GB.QuadFace(i6, i7, i8, i5)
+          ]
+        end
+      )
+    end
+    Makie.map!(GB.Mesh, plot, [:verts, :quads], :mesh)
+    Makie.mesh!(plot, plot[:mesh], color=plot[:colorant], shading=true)
```
juliohm:
```suggestion
    Makie.mesh!(plot, plot.mesh, color=plot.colorant, shading=true)
```

ext/grid/cartesian.jl
```diff
@@ -45,54 +50,69 @@ end
...
+    end
+    Makie.map!(plot, [:xyz], :zs) do xyz
+      xyz[3][(begin + 1):end]
+    end
+    Makie.map!(plot, [:sp], :rect) do sp
+      Makie.Rect3(-1 .* sp, sp)
+    end
+    Makie.map!(plot, [:xs, :ys, :zs], :coords) do xs, ys, zs
+      [(x, y, z) for z in zs for y in ys for x in xs]
+    end
```
juliohm: Please merge these blocks so that `:xyz` is the input and `:coords` is the output.

ext/grid/cartesian.jl
```diff
@@ -45,54 +50,69 @@ end
...
+    Makie.meshscatter!(plot, plot[:coords], marker=plot[:rect], markersize=1, color=plot[:colorant])
```
juliohm:
```suggestion
    Makie.meshscatter!(plot, plot.coords, marker=plot.rect, markersize=1, color=plot.colorant)
```

ext/grid/cartesian.jl
```diff
@@ -101,32 +121,40 @@ function vizgrid!(plot::Viz{<:Tuple{CartesianGrid}}, ::Type{<:𝔼}, ::Val{3}, :
...
     xysegments(ustrip.(x), ustrip.(y))
   end
-  x = Makie.@lift $xy[1]
-  y = Makie.@lift $xy[2]
+  Makie.map!(plot, [:facets_xy], :facets_x) do xy
+    xy[1]
+  end
+  Makie.map!(plot, [:facets_xy], :facets_y) do xy
+    xy[2]
+  end
```
juliohm: Please merge these blocks into one. Input is `:object` and output is `[:x, :y]`.

ext/grid/rectilinear.jl
```diff
@@ -3,35 +3,40 @@
...
 function vizgrid!(plot::Viz{<:Tuple{RectilinearGrid}}, M::Type{<:𝔼}, pdim::Val{2}, edim::Val{2})
-  grid = plot[:object]
-  color = plot[:color]
-  alpha = plot[:alpha]
-  colormap = plot[:colormap]
-  colorrange = plot[:colorrange]
   showsegments = plot[:showsegments]
 
-  if crs(grid[]) <: Cartesian
+  if crs(plot[:object][]) <: Cartesian
```
juliohm:
```suggestion
  if crs(plot.object[]) <: Cartesian
```

ext/grid/rectilinear.jl
```diff
@@ -3,35 +3,40 @@
...
-    ys = Makie.@lift $xyz[2]
+    Makie.map!(plot, [:object], :xyz) do grid
+      map(x -> ustrip.(x), Meshes.xyz(grid))
+    end
+    Makie.map!(plot, [:xyz], :xs) do xyz
+      xyz[1]
+    end
+    Makie.map!(plot, [:xyz], :ys) do xyz
+      xyz[2]
+    end
```
juliohm: Merge blocks.

ext/grid/rectilinear.jl
```diff
@@ -3,35 +3,40 @@
...
 
-    if nc[] == nv[]
+    if plot[:nc][] == plot[:nv][]
```
juliohm: Please replace all occurrences of `plot[:foo]` by `plot.foo`.

juliohm (changes requested):
Minor suggestions that will help us in future rounds of review.

cserteGT3:
Thank you for the review!
I've:
- reverted the commit that added the comment in the workflow file
- merged all `map!` do-blocks I thought should be merged
- replaced all occurrences of `plot[:foo]` with `plot.foo`

ext/grid/cartesian.jl
```diff
@@ -3,39 +3,38 @@
...
+    Makie.map!(plot, [:colorant, :sz], :C) do colorant, sz
+      reshape(colorant, sz .+ 1)
+    end
+    Makie.image!(plot, plot.x, plot.y, plot.C, interpolate=true)
   else
     # visualize as built-in image without interpolation
-    C = Makie.@lift $nc == 1 ? fill($colorant, $sz) : reshape($colorant, $sz)
-    Makie.image!(plot, x, y, C, interpolate=false)
+    Makie.map!(plot, [:nc, :colorant, :sz], :C) do nc, colorant, sz
+      nc == 1 ? fill(colorant, sz) : reshape(colorant, sz)
```
juliohm:
```suggestion
    Makie.map!(plot, [:colorant, :sz, :nc], :C) do colorant, sz, nc
      nc == 1 ? fill(colorant, sz) : reshape(colorant, sz)
```

ext/grid/cartesian.jl
```diff
@@ -45,54 +44,63 @@ end
...
 
   # spacing and coordinates
-  sp = Makie.@lift ustrip.(spacing($grid))
-  xyz = Makie.@lift map(x -> ustrip.(x), Meshes.xyz($grid))
+  Makie.map!(plot, [:object], :sp) do grid
+    ustrip.(spacing(grid))
+  end
+  Makie.map!(plot, [:object], :xyz) do grid
+    map(x -> ustrip.(x), Meshes.xyz(grid))
+  end
```
juliohm: Could you please confirm that `:sp` and `:xyz` need to be outside the following `if-else` branch? If they only appear in a specific branch, we should probably move the definition there, and merge more `map!` blocks.

ext/grid/cartesian.jl
```diff
@@ -101,32 +109,25 @@ function vizgrid!(plot::Viz{<:Tuple{CartesianGrid}}, ::Type{<:𝔼}, ::Val{3}, :
...
 function vizgridfacets!(plot::Viz{<:Tuple{CartesianGrid}}, ::Type{<:𝔼}, ::Val{2}, ::Val{2})
-  grid = plot[:object]
-  segmentcolor = plot[:segmentcolor]
-  segmentsize = plot[:segmentsize]
+  segmentcolor = plot.segmentcolor
+  segmentsize = plot.segmentsize
 
-  xy = Makie.@lift let
-    x, y = Meshes.xyz($grid)
+  Makie.map!(plot, [:object], [:facets_x, :facets_y]) do grid
```
juliohm: Please rename `:facets_x` to `:xfacets`. Same for `:yfacets`.

ext/grid/rectilinear.jl
```diff
@@ -3,35 +3,33 @@
...
+      colorant isa AbstractVector ? length(colorant) : 1
+    end
 
     # grid coordinates
-    xyz = Makie.@lift map(x -> ustrip.(x), Meshes.xyz($grid))
-    xs = Makie.@lift $xyz[1]
-    ys = Makie.@lift $xyz[2]
-
-    if nc[] == nv[]
+    Makie.map!(plot, [:object], [:xs, :ys]) do grid
```
juliohm: Let's rename `:xs` and `:ys` to `:x` and `:y` to be consistent with other parts of the code.

ext/grid/rectilinear.jl
```diff
@@ -43,16 +41,13 @@ function vizgrid!(plot::Viz{<:Tuple{RectilinearGrid}}, M::Type{<:𝔼}, pdim::Va
...
 function vizgridfacets!(plot::Viz{<:Tuple{RectilinearGrid}}, ::Type{<:𝔼}, ::Val{2}, ::Val{2})
-  grid = plot[:object]
-  segmentcolor = plot[:segmentcolor]
-  segmentsize = plot[:segmentsize]
+  segmentcolor = plot.segmentcolor
+  segmentsize = plot.segmentsize
 
-  xy = Makie.@lift let
-    x, y = Meshes.xyz($grid)
+  Makie.map!(plot, [:object], [:facets_x, :facets_y]) do grid
```
juliohm: `:xfacets` and `:yfacets`

ext/grid/structured.jl
```diff
@@ -41,13 +42,14 @@ function vizgrid!(plot::Viz{<:Tuple{StructuredGrid}}, M::Type{<:𝔼}, pdim::Val
...
+  segmentcolor = plot.segmentcolor
+  segmentsize = plot.segmentsize
 
-  tup = Makie.@lift structuredsegments($grid)
-  x, y = Makie.@lift($tup[1]), Makie.@lift($tup[2])
-  Makie.lines!(plot, x, y, color=segmentcolor, linewidth=segmentsize)
+  Makie.map!(structuredsegments, plot, [:object], :facets_tup)
+  Makie.map!(plot, [:facets_tup], [:facets_x, :facets_y]) do tup
+    (tup[1], tup[2])
+  end
```
juliohm: Merge two blocks and output `:xfacets` and `:yfacets` directly.

ext/utils.jl
```diff
@@ -5,11 +5,15 @@
...
+  pointsize = plot.pointsize
+  segmentsize = plot.segmentsize
 
-  rmesh = Makie.@lift reduce(merge, $meshes)
-  color = Makie.@lift [$colors[i] for (i, m) in enumerate($meshes) for _ in 1:nelements(m)]
+  rmesh = Makie.map(meshes) do ms
+    reduce(merge, ms)
+  end
+  color = Makie.map(colors, meshes) do cs, ms
+    [cs[i] for (i, m) in enumerate(ms) for _ in 1:nelements(m)]
```
juliohm: What is the difference between using `map!` and `map`? Was `map` intentional here?

cserteGT3: I think we used `map` here, because the variables aren't observables and `map!` works on observables. Though I'm not sure.
I'll look into it later.

cserteGT3:
I think I've addressedd all your comments. Tests passed on my machine, so I expect CI to pass as well. I don't know why building the documentation fails though.
*Edit: documentation build works locally.

ext/utils.jl
```diff
@@ -5,13 +5,19 @@
...
-  color = Makie.@lift [$colors[i] for (i, m) in enumerate($meshes) for _ in 1:nelements(m)]
+  rmesh = Symbol(meshes, :_rmeshes)
+  rcolor = Symbol(meshes, :_rcolor)
 
-  viz!(plot, rmesh, color=color, pointsize=pointsize, segmentsize=segmentsize)
+  Makie.map!(plot, [meshes, colors], [rmesh, rcolor]) do ms, cs
+    rmes = reduce(merge, ms)
+    color = [cs[i] for (i, m) in enumerate(ms) for _ in 1:nelements(m)]
+    (rmes, color)
+  end
```
juliohm: I am bit confused by this section. Why can't we use `map!` as we did in other places? Why the `Symbol` creation, etc.?

cserteGT3: It possible, that `vizmany!` is called with different arguments for the same plot and we cannot use the same observables, so we need to create new ones.
Consider the following example:
```julia
b1 = Box((0.0, 0.0, 0.0), (1.0, 1.0, 1.0))
b2 = Box((2.0, 1.0, 0.0), (3.0, 2.0, 1.0))
s1 = Ball((3.0, 0.0, 3.0), 1.0)
s2 = Ball((-1.0, 0.0, -1.0), 1.0)
m1 = Multi([b1, b2])
m2 = Multi([s1, s2])
g = GeometrySet([m1, m2])
```
I put a `@show rcolor` command in `vizmany!` so we can see that it is called two times:

```julia
julia> viz(g)
rcolor = Symbol("Computed(:gvec_1_parents_meshes, #undef)_rcolor")
rcolor = Symbol("Computed(:gvec_2_parents_meshes, #undef)_rcolor")
```

If the ouput variables are the same, it errors on its second call, as we cannot map different inputs to the same output.

juliohm:
@cserteGT3 I've reviewed and adjusted the code in all files, except for `ext/geomset.jl` and `ext/utils.jl`. Could you try to minimize the changes in `ext/geomset.jl`? The changes seem to affect the logic of the previous implementation.

juliohm:
@cserteGT3 I've attempted to move the `:colorant` calculation up in the `viz` stack so that it is only computed once before any logic is performed with the `:object` being visualized. I am not sure if tests will pass because we still need to fix the `GeometrySet` methods.

ext/grid/transformed.jl
```diff
@@ -3,23 +3,20 @@
...
+  end
 
-  if isoptimized(trans[]) # visualize parent grid and transform visualization
-    grid = Makie.@lift parent($tgrid)
-    viz!(plot, grid; color, alpha, colormap, colorrange, showsegments, segmentcolor, segmentsize)
-    makietransform!(plot, trans)
-  else # fallback to full grid visualization
+  if isoptimized(plot.trans[])
+    # visualize parent grid and transform visualization
+    plot.object = plot.pgrid
```
juliohm: I am assuming that compute graphs are mutable in order to dispatch the appropriate method for the parent grid. @cserteGT3 could you please confirm this is a valid approach?

ext/geomset.jl
```diff
@@ -2,43 +2,49 @@
...
-  alpha = plot[:alpha]
-  colormap = plot[:colormap]
-  colorrange = plot[:colorrange]
+  gset = plot.object[]
 
   # process color spec into colorant
-  colorant = facets ? nothing : Makie.@lift(process($color, $colormap, $colorrange, $alpha))
+  if !facets
+    Makie.map!(process, plot, [:color, :colormap, :colorrange, :alpha], :colorant)
+  end
```
juliohm: @cserteGT3 if you could try to follow the same code pattern we have for `Grid`, `Mesh` and `SubDomain`, that would be great. Basically, we need to move the calculation of `:colorant` to the `plot!` method at the top of the file, and then adjust the `vizgset!` methods.

I understand that the main challenge with the `GeometrySet` methods is the need to create new nodes in the compute graph with unique identifiers. Do you think we can do that without modifying the original signature of the `vizgset!` methods?

ext/grid/transformed.jl
juliohm: I believe it is valid. I found `Makie.update!` to set nodes explicitly in the graph.

ext/geomset.jl
cserteGT3: I'm looking at it for a couple of minutes by now and I'm still a bit confused.

Do we need the `facets` kwarg in `vizgset!`? Because it's never called with `true`. Or `vizfacets!(plot::Viz{<:Tuple{GeometrySet}}) = vizgset!(plot, facets=false)` meant to be `facets=true`?

ext/subdomain.jl
```diff
@@ -2,96 +2,58 @@
...
-    segmentsize,
-    showpoints,
-    pointmarker,
-    pointcolor,
-    pointsize
-  )
+function vizsubdomfallback!(plot, ::Type, ::Val, ::Val)
+  # visualize as geometry set
+  gset = GeometrySet(collect(plot.object[]))
+  Mke.update!(plot, object=gset)
```
juliohm:
```suggestion
  Makie.update!(plot, object=gset)
```

ext/geomset.jl
juliohm: If you feel that the `facets` kwarg is not needed, please feel free to get rid of it. We refactored this method so many times in the past that it might be simply an outdated option that is no longer used. The logic with `vizgset!` and `vizgsetfacets!` should be similar to that of `Mesh`, `Grid`, ...

juliohm:
We are still getting an error with `Makie.update!`. It says that `:object` is not a node in the graph, but it should be there, right?

ext/geomset.jl
cserteGT3: Ok, I'll give it go

cserteGT3: I've done what I could about this. Tests passed locally (except the `Makie.update!` issue discussed separately).

ext/geomset.jl
```diff
@@ -61,159 +49,230 @@ function vizgset!(plot, M::Type, pdim::Val, ::Val, geoms::ObservableVector{<:Geo
...
+    Makie.poly!(plot, plot[polys_sym], color=plot[colors])
   end
 end
 
 # -------
 # FACETS
 # -------
 
-vizfacets!(plot::Viz{<:Tuple{GeometrySet}}) = vizgset!(plot, facets=false)
+vizfacets!(plot::Viz{<:Tuple{GeometrySet}}) = vizfacets!(plot)
```
juliohm: Isn't it an infinite loop?

cserteGT3:
> We are still getting an error with `Makie.update!`. It says that `:object` is not a node in the graph, but it should be there, right?

I also expected to work like this. We could update it like this:
```julia
  gset = GeometrySet(collect(plot.object[]))
  plot.object[]=gset
  vizgset!(plot)
```
But then the object types don't match (expects a `SubDomain` but gets a `GeometrySet`). I've yet to have a solution...

ext/geomset.jl
cserteGT3: Yet, but it never gets called. I'll fix it

juliohm:
> But then the object types don't match (expects a `SubDomain` but gets a `GeometrySet`). I've yet to have a solution...

We can go back to calling the generic `viz!` with explicit `kwargs` forwarded too. I can take a look into it after you finish the `GeometrySet` case.

ext/geomset.jl
cserteGT3: I've pushed a refactor, that should take care of it

ext/grid/transformed.jl
```diff
@@ -3,23 +3,17 @@
...
+  Makie.map!(transformation, plot, :object, :trans)
 
-  if isoptimized(trans[]) # visualize parent grid and transform visualization
-    grid = Makie.@lift parent($tgrid)
-    viz!(plot, grid; color, alpha, colormap, colorrange, showsegments, segmentcolor, segmentsize)
-    makietransform!(plot, trans)
-  else # fallback to full grid visualization
+  if isoptimized(plot.trans[])
+    # visualize parent grid and transform visualization
+    pgrid = parent(plot.object[])
```
juliohm: @cserteGT3 I wonder when it is necessary to update the compute graph. If calling `plot.object[]` forces the graph to update the object provided by the user (e.g., observable), then we can assume that `pgrid` will be up-to-date? Or we need to add it to the graph to get interactivity?

Answering that question might also help cleanup the `GeometrySet` implementation.

juliohm: Nevermind, I think the syntax `plot.object[]` is just a snapshot of the current graph. Makie probably needs the node itself in the calls to keep track of any update automatically.

juliohm:
I've reviewed the `GeometrySet` methods and removed the `vizmany!` utility function. We are almost there.

juliohm (approved):
That is an amazing contribution @cserteGT3 ❤️ Thank you for pushing it!

juliohm:
Test failing on Windows-float32 is a random issue on GitHub (cc: @JoshuaLampert)

@JoshuaLampert do you have any suggestion before we merge this? I confirm it fixes #1364 

After the PR is merged we can add that MWE to MeshVizTests.jl

JoshuaLampert:
I don't have the time to review this PR, but I am fine with merging. The random Windows issue becomes really annoying...

cserteGT3:
Thank you @juliohm for the active work and for pushing it to the finish line! It's amazing to work with you on this project as a contributor. I hope that you feel as appreciated for working on Meshes.jl (and its ecosystem) as I feel when I'm bringing in these ideas and half-implemented PRs :)

juliohm:
I'm super happy to work together on these features! This project is a lot
of fun!

Em seg., 8 de jun. de 2026, 16:16, cserteGT3 ***@***.***>
escreveu:

Issue #1304, opened 2025-12-09
Review `cartesianrange` for `Grid` and `Box` in non-Euclidean manifold


juliohm:
The current implementation is not covered by our tests, and it looks like it is buggy. I am opening the issue to not forget to fix it.

PR #1369, opened 2026-06-05
Add cartesianrange methods for LatLon grids

anj1:
This PR works towards a fix for https://github.com/JuliaGeometry/Meshes.jl/issues/1304

I added explicit geodesic `cartesianrange` / `indices` methods for `RegularGrid{🌐,<:LatLon}` and `RectilinearGrid{🌐,<:LatLon}`. 

We dispatch the fast path only for axis-separable LatLon grids. I updated `Slice` dispatch so geodesic slicing uses the new fast path only when the grid type supports it, and falls back to general `indices` otherwise.

I also added regression tests for interior geodesic boxes, edge clamping at the grid boundary, wrap-around, out of bounds queries, and narrow wrapped boxes.


src/transforms/slice.jl
```diff
@@ -43,7 +43,9 @@ _slice(g::Grid, inds::CartesianIndices) = getindex(g, inds)
 _sliceinds(d::Domain, b) = indices(d, b)
 _sliceinds(g::OrthoRegularGrid, b) = cartesianrange(g, b)
 _sliceinds(g::OrthoRectilinearGrid, b) = cartesianrange(g, b)
-_sliceinds(g::Grid{🌐}, b::Box{🌐}) = cartesianrange(g, b)
+_sliceinds(g::RegularGrid{🌐,<:LatLon}, b::Box{🌐}) = cartesianrange(g, b)
+_sliceinds(g::RectilinearGrid{🌐,<:LatLon}, b::Box{🌐}) = cartesianrange(g, b)
+_sliceinds(g::Grid{🌐}, b::Box{🌐}) = indices(g, b)
```
juliohm: Let's leave these changes to a separate PR. We need to make sure that we have the correct `indices` and `cartesianrange` methods in place before adjusting  the `Slice` transform.

src/indices.jl
```diff
@@ -97,6 +97,16 @@ function indices(grid::OrthoRectilinearGrid, box::Box)
   LinearIndices(size(grid))[range] |> vec
 end
 
+function indices(grid::RegularGrid{🌐,<:LatLon}, box::Box{🌐})
```
juliohm:
```suggestion
function indices(grid::RegularGrid{🌐}, box::Box{🌐})
```

src/indices.jl
```diff
@@ -97,6 +97,16 @@ function indices(grid::OrthoRectilinearGrid, box::Box)
...
+  range = cartesianrange(grid, box)
+  LinearIndices(size(grid))[range] |> vec
+end
+
+function indices(grid::RectilinearGrid{🌐,<:LatLon}, box::Box{🌐})
```
juliohm:
```suggestion
function indices(grid::RectilinearGrid{🌐}, box::Box{🌐})
```

src/indices.jl
```diff
@@ -108,9 +118,14 @@ Return the Cartesian range of the elements of the `grid` that intersect with the
...
 
+cartesianrange(grid::RegularGrid{🌐,<:LatLon}, box::Box{🌐}) = _geodesicrange(grid, box)
+
+cartesianrange(grid::RectilinearGrid{🌐,<:LatLon}, box::Box{🌐}) = _geodesicrange(grid, box)
+
 _manifoldrange(::Type{<:𝔼}, grid::Grid, box::Box) = _euclideanrange(grid, box)
 
-_manifoldrange(::Type{<:🌐}, grid::Grid, box::Box) = _geodesicrange(grid, box)
+_manifoldrange(::Type{<:🌐}, grid::Grid, box::Box) =
+  throw(ArgumentError("cartesianrange for geodesic boxes requires a regular or rectilinear grid in LatLon coordinates"))
```
juliohm:
```suggestion
_manifoldrange(::Type{<:🌐}, grid::Grid, box::Box) = _geodesicrange(grid, box)
```

src/indices.jl
```diff
@@ -108,9 +118,14 @@ Return the Cartesian range of the elements of the `grid` that intersect with the
 """
 cartesianrange(grid::Grid{M}, box::Box{M}) where {M} = _manifoldrange(M, grid, box)
 
+cartesianrange(grid::RegularGrid{🌐,<:LatLon}, box::Box{🌐}) = _geodesicrange(grid, box)
+
+cartesianrange(grid::RectilinearGrid{🌐,<:LatLon}, box::Box{🌐}) = _geodesicrange(grid, box)
+
```
juliohm:
```suggestion
```

juliohm (changes requested):
Thank you @anj1! Looking forward to reviewing this PR further.

anj1:
Yes I see what you're getting at, this is closer to the core design, we delegate cartesianrange through manifoldrange already, no need for the extra side paths.

src/indices.jl
```diff
@@ -150,66 +156,78 @@ function _euclideanrange(grid::OrthoRectilinearGrid, box::Box)
...
-  a = convert(LatLon, coords(vertex(grid, (1, 1))))
-  b = convert(LatLon, coords(vertex(grid, (nlon, 1))))
-  c = convert(LatLon, coords(vertex(grid, (1, nlat))))
+# -----------------
+# HELPER FUNCTIONS
+# -----------------
 
-  swaplon = a.lon > b.lon
-  swaplat = a.lat > c.lat
+function _regularaxisrange(vals, lo, hi)
```
juliohm, line 181: Should `_regularaxisrange` be renamed to `_regularlatrange`? What is the rationale for the name?

src/indices.jl
```diff
@@ -150,66 +156,78 @@ function _euclideanrange(grid::OrthoRectilinearGrid, box::Box)
...
-function _geodesicrange(grid::Grid, box::Box)
-  nlon, nlat = vsize(grid)
+function cartesianrange(grid::RegularGrid{🌐}, box::Box{🌐})
+  lat, lon = xyz(grid |> Proj(LatLon))
+  boxmin = convert(LatLon, coords(minimum(box)))
+  boxmax = convert(LatLon, coords(maximum(box)))
+  latrange = _regularaxisrange(lat, boxmin.lat, boxmax.lat)
+  lonrange = _regularlonrange(lon, boxmin.lon, boxmax.lon)
+  CartesianIndices((latrange, lonrange))
+end
```
juliohm, line 166: I believe we can reuse the idea used in the `OrthoRegularGrid` implementation where we compute the intersection with the boundingbox of the grid, select the corners of the resulting box, and derive `ijkₛ` and `ijkₑ`. The only difference lies in the method used to derive these two variables.

In the `OrthoRegularGrid` case we use linear algebra (e.g., `(lo - or) ./ sp`) to find the corners. In the `RegularGrid{🌐}` case, we can do the same operation with `LatLon` coordinates.

Could you please refactor the code accordingly? Notice that your current implementation calls `xyz` which is expensive.

src/indices.jl
```diff
@@ -150,66 +156,78 @@ function _euclideanrange(grid::OrthoRectilinearGrid, box::Box)
...
 
+function cartesianrange(grid::RectilinearGrid{🌐}, box::Box{🌐})
+  lat, lon = xyz(grid |> Proj(LatLon))
   boxmin = convert(LatLon, coords(minimum(box)))
   boxmax = convert(LatLon, coords(maximum(box)))
+  latrange = _rectilinearaxisrange(lat, boxmin.lat, boxmax.lat)
+  lonrange = _rectilinearlonrange(lon, boxmin.lon, boxmax.lon)
+  CartesianIndices((latrange, lonrange))
+end
```
juliohm, line 175: Similar comment for `RectilinearGrid{🌐}` in the sense that the we could reuse the method for `OrthoRectilinearGrid`, just changing the method to compute the `ijkₛ` and `ijkₑ`.

src/indices.jl
```diff
@@ -150,66 +156,78 @@ function _euclideanrange(grid::OrthoRectilinearGrid, box::Box)
...
-    p = vertex(grid, (1, i))
-    c = convert(LatLon, coords(p))
-    c.lat ≤ latmin
-  end
-  jₑ = findfirst(latinds) do i
-    p = vertex(grid, (1, i))
-    c = convert(LatLon, coords(p))
-    c.lat ≥ latmax
-  end
+function _lonrange(vals, lo, hi, axisrange)
```
juliohm, line 212: I understand this helper function was added to handle the periodicity of the `lon` coordinate. I believe we can avoid it if we have a correct implementation of intersection of boxes over `🌐`.

Perhaps we could start a separate PR where we first make sure that the intersection of boxes over the globe is producing the expected results What do you think?

juliohm:
@anj1 what is the status here?

PR #1374, opened 2026-06-13
Update to Colorfy.jl v2.0

juliohm:
To support more colorful representations of types.


juliohm:
The visual test is failing as expected. The new figure is correct. I will update it in MeshesVizTests.jl.

PR #1375, opened 2026-06-13
Use Makie docstring system for viz

juliohm:
I tried to follow the Makie documentation on recipes, but the docstring for each attribute is not showing up even after loading a Makie backend in the same session.

If anyone knows what is happening, please feel free to suggest changes.


juliohm:
Fixed the issue. I was using comments instead of strings.

PR #1376, opened 2026-06-20
WIP: Add support for optional shading

grero:
This adds a keyword `shading` to the `viz` recipe controlling whether a 3D mesh should use shading.

### DIFF
--- a/ext/MeshesMakieExt.jl
+++ b/ext/MeshesMakieExt.jl
@@ -33,7 +33,8 @@ Makie.@recipe(Viz, object) do scene
     showpoints=false,
     pointmarker=:circle,
     pointcolor=:gray30,
-    pointsize=4
+    pointsize=4,
+    shading=false,
   )
 end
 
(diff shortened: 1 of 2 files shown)


juliohm:
Hi @grero. Could you please share an example where shading wouldn't be a good setting in 3D?

Also, the PR has conflicts and if we really want to consider this option, we need to make sure that it is passed everywhere, not just with the `SimpleMesh` case.

grero:
My particular use case is that I want to visualize what is essentially a histogram computed using bins that exist on a 3D mesh. There, I just want to see the colors, with no effect of lighting at all. That being said, I apologize for not labeling this PR a work in progress, which it very much is. I'll try to resolve the conflicts so that it at last can be merge cleanly. It is of course up to the maintainers to decide if this is something that would be interesting to everyone using this package.

juliohm:
> My particular use case is that I want to visualize what is essentially a histogram computed using bins that exist on a 3D mesh.

Do you have a visual example that you can share?

PR #1378, opened 2026-06-21
Add precompile workloads for viz

juliohm:
I am seeing 2x speedup regarding the time-to-first-viz on my machine.

The precompilation takes approximately 40s, which I think is acceptable given the speedup.


juliohm:
Our downstream visual tests in MeshesVizTests.jl went from 14min to 8min, which is consistent with what I am measuring locally 🚀 I will go ahead and merge the PR after all tests pass.

PR #1387, opened 2026-07-01
Review centroid definitions

### LINKED ISSUE #1385: Wrong `centroid` calculation?
Hello,

I am getting wrong results when calculating `centroid`.
```
julia>  v = [
        Meshes.Point(1.20768412712354e6 , 440894.51013951),
         Meshes.Point(1.2074204326973e6 , 440608.763531389),
        Meshes.Point(1.20739595705862e6 , 440582.321626554),
        Meshes.Point(1.20718830292167e6 , 440357.136257657),
        Meshes.Point(1.20712196240754e6 , 440285.477674712),
        Meshes.Point(1.20679979364406e6 , 439935.793088486),
        Meshes.Point(1.20506571504692e6 , 438056.784455802),
        Meshes.Point(1.19570432824241e6 , 446330.054193162),
        Meshes.Point(1.19707924287483e6 , 447774.530449439),
        Meshes.Point(1.19824467667019e6 , 449000.26154475),
        Meshes.Point(1.19971692661997e6 , 450548.653675972),
        Meshes.Point(1.20921750106109e6 , 442557.765689318)
        ];

julia> control = Ring(v...)
Ring
├─ Point(x: 1.20768412712354e6 m, y: 440894.51013951 m)
├─ Point(x: 1.2074204326973e6 m, y: 440608.763531389 m)
├─ Point(x: 1.20739595705862e6 m, y: 440582.321626554 m)
├─ Point(x: 1.20718830292167e6 m, y: 440357.136257657 m)
├─ Point(x: 1.20712196240754e6 m, y: 440285.477674712 m)
⋮
├─ Point(x: 1.19570432824241e6 m, y: 446330.054193162 m)
├─ Point(x: 1.19707924287483e6 m, y: 447774.530449439 m)
├─ Point(x: 1.19824467667019e6 m, y: 449000.26154475 m)
├─ Point(x: 1.19971692661997e6 m, y: 450548.653675972 m)
└─ Point(x: 1.20921750106109e6 m, y: 442557.765689318 m)

julia> centroid(control)
├─ x: 1.204053247197345e6 m
└─ y: 443077.6710272292 m

julia> realcentroid
├─ x: 1.2024640269628512e6 m
└─ y: 444335.8162640141 m
```

the `realcentroid` was manually calculated using first area moments for the ring (can provide the implementation if needed).

Here is visualized (red `centroid` output, green the `realcentroid`)

Image

with this code:
```
julia> viz(control);

julia> viz(control)

julia> viz!(centroid(control); color=:red, pointsize=10);

julia> viz!(realcentroid; color=:green, pointsize=10);
```

juliohm:
Draft to fix #1385 

Appreciate it if someone with more time could review and adjust the tests. We have some errors in tests when they attempt to convert ForwardDiff.Dual into AbstractFloat and other numerical error due to the new definitions.

### DIFF
--- a/src/centroid.jl
+++ b/src/centroid.jl
@@ -43,15 +43,15 @@ end
 
 centroid(t::Torus) = center(t)
 
-centroid(s::Segment) = s(1//2)
+centroid(s::Segment) = s(1 // 2)
 
-centroid(t::Triangle) = t(1//3, 1//3)
+centroid(t::Triangle) = t(1 // 3, 1 // 3)
 
-centroid(q::Quadrangle) = q(1//2, 1//2)
+centroid(q::Quadrangle) = q(1 // 2, 1 // 2)
 
-centroid(t::Tetrahedron) = t(1//4, 1//4, 1//4)
+centroid(t::Tetrahedron) = t(1 // 4, 1 // 4, 1 // 4)
 
-centroid(h::Hexahedron) = h(1//2, 1//2, 1//2)
+centroid(h::Hexahedron) = h(1 // 2, 1 // 2, 1 // 2)
 
 centroid(m::Multi) = centroid(GeometrySet(parent(m)))
 
--- a/src/centroid.jl
+++ b/src/centroid.jl
@@ -11,8 +11,6 @@ centroid(g::Geometry) = withcrs(g, integral(to, g) / measure(g))
 
 centroid(p::Point) = p
 
-centroid(s::Segment) = center(s)
-
 centroid(p::Plane) = center(p)
 
 centroid(b::Box) = center(b)
@@ -42,6 +40,18 @@ function centroid(p::ParaboloidSurface)
   y = zero(z)
   c + Vec(x, y, z / 2)
 end
+
+centroid(t::Torus) = center(t)
+
+centroid(s::Segment) = s(1//2)
+
+centroid(t::Triangle) = t(1//3, 1//3)
+
+centroid(q::Quadrangle) = q(1//2, 1//2)
+
+centroid(t::Tetrahedron) = t(1//4, 1//4, 1//4)
+
+centroid(h::Hexahedron) = h(1//2, 1//2, 1//2)
 
 centroid(m::Multi) = centroid(GeometrySet(parent(m)))
 
--- a/src/centroid.jl
+++ b/src/centroid.jl
@@ -7,19 +7,23 @@
 
 The centroid of the `geometry`.
 """
-centroid(g::Geometry) = center(g) # some geometries have a natural center
+centroid(g::Geometry) = withcrs(g, integral(to, g) / measure(g))
 
 centroid(p::Point) = p
 
-centroid(p::Polygon) = centroid(first(rings(p)))
+centroid(s::Segment) = center(s)
 
-centroid(n::Ngon) = coordmean(vertices(n))
+centroid(p::Plane) = center(p)
 
-centroid(p::Polytope) = coordmean(vertices(p))
+centroid(b::Box) = center(b)
 
-centroid(b::Box) = coordmean(extrema(b))
+centroid(b::Ball) = center(b)
 
-centroid(p::Plane) = p(0, 0)
+centroid(s::Sphere) = center(s)
+
+centroid(d::Disk) = center(d)
+
+centroid(c::Circle) = center(c)
 
 centroid(c::Cylinder) = centroid(boundary(c))
 
--- a/src/geometries/polytopes/segment.jl
+++ b/src/geometries/polytopes/segment.jl
@@ -21,7 +21,7 @@ Base.maximum(s::Segment) = s.vertices[2]
 
 Base.extrema(s::Segment) = s.vertices[1], s.vertices[2]
 
-center(s::Segment) = centroid(s)
+center(s::Segment) = coordmean(extrema(s))
 
 ==(s₁::Segment, s₂::Segment) = s₁.vertices == s₂.vertices
 
--- a/src/geometries/primitives/box.jl
+++ b/src/geometries/primitives/box.jl
@@ -51,6 +51,8 @@ Base.maximum(b::Box) = b.max
 
 Base.extrema(b::Box) = minimum(b), maximum(b)
 
+center(b::Box) = coordmean(extrema(b))
+
 diagonal(b::Box{<:𝔼}) = norm(maximum(b) - minimum(b))
 
 sides(b::Box{<:𝔼}) = Tuple(maximum(b) - minimum(b))
--- a/src/geometries/primitives/plane.jl
+++ b/src/geometries/primitives/plane.jl
@@ -40,6 +40,8 @@ paramdim(::Type{<:Plane}) = 2
 
 normal(p::Plane) = unormalize(ucross(p.u, p.v))
 
+center(p::Plane) = p(0, 0)
+
 ==(p₁::Plane, p₂::Plane) =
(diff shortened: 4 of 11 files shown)


ghyatzo (commented):
Looks ok, from a quick cursory look to the tests it seems that the generic fallback has some numerical issues and most of the equalities will need to be turned into `approx`.
What this showed me though is that probably the fallback should be just that and we should add analytical formulations where possible for both speed and precision. 

Also the LatLon centroid seems much more off than just rounding errors, probably worth it to investigate what happens in a non flat manifold.

I'll try to find some time the next few days to add specific implementations for rings and polygons.

juliohm:
I think the safest route here is to make sure that the fallback is producing the expected results and fix the cases that are still problematic. Then we could start fresh PRs with specific optimizations with proper benchmarks.jl

ghyatzo:
Yes agreed it would have to be on top of this.

ghyatzo:
Ok, some additional findings:
There currently is an issue with `Ring`s. the fallback formula computes the centroid of the perimeter points, which is different than the centroid of the enclosed area.
As an example:
```
julia> rtrig
Ring
├─ Point(x: 0.0 m, y: 0.0 m)
├─ Point(x: 1.0 m, y: 0.0 m)
└─ Point(x: 0.0 m, y: 1.0 m)

julia> centroid(rtrig) # perimeter centroid
├─ x: 0.35355339059327373 m
└─ y: 0.35355339059327373 m

julia> first_area_moment(rtrig) # area centroid
2-element SVector{2, Unitful.Quantity{Float64, 𝐋, Unitful.FreeUnits{(m,), 𝐋, nothing}}} with indices SOneTo(2):
 0.3333333333333333 m
 0.3333333333333333 m
```

I think rings at the moment are doing double duty of both being the perimeter of a polygon and enclosing its area. The `measure`/`integral` machinery see it as the perimeter. We must decide what we want to centroid of the Ring to be: The centroid of the enclosing area or of the point set of the vertices (close to the old behaviour).
Probably for a separate PR, but I could see the use for a `enclosedarea` alias in `measures.jl` specifically for `Ring`s. But I digress.

I'd lean towards adding a `centroid(r::Ring) = centroid(PolyArea(r))` approach (since usually when you work with rings you are working with the enclosed area), you can retrieve the centroid of the perimeter vertices with `centroid(PointSet(vertices(r)))` if you really want, but I think that should be explicit since it's a bit more specialised.

Then in later PRs we could add specialisations of `centroid(p::PolyArea)` to use analytical formulations.


Regarding the tests:
- most test errors as just floating point errors that an `approx` should fix.
- the `latlon` test produce the correct centroid, but the expected answer is wrong (it's a linear interpolation in lat lon space, which is not the same as the true area centroid)
- the Dual number method errors I am not sure how to approach them, since most seems to be within Base itself which this new `integral` path exercises.

juliohm:
You are misinterpreting rings as areas, they are 1D geometries. There is no single place in the code base where they are treated as areas.

Regarding the other issues, could you share suggestions in the PR for the isapprox changes? I could apply the suggestions and look into the other issues with more time later.

ghyatzo:
I know they are 1 dimension objects.
But while being 1D they also uniquely identify an area in the plane. 

I'm just arguing that returning the perimeter centroid could be source of potential confusion.
Maybe just a warning in the docs?

Regarding suggestions I can't since the test files are not in the PR and I don't have the right permissions to commit to this PR directly.

juliohm:
> But while being 1D they also uniquely identify an area in the plane.

I understand where you are coming from, but that is inconsistent with the definition we discussed in the issue. We could make similar arguments for other geometries like saying that the `Sphere` identifies the `boundary` of the `Ball` and therefore we should implement methods for the `Sphere` as if it was a `Ball`. That quickly becomes unmanageable.

> Regarding suggestions I can't since the test files are not in the PR and I don't have the right permissions to commit to this PR directly.

It is unfortunate that GitHub only allows suggestions in modified files. I will try to look into it as soon as I am done with some tasks here.

ghyatzo:
I do have some fixes locally for the `isapprox` errors. I can share the patches as soon as I get back to a pc.

Here are the patches
<details><summary>Details</summary>
<p>

```
diff --git a/test/geomsets.jl b/test/geomsets.jl
index 9e3f0e81fd..55efffed58 100644
--- a/test/geomsets.jl
+++ b/test/geomsets.jl
@@ -5,7 +5,7 @@
   gset = GeometrySet([s, t, p])
   @test crs(gset) <: Cartesian{NoDatum}
   @test Meshes.lentype(gset) == ℳ
-  @test [centroid(gset, i) for i in 1:3] == cart.([(1 / 2, 1 / 2), (1 / 3, 1 / 3), (1 / 2, 1 / 2)])
+  @test all([centroid(gset, i) for i in 1:3] .≈ cart.([(1 / 2, 1 / 2), (1 / 3, 1 / 3), (1 / 2, 1 / 2)]))
 
   s = Segment(cart(0, 0), cart(1, 1))
   t = Triangle(cart(0, 0), cart(1, 0), cart(0, 1))
diff --git a/test/meshes.jl b/test/meshes.jl
index 50815fa0c5..bf0bbacaf1 100644
--- a/test/meshes.jl
+++ b/test/meshes.jl
@@ -258,7 +258,7 @@
     cart(5, 5, 5),
     cart(0, 5, 5)
   )
-  @test all(centroid(grid, i) == centroid(grid[i]) for i in 1:nelements(grid))
+  @test all(centroid(grid, i) ≈ centroid(grid[i]) for i in 1:nelements(grid))
 
   # mixed units
   grid = CartesianGrid((T(0) * u"m", T(0) * u"cm"), (T(100) * u"cm", T(1) * u"m"), GridTopology(10, 10))
diff --git a/test/sorting.jl b/test/sorting.jl
index ad5aad24a7..f3f6376c63 100644
--- a/test/sorting.jl
+++ b/test/sorting.jl
@@ -1,7 +1,7 @@
 @testitem "DirectionSort" setup = [Setup] begin
   g = cartgrid(3, 3)
   s = sort(g, DirectionSort((T(1), T(1))))
-  @test centroid.(s) ==
+  @test all(centroid.(s) .≈
         cart.([
     (0.5, 0.5),
     (1.5, 0.5),
@@ -12,5 +12,5 @@
     (2.5, 1.5),
     (1.5, 2.5),
     (2.5, 2.5)
-  ])
+  ]))
 end
diff --git a/test/subdomains.jl b/test/subdomains.jl
index bbdc041d60..d92688d336 100644
--- a/test/subdomains.jl
+++ b/test/subdomains.jl
@@ -20,7 +20,7 @@
   for i in 1:3
     e = grid[inds[i]]
     @test v[i] == e
-    @test centroid(v, i) == centroid(e)
+    @test centroid(v, i) ≈ centroid(e)
   end
 
   points = cart.([(0, 0), (1, 0), (0, 1), (1, 1), (0.5, 0.5)])
```

</p>
</details>

juliohm:
All tests pass now with a few additional methods. We can review these methods in a separate PR after this one is merged and a patch is released. I am still trying to figure out how to hunt down two warnings from the h-adaptive integration routines:

```julia
┌ Warning: maximum number of subdivisions reached `maxsubdiv=32768`, try increasing the keyword argument `maxsubdiv`.
┌ Warning: maximum number of subdivisions reached `maxsubdiv=32768`, try increasing the keyword argument `maxsubdiv`.
```

Do you know how we can print the exact line of the tests that emitted these warnings? I will run testitems manually in the meantime to find out.

ghyatzo:
No idea unfortunately, my instinct is the two lat Lon tests, since that is the more complex integration domain and the results are still slightly off from the true result. So maybe it could be a good first candidate to check

juliohm:
We have a (inevitable) regression in the GJK intersection algorithm because it relies on the `centroid` of geometries to find support points. We can look into optimizations for `Triangle` and `Tetrahedron` in a separate PR.

juliohm:
Merging and releasing as a hot fix.

PR #1389, opened 2026-07-03
Fix StackOverflowError in atol/rtol for non-Float

### LINKED ISSUE #1388: StackOverflowError when calling `atol`/`rtol` on `BigFloat` (e.g. via `integral` on a `Box{BigFloat}`)
`atol` and `rtol` in src/tolerances.jl are only defined for `Float32` and `Float64`. Any other `AbstractFloat` type (e.g. `BigFloat` or `Float64x2` from `MultiFloats`) falls through to the generic fallback:
```julia
atol(x) = atol(typeof(x))
rtol(x) = rtol(typeof(x))
```
Since there's no `atol(::Type{Float64x2})`/`rtol(::Type{Float64x2})` method, this calls `atol(typeof(Float64x2)) = atol(DataType)`, which again falls through to `atol(typeof(DataType)) = atol(DataType)` — infinite recursion, causing a `StackOverflowError` instead of a `MethodError`.
**MWE:**
```julia
using MultiFloats
using Meshes
f(x) = x[1]^(3/2)
geom = Meshes.Box((Float64x2(0.0),), (Float64x2(1.0),))
Meshes.integral(p -> f(Meshes.ustrip.(Meshes.to.(p))), geom)
```
**Error:**
```
ERROR: StackOverflowError:
Stacktrace:
 [1] rtol(x::Type) (repeats 26660 times)
   @ Meshes ~/.julia/dev/Meshes/src/tolerances.jl:39
```
**Environment :**
```julia
julia> versioninfo()
Julia Version 1.10.11
Commit a2b11907d7b (2026-03-09 14:59 UTC)
Build Info:
  Official https://julialang.org/ release
Platform Info:
  OS: macOS (arm64-apple-darwin22.4.0)
  CPU: 8 × Apple M3
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-15.0.7 (ORCJIT, apple-m1)
Threads: 1 default, 0 interactive, 1 GC (on 4 virtual cores)

julia> using Pkg; Pkg.status()
Status `~/.julia/environments/dev/Project.toml`
  [eacbb407] Meshes v0.57.21 `~/.julia/dev/Meshes`
```
**Proposed fix:**
Add a generic fallback for `AbstractFloat subtypes, mirroring the formula already used for `Float32`/`Float64`:
```julia
Meshes.atol(::Type{T}) where {T<:AbstractFloat} = eps(T)^(3 // 4)
Meshes.rtol(::Type{T}) where {T<:AbstractFloat} = eps(T)^(1 // 3)
```
placed after the existing `Float32`/`Float64` methods, so those remain the ones actually hit for the common cases, while `BigFloat` (and any other `AbstractFloat` subtype) gets a sensible default instead of recursing.

XamLe:
Closes #1388

**Problem**

`atol` and `rtol` only had concrete methods for `Float32` and `Float64`. Any other `AbstractFloat` subtype (e.g. `BigFloat` or `Float64x2` from [MultiFloats.jl](https://github.com/dzhang314/MultiFloats.jl/)) fell through to the value-based generic:
```
atol(x) = atol(typeof(x))
```
With no matching `atol(::Type{SomeOtherFloat})` method, calling e.g. `atol(BigFloat)` dispatches to `atol(x::Type)` → `atol(typeof(DataType))` → `atol(DataType)` → ... infinite recursion → `StackOverflowError`.

**Fix**

Add generic `AbstractFloat` fallbacks in src/tolerances.jl, placed after the `Float32`/`Float64` methods:
```
atol(::Type{T}) where {T<:AbstractFloat} = eps(T)^(3 // 4)
rtol(::Type{T}) where {T<:AbstractFloat} = eps(T)^(1 // 3)
```
Julia's most-specific dispatch ensures Float32 and Float64 continue to hit their cached-constant methods; all other `AbstractFloat` subtypes now get a sensible value using the same formula already used for the common types.

**Tests**

Four tests added to test/tolerances.jl using `BigFloat` (stdlib, no new dependency):

- `atol(BigFloat)` / `rtol(BigFloat)` return the expected eps-based values
- `atol(zero(BigFloat))` / `rtol(zero(BigFloat))` correctly dispatch via the value-based fallback

### DIFF
--- a/src/tolerances.jl
+++ b/src/tolerances.jl
@@ -21,6 +21,7 @@ isapprox(a::T, b::T, atol=atol(T))
 atol(x) = atol(typeof(x))
 atol(::Type{Float32}) = ATOL32[]
 atol(::Type{Float64}) = ATOL64[]
+atol(::Type{T}) where {T<:AbstractFloat} = eps(T)^(3 // 4)
 atol(𝒬::Type{<:Quantity}) = atol(numtype(𝒬)) * unit(𝒬)
 atol(𝒟::Type{<:ForwardDiff.Dual}) = atol(ForwardDiff.valtype(𝒟))
 
@@ -39,6 +40,7 @@ source code for numerical integration for example.
 rtol(x) = rtol(typeof(x))
 rtol(::Type{Float32}) = RTOL32[]
 rtol(::Type{Float64}) = RTOL64[]
+rtol(::Type{T}) where {T<:AbstractFloat} = eps(T)^(1 // 3)
 rtol(𝒬::Type{<:Quantity}) = rtol(numtype(𝒬))
 rtol(𝒟::Type{<:ForwardDiff.Dual}) = rtol(ForwardDiff.valtype(𝒟))
 
--- a/test/tolerances.jl
+++ b/test/tolerances.jl
@@ -30,6 +30,12 @@
   @inferred Meshes.rtol(𝒜)
   @inferred Meshes.rtol(𝒱)
 
+  # generic AbstractFloat fallback (e.g. BigFloat)
+  @test Meshes.atol(BigFloat) == eps(BigFloat)^(3 // 4)
+  @test Meshes.rtol(BigFloat) == eps(BigFloat)^(1 // 3)
+  @test Meshes.atol(zero(BigFloat)) == Meshes.atol(BigFloat)
+  @test Meshes.rtol(zero(BigFloat)) == Meshes.rtol(BigFloat)
+
   # maximum length for discretization
   @test Meshes.maxlen() == 500u"km"
 end


JoshuaLampert (approved):
Looks good to me. Thanks!

juliohm (approved):
Awesome!

juliohm:
Some of the benchmarks are affected apparently? Could you please confirm the new method is not overwriting the existing ones? The syntax `T<:AbstractFloat` might by counter-intuitive.

JoshuaLampert:
> Some of the benchmarks are affected apparently?

Are they? It looks like only `clipping/SutherlandHodgman` under julia v1.10 is a bit slower in the bechmark result, but it is within the standard deviation. On julia v1.12 the ratio is exactly 1.

juliohm:
Regressions on Julia LTS:

```
differentiation/triangle	0.371 ± 0.019 μs	0.411 ± 0.02 μs	0.903 ± 0.064
integration/triangle	9.4 ± 0.36 μs	10.2 ± 0.35 μs	0.922 ± 0.048
```

If we use the `+` we still get a number that is smaller than 1.

juliohm:
Tested locally and the right method is called:

```julia
julia> f(::Type{Float32}) = 1
f (generic function with 3 methods)

julia> f(::Type{Float64}) = 2
f (generic function with 4 methods)

julia> f(::Type{T}) where {T<:AbstractFloat} = 3
f (generic function with 5 methods)

julia> f(Float32)
1

julia> f(Float64)
2

julia> f(BigFloat)
3
```

What could explain the regression on LTS? Just random noise?

JoshuaLampert:
Yes, I would guess it's random noise. With v1.12 at least `integration/triangle` looks fine.

juliohm:
Let's restart the benchmark action on LTS to see if the regression is gone.

JoshuaLampert:
Now it is within the standard deviation :+1:

juliohm:
Good to know these benchmarks are suitable to considerable noise to the point of falsely pointing regressions.

Thank you @XamLe and @JoshuaLampert for the contribution and review!

juliohm:
Patch released: https://github.com/JuliaRegistries/General/pull/159942

JoshuaLampert:
Yes, I would only trust these benchmarks if the ratio significantly deviates from 1, say if `ratio + ratio_err < 0.8`. This is also the threshold used in this PR to give some indication for performance regressions or gains: https://github.com/MilesCranmer/AirspeedVelocity.jl/pull/91.

Issue #1383, opened 2026-06-23
`area`/`measure` of `PolyArea` incorrectly adds the area of the inner rings.


ghyatzo:
example:

```
julia> r
Ring
├─ Point(x: 0.0 m, y: 0.0 m)
├─ Point(x: 1.0 m, y: 0.0 m)
├─ Point(x: 1.0 m, y: 1.0 m)
└─ Point(x: 0.0 m, y: 1.0 m)

julia> ir
Ring
├─ Point(x: 0.25 m, y: 0.25 m)
├─ Point(x: 0.75 m, y: 0.25 m)
├─ Point(x: 0.75 m, y: 0.75 m)
└─ Point(x: 0.25 m, y: 0.75 m)

julia> pa = PolyArea(r, ir)
PolyArea
  outer
  └─ Ring((x: 0.0 m, y: 0.0 m), ..., (x: 0.0 m, y: 1.0 m))
  inner
  └─ Ring((x: 0.25 m, y: 0.25 m), ..., (x: 0.25 m, y: 0.75 m))

julia> area(pa)
1.25 m^2

julia> measure(pa)
1.25 m^2

julia> enclosedarea(r)
1.0 m^2

julia> enclosedarea(ir)
0.25 m^2
```

I'd expect `0.75 m^2` since the inner ring should represent a hole in the outer polygon.
(`enclosedarea` is a custom function)

ghyatzo:
I suspect this hits the generic `sum(measure, simplexify(geometry))` method

juliohm:
Is it another instance of wrong orientation of rings? People often open
this issue without reading the docstring. The inner rings must have CW
orientation (negative area).

Em ter., 23 de jun. de 2026, 15:30, cschen ***@***.***>
escreveu:

ghyatzo:
yup, my bad apologies

juliohm:
Please feel free to improve the docs to make this more visible. More people
will open the same issue in the future otherwise. I thought we had a
sentence in the PolyArea docstring but apparently not.

Em ter., 23 de jun. de 2026, 15:34, cschen ***@***.***>
escreveu:

Issue #1381, opened 2026-06-22
`area` on a ring should either error or return the enclosed area (with proper units)


ghyatzo:
as title,

motivating example:
```
julia> r
Ring
├─ Point(x: 0.0 m, y: 0.0 m)
├─ Point(x: 1.0 m, y: 0.0 m)
├─ Point(x: 1.0 m, y: 1.0 m)
└─ Point(x: 0.0 m, y: 1.0 m)

julia> area(r)
4.0 m # ???
```

I'd lean towards an error.

ghyatzo:
Now that I check, I also found that `rope` doesn't behave.
I guess that the `area` alias for `measure` needs to be a bit more specific on what can be called on.

juliohm:
Rope and Ring have parametric dimension 1 so the measure is the length. The
area can be called with geometries of parametric dimension 2. Check
paramdim.

The function measure is recommended if you need to be agnostic to
parametric dimension.

Em seg., 22 de jun. de 2026, 20:57, cschen ***@***.***>
escreveu:

juliohm:
In other words, length, area and volume are aliases of measure for specific
dimensions. I believe we throw an informative error message already, but
maybe some code path is being reached that skips the error message?

Em seg., 22 de jun. de 2026, 23:07, Júlio Hoffimann <
***@***.***> escreveu:

ghyatzo:
Why is this invalid?
Calling "area" on a 1 dimensional object should error.

I understand that measure is the generic dimension agnostic implementation and that works. But the area alias is not as generic. It should error and not return silently the wrong thing. 

Clearly the checks currently in place are not working properly.

ghyatzo:
The error is not being raised because `isparametrized` for rings returns `false`. Even though `paramdim` correctly returns `1`

juliohm:
> The error is not being raised because `isparametrized` for rings returns `false`. Even though `paramdim` correctly returns `1`

I just saw that. Opened https://github.com/JuliaGeometry/Meshes.jl/pull/1382

PR #1392, opened 2026-07-09
adds specialized version of area and centroid for 2D polygons

ghyatzo:
Hello,

This PR introduces a few specializations for 2D Polygons. The formula use the shoelace formula (for the area) and it's extention (through green's theorem) to also the centroid (when seen as the first area moment of a shape).

This formulation to keep things fast and non allocating works on rings as the primary building block, so this introduces a private `signedenclosedarea` that works on rings (for polygons there is the usual `area` or `measure`).  The presence of this utility is needed at the ring level because we need to calculate the area enclosed by a ring to calculate the centroid of a polygon with holes. and instead of constructing a polygon each time we just use the ring directly. and also because we need a way to compute the signed area of a polygon (`measure` returns alwasy positive and it was kept like that).

the centroid calculation can maybe be extended  to 3D polygons using `proj2D` for the time being, but 

this approach is quite a bit faster (~1000x) than the current approach, is non allocating and fixes some numerical instabilities due to the generic integration approach for centroid:

This PR:
```
julia> @btime centroid($poly)
  42.298 ns (0 allocations: 0 bytes)
├─ x: 0.5 m
└─ y: 0.0 m

julia> @btime area($poly)
  21.104 ns (0 allocations: 0 bytes)
6.0 m²
```

master:
```
julia> @btime centroid($poly)
  39.792 μs (2018 allocations: 59.89 KiB)
├─ x: 0.5000000000000001 m
└─ y: 0.0 m

julia> @btime area($poly)
  870.000 ns (17 allocations: 1.17 KiB)
6.0 m²
```

Additionally adds a tests to catch the case where the centroid of a CW oriented polygon is not properly signed. (Was good before, just is a nice thing to check just in case)

No LLM used for this PR.

### DIFF
--- a/src/centroid.jl
+++ b/src/centroid.jl
@@ -25,6 +25,32 @@ centroid(c::Circle) = center(c)
 
 centroid(c::Cylinder) = centroid(boundary(c))
 
+function centroid(p::Polygon{𝔼{2}})
+  # weighted average of the centroids by their
+  # signed (!!!) area (would be the mass, but it's uniform)
+  res = sum(rings(p)) do r
+    sum(segments(r)) do s
+      # calculates the centroid of the enclosed area by the ring
+      p₁, p₂ = to.(vertices(s))
+      x = (p₁[1], p₂[1])
+      y = (p₁[2], p₂[2])
+
+      k = x[1]*y[2] - x[2]*y[1]
+      SA[(x[2]+x[1])*k, (y[2]+y[1])*k]
+    end * 1/6 #= * 1/signedenclosedarea(r) * signedenclosedarea(r) =#
+    # the actual calculation needs dividing
+    # by  the enclosed area, but since we're
+    # doing a weighted average where the
+    # weight is the enclosed area we can simplify
+  end / sum(signedenclosedarea, rings(p))
+
+  # There is a subtle issue, if area is defined with "abs", then
+  # the centroid for a CW ring will be of flipped
+  # sign, while a signed area approach.
+
+  withcrs(p, res)
+end
+
 function centroid(c::CylinderSurface)
   a = centroid(bottom(c))
   b = centroid(top(c))
--- a/src/utils/misc.jl
+++ b/src/utils/misc.jl
@@ -13,6 +13,21 @@ function signarea(A::Point, B::Point, C::Point)
 end
 
 """
+    signedenclosedarea(r)
+
+Computes the enclosed area of a ring using the shoelace formula on its vertices.
+"""
+function signedenclosedarea(r::Ring{𝔼{2}})
+  sum(segments(r)) do s
+    p₁, p₂ = to.(vertices(s))
+    x = (p₁[1], p₂[1])
+    y = (p₁[2], p₂[2])
+
+    (x[2]+x[1])*(y[2]-y[1])/2
+  end
+end
+
+"""
     householderbasis(n)
 
 Returns a pair of orthonormal tangent vectors `u` and `v` from a normal `n`,
--- a/test/polytopes.jl
+++ b/test/polytopes.jl
@@ -650,9 +650,10 @@ end
   @test centroid(poly) ≈ cart(0.5, 0.5)
 
   # centroid is orientation-invariant
-  # (CW-oriented polygon should give the same centroid as CCW)
-  polyCW = PolyArea(reverse(vertices(Ring(cart.([(0, 0), (1, 0), (1, 1), (0, 1)])))))
-  @test centroid(polyCW) ≈ cart(0.5, 0.5)
+  verts = cart.([(0, 0), (1, 0), (1, 1), (0, 1)])
+  poly1 = PolyArea(verts)
+  poly2 = PolyArea(reverse(verts))
+  @test centroid(poly1) ≈ centroid(poly2)
 
   # single vertex access
   outer = cart.([(0, 0), (1, 0), (1, 1), (0, 1)])
--- a/test/polytopes.jl
+++ b/test/polytopes.jl
@@ -649,6 +649,11 @@ end
   poly = PolyArea(cart.([(0, 0), (1, 0), (1, 1), (0, 1)]))
   @test centroid(poly) ≈ cart(0.5, 0.5)
 
+  # centroid is orientation-invariant
+  # (CW-oriented polygon should give the same centroid as CCW)
+  polyCW = PolyArea(reverse(vertices(Ring(cart.([(0, 0), (1, 0), (1, 1), (0, 1)])))))
+  @test centroid(polyCW) ≈ cart(0.5, 0.5)
+
   # single vertex access
   outer = cart.([(0, 0), (1, 0), (1, 1), (0, 1)])
   hole1 = cart.([(0.2, 0.2), (0.4, 0.2), (0.4, 0.4), (0.2, 0.4)])
--- a/src/measures.jl
+++ b/src/measures.jl
@@ -89,15 +89,7 @@ function measure(t::Tetrahedron)
   abs((A - D) ⋅ ((B - D) × (C - D))) / 6
 end
 
-function measure(p::Polygon{𝔼{2}})
-  Σ = sum(rings(p)) do r
-    v = vertices(r)
-    n = nvertices(r)
-    c = centroid(r)
-    sum(signarea(c, v[i], v[i + 1]) for i in 1:n)
-  end
-  abs(Σ)
-end
+measure(p::Polygon{𝔼{2}}) = abs(sum(signedenclosedarea, rings(p)))
 
 measure(m::Multi{<:𝔼}) = sum(measure, parent(m))
 
--- a/src/orientation.jl
+++ b/src/orientation.jl
@@ -28,6 +28,11 @@ end
 
 orientation(r::Ring{𝔼{3}}) = orientation(proj2D(r))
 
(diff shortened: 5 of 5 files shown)


src/utils/misc.jl
```diff
@@ -12,6 +12,21 @@ function signarea(A::Point, B::Point, C::Point)
   ((B - A) × (C - A)) / 2
 end
 
+"""
+    signedenclosedarea(r)
+
+Computes the enclosed area of a ring using the shoelace formula on its vertices.
+"""
+function signedenclosedarea(r::Ring{𝔼{2}})
```
juliohm, line 20: Let's make this a new method of the existing `signedarea` that takes a `AbstractVector{<:Point}` as input. You can then call `signedarea(vertices(ring))` to recover the desired behavior in the algorithms for `area` and `centroid`.

src/utils/misc.jl
```diff
@@ -12,6 +12,21 @@ function signarea(A::Point, B::Point, C::Point)
...
+  sum(segments(r)) do s
+    p₁, p₂ = to.(vertices(s))
+    x = (p₁[1], p₂[1])
+    y = (p₁[2], p₂[2])
+
+    (x[2]+x[1])*(y[2]-y[1])/2
```
juliohm, line 26: Try to express this operation like we do in the other method above. Instead of manipulating raw Cartesian coordinates, let the product (B - A) x (C - A) do the low-level conversion. We use StaticArrays.jl for these points and vectors, so the compiler should be able to optimize everything equally well?

src/centroid.jl
```diff
@@ -25,6 +25,32 @@ centroid(c::Circle) = center(c)
 
 centroid(c::Cylinder) = centroid(boundary(c))
 
+function centroid(p::Polygon{𝔼{2}})
+  # weighted average of the centroids by their
+  # signed (!!!) area (would be the mass, but it's uniform)
+  res = sum(rings(p)) do r
+    sum(segments(r)) do s
```
juliohm, line 32: Can we re-express this low-level manipulation of Cartesian coordinates in terms of higher-level constructs like `signedarea`? I know it is tempting to extract two generic numbers `x` and `y` and do stuff, but we avoid it in order to support more CRS types. If a utility is missing, please feel free to add.

Ultimately, it is much easier specialize the utility for different CRS in a single place than to go over all algorithms that rely on the utility and rewrite the logic to support LatLon, Cartesian, Projected, ... systems.

test/polytopes.jl
```diff
@@ -649,6 +649,11 @@ end
   poly = PolyArea(cart.([(0, 0), (1, 0), (1, 1), (0, 1)]))
   @test centroid(poly) ≈ cart(0.5, 0.5)
 
+  # centroid is orientation-invariant
+  # (CW-oriented polygon should give the same centroid as CCW)
+  polyCW = PolyArea(reverse(vertices(Ring(cart.([(0, 0), (1, 0), (1, 1), (0, 1)])))))
+  @test centroid(polyCW) ≈ cart(0.5, 0.5)
```
```suggestion
  verts = cart.([(0, 0), (1, 0), (1, 1), (0, 1)])
  poly1 = PolyArea(verts)
  poly2 = PolyArea(reverse(verts))
  @test centroid(poly1) ≈ centroid(poly2)
```

juliohm (changes requested):
Thank you @ghyatzo ! 💯 Attached is a first round of review with suggestions to improve the adherence to the project.

src/utils/misc.jl
ghyatzo: Uhmm, I'll try and see If I can rearrange the calculations. What exactly is that `x` product? is it the cross product?

src/centroid.jl
ghyatzo: The formula seems similar to the shoelace one, but I don't think it can be expressed solely as a function of the `signedarea`. I understand your point, but with the current derivation, the algorithm exclusively works in 2D euclidean coordinates by construction. When I derived it I didn't need much else than that.

This approach can definitely be generalized to higher dimensions and different manifolds (using generalized stokes theorem, of which green's theorem is a special case). But that would require quite a bit of differential geometry machinery to handle differential forms, exterior derivatives and all that jazz. Maybe an alternative is going the geometric algebra way, but that as well would require quite a bit of machinery to set up.

It sounds like a super interesting problem to dive into, but at the moment I don't have the bandwidth for that kind of work unfortunately.

src/utils/misc.jl
ghyatzo: I see where this is coming from and I tried to write an implementation using just points. but it's a rather ugly exercise in index juggling, since we need to wrap around for the last connecting segment. Something that `segments` handles already very well, since it enforces the fact that rings are closed.

Haven't gotten around to see if it possible to implement the same thing using the `x` product approach yet (from the other comment), which would probably make the vertices approach more feasible. But for the time being I do think that it's just simpler and more robust to just be able to iterate over segments.

I'll revisit this as soon as I figure out the next comment.

juliohm: Are you aware that `vertices` of `Ring` are `CircularVector` and you can index pass the last point without problem?

src/utils/misc.jl
juliohm: I would try to come back to the abstraction of integrals. If the formula is a special case of some integral over some geometry, then we just need to specialize the method for `Cartesian` and `Projected` CRS.

The more you zoom out of these low-level coordinate details, the more you can reuse existing machinery.

src/centroid.jl
juliohm: Similar comment here. We don't need to support all CRS, but we need to know the concept behind the implementation to specialize methods for specific CRS. For  example, if the formula represents an integral over a `Ring` or `Polygon` with `Cartesian` coordinates, then we can just specialize that case. Let the Julia compiler pick the method for different CRS. Maybe this optimization should only exist for `Cartesian` and `Projected` because of the assumptions used in the derivation.

src/utils/misc.jl
ghyatzo: No, I missed that detail. Thanks for pointing that out.

juliohm:
@ghyatzo I implemented an optimization in #1393 and added a benchmark to track future changes.

Please start a fresh PR if you think that we can optimize this further.

ghyatzo:
So that also can be closed and reimplemented without attribution?

Sure

juliohm:
> So that also can be closed and reimplemented without attribution?
> 
> Sure

Oh man, this is such an unnecessary reaction. It is so difficult to do open source with this kind of feedback. We do our best to share high-quality software in our free time, and people only care about attribution. Even after we explicitly add their github handle to our [contributors list](https://github.com/JuliaEarth/GeoStats.jl#contributors).

@ghyatzo let me make it clear: you submitted a PR with poor, incomplete changes that would never be merged as per the Meshes.jl code standards. I've waited for a week to give you time to improve the work. Yesterday, I've started a fresh PR with a cleaner implementation of this idea (which is older than both of us), and I did it because improving the software quickly is more important than attributing such a trivial change to you.

I hope you can change your mindset in the future, or at least adjust your expectations. We are not here to praise contributors. We are here to evolve software together in our free time.

ghyatzo:
It's just a matter of communication.
As you mentioned, it's free time. There was no rush and I had a million other things to deal with. I had set aside time this weekend to address this and other issues.

I am not expecting compensation, praise or being hailed as the next Buddha. A thanks and a heads up that you were apparently in a rush was all that was needed. 

The approach might be older than both of us but clearly wasn't on your radar otherwise you would have used it already. 

The change you made, is not so trivial as you think it is. Since you're not just reformatting things. You're foundamentally changing the algorithm, which should require at least a smidge of formal verification to be sure it works. lest we end up again with a broken implementation that just seems right because we don't have the right set of tests. Maybe you did, I don't know.

In any case, glad this got in. Goodbye.

juliohm:
> A thanks and a heads up that you were apparently in a rush was all that was needed.

I could start requesting some gratitude from you as well, but I don't need that. You opened https://github.com/JuliaGeometry/Meshes.jl/issues/1381 and I fixed it within 24h. You opened https://github.com/JuliaGeometry/Meshes.jl/issues/1383 and I replied after 6min. I then adjusted the PR you submitted immediately so that the improvement to the documentation could reach more people, more quickly (https://github.com/JuliaGeometry/Meshes.jl/pull/1384) and I always try to say [Thank you!](https://github.com/JuliaGeometry/Meshes.jl/pull/1384#pullrequestreview-4556296705) out loud by the way. You opened https://github.com/JuliaGeometry/Meshes.jl/issues/1385 and I replied within 8min... When people set their mind to constant credit issues, a "thank you" is never enough. I hope you can put this conflict aside, and can continue contributing to the project. Your contributions are very welcome.

ghyatzo:
I do appreciate the effort you put in and you're right, I haven't said thank you myself. Let's drop it here, waste of energy. I shouldn't have replied while jet lagged and sleep deprived.

(My uneasiness about the new implementation remains though, I'll try and take a formal look at it this weekend anyway)

juliohm:
Thank you! Looking forward to further improvements. I believe the case with
concave polygons with extreme concentration of vertices in the concave ear
is the one that might need our attention. The signed area (and winding
number) is quite general in the sense that it only needs a reference point
inside the ring. I've chosen coordmean but any other point would produce a
similar result.

Em qui., 16 de jul. de 2026, 11:30, cschen ***@***.***>
escreveu:

PR #1393, opened 2026-07-15
Optimize centroid and measure of Polygon{E{2}}

juliohm:
Related to #1392


juliohm:
~15x speedup in Julia LTS and ~25x speedup in Julia latest release.

juliohm:
Jumped to ~92x speedup in Julia LTS and ~254x speedup in Julia latest release.

juliohm:
Jumped to ~192x speedup in Julia LTS and ~497x speedup in Julia latest release.

src/centroid.jl
```diff
@@ -53,6 +53,20 @@ centroid(t::Tetrahedron) = t(1 // 4, 1 // 4, 1 // 4)
 
 centroid(h::Hexahedron) = h(1 // 2, 1 // 2, 1 // 2)
 
+function centroid(p::Polygon{𝔼{2}})
+  c = sum(rings(p)) do r
+    v = vertices(r)
+    n = nvertices(r)
+    o = coordmean(v)
```
ghyatzo, line 60: You are reintroducing the average coordinate which was causing the issues before since it's not stable.

Have you formally verified that the sum of the triangles is independent from the choice of such center point?

juliohm:
I've added a test with the MWE of that issue. The reference point inside
the ring shouldn't be a problem as it is only used to form triangles in the
discretization. But I didn't prove it on paper.

Em qui., 16 de jul. de 2026, 00:59, cschen ***@***.***>
escreveu:

ghyatzo:
Confirm that the choice of third point in the triangle is completely arbitrary as long as it is the same for all segments.

PR #1394, opened 2026-07-16
Move Polygon{E{2}} up in the list of centroid specializations

juliohm:
To adhere to the convention in other source files.


juliohm:
Noticed that we already have other cases where we place polytopes of higher parametric dimension first. Closing it.

PR #1396, opened 2026-07-31
Use TileIterator to plot large grids

eliascarv:
This PR implements the `vizgrid!` using separate tiles for cases where the grid is very large, optimizing GPU memory usage.
@juliohm we need to define the minimum size for the grid to use this technique, as well as the tile size.

### DIFF
--- a/ext/grid/cartesian.jl
+++ b/ext/grid/cartesian.jl
@@ -4,9 +4,10 @@
 
 function vizgrid!(plot::Viz{<:Tuple{CartesianGrid}}, ::Type{<:𝔼}, ::Val{2}, ::Val{2})
   # visualize as built-in image with or without interpolation
-  Makie.map!(plot, [:object, :colorant], [:x, :y, :C, :interpolate]) do grid, colorant
+  Makie.map!(plot, [:object, :colorant], [:x, :y, :C, :interpolate, :usetiles]) do grid, colorant
     sz = size(grid)
     nv = nvertices(grid)
+    ne = nelements(grid)
     nc = colorant isa AbstractVector ? length(colorant) : 1
 
     x, y = map(Meshes.xyz(grid)) do c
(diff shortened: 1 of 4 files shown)


ext/grid/cartesian.jl
```diff
@@ -21,9 +22,30 @@ function vizgrid!(plot::Viz{<:Tuple{CartesianGrid}}, ::Type{<:𝔼}, ::Val{2}, :
...
 
-    x, y, C, (nc == nv)
+    x, y, C, (nc == nv), (ne ≥ 100000)
+  end
+
+  if plot.usetiles[]
+    Makie.map!(plot, [:object], [:xcoord, :ycoord]) do grid
+      map(c -> ustrip.(c), Meshes.xyz(grid))
+    end
+    for (i, tile) in enumerate(TileIterator(axes(plot.C[]), (100, 100)))
```
juliohm, line 32: I've tried with and without the tile technique and the speed is basically the same. Also tried playing with the tile size (256, 512,1024), and the performance isn't improved. I asked AI agents and they recommended a single call to `Makie.image!` because most of the time is spent in sending the texture to the GPU and multiple calls to `Makie.image!` can overload the scene (and are typically slower). I have a script where plotting tiles is much quicker and I will double check it to see if I am missing something.

juliohm:
@eliascarv I believe this optimization doesn't help us in terms of speed, but it might help in the case of RAM constraints. Though I believe we need to look into `Makie.Resampler` and `Makie.Pyramid` before committing to this manual tiling approach: https://docs.makie.org/stable/reference/plots/heatmap#Plotting-large-Heatmaps

juliohm:
@eliascarv did you have a chance to look Resampler or Pyramid? Should this PR be closed?

eliascarv:
@juliohm I'll close this PR. I'll check Resampler and Pyramid in another PR when I have free time.

PR #1397, opened 2026-07-31
Add `Segment`-`Polygon` intersection method

phchavesmaia:
I recently noticed that `intersection` does not implement a method for `Chain`–`Polygon` intersections. Currently, when no matching method is found, the function falls back to swapping the argument order. However, because there is also no `Polygon`–`Chain` method, it repeatedly swaps the arguments back and forth, resulting in a stack overflow. It follows a reproducible example:
```julia
using Meshes

rope = Rope([
    Point(0,0),
    Point(2,0)
])

poly = PolyArea([
    Point(1,-1),
    Point(3,-1),
    Point(3,1),
    Point(1,1)
])

intersection(rope, poly)
```

While working on a fix, I realized that the first step was to implement `Segment`–`Polygon` intersections, since `Chain`–`Polygon` intersections can naturally be built on top of them. This PR covers that first step.

I implemented this method by leveraging as much as I could on the previously implemented `Segment`-`Segment` intersection. This may not be the most efficient way to go about it, but it was the most immediate one to me right now.

**Result**: For my simple tests, the method is working just fine, but I have not stress tested it. When this PR gets completed, I hope to start working on the `Chain`-`Polygon` part.

### DIFF
--- a/src/intersections/segments.jl
+++ b/src/intersections/segments.jl
@@ -224,7 +224,7 @@ function intersection(f, seg::Segment{𝔼{2}}, poly::Polygon{𝔼{2}})
       intersects(sbox, boundingbox(edge)) || continue
       intersection(seg, edge) do I
         if type(I) == NotIntersecting
-          continue
+          return nothing
         elseif type(I) == Overlapping
           s = get(I)
           push!(splitpoints, vertices(s)...)
--- a/src/intersections/segments.jl
+++ b/src/intersections/segments.jl
@@ -209,7 +209,7 @@ end
 # 3. the segment is degenerate and is inside the polygon (Touching -> Point)
 # 4. the segment is degenerate and is outside the polygon (NotIntersecting -> Nothing)
 # 5. do not overlap nor intersect (NotIntersecting -> Nothing)
-function intersection(f, seg::Segment, poly::Polygon)
+function intersection(f, seg::Segment{𝔼{2}}, poly::Polygon{𝔼{2}})
 
   a, b = vertices(seg)
 
@@ -229,25 +229,27 @@ function intersection(f, seg::Segment, poly::Polygon)
   boundarypoints = Point[]
   for ring in rings(poly)
     for edge in segments(ring)
-      I = intersection(seg, edge)
-      itype = type(I)
-      # no intersection
-      if itype == NotIntersecting
-        continue
-      end
-      # some intersection
-      geom = get(I)
-      if geom isa Point # crosses or touches 
-        push!(boundarypoints, geom)
-        λ = ustrip((geom-a) ⋅ v / (v ⋅ v))
-        push!(segλs, λ)
-      elseif geom isa Segment # overlaps
-        c, d = vertices(geom)
-        λc = ustrip((c-a) ⋅ v / (v ⋅ v))
-        λd = ustrip((d-a) ⋅ v / (v ⋅ v))
-        push!(segλs, λc, λd)
-      else
-        @error "Unexpected Segment × Polygon edge intersection geometry: " * "$(typeof(geometry))"
+      if intersects(boundingbox(seg), boundingbox(edge))
+        intersection(seg, edge) do I
+          # no intersection
+          if type(I) == NotIntersecting
+            return
+          end
+          # some intersection
+          geom = get(I)
+          if geom isa Point # Crossing, EdgeTouching, CornerTouching, etc.
+            push!(boundarypoints, geom)
+            λ = ustrip((geom-a) ⋅ v / (v ⋅ v))
+            push!(segλs, λ)
+          elseif geom isa Segment # Overlapping collinear segments
+            c, d = vertices(geom)
+            λc = ustrip((c-a) ⋅ v / (v ⋅ v))
+            λd = ustrip((d-a) ⋅ v / (v ⋅ v))
+            push!(segλs, λc, λd)
+          else
+            @error "Unexpected Segment × Polygon edge intersection geometry: " * "$(typeof(geom))"
+          end
+        end
       end
     end
   end
@@ -270,7 +272,7 @@ function intersection(f, seg::Segment, poly::Polygon)
     geometry = length(pieces) == 1 ? only(pieces) : Multi(pieces) 
     return @IT Overlapping geometry f # Case 1
   elseif !isempty(boundarypoints) # no segments, but some intersection
-    geometry = length(boundarypoints) == 1 ? only(pieces) : first(pieces) # it gotta be a single point!
+    geometry = length(boundarypoints) == 1 ? only(boundarypoints) : first(boundarypoints) # it gotta be a single point!
     return @IT Touching geometry f # Case 2
   else
     return @IT NotIntersecting nothing f # Case 5 
--- a/src/intersections/segments.jl
+++ b/src/intersections/segments.jl
@@ -215,10 +215,13 @@ function intersection(f, seg::Segment{𝔼{2}}, poly::Polygon{𝔼{2}})
 
   # check for degenerate segments
   a, b = vertices(seg)
-  if a ≈ b && a ∈ poly && !any(v -> a ≈ v, vertices(poly))
-    return @IT Touching a f
-  elseif a ≈ b && any(v -> a ≈ v, vertices(poly))
-    return @IT CornerTouching a f
+  if a ≈ b
+    isvertex = any(≈(a), eachvertex(poly))
+    if a ∈ poly && !isvertex
+      return @IT Touching a f
+    elseif isvertex
+      return @IT CornerTouching a f
+    end
   end
 
   # extract flat xy coordinates
--- a/src/intersections/segments.jl
+++ b/src/intersections/segments.jl
@@ -268,21 +268,18 @@ function intersection(f, seg::Segment{𝔼{2}}, poly::Polygon{𝔼{2}})
   end
   resize!(events, i)
 
-  # create geometry from pieces
-  piecegeometry(ps) = length(ps) == 1 ? only(ps) : Multi(ps)
+  # glue pieces together into a single geometry
+  glue(pieces) = length(pieces) == 1 ? only(pieces) : Multi(pieces)
 
-  # create auxiliary variables
-  ncross = count(e -> e[2], events)
-  hasinterior = any(isinterior)
+  # number of crossing events
+  ncrosses = count(e -> e[2], events)
 
   # classify intersection
-  if hasinterior
-    geometry = piecegeometry(pieces)
-    return @IT Intersecting geometry f
+  if any(isinterior)
+    return @IT Intersecting glue(pieces) f
   elseif !isempty(pieces)
-    geometry = piecegeometry(pieces)
-    return @IT EdgeTouching geometry f
-  elseif ncross == 1
+    return @IT EdgeTouching glue(pieces) f
+  elseif ncrosses == 1
     i = findfirst(e -> e[2], events)
     p = events[i][1]
     isendpoint = p ≈ a || p ≈ b
@@ -292,7 +289,7 @@ function intersection(f, seg::Segment{𝔼{2}}, poly::Polygon{𝔼{2}})
     else
       return @IT CornerTouching p f
     end
-  elseif ncross > 1
+  elseif ncrosses > 1
     points = typeof(a)[e[1] for e in events if e[2]]
     return @IT Intersecting Multi(points) f
   else
--- a/src/intersections/segments.jl
+++ b/src/intersections/segments.jl
@@ -257,11 +257,11 @@ function intersection(f, seg::Segment{𝔼{2}}, poly::Polygon{𝔼{2}})
       # event is fully merged, connect it with the next event
       depth += eᵢ[3]
       piece = Segment(eᵢ[1], eⱼ[1])
-      if depth > 0 
-        push!(pieces, piece) 
-      elseif center(piece) ∈ poly 
-        inpoly = true 
-        push!(pieces, piece) 
+      if depth > 0
+        push!(pieces, piece)
+      elseif center(piece) ∈ poly
+        inpoly = true
+        push!(pieces, piece)
       end
       i += 1
       events[i] = eⱼ
--- a/src/intersections/segments.jl
+++ b/src/intersections/segments.jl
@@ -223,8 +223,9 @@ function intersection(f, seg::Segment{𝔼{2}}, poly::Polygon{𝔼{2}})
       intersects(sbox, boundingbox(edge)) || continue
       intersection(seg, edge) do I
         itype = type(I)
-        itype == NotIntersecting && return
-        if itype == Overlapping
+        if itype == NotIntersecting
+          return nothing
+        elseif itype == Overlapping
           c, d = vertices(get(I))
           if λ(d) < λ(c)
             c, d = d, c
--- a/src/intersections/segments.jl
+++ b/src/intersections/segments.jl
@@ -209,8 +209,6 @@ end
 # 5. the segment is degenerate and is inside the polygon (Touching -> Point)
 # 6. do not overlap nor intersect (NotIntersecting -> Nothing)
 function intersection(f, seg::Segment{𝔼{2}}, poly::Polygon{𝔼{2}})
-  a, b = vertices(seg)
-
   # check bounding boxes for early exit
   sbox = boundingbox(seg)
   pbox = boundingbox(poly)
@@ -219,6 +217,7 @@ function intersection(f, seg::Segment{𝔼{2}}, poly::Polygon{𝔼{2}})
   end
 
   # check for degenerate segments
+  a, b = vertices(seg)
   if a ≈ b && a ∈ poly && !any(v -> a ≈ v, vertices(poly))
     return @IT Touching a f
   elseif a ≈ b && any(v -> a ≈ v, vertices(poly))
--- a/src/intersections/segments.jl
+++ b/src/intersections/segments.jl
@@ -213,28 +213,22 @@ function intersection(f, seg::Segment{𝔼{2}}, poly::Polygon{𝔼{2}})
   pbox = boundingbox(poly)
   intersects(sbox, pbox) || return @IT NotIntersecting nothing f
 
-  # assess intersections
-  a, b = vertices(seg)
-  splitpoints = typeof(a)[a, b]
-  boundarypoints = typeof(a)[]
+  # store and classify intersection points
+  splitpoints = collect(eachvertex(seg))
+  boundarypoints = empty(splitpoints)
   boundaryoverlaps = typeof(seg)[]
-
   for ring in rings(poly)
     rbox = boundingbox(ring)
-    # check for obvious no intersection case beforehand
     intersects(sbox, rbox) || continue
     for edge in segments(ring)
-      # the selected axis ranges overlap, but the full bounding boxes may still be separated along the other axis
       intersects(sbox, boundingbox(edge)) || continue
       intersection(seg, edge) do I
-        itype = type(I)
-        if itype == NotIntersecting
-          return nothing
-        elseif itype == Overlapping
-          overlap = get(I)
-          c, d = vertices(overlap)
-          push!(splitpoints, c, d)
-          push!(boundaryoverlaps, overlap)
+        if type(I) == NotIntersecting
+          continue
+        elseif type(I) == Overlapping
+          s = get(I)
+          push!(splitpoints, vertices(s)...)
+          push!(boundaryoverlaps, s)
         else
           p = get(I)
           push!(splitpoints, p)
@@ -243,10 +237,14 @@ function intersection(f, seg::Segment{𝔼{2}}, poly::Polygon{𝔼{2}})
       end
     end
   end
-  unique!(boundarypoints)
-  dir = normalize(b-a)
-  λ(p) = ustrip((p - a) ⋅ dir)
+
+  # sort intersection points along the segment
+  a, b = vertices(seg)
+  λ(p) = (p - a) ⋅ (b - a)
   sort!(splitpoints, by=λ)
+
+  # remove duplicate points in boundary
+  unique!(boundarypoints)
 
   # create resulting intersection segments from base intersection points
   pieces = typeof(seg)[]
--- a/src/intersections/segments.jl
+++ b/src/intersections/segments.jl
@@ -204,7 +204,7 @@ function intersection(f, seg::Segment, line::Line)
 end
 
 # intersection between a segment and a polygon
-# 1. overlap of line and polygon (Overlapping -> Segment)
+# 1. overlap of segment and polygon (Overlapping -> Segment)
 # 2. intersect at one an point, exactly (Touching -> Point)
 # 3. the segment is degenerate and is inside the polygon (Touching -> Point)
 # 4. the segment is degenerate and is outside the polygon (NotIntersecting -> Nothing)
--- a/src/intersections/segments.jl
+++ b/src/intersections/segments.jl
@@ -208,7 +208,6 @@ end
 # 4. intersect at a single non-vertex point (Touching -> Point)
 # 5. do not overlap nor intersect (NotIntersecting -> Nothing)
 function intersection(f, seg::Segment{𝔼{2}}, poly::Polygon{𝔼{2}})
-  # early exit if bounding boxes do not intersect
   sbox = boundingbox(seg)
   pbox = boundingbox(poly)
   intersects(sbox, pbox) || return @IT NotIntersecting nothing f
@@ -236,7 +235,6 @@ function intersection(f, seg::Segment{𝔼{2}}, poly::Polygon{𝔼{2}})
         p = get(I)
         push!(events, (p, true, Int8(0)))
       end
-    end
     end
   end
 
--- a/src/intersections/segments.jl
+++ b/src/intersections/segments.jl
@@ -203,6 +203,203 @@ function intersection(f, seg::Segment, line::Line)
   end
 end
 
+# intersection between a segment and a polygon
+# 1. overlap with the polygon interior (Intersecting -> Segment or Multi)
+# 2. overlap with a polygon edge only (EdgeTouching -> Segment or Multi)
+# 3. intersect at an interior point of the segment and a polygon vertex (CornerTouching -> Point)
+# 4. intersect only at an endpoint of the segment (Touching -> Point)
+# 5. the segment is degenerate and is inside the polygon (Touching -> Point)
+# 6. do not overlap nor intersect (NotIntersecting -> Nothing)
+function intersection(f, seg::Segment{𝔼{2}}, poly::Polygon{𝔼{2}})
+  a, b = vertices(seg)
+  segbbox = boundingbox(seg)
+  polybbox = boundingbox(poly)
+
+  # check for degenerate segments
+  if a ≈ b && a ∈ poly
+    return @IT Touching a f # Case 5
+  end
+
+  # assess obvious no intersection case beforehand
+  if !intersects(segbbox, polybbox)
+    return @IT NotIntersecting nothing f # Case 6
+  end
+
+  # segment endpoints as scalars + segment vector
+  segλs = [0.0, 1.0]
+  v = b - a
+  v² = v ⋅ v
+
+  # extract the first Cartesian coordinate
+  x(p) = flat(coords(p)).x
+  y(p) = flat(coords(p)).y
+
+  # choose the polygon's longest bounding-box axis
+  polymin, polymax = extrema(polybbox)
+  xside = x(polymax) - x(polymin)
+  yside = y(polymax) - y(polymin)
+  axiscoord = xside ≥ yside ? x : y
+
+  # segment range along the chosen axis
+  segmin, segmax = extrema(segbbox)
+  segstart = axiscoord(segmin)
+  segstop = axiscoord(segmax)
+
+  # assess intersections
+  boundarypoints = typeof(a)[]
+  boundaryoverlaps = Tuple{Float64,Float64}[]
+
+  for ring in rings(poly)
+    ringbbox = boundingbox(ring)
+
+    # check for obvious no intersection case beforehand
+    intersects(segbbox, ringbbox) || continue
+
+    edges = collect(segments(ring))
+    boxes = boundingbox.(edges)
+
+    # edge ranges
+    starts = map(boxes) do box
+      pmin, _ = extrema(box)
+      axiscoord(pmin)
+    end
+    stops = map(boxes) do box
+      _, pmax = extrema(box)
+      axiscoord(pmax)
+    end
+
+    # Sort by minimum coordinate
+    inds = sortperm(starts)
+    edges = edges[inds]
+    boxes = boxes[inds]
+    starts = starts[inds]
+    stops = stops[inds]
+
+    for i in eachindex(edges)
+      # edge ends before the segment begins. Later edges may still overlap, so skip only this edge.
+      stops[i] < segstart && continue
+
+      # edge starts after the segment ends. Since edges are sorted by the selected axis minimum, all later edges are also too far right.
+      starts[i] > segstop && break
+
+      # the selected axis ranges overlap, but the full bounding boxes may still be separated along the other axis
+      intersects(segbbox, boxes[i]) || continue
+
+      edge = edges[i]
+      intersection(seg, edge) do I
+        itype = type(I)
+        # no intersection
+        if itype == NotIntersecting
+          return
+        elseif itype == Overlapping
+          # Overlapping segments
+          overlap = get(I)
+          c, d = vertices(overlap)
+          λc = ustrip((c - a) ⋅ v / v²)
+          λd = ustrip((d - a) ⋅ v / v²)
+          λ₁, λ₂ = minmax(λc, λd)
+          push!(segλs, λ₁, λ₂)
+          push!(boundaryoverlaps, (λ₁, λ₂))
+        else
+          # Crossing, EdgeTouching, CornerTouching, etc.
+          p = get(I)
+          push!(boundarypoints, p)
+          λ = ustrip((p - a) ⋅ v / v²)
+          push!(segλs, λ)
+        end
+      end
+    end
+  end
+
+  # get unique boundary points
+  uniquepoints = typeof(a)[]
+  for p in boundarypoints
+    any(q -> q ≈ p, uniquepoints) || push!(uniquepoints, p)
+  end
+  boundarypoints = uniquepoints
+
+  # remove duplicate λ values and sort them
+  sort!(segλs)
+  for i in 2:length(segλs)
+    segλs[i] = mayberound(segλs[i], segλs[i - 1])
+  end
+  unique!(segλs)
+
+  # check if a λ value lies in a boundary overlap
+  function inboundaryoverlap(λ)
+    any(boundaryoverlaps) do (λ₁, λ₂)
+      λ₁ < λ < λ₂ || λ ≈ λ₁ || λ ≈ λ₂
+    end
+  end
+
+  # create segments from the λ values
+  pieces = typeof(seg)[]
+  interiorpieces = typeof(seg)[]
+  boundarypieces = typeof(seg)[]
+  for (λ₁, λ₂) in zip(segλs[1:(end - 1)], segλs[2:end])
+    λ₁ ≈ λ₂ && continue
+    λmid = (λ₁ + λ₂) / 2
+    piece = Segment(seg(λ₁), seg(λ₂))
+
+    if inboundaryoverlap(λmid)
+      push!(pieces, piece)
+      push!(boundarypieces, piece)
+    elseif seg(λmid) ∈ poly
+      push!(pieces, piece)
+      push!(interiorpieces, piece)
+    end
+  end
+
+  # merge continuous segments
+  function mergepieces(pieces)
+    length(pieces) ≤ 1 && return pieces
+
+    merged = eltype(pieces)[]
+    for piece in pieces
+      if isempty(merged)
+        push!(merged, piece)
+      else
+        previous = last(merged)
+        p₁, p₂ = vertices(previous)
+        q₁, q₂ = vertices(piece)
+        if p₂ ≈ q₁
+          merged[end] = Segment(p₁, q₂)
+        else
+          push!(merged, piece)
+        end
+      end
+    end
+    merged
+  end
+
+  pieces = mergepieces(pieces)
+  interiorpieces = mergepieces(interiorpieces)
+  boundarypieces = mergepieces(boundarypieces)
+
+  # create geometry from pieces
+  piecegeometry(pieces) = length(pieces) == 1 ? only(pieces) : Multi(pieces)
+
+  # classifying intersections
+  if !isempty(interiorpieces)
+    geometry = piecegeometry(pieces)
+    return @IT Intersecting geometry f # Case 1
+  elseif !isempty(boundarypieces)
+    geometry = piecegeometry(boundarypieces)
+    return @IT EdgeTouching geometry f # Case 2
+  elseif length(boundarypoints) == 1
+    p = only(boundarypoints)
+    if p ≈ a || p ≈ b
+      return @IT Touching p f # Case 4
+    else
+      return @IT CornerTouching p f # Case 3
+    end
+  elseif !isempty(boundarypoints)
+    return @IT Intersecting Multi(boundarypoints) f # Case 1
+  else
+    return @IT NotIntersecting nothing f # Case 6
+  end
+end
+
 # Algorithm 4 of Jiménez, J., Segura, R. and Feito, F. 2009.
 # (https://www.sciencedirect.com/science/article/pii/S0925772109001448?via%3Dihub)
 function intersection(f, seg::Segment{𝔼{3}}, tri::Triangle{𝔼{3}})
@@ -329,156 +526,6 @@ function intersection(f, seg::Segment{𝔼{3}}, tri::Triangle{𝔼{3}})
   p = Segment(Q1, Q2)(λ)
 
   return @IT Intersecting p f
-end
-
-# intersection between a segment and a polygon
-# 1. overlap of segment and polygon (Overlapping -> Segment)
-# 2. intersect at one point, exactly (Touching -> Point)
-# 3. the segment is degenerate and is inside the polygon (Touching -> Point)
-# 4. do not overlap nor intersect (NotIntersecting -> Nothing)
-function intersection(f, seg::Segment{𝔼{2}}, poly::Polygon{𝔼{2}})
-  a, b = vertices(seg)
-  segbbox = boundingbox(seg)
-  polybbox = boundingbox(poly)
-
-  # check for degenerate segments
-  if a ≈ b && a ∈ poly
-    return @IT Touching a f # Case 3
-  end
-
-  # assess obvious no intersection case beforehand
-  if !intersects(segbbox, polybbox)
-    return @IT NotIntersecting nothing f # Case 4
-  end
-
-  # segment endpoints as scalars + segment vector
-  segλs = [0.0, 1.0]
-  v = b - a
-
-  # Extract the first Cartesian coordinate
-  x(p) = flat(coords(p)).x
-  y(p) = flat(coords(p)).y
-
-  # Choose the polygon's longest bounding-box axis
-  polymin, polymax = extrema(polybbox)
-  xside = x(polymax) - x(polymin)
-  yside = y(polymax) - y(polymin)
-  axiscoord = xside ≥ yside ? x : y
-
-  # Segment range along the chosen axis
-  segmin, segmax = extrema(segbbox)
-  segstart = axiscoord(segmin)
-  segstop = axiscoord(segmax)
-
-  # assess intersections
-  boundarypoints = typeof(a)[]
-  for ring in rings(poly)
-    ringbbox = boundingbox(ring)
-    # check for obvious no intersection case beforehand
-    intersects(segbbox, ringbbox) || continue
-    edges = collect(segments(ring))
-    boxes = boundingbox.(edges)
-    # Edge ranges
-    starts = map(boxes) do box
-      pmin, _ = extrema(box)
-      axiscoord(pmin)
-    end
-    stops = map(boxes) do box
-      _, pmax = extrema(box)
-      axiscoord(pmax)
-    end
-
-    # Sort by minimum coordinate
-    inds = sortperm(starts)
-    edges = edges[inds]
-    boxes = boxes[inds]
-    starts = starts[inds]
-    stops = stops[inds]
-    for i in eachindex(edges)
-      # Edge ends before the segment begins. Later edges may still overlap, so skip only this edge.
-      stops[i] < segstart && continue
-
-      # Edge starts after the segment ends. Since edges are sorted by the selected axis minimum, all later edges are also too far right.
-      starts[i] > segstop && break
-
-      # The selected axis ranges overlap, but the full bounding boxes may still be separated along the other axis
-      intersects(segbbox, boxes[i]) || continue
-
-      edge = edges[i]
-      intersection(seg, edge) do I
-        itype = type(I)
-        # no intersection
-        if itype == NotIntersecting
-          return
-        elseif itype == Overlapping
-          # Overlapping segments
-          seg = get(I)
-          c, d = vertices(seg)
-          λc = ustrip((c-a) ⋅ v / (v ⋅ v))
-          λd = ustrip((d-a) ⋅ v / (v ⋅ v))
-          push!(segλs, λc, λd)
-        else
-          # Crossing, EdgeTouching, CornerTouching, etc.
-          p = get(I)
-          push!(boundarypoints, p)
-          λ = ustrip((p-a) ⋅ v / (v ⋅ v))
-          push!(segλs, λ)
-        end
-      end
-    end
-  end
-
-  # get unique boundary points
-  unique!(boundarypoints)
-  # remove duplicate λ values and sort them
-  sort!(segλs)
-  for i in 2:length(segλs)
-    segλs[i] = mayberound(segλs[i], segλs[i - 1])
-  end
-  unique!(segλs)
-  # create segments from the λ values that are inside the polygon
-  pieces = typeof(seg)[]
-  for (λ₁, λ₂) in zip(segλs[1:(end - 1)], segλs[2:end])
-    λ₁ ≈ λ₂ && continue
-    λmid = (λ₁ + λ₂) / 2
-    if seg(λmid) ∈ poly
-      push!(pieces, Segment(seg(λ₁), seg(λ₂)))
-    end
-  end
-  # merge continuous segments
-  if length(pieces) > 1
-    merged = typeof(seg)[]
-    for piece in pieces
-      if isempty(merged)
-        push!(merged, piece)
-      else
-        previous = last(merged)
-        if !intersects(boundingbox(previous), boundingbox(piece))
-          push!(merged, piece)
-          continue
-        end
-        p₁, p₂ = vertices(previous)
-        q₁, q₂ = vertices(piece)
-        if p₂ ≈ q₁
-          merged[end] = Segment(p₁, q₂)
-        else
-          push!(merged, piece)
-        end
-      end
-    end
-    pieces = merged
-  end
-
-  # classifying intersections
-  if !isempty(pieces) # positive-length intersection
-    geometry = length(pieces) == 1 ? only(pieces) : Multi(pieces)
-    return @IT Overlapping geometry f # Case 1
-  elseif !isempty(boundarypoints) # no segments, but some intersection
-    geometry = length(boundarypoints) == 1 ? only(boundarypoints) : Multi(boundarypoints)
-    return @IT Touching geometry f # Case 2
-  else
-    return @IT NotIntersecting nothing f # Case 5 
-  end
 end
 
 # sorts four numbers using a sorting network 
--- a/src/intersections/segments.jl
+++ b/src/intersections/segments.jl
@@ -283,8 +283,7 @@ function intersection(f, seg::Segment{𝔼{2}}, poly::Polygon{𝔼{2}})
   elseif ncrosses == 1
     i = findfirst(e -> e[2], events)
     p = events[i][1]
-    isvertex = any(v -> p ≈ v, eachvertex(poly))
-    if isvertex
+    if any(≈(p), eachvertex(poly))
       return @IT CornerTouching p f
     else
       return @IT Touching p f
--- a/src/intersections/segments.jl
+++ b/src/intersections/segments.jl
@@ -208,7 +208,7 @@ end
 # 4. intersect at a single non-vertex point (Touching -> Point)
 # 5. do not overlap nor intersect (NotIntersecting -> Nothing)
 function intersection(f, seg::Segment{𝔼{2}}, poly::Polygon{𝔼{2}})
-  # check bounding boxes for early exit
+  # early exit if bounding boxes do not intersect
   sbox = boundingbox(seg)
   pbox = boundingbox(poly)
   intersects(sbox, pbox) || return @IT NotIntersecting nothing f
@@ -224,18 +224,16 @@ function intersection(f, seg::Segment{𝔼{2}}, poly::Polygon{𝔼{2}})
     end
   end
 
-  # extract flat xy coordinates
+  # coordinate with largest bounding box side
+  xside, yside = sides(pbox)
   x(p) = flat(coords(p)).x
   y(p) = flat(coords(p)).y
-
-  # choose the polygon's longest bounding box axis
-  xside, yside = sides(pbox)
-  axiscoord = xside ≥ yside ? x : y
+  coord = xside ≥ yside ? x : y
 
   # segment range along the chosen axis
   smin, smax = extrema(sbox)
-  sstart = axiscoord(smin)
-  sstop = axiscoord(smax)
+  sstart = coord(smin)
+  sstop = coord(smax)
 
   # assess intersections
   boundarypoints = typeof(a)[]
@@ -257,11 +255,11 @@ function intersection(f, seg::Segment{𝔼{2}}, poly::Polygon{𝔼{2}})
     # edge ranges
     estart = map(ebox) do box
       pmin, _ = extrema(box)
-      axiscoord(pmin)
+      coord(pmin)
     end
     estop = map(ebox) do box
       _, pmax = extrema(box)
-      axiscoord(pmax)
+      coord(pmax)
     end
 
     # sort by minimum coordinate
--- a/src/intersections/segments.jl
+++ b/src/intersections/segments.jl
@@ -283,7 +283,7 @@ function intersection(f, seg::Segment{𝔼{2}}, poly::Polygon{𝔼{2}})
   elseif ncrosses == 1
     i = findfirst(e -> e[2], events)
     p = events[i][1]
-    isvertex = any(v -> p ≈ v, vertices(poly))
+    isvertex = any(v -> p ≈ v, eachvertex(poly))
     if isvertex
       return @IT CornerTouching p f
     else
--- a/src/intersections/segments.jl
+++ b/src/intersections/segments.jl
@@ -215,20 +215,19 @@ function intersection(f, seg::Segment{𝔼{2}}, poly::Polygon{𝔼{2}})
   sbox = boundingbox(seg)
   pbox = boundingbox(poly)
   if !intersects(sbox, pbox)
-    return @IT NotIntersecting nothing f # Case 6
+    return @IT NotIntersecting nothing f
   end
 
   # check for degenerate segments
   if a ≈ b && a ∈ poly && !any(v -> a ≈ v, vertices(poly))
-    return @IT Touching a f # Case 5
+    return @IT Touching a f
   elseif a ≈ b && any(v -> a ≈ v, vertices(poly))
-    return @IT CornerTouching a f # Case 3
+    return @IT CornerTouching a f
   end
 
   # segment endpoints as scalars + segment vector
-  segλs = [0.0, 1.0]
-  v = b - a
-  v² = v ⋅ v
+  λs = [0.0, 1.0]
+  v = normalize(b - a)
 
   # extract flat xy coordinates
   x(p) = flat(coords(p)).x
@@ -268,7 +267,7 @@ function intersection(f, seg::Segment{𝔼{2}}, poly::Polygon{𝔼{2}})
       axiscoord(pmax)
     end
 
-    # Sort by minimum coordinate
+    # sort by minimum coordinate
     inds = sortperm(starts)
     edges = edges[inds]
     boxes = boxes[inds]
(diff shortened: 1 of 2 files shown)


juliohm:
Thank you @phchavesmaia ! I will try to review it tomorrow 💯

src/intersections/segments.jl
```diff
@@ -203,9 +203,84 @@ function intersection(f, seg::Segment, line::Line)
...
+  else
+    return @IT NotIntersecting nothing f # Case 5 
+  end 
+
+end
+
 # Algorithm 4 of Jiménez, J., Segura, R. and Feito, F. 2009.
 # (https://www.sciencedirect.com/science/article/pii/S0925772109001448?via%3Dihub)
-function intersection(f, seg::Segment, tri::Triangle)
+function intersection(f, seg::Segment{𝔼{3}}, tri::Triangle{𝔼{3}})
```
juliohm, line 325: Do you mean this method assumes objects embedded in E{3}? I have to read the paper again to remember the algorithm. If this change is unrelated to the new method, let's undo it.

src/intersections/segments.jl
```diff
@@ -203,9 +203,84 @@ function intersection(f, seg::Segment, line::Line)
...
+      elseif geom isa Segment # overlaps
+        c, d = vertices(geom)
+        λc = ustrip((c-a) ⋅ v / (v ⋅ v))
+        λd = ustrip((d-a) ⋅ v / (v ⋅ v))
+        push!(segλs, λc, λd)
+      else
+        @error "Unexpected Segment × Polygon edge intersection geometry: " * "$(typeof(geometry))"
+      end
+    end
+  end
```
juliohm: From what I am understanding, you are trying to obtain all points of intersection between the query segment and the segments of the rings of the polygon. Take a look at the `pairwiseintersect` utility function added by @souma4. It uses a "sweeping" algorithm that has very good performance.

juliohm (commented):
@phchavesmaia thank you for another awesome contribution. I've left some comments/questions for your assessment.

src/intersections/segments.jl
phchavesmaia: Yes, I do. My first attempt at implementing the `Segment`-`Polygon` intersection relied on this `Triangle`-`Segment` method. My idea was to decompose an arbitrary *n*-gon into triangles, making the polygon case a straightforward generalization of the triangle one. Unfortunately, as far as I understand, this method assumes that the segment and the triangle are not coplanar. Since that assumption does not hold in my case, I had to pursue a different approach.

I though about adding it into this PR since the `Polygon` call may conflict with the `Triangle` one (or vice-versa in the E{2} case).

src/intersections/segments.jl
phchavesmaia: I will look into it!

phchavesmaia: I read the function and, as far as I understood it, it is a great way to reduce a O(m⋅n) problem to a O((m+n+k)⋅log(m+n)) one, where m and n are the number of segments from the two sets and k is the number of intersections. In fact, this gives me ideas on implementing it when the time to make a `Chain`-`Polygon` algorithm comes, although I would probably have to adapt it in some way (I don't think it catches overlapping segments, but I may be wrong). 

Going back to this PR, the `Segment`-`Polygon` problem is a O(1⋅n) problem, so there should not be many (if any) gains to be made in implementing `pairwiseintersect` here. I may be missing something, though.

src/intersections/segments.jl
juliohm: Thank you for checking the assumptions behind these methods. We should probably add comments to make these assumptions explicit.

src/intersections/segments.jl
juliohm: Good observation. I think the `pairwiseintersect` won't be very useful in this case. You could still leverage the "sweep line" trick and only test segments that are within the xmin and xmax of the query segment for example. If the segment is beyond this interval, then it surely does not intersect.

src/intersections/segments.jl
```diff
@@ -203,9 +203,94 @@ function intersection(f, seg::Segment, line::Line)
...
+
+  # assess obvious no intersection case beforehand
+  if !intersects(segbbox, polybbox)
+    return @IT NotIntersecting nothing f # Case 5
+  end
+  
+  # segment endpoints as scalars + segment vector
+  segλs = [0.0, 1.0]; v = b-a;
```
juliohm: Try to press `Ctrl+Shift+I` in your VSCode session (if you are using this IDE) to apply the formatting settings of the project. We store a format file in the repository.

src/intersections/segments.jl
```diff
@@ -203,9 +203,94 @@ function intersection(f, seg::Segment, line::Line)
...
 
+# intersection between a segment and a polygon
+# 1. overlap of line and polygon (Overlapping -> Segment)
+# 2. intersect at one an point, exactly (Touching -> Point)
+# 3. the segment is degenerate and is inside the polygon (Touching -> Point)
+# 4. the segment is degenerate and is outside the polygon (NotIntersecting -> Nothing)
+# 5. do not overlap nor intersect (NotIntersecting -> Nothing)
+function intersection(f, seg::Segment{𝔼{2}}, poly::Polygon{𝔼{2}})
+
+  a, b = vertices(seg); segbbox = boundingbox(seg); polybbox = boundingbox(poly)
```
juliohm: Try to define one variable per line. In this case, the definition of `segbbox` and `polybbox` deserve a separate line each.

src/intersections/segments.jl
```diff
@@ -203,9 +203,94 @@ function intersection(f, seg::Segment, line::Line)
...
+  segλs = [0.0, 1.0]; v = b-a;
+
+  # assess intersections with the edges
+  boundarypoints = Point[]
+  for ring in rings(poly)
+    ringbbox = boundingbox(ring)
+    for edge in segments(ring)
+      intersects(segbbox, ringbbox) || continue
+      edgebbox = boundingbox(edge)
+      if intersects(segbbox, edgebbox)
```
juliohm: Nice. So you adopted a bounding-box optimization here too. I wonder if we gain more performance by just checking the `x` coordinates of the vertices. Probably not worth it. The `pairwiseintersect` function I mentioned does an `x`-sort to only scan relevant segments that are "near" the query segment.

src/intersections/segments.jl
```diff
@@ -203,9 +203,94 @@ function intersection(f, seg::Segment, line::Line)
   end
 end
 
+# intersection between a segment and a polygon
+# 1. overlap of line and polygon (Overlapping -> Segment)
```
```suggestion
# 1. overlap of segment and polygon (Overlapping -> Segment)
```

src/intersections/segments.jl
```diff
@@ -203,9 +203,94 @@ function intersection(f, seg::Segment, line::Line)
...
+      edgebbox = boundingbox(edge)
+      if intersects(segbbox, edgebbox)
+        intersection(seg, edge) do I
+          # no intersection
+          if type(I) == NotIntersecting
+            return
+          end
+          # some intersection
+          geom = get(I)
+          if geom isa Point # Crossing, EdgeTouching, CornerTouching, etc.
```
juliohm: Here we could test for the `type(I)` instead in an "else" branch. It is more idiomatic to test for the type of intersection (an enum) and then call `get(I)` knowing what the geometry type is.

src/intersections/segments.jl
```diff
@@ -203,9 +203,94 @@ function intersection(f, seg::Segment, line::Line)
...
+            push!(boundarypoints, geom)
+            λ = ustrip((geom-a) ⋅ v / (v ⋅ v))
+            push!(segλs, λ)
+          elseif geom isa Segment # Overlapping collinear segments
+            c, d = vertices(geom)
+            λc = ustrip((c-a) ⋅ v / (v ⋅ v))
+            λd = ustrip((d-a) ⋅ v / (v ⋅ v))
+            push!(segλs, λc, λd)
+          else
+            @error "Unexpected Segment × Polygon edge intersection geometry: " * "$(typeof(geom))"
```
juliohm: If this branch is not reachable, we can eliminate it to help the compiler a bit.

src/intersections/segments.jl
phchavesmaia: Oh, did not know about this, I will try it!

src/intersections/segments.jl
phchavesmaia: I tried implementing this idea and observed a slight increase in efficiency when using the new algorithm in more complex polygons. When it comes to simpler polygons, the opposite holds. 

That said, the changes are **really** small. It was probably not worth the effort, but it was a nice learning experience nonetheless! 

I think a more robust way to accelerate these intersection checks would be to add a BVH spatial indexing utility in `Meshes.jl`. Besides improving these intersection queries, it would also be extremely useful for functions such as `geojoin` in `GeoTables.jl`, along with many others that could benefit from a static spatial index. I may work on a simple implementation in the future, as it is a much more manageable project than implementing an R-tree spatial index.

src/intersections/segments.jl
```diff
@@ -203,9 +203,134 @@ function intersection(f, seg::Segment, line::Line)
   end
 end
 
+# intersection between a segment and a polygon
+# 1. overlap of segment and polygon (Overlapping -> Segment)
+# 2. intersect at one an point, exactly (Touching -> Point)
+# 3. the segment is degenerate and is inside the polygon (Touching -> Point)
+# 4. the segment is degenerate and is outside the polygon (NotIntersecting -> Nothing)
+# 5. do not overlap nor intersect (NotIntersecting -> Nothing)
```
juliohm: Cases (4) and (5) are the same with difference that (4) is about degenerate segments, right? Maybe we should merge the two into the description (5)?

src/intersections/segments.jl
```diff
@@ -203,9 +203,134 @@ function intersection(f, seg::Segment, line::Line)
...
+  polybbox = boundingbox(poly)
+
+  # check for degenerate segments
+  if a ≈ b
+    if a ∈ poly
+      return @IT Touching a f # Case 3
+    else
+      return @IT NotIntersecting nothing f # Case 4
+    end
+  end
```
juliohm: Just double-checking that we need to be explicit about degenerate segments because the algorithm that comes next fails to catch this case. Is that right? I see you compute `v=b-a` and then divide by `dot(v, v)` in a few branches. Could we just handle this as a special case with `Touching` intersection type instead? If the situation is more complex, please ignore this comment. The two degenerate branches are `Touching` and `NotIntersecting`. The latter intersection type returns `nothing` so no division by `dot(v,v)` needed.

src/intersections/segments.jl
```diff
@@ -203,9 +203,134 @@ function intersection(f, seg::Segment, line::Line)
...
+
+  # assess obvious no intersection case beforehand
+  if !intersects(segbbox, polybbox)
+    return @IT NotIntersecting nothing f # Case 5
+  end
+
+  # segment endpoints as scalars + segment vector
+  segλs = [0.0, 1.0];
+  v = b-a;
```
juliohm: For some reason the formatter is not picking this extra semi-colon and the whitespace between b-a. I can adjust manually after we converge with the implementation.

src/intersections/segments.jl
```diff
@@ -203,9 +203,134 @@ function intersection(f, seg::Segment, line::Line)
...
+
+  # Extract the first Cartesian coordinate
+  x(p) = flat(coords(p)).x
+  y(p) = flat(coords(p)).y
+
+  # Choose the polygon's longest bounding-box axis
+  polymin, polymax = extrema(polybbox)
+  xspan = x(polymax) - x(polymin)
+  yspan = y(polymax) - y(polymin)
+  axiscoord = xspan ≥ yspan ? x : y
```
juliohm: Technically, these subtractions are "sides" and not "span" (which are ranges), I would rename them to `xside` and `yside`.

src/intersections/segments.jl
```diff
@@ -203,9 +203,134 @@ function intersection(f, seg::Segment, line::Line)
...
+
+  # Segment range along the chosen axis
+  segmin, segmax = extrema(segbbox)
+  segstart = axiscoord(segmin)
+  segstop = axiscoord(segmax)
+
+  # assess intersections
+  boundarypoints = Point[]
```
juliohm: We need to make sure that we are propagating the concrete point type within the geometries. The type `Point` is abstract, which can compromise performance. You could ask for the `typeof` some reference point like `segmin` or `segmax` above for the initialization of the list.

src/intersections/segments.jl
```diff
@@ -203,9 +203,134 @@ function intersection(f, seg::Segment, line::Line)
...
+  for (λ₁, λ₂) in zip(segλs[1:(end - 1)], segλs[2:end])
+    λ₁ ≈ λ₂ && continue
+    λmid = (λ₁ + λ₂) / 2
+    if seg(λmid) ∈ poly
+      push!(pieces, Segment(seg(λ₁), seg(λ₂)))
+    end
+  end
+
+  # classifying intersections
+  if !isempty(pieces) # Positive-length intersection
```
```suggestion
  if !isempty(pieces) # positive-length intersection
```

We try to avoid capital case in comments, just a minor detail

juliohm (changes requested):
Looking great! Attached a second round of review.

src/intersections/segments.jl
juliohm: It would be wonderful if we had a BVH utility! ❤️

phchavesmaia (commented):
Code after 2nd round of review. I added some tests and it is passing, though they are not as thorough as possible.

test/intersections.jl
```diff
@@ -451,6 +451,65 @@ end
...
+  s = Segment(cart(-1, 2), cart(5, 2))
+  @test s ∩ p ≈ p ∩ s ≈ Segment(cart(0, 2), cart(4, 2))
+  @test intersection(s, p) |> type == Overlapping
+  @test intersection(p, s) |> type == Overlapping
+
+  # segment lies on polygon boundary
+  s = Segment(cart(-1, 0), cart(5, 0))
+  @test s ∩ p ≈ p ∩ s ≈ Segment(cart(0, 0), cart(4, 0))
+  @test intersection(s, p) |> type == Overlapping
+  @test intersection(p, s) |> type == Overlapping
```
juliohm: I believe this test should return `EdgeTouching` instead of `Overlapping` as the segment is touching an edge of the polygon:

Image

juliohm: The hardest part is to converge on these definitions of intersections. Let me try to draw something here...

phchavesmaia: Oh, I see! I went with the segment-segment definition, but you are right.

juliohm: What do you think of the following definitions?

Image

I am considering the polygon as the reference geometry, and thinking about configurations that would be useful to identify in applications. From the point of view of someone interested in these intersections, it seems useful to know when the segment is touching the boundary of the polygon in certain ways (`CornerTouching`, `EdgeTouching` or just `Touching`). The `Overlapping` case could also be replaced by the generic `Intersecting`  case as there is nothing special about it.

juliohm: We could formalize these definitions as follows:

- `CornerTouching`: intersect at any point of the segment and a vertex of the polygon
- `EdgeTouching`: intersect at multiple points on a single edge of the polygon
- `Touching`: intersect an one endpoint of the segment
- `Intersecting`: intersect in a more complex way
- `NotIntersecting`: does not intersect in any way

If some of these cases are expensive to compute, we could simplify the list. At the end of the day this will be a compromise between specificity of intersection type and utility in applications.

phchavesmaia: These are great! I just implemented a version based on this typology. I particularly like using `Intersecting` as a catch-all for the more complex cases, since otherwise we'd end up introducing more and more edge cases (n-point touching intersections, for example).

test/intersections.jl
```diff
@@ -424,32 +451,73 @@ end
...
+  @test s ∩ p ≈ p ∩ s ≈ cart(0, 0)
+  @test intersection(s, p) |> type == CornerTouching
+  @test intersection(p, s) |> type == CornerTouching
+  s = Segment(cart(-1, 2), cart(0, 2))
+  p = PolyArea([cart(0, 0), cart(6, 0), cart(6, 6), cart(4, 6), cart(4, 2), cart(2, 2), cart(2, 6), cart(0, 6)])
+  @test s ∩ p ≈ p ∩ s ≈ cart(0, 2)
+  @test intersection(s, p) |> type == Touching
+  @test intersection(p, s) |> type == Touching
+  s = Segment(cart(-1, -1), cart(0, 0))
+  p = PolyArea([cart(0.0, 0.0), cart(1.0, 0.0), cart(1.0, 1.0), cart(0.0, 1.0)])
```
juliohm, line 486: Regarding this test, which has the following figure:

Image

Should this case be considered `CornerTouching` or remain as a generic `Touching` case? The current definition of `Touching` is an intersection at any endpoint of the segment. The current definition of `CornerTouching` is an intersection at any vertex of the polygon + the point must be in the interior of segment. This requirement that the point must be in the interior of the segment is the one that we need to make sure is the most useful in applications. If we remove this requirement, the point automatically becomes `CornerTouching`.

juliohm (commented):
I've moved the tests within the testset to align with the method definitions in the source. I also added a "type stability tests" section following the same pattern of other tests.

After we settle on the `CornerTouching` versus `Touching` definition, I can take a look into the source code again to provide additional suggestions.

test/intersections.jl
phchavesmaia: I think removing the interior-point requirement and considering this `CornerTouching` would make the code a marginally more complicated. That said, I don't have a strong opinion on which definition is more useful in practice, since my geo workflows generally don't care about the specific intersection type (I'm mostly doing geojoins/overlays and calculating areas).

juliohm: > I think removing the interior-point requirement and considering this `CornerTouching` would make the code a marginally more complicated.

If removing the requirement makes the code more complicated, that is probably not the right thing to do 🙂 I will try to look into the algorithm tomorrow.

phchavesmaia: Oh, please don't worry about it. As I've said, it would be only a marginal difference. If it does make the intersection types more useful, I think it will more than worth it.

phchavesmaia: I just added the `CornerTouching` case. If you feel like it would be better to keep the previous one, we can just roll back these latest commits 🙂.

src/intersections/segments.jl
```diff
@@ -203,9 +201,209 @@ function intersection(f, seg::Segment, line::Line)
   end
 end
 
+# The intersection type can be one of six types:
+# 1. overlap with the polygon interior (Intersecting -> Segment or Multi)
+# 2. overlap with a polygon edge (EdgeTouching -> Segment or Multi)
+# 3. intersect at a polygon vertex (CornerTouching -> Point)
+# 4. intersect at an endpoint of the segment (Touching -> Point)
+# 5. the segment is degenerate and is inside the polygon (Touching -> Point)
```
juliohm: I find it a bit strange that we deal with this special case of degenerate segments as a separate case. Shouldn't this be another instance of the generic `Intersecting` condition? The `Touching` condition is more about one endpoint. If the degenerate segment is inside the polygon, then it overlaps with the interior.

src/intersections/segments.jl
```diff
@@ -203,9 +201,204 @@ function intersection(f, seg::Segment, line::Line)
...
+    return @IT NotIntersecting nothing f
+  end
+
+  # check for degenerate segments
+  a, b = vertices(seg)
+  if a ≈ b && a ∈ poly && !any(v -> a ≈ v, vertices(poly))
+    return @IT Touching a f
+  elseif a ≈ b && any(v -> a ≈ v, vertices(poly))
+    return @IT CornerTouching a f
+  end
```
juliohm: What happens if we remove this special treatment for degenerate segments? The code crashes with unexpected division by zero somewhere?

src/intersections/segments.jl
```diff
@@ -203,9 +201,204 @@ function intersection(f, seg::Segment, line::Line)
...
+
+  # segment endpoints as scalars + segment vector
+  λs = [0.0, 1.0]
```
juliohm: Could you please refactor the PR so that the variables are introduced near their paragraph of usage? For example, these two variables `λs` and `v` are defined in the beginning of the function, but only used later after multiple paragraphs. That makes the logic a bit more challenging to digest.

Also, if we could adopt a naming convention for the points derived from segments, polygons, rings, that would be great. For example, I renamed `segbox` to `sbox` and `polybox` to `pbox`. We could then introduce `smin, smax = extrema(sbox)` and `pmin, pmax = extrema(pbox)` in the rest of the algorithm instead of generic names that are harder to track.

src/intersections/segments.jl
```diff
@@ -203,9 +201,204 @@ function intersection(f, seg::Segment, line::Line)
...
+    any(boundaryoverlaps) do (λ₁, λ₂)
+      λ₁ < λ < λ₂ || λ ≈ λ₁ || λ ≈ λ₂
+    end
+  end
+
+  # create segments from the λ values
+  pieces = typeof(seg)[]
+  interiorpieces = typeof(seg)[]
+  boundarypieces = typeof(seg)[]
+  for (λ₁, λ₂) in zip(λs[1:(end - 1)], λs[2:end])
```
juliohm: Here you can make use of `CircularVector(λs)`. That way you can index with `i` and `i+1` for any `i` even if it goes beyond the `length` of the vector.

src/intersections/segments.jl
```diff
@@ -203,9 +201,204 @@ function intersection(f, seg::Segment, line::Line)
...
+      q₁, q₂ = vertices(piece)
+      if p₂ ≈ q₁
+        merged[end] = Segment(p₁, q₂)
+      else
+        push!(merged, piece)
+      end
+    end
+  end
+  merged
+end
```
juliohm, line 321: Thank you for refactoring the logic into an auxiliary function! It helps a lot with the review!

juliohm (changes requested):
@phchavesmaia attached you can find another round of review. Thank you for cleaning up the code for further review.

src/intersections/segments.jl
phchavesmaia: I think it will, but I did not check it. Since this `CornerTouching` is inside the polygon it should go through the branch where it is diving by `v`.

src/intersections/segments.jl
phchavesmaia: I moved `λs` and `v` somewhat closer to where they are used, but they still need to be defined relatively early for efficiency. For instance, storing `v` is not strictly necessary, but since its value can be reused multiple times throughout the loops, I thought it would be better to compute it only once. Similarly, `λs` needs to be initialized before the first ring loop because values are accumulated into it across iterations. 

I also tried to make the naming scheme more consistent throughout the function.

src/intersections/segments.jl
phchavesmaia: I was thinking along the lines of it being a single point intersection, hence the `Touching` category. That said, it could certainly be a `Intersecting` type. Maybe we should rewrite the definitions as:
`# 3. intersect at a polygon vertex (CornerTouching -> Point)`
`# 4. intersect at a single non-vertex point (Touching -> Point)`

Removing the definition `#5 .`

src/intersections/segments.jl
phchavesmaia: You are welcome! I think this function will also help when I get to work with the `Chain`-`Polygon` method.

src/intersections/segments.jl
phchavesmaia: I may be missing something, but I don't think there is any circularity in this loop, so I'm concerned that using `CircularVector` here could introduce unwanted wraparound behavior.

src/intersections/segments.jl
```diff
@@ -215,20 +215,19 @@ function intersection(f, seg::Segment{𝔼{2}}, poly::Polygon{𝔼{2}})
...
-    return @IT CornerTouching a f # Case 3
+    return @IT CornerTouching a f
   end
 
   # segment endpoints as scalars + segment vector
-  segλs = [0.0, 1.0]
-  v = b - a
-  v² = v ⋅ v
+  λs = [0.0, 1.0]
+  v = normalize(b - a)
```
phchavesmaia: I believe this leads to an error, since we would have that `v = (b-a) / sqrt((b-a) ⋅ (b-a))`. I propose changing it to `v = (b-a) / ((b-a) ⋅ (b-a))`.

src/intersections/segments.jl
juliohm: You could still simplify the loop removing the `zip` if you only need `i in 1:n-1`:

```julia
for i in 1:(length(λs)-1)
  λ₁, λ₂ = λs[i], λs[i+1]
end
```

I remember seeing cases where `zip` inhibited type inference.

juliohm: Fixed it.

src/intersections/segments.jl
```diff
@@ -224,64 +223,64 @@ function intersection(f, seg::Segment{𝔼{2}}, poly::Polygon{𝔼{2}})
...
+  sstart = axiscoord(smin)
+  sstop = axiscoord(smax)
 
   # assess intersections
   boundarypoints = typeof(a)[]
   boundaryoverlaps = Tuple{Float64,Float64}[]
 
+  # segment endpoints as scalars + segment vector
+  λs = [0.0, 1.0]
+  v = (b-a) / ((b-a) ⋅ (b-a))
```
juliohm: Could you please explain again why `normalize(b - a)` or `unormalize(b - a)` isn't enough? My understanding is that you just want to get the direction of the segment to perform projections.

src/intersections/segments.jl
```diff
@@ -224,64 +223,64 @@ function intersection(f, seg::Segment{𝔼{2}}, poly::Polygon{𝔼{2}})
...
-    ringbbox = boundingbox(ring)
+    rbox = boundingbox(ring)
 
     # check for obvious no intersection case beforehand
-    intersects(sbox, ringbbox) || continue
+    intersects(sbox, rbox) || continue
 
     edges = collect(segments(ring))
-    boxes = boundingbox.(edges)
+    ebox = boundingbox.(edges)
```
juliohm: Notice that we can avoid intermediate allocations and avoid `collect`:

```julia
eboxes = map(boundingbox, segments(ring))
```

src/intersections/segments.jl
```diff
@@ -203,9 +201,204 @@ function intersection(f, seg::Segment, line::Line)
...
+  sbox = boundingbox(seg)
+  pbox = boundingbox(poly)
+  intersects(sbox, pbox) || return @IT NotIntersecting nothing f
+
+  # check for degenerate segments
+  a, b = vertices(seg)
+  if a ≈ b
+    isvertex = any(≈(a), eachvertex(poly))
+    if a ∈ poly && !isvertex
+      return @IT Touching a f
```
juliohm: Isn't this case the generic `Intersects` case?

src/intersections/segments.jl
```diff
@@ -203,9 +201,204 @@ function intersection(f, seg::Segment, line::Line)
...
+    elseif isvertex
+      return @IT CornerTouching a f
+    end
+  end
+
+  # coordinate with largest bounding box side
+  xside, yside = sides(pbox)
+  x(p) = flat(coords(p)).x
+  y(p) = flat(coords(p)).y
+  coord = xside ≥ yside ? x : y
```
juliohm: Could we simply pick `x` regardless of the largest bounding box side? I am assuming that most polygons will be "balanced" in x and y. If you could share a sample polygon where this is beneficial in practice, that would be super helpful.

If the algorithm works just fine with `x` or `y`, I would simply pick `x` for simplicity.

src/intersections/segments.jl
```diff
@@ -203,9 +201,204 @@ function intersection(f, seg::Segment, line::Line)
...
+          push!(boundaryoverlaps, (λ₁, λ₂))
+        else
+          p = get(I)
+          push!(boundarypoints, p)
+          λ = ustrip((p - a) ⋅ v)
+          push!(λs, λ)
+        end
+      end
+    end
+  end
```
juliohm, line 239: I am understanding that the algorithm is storing relevant intersection points between the segment and the edge. The storage is in terms of the lambda parameters. Later on, you loop over the vector of lambda to construct the sub-segments.

Couldn't we store the `Point`s directly instead of the lambdas?

src/intersections/segments.jl
```diff
@@ -203,9 +201,204 @@ function intersection(f, seg::Segment, line::Line)
...
+
+  # create segments from the λ values
+  pieces = typeof(seg)[]
+  interiorpieces = typeof(seg)[]
+  boundarypieces = typeof(seg)[]
+  for i in 1:(length(λs) - 1)
+    λ₁ = λs[i]
+    λ₂ = λs[i + 1]
+
+    λ₁ ≈ λ₂ && continue
```
juliohm: Isn't this case unreachable? I saw you called `sort!`, `mayberound` and then `unique!` in the vector of lambdas.

juliohm (changes requested):

src/intersections/segments.jl
juliohm: `normalize` will return a unitless static vector and `unormalize` will return a `Vec` with meter units. In both cases, we get a normalized direction to use in projections.

src/intersections/segments.jl
phchavesmaia: I was thinking along the lines of it being a single point intersection, hence the `Touching` category. I was thinking about rewriting the definitions as:

`# 3. intersect at a polygon vertex (CornerTouching -> Point)`
`# 4. intersect at a single non-vertex point (Touching -> Point)`

That said, it could certainly be a `Intersecting` type. As I mentioned before, I don't really know what would be the most useful typology for other projects, so I ultimately defer the choice to you.

src/intersections/segments.jl
phchavesmaia: I investigated this further and tried a few different polygon configurations, but I couldn’t find a case where the pruning provided a real performance benefit. The bounding-box intersection checks are already cheap enough that the additional sorting and materialization costs seem to outweigh any gains from pruning. Furthermore, removing the pruning improves code readability.

If you are ok with it, I will remove it from this new commit, but we can go back to it.

src/intersections/segments.jl
phchavesmaia: In the current version, each `λ ∈ [0, 1]` is a scaling factor along the segment `s`, such that `s(λ)` is either an endpoint of `s` or a point where `s` intersects the boundary of the polygon `p`. In particular, observe that `s(0)=a` and `s(1)=b`. 

If we use `v = normalize(b - a)`, then `norm(v) = 1`, meaning that `(q - a) ⋅ v` represents the distance from `a` to `q` along the segment, rather than a scaling factor in `[0, 1]`. Consequently, these values should not be passed directly to `s(λ)`.

src/intersections/segments.jl
phchavesmaia: Yes, it should be unreachable after the `mayberound` operation. We should remove the `mayberound` pass, since this one is more important.

src/intersections/segments.jl
phchavesmaia: We can! I’m currently playing with a version that does exactly this, but I’ve found that it loses a little efficiency. My guess is that reducing the problem to scalars simply gives Julia less work to do than carrying the geometries around. That said, I think the code gains a lot of readability. I will post this version here so you can compare.

src/intersections/segments.jl
phchavesmaia: We can use `eboxes = map(boundingbox, segments(rings))`, though I don't really see how to avoid materializing `edges` in a vector since we will loop over a subsection of them (and `segments` doesn't support indexing). I think I am missing something....

src/intersections/segments.jl
juliohm: The new version is more readable for sure!

src/intersections/segments.jl
juliohm: The new version addressed this issue 👍🏽

src/intersections/segments.jl
```diff
@@ -203,9 +201,128 @@ function intersection(f, seg::Segment, line::Line)
...
+          push!(splitpoints, vertices(s)...)
+          push!(boundaryoverlaps, s)
+        else
+          p = get(I)
+          push!(splitpoints, p)
+          push!(boundarypoints, p)
+        end
+      end
+    end
+  end
```
juliohm, line 239: @phchavesmaia could you please describe the suggested algorithm with simple words? My understanding is that you are storing the intersections along the segment (depicted as red dots below) and an additional flag to identify sub-segments:

Image

Later on, you compute the mid points (depicted as black crosses) to check if the sub-segments are inside or outside the polygon. How do you use the information stored in the auxiliary variables to merge the sub-segments and points into the final intersection result?

Could we further simplify the algorithm so that we only need to store the `splitpoints`? Does it take polygon holes (triangle in the figure) into account already?

phchavesmaia: Of course!

The idea is to construct the segment-polygon intersection from the intersections between the segment and each polygon edge. Point intersections are stored in `boundarypoints`, overlapping edges in `boundaryoverlaps`, and all relevant intersection points, together with the segment endpoints, in `splitpoints`.

The `splitpoints` are then sorted along the segment, and consecutive points define candidate pieces. I check the midpoint of each piece to determine whether it lies inside the polygon or on one of the boundary overlaps. This also handles polygons with holes, since they generate `splitpoints` (points `M` and `N`, in your example) denoting a sub-segment outside of the polygon (point `Q`, in your example). The resulting pieces are stored separately as `interiorpieces` and `boundarypieces`, while `pieces` contains both so that the complete intersection geometry can be returned. The resulting geometries are merged with the `_mergepieces` and `piecegeometry` auxiliary functions.

If there are no segment pieces, `boundarypoints` is used to classify isolated intersections as `Touching` or `CornerTouching`.

I think `boundarypoints` could be removed by keeping the segment endpoints separate from `splitpoints` and adding them only when constructing the candidate pieces. That would let `splitpoints` represent only actual boundary intersections and serve the role currently played by `boundarypoints`. I don't expect a significant performance improvement, though, since maintaining `boundarypoints` is quite cheap. Removing the other auxiliary collections would require a more substantial redesign of the classification logic.

juliohm: Thank you, but I am curious about the specific algorithm used here:

> The `splitpoints` are then sorted along the segment, and consecutive points define candidate pieces. I check the midpoint of each piece to determine whether it lies inside the polygon or on one of the boundary overlaps.

After you compute the mid point you perform specific checks. How do you use the result of these checks to decide the type of sub-segment? I am looking for something high-level like:

```julia
if midpoint_is_inside_polygon
  # do this
elseif midpoint_is_on_boundary_of_polygon
  # do that
...
```

We have `\in`, `sideof` and other utilities that could be handy if the algorithm is clearly stated in English.

My understanding is that there is a classification algorithm followed by a merge algorithm. It would be nice to learn more about them to provide more feedback.

phchavesmaia: The part you are looking for is this (lines 262-268):
```julia
if any(overlap -> midpoint ∈ overlap, boundaryoverlaps) # check if the midpoint lies in a polygon boundary
      push!(pieces, piece)
      push!(boundarypieces, piece)
elseif midpoint ∈ poly # check if the midpoint is lies inside the polygon
      push!(pieces, piece)
      push!(interiorpieces, piece)
```
I think using `boundaryoverlaps` instead of `sideof`, for example, should be more efficient because it only contains the boundary segments that were already identified as relevant to `s`. In addition, `any` short-circuits, so the search stops as soon as a matching overlap is found.

The intersection type is determined only at the end, based on how the auxiliary variables were populated during the previous steps.

juliohm: Thank you @phchavesmaia ! Do you think it would make sense to build `pieces`, `boundarypieces` and `interiorpieces` directly, without allocating the intermediate `splitpoints`, `boundarypoints` and `boundaryoverlaps`? My understanding is that the algorithm branches on the intersection type between `seg` and `edge` and then it could push the `pieces` using additional checks like the mid point check, etc. If the code remains readable, that would save us some allocations. It can potentially simplify the implementation too.

phchavesmaia: You’re welcome, @juliohm! I don't think `pieces`, `boundarypieces`, and `interiorpieces` can always be built directly while iterating over the polygon edges. A single edge intersection usually gives only one boundary event, and we still need to know the next event along `s` before we can determine the corresponding sub-segment. This is especially relevant for concave polygons and polygons with holes, where there may be several crossings.

I experimented with a few alternatives, including storing the intersection type with each point and using an all-encompassing `events` vector that also stores the actual overlapping intersection geometries. However, these approaches became more convoluted, as they required additional loops to organize the events without reducing intermediate allocations (and, in some cases, actually increasing them).

I therefore believe that some form of “collect all events → sort them along `s` → classify the intervals between consecutive events” is still necessary. That said, I may be missing a simpler alternative.

juliohm: Got it. What about replacing `boundarypoints` and `boundaryoverlaps` by simple boolean vectors `docross` and `dooverlap` of the same length of `splitpoints`? I've checked the usage of these variables in the following paragraphs and they are basically flags that you use to merge the pieces together.

In the first loop you could fill `splitpoints` and flag each point as either a point that `docross`es an edge or a point that `dooverlap`s with an edge. You could also introduce a third flag based on the mid point if necessary. My intuition is that this could potentially simplify the code, besides reducing allocations with multiple copies of points in 3 vectors. The checks later could be as simple as:

```julia
if dooverlap[i] && dooverlap[i+1]
  p1, p2 = splitpoints[i:i+1]
  push!(pieces, Segment(p1, p2))
elseif docross[i]
  ...
end
```

Does it make sense?

phchavesmaia: Yes, it does! I think I arrived at something along those lines, and it provided a measurable efficiency improvement. I’ll post it here so you can take a look. The main differences from your suggestion are:

1. I store the information in a tuple, since that turned out to be the most efficient way to keep the point and flags aligned.
2. `dooverlap` is not a Boolean, because I need to distinguish the beginning and end of an overlap (otherwise, two separate overlaps could be incorrectly connected).

Unfortunately, I think the code became somewhat less readable. I also tried using a `NamedTuple`, since that would make the fields much clearer, but for reasons I still don’t fully understand, it eliminated the efficiency gains. In fact, that version was slower and allocated more memory than the current implementation.

src/intersections/segments.jl
```diff
@@ -203,9 +201,136 @@ function intersection(f, seg::Segment, line::Line)
...
+# 4. intersect at a single non-vertex point (Touching -> Point)
+# 5. do not overlap nor intersect (NotIntersecting -> Nothing)
+function intersection(f, seg::Segment{𝔼{2}}, poly::Polygon{𝔼{2}})
+  sbox = boundingbox(seg)
+  pbox = boundingbox(poly)
+  intersects(sbox, pbox) || return @IT NotIntersecting nothing f
+
+  # collect boundary events
+  a, b = vertices(seg)
+  events = Tuple{typeof(a),Bool,Int8}[(a, false, Int8(0)), (b, false, Int8(0))] # event = (point, docross, overlap)
```
```suggestion
  events = [(a, false, Int8(0)), (b, false, Int8(0))] # event = (point, docross, overlap)
```

Initialization is enough for type inference.

src/intersections/segments.jl
```diff
@@ -203,9 +201,136 @@ function intersection(f, seg::Segment, line::Line)
...
+  λ(p) = (p - a) ⋅ (b - a)
+  for ring in rings(poly)
+    intersects(sbox, boundingbox(ring)) || continue
+    for edge in segments(ring)
+      intersects(sbox, boundingbox(edge)) || continue
+      I = intersection(seg, edge)
+      itype = type(I)
+      if itype == NotIntersecting
```
juliohm: I am noticing that you are undoing some of the changes I've made to this PR. Could you please double check your are pulling the changes locally before working on new commits?

src/intersections/segments.jl
```diff
@@ -203,9 +201,136 @@ function intersection(f, seg::Segment, line::Line)
...
+      piece = Segment(eⱼ[1], eᵢ[1])
+      if overlapdepth > 0
+        push!(pieces, piece)
+        push!(boundarypieces, piece)
+      else
+        midpoint = center(piece)
+        if midpoint ∈ poly
+          push!(pieces, piece)
+          push!(interiorpieces, piece)
+        end
```
juliohm: I ask the same question here about using flags instead of duplicating the data. Instead of copying the `piece` to `interiorpieces` and `boundarypieces`, I think it should be enough to flag the piece using a boolean vector.

juliohm (changes requested):
Thank you @phchavesmaia , attached is another round of review.

src/intersections/segments.jl
juliohm: Also, is there a reason why we need `Int8`? Do we gain any considerable space compared to a simple `Int`?

phchavesmaia: Yes, we do save some memory (around 6% in my test file).

src/intersections/segments.jl
phchavesmaia: I will double check it.

src/intersections/segments.jl
phchavesmaia: Yes, it is possible, and we do save some allocations! The flag construction is not entirely straightforward, so I had to replace `_mergepieces`  with `_pushpiece!`, a function that merges when necessary, pushes the resulting piece, and tracks whether any section of the merged piece is interior. I’ll upload a version incorporating your suggestion.

src/intersections/segments.jl
```diff
@@ -203,9 +201,125 @@ function intersection(f, seg::Segment, line::Line)
...
+function intersection(f, seg::Segment{𝔼{2}}, poly::Polygon{𝔼{2}})
+  # early exit if bounding boxes do not intersect
+  sbox = boundingbox(seg)
+  pbox = boundingbox(poly)
+  intersects(sbox, pbox) || return @IT NotIntersecting nothing f
+
+  # collect and classify boundary events
+  a, b = vertices(seg)
+  λ(p) = (p - a) ⋅ (b - a)
+  events = [(a, false, 0), (b, false, 0)] # event = (point, crosses, overlaps)
```
juliohm: @phchavesmaia wouldn't it make sense to split this complex vector of tuples into 3 simple vectors? If the split compromises performance, I think we can still improve readability by replacing tuples with named tuples. Each event could be something like `(point=a, crosses=false,  noverlaps=0)`. That way one could access information in the event later without figuring out indices.

src/intersections/segments.jl
```diff
@@ -203,9 +201,125 @@ function intersection(f, seg::Segment, line::Line)
...
+    end
+  end
+
+  # sort events along the segment
+  sort!(events, by=e -> λ(e[1]))
+
+  # collect unique intersection pieces
+  j = 1
+  overlapdepth = 0
+  pieces = Tuple{typeof(seg),Bool}[] # pieces = (piece, isinterior)
```
juliohm: Similar question here. Isn't more idiomatic to split this into two simple vectors, one containing sements and another containing the `isinterior` flag?

src/intersections/segments.jl
```diff
@@ -203,9 +201,125 @@ function intersection(f, seg::Segment, line::Line)
...
+  for i in 2:length(events)
+    eᵢ = events[i]
+    eⱼ = events[j]
+    if eᵢ[1] ≈ eⱼ[1]
+      # merge coincident events
+      events[j] = (eⱼ[1], eⱼ[2] || eᵢ[2], eⱼ[3] + eᵢ[3])
```
juliohm: With named tuples we could rewrite this to

```julia
(ej.point, ej.crosses || ei.crosses, ej.noverlaps + ei.noverlaps)
```

src/intersections/segments.jl
```diff
@@ -203,9 +201,125 @@ function intersection(f, seg::Segment, line::Line)
...
+    return @IT Intersecting Multi(points) f
+  else
+    return @IT NotIntersecting nothing f
+  end
+end
+
+# merge the continuous segments of a segment list (`pieces`) into a single geometry and push then, indicating whether they are an interior segment
+function _pushpiece!(pieces, piece, isinterior)
+  if !isempty(pieces)
+    previous, previous_isinterior = last(pieces)
```
juliohm: We don't use underscore in our code style

src/intersections/segments.jl
```diff
@@ -203,9 +201,125 @@ function intersection(f, seg::Segment, line::Line)
...
+    p₁, p₂ = vertices(previous)
+    q₁, q₂ = vertices(piece)
+    if p₂ ≈ q₁
+      mergedpiece = Segment(p₁, q₂)
+      pieces[end] = (mergedpiece, previous_isinterior || isinterior)
+      return nothing
```
juliohm: Nor explicit `return`, unless it is necessary for premature termination of the function. We use `return` in the `@IT` calls because we are explicitly terminating the function in different branches.

src/intersections/segments.jl
phchavesmaia: The problem with splitting this vector is the indexing, since the sorting has to remain consistent across all parallel vectors. Re-indexing them introduces enough overhead that the previous version, which stored the geometries directly, becomes faster.

I also tried using named tuples (see [this reply](https://github.com/JuliaGeometry/Meshes.jl/pull/1397/changes#r3772058678)), since I agree that readability is somewhat compromised here. However, named tuples are substantially slower than regular tuples, to the point that storing the geometries, as in commit [769b0ca](https://github.com/JuliaGeometry/Meshes.jl/commit/769b0ca5ec17292b3842372a2fadb5b9d309b9a4), becomes the faster alternative.

To be clear, named tuples save memory compared with the geometry-based version, however, they are about 15% slower. Compared with regular indexed tuples, the memory usage is essentially identical, but named tuples are about 20% slower.

As I see it, there are three possible directions from here, given that the `pieces` tuple issue is already solved by using parallel vectors, which improves both readability and efficiency:

1. Use named tuples for the events, improving readability at the cost of runtime performance due to the additional allocations (~20% slower in my tests).
2. Keep regular indexed tuples, sacrificing some readability in exchange for better performance.
3. Look for an alternative based on parallel vectors that preserves both runtime and memory efficiency. The current alternative I was working with takes double the time of the regular indexed tuple approach.

That said, perhaps because I have been reading this function quite a lot, I don't find the current tuple-based version particularly difficult to follow. The tuple ordering is already documented in the comment, and we could reinforce that ordering in a few additional comments where the fields are accessed if we want to make the code easier to read.

What do you think?

src/intersections/segments.jl
phchavesmaia: Got it.

src/intersections/segments.jl
phchavesmaia: Got it.

src/intersections/segments.jl
phchavesmaia: Why not use `Int8`? I think it would be better to keep it, as it reduces approximately 10% of the memory usage without compromising readability.

src/intersections/segments.jl
phchavesmaia: I ran some tests and since we do not re-index this vector we can actually increase memory efficiency by splitting the tuple into parallel vectors. It follows in the review a version with this implementation.

src/intersections/segments.jl
phchavesmaia: See the other reply. When we agree on the direction to take, I will change this accordingly.

src/intersections/segments.jl
juliohm: Could you please share the benchmark code you are using so that I can test it locally too?

juliohm: Are we assuming that the upper bound on the number of overlaps is 127?

```julia
julia> typemax(Int8)
127
```

In my next round of review I will try to digest the logic behind these integers. I am imagining large polygons (e.g., entire continents) with a lot of "zig-zag" at the borders. If you place a segment in such a way that it crosses more than 127 segments, then the implementation with `Int8` might fail?

phchavesmaia: I may be understanding this wrong, but what I get from `Int8` is that it allows for integer values that range between -128 and +128. Since we only use -1, 0, and 1 everything should be fine independently from the number of crosses.

juliohm:
Ok, then it is my understanding of the code that is wrong. I didn't track
the increments to know the only possible values were -1, 0, +1. I'll
document it better in my next review.

Em ter., 18 de ago. de 2026, 12:47, Pedro H. Chaves Maia <
***@***.***> escreveu:

src/intersections/segments.jl
phchavesmaia: I was initially benchmarking only a complex polygon intersection, which is admittedly a worst-case scenario for named tuples in this comparison. I therefore expanded the benchmarks to cover different intersection types and simpler polygons.

I also noticed substantial variation across repeated runs, so I added bootstrapping to estimate a confidence interval for the median runtime ratio. The reproducible benchmark is below:

```julia
using Statistics, BenchmarkTools, Meshes, Printf

# Benchmark cases
square = PolyArea([
  Point(-1.0, -1.0),
  Point(1.0, -1.0),
  Point(1.0, 1.0),
  Point(-1.0, 1.0),
])

complexpoly = PolyArea([
  Point(-10.0, -3.0),
  Point(-9.0, 2.0),
  Point(-8.0, -2.0),
  Point(-7.0, 3.0),
  Point(-6.0, -3.0),
  Point(-5.0, 2.0),
  Point(-4.0, 0.0),
  Point(-3.0, 0.0), # boundary overlap
  Point(-2.0, -3.0),
  Point(-1.0, 3.0),
  Point(0.0, -2.0),
  Point(1.0, 3.0),
  Point(2.0, -3.0),
  Point(3.0, 0.0), # vertex on segment
  Point(4.0, 2.0),
  Point(5.0, -3.0),
  Point(6.0, 3.0),
  Point(7.0, -2.0),
  Point(8.0, 2.0),
  Point(9.0, -3.0),
  Point(10.0, -3.0),
  Point(10.0, 5.0),
  Point(-10.0, 5.0),
])

cases = [
  (
    name="bbox disjoint",
    seg=Segment(Point(-5.0, 0.0), Point(-4.0, 0.0)),
    poly=square,
  ),
  (
    name="endpoint touch",
    seg=Segment(Point(-2.0, 0.0), Point(-1.0, 0.0)),
    poly=square,
  ),
  (
    name="corner touch",
    seg=Segment(Point(-2.0, -2.0), Point(-1.0, -1.0)),
    poly=square,
  ),
  (
    name="edge overlap",
    seg=Segment(Point(-2.0, -1.0), Point(2.0, -1.0)),
    poly=square,
  ),
  (
    name="simple crossing",
    seg=Segment(Point(-2.0, 0.0), Point(2.0, 0.0)),
    poly=square,
  ),
  (
    name="fully interior",
    seg=Segment(Point(-0.5, 0.0), Point(0.5, 0.0)),
    poly=square,
  ),
  (
    name="complex overlap",
    seg=Segment(Point(-11.0, 0.0), Point(11.0, 0.0)),
    poly=complexpoly,
  ),
]

# Benchmark runner
function benchmark_all(cases; rounds=30, nboot=10_000, level=0.95)
  results = map(cases) do case
    seg = case.seg
    poly = case.poly
    ratios = Vector{Float64}(undef, rounds)
    regular_times = similar(ratios)
    named_times = similar(ratios)
    regular_mem = 0
    named_mem = 0
    regular_allocs = 0
    named_allocs = 0
    for i in 1:rounds
      b_regular = @benchmarkable intersection($identity, $seg, $poly)
      b_named = @benchmarkable Meshes.intersection_named($identity, $seg, $poly)

      # Alternate execution order to reduce systematic bias
      if iseven(i)
        t_named = run(b_named)
        t_regular = run(b_regular)
      else
        t_regular = run(b_regular)
        t_named = run(b_named)
      end
      m_regular = median(t_regular)
      m_named = median(t_named)
      regular_times[i] = m_regular.time
      named_times[i] = m_named.time
      ratios[i] = m_named.time / m_regular.time

      # Allocation counts should be stable, so save them once
      if i == 1
        regular_mem = m_regular.memory
        named_mem = m_named.memory
        regular_allocs = m_regular.allocs
        named_allocs = m_named.allocs
      end
    end

    # Bootstrap confidence interval for the median runtime ratio
    n = length(ratios)
    boot = Vector{Float64}(undef, nboot)
    for i in 1:nboot
      sample = ratios[rand(1:n, n)]
      boot[i] = median(sample)
    end
    α = (1 - level) / 2
    ratio_lo = quantile(boot, α)
    ratio_hi = quantile(boot, 1 - α)
    (
      case=case.name,
      regular_ns=median(regular_times),
      named_ns=median(named_times), time_ratio=median(ratios),
      ratio_lo=ratio_lo,
      ratio_hi=ratio_hi, regular_mem=regular_mem,
      named_mem=named_mem, regular_allocs=regular_allocs,
      named_allocs=named_allocs,
    )
  end

  println()
  @printf(
    "%-18s %11s %11s %9s %20s %10s %10s %9s %8s %8s\n",
    "case",
    "regular μs",
    "named μs",
    "Δ time",
    "     95% CI      ",
    "regular KB",
    "named KB",
    "Δ memory",
    "alloc R",
    "alloc N",
  )
  println("-"^130)

  for r in results
    timepct = 100 * (r.time_ratio - 1)
    timelo = 100 * (r.ratio_lo - 1)
    timehi = 100 * (r.ratio_hi - 1)
    memratio = r.named_mem / max(r.regular_mem, 1)
    mempct = 100 * (memratio - 1)
    @printf(
      "%-18s %11.3f %11.3f %+8.1f%% [%+6.1f%%, %+6.1f%%] %10.2f %10.2f %+8.1f%% %8d %8d\n",
      r.case,
      r.regular_ns / 1_000,
      r.named_ns / 1_000,
      timepct,
      timelo,
      timehi,
      r.regular_mem / 1024,
      r.named_mem / 1024,
      mempct,
      r.regular_allocs,
      r.named_allocs,
    )
  end
  results
end

results = benchmark_all(cases; rounds=100);
```
Where `Meshes.intersection_named` is the named-tuple implementation. I simply replaced the regular tuples with named tuples and made the corresponding adjustments throughout the function. On my machine, the results were:
```
case                regular μs    named μs    Δ time         95% CI       regular KB   named KB  Δ memory  alloc R  alloc N
----------------------------------------------------------------------------------------------------------------------------------
bbox disjoint            0.645       0.645     +0.7% [  +0.6%,   +0.8%]       0.64       0.64     +0.0%       26       26
endpoint touch           1.226       1.336     +8.9% [  +8.6%,   +9.1%]       1.52       1.55     +2.1%       51       52
corner touch             1.357       1.570    +17.3% [ +16.7%,  +17.8%]       1.59       1.66     +3.9%       57       59
edge overlap             1.086       1.333    +19.7% [ +19.3%,  +20.1%]       1.45       1.52     +4.3%       47       49
simple crossing          0.930       1.131    +21.6% [ +21.5%,  +21.8%]       1.25       1.31     +5.0%       38       40
fully interior           0.664       0.670     +0.9% [  +0.8%,   +1.1%]       0.84       0.84     +0.0%       29       29
complex overlap          8.222      10.103    +22.7% [ +22.6%,  +22.9%]       7.91       8.53     +7.9%      260      280
```
I know the runtimes are already quite small, but I think this is still an important discussion, especially since this method will be used to implement the `Rope`-`Polygon` intersection. I’m particularly worried in how these differences might scale in more realistic workloads, such as computing intersections between rivers and municipalities.

juliohm: Really appreciate the careful analysis of the implementation! I believe the solution with tuples is fine in this new version, which is very easy to read now.

juliohm: Answered in the global thread by accident. I understood the semantic of this variable now.

src/intersections/segments.jl
```diff
@@ -203,9 +201,123 @@ function intersection(f, seg::Segment, line::Line)
...
+function intersection(f, seg::Segment{𝔼{2}}, poly::Polygon{𝔼{2}})
+  # early exit if bounding boxes do not intersect
+  sbox = boundingbox(seg)
+  pbox = boundingbox(poly)
+  intersects(sbox, pbox) || return @IT NotIntersecting nothing f
+
+  # collect and classify boundary events, which are of the
+  # form (point, iscrossing, enteringmode). The enteringmode
+  # is +1 if the segment is entering the polygon, -1 if it is
+  # leaving the polygon, and 0 if it is neither (e.g. crossing)
```
juliohm: I've added this comment based on the code. Could you please confirm I understood the algorithm correctly?

src/intersections/segments.jl
```diff
@@ -203,9 +201,123 @@ function intersection(f, seg::Segment, line::Line)
...
+  resize!(events, i)
+
+  # glue pieces together into a single geometry
+  glue(pieces) = length(pieces) == 1 ? only(pieces) : Multi(pieces)
+
+  # number of crossing events
+  ncrosses = count(e -> e[2], events)
+
+  # classify intersection
+  if any(isinterior)
```
juliohm: It seems that the variable `isinterior` is only used here with the predicate `any`. What about defining a variable `anyinterior::Bool` that is updated within the "collect intersection pieces" loop instead? That could save us some allocations and could simplify the `_pushpieces!` logic a bit. I am thinking of something like:

```julia
anyinterior =  false
for j in 2:length(events)
  ...
  if center(piece)  \in poly
    anyinterior = true
   ...
  end
end
```

src/intersections/segments.jl
```diff
@@ -203,9 +201,123 @@ function intersection(f, seg::Segment, line::Line)
...
+  else
+    return @IT NotIntersecting nothing f
+  end
+end
+
+# merge the continuous segments of a segment list (`pieces`) into
+# a single geometry and push then, indicating whether they are an
+# interior segment in the `isinterior` vector.
+function _pushpiece!(pieces, isinterior, piece, isint)
+  if !isempty(pieces)
```
juliohm: I understand that this check is performed because `pieces` is empty in the first iteration of the loop. Perhaps we could initialize `pieces` before the loop in such a way that this check is not needed?

juliohm (commented):
This new version is very nice! Super readable! 👏🏽 💯

src/intersections/segments.jl
phchavesmaia: This is correct!

src/intersections/segments.jl
phchavesmaia: Oh, sure! That's a straight up improvement!

src/intersections/segments.jl
phchavesmaia: Exactly. We could do it, however, I don't think it's worth it. Notice that there are checks in the function that depend on whether `pieces` is populated, so initializing it with a candidate segment would require refactoring the classification scheme. Moreover, besides merging pieces, this loop also deduplicates events, which prevents us from generating degenerate segments.

In the end, removing the `!isempty(pieces)` check would likely introduce additional conditionals in the loop. I don't think that would improve efficiency by any significant amount, and it would most likely hurt readability.

src/intersections/segments.jl
```diff
@@ -203,9 +201,121 @@ function intersection(f, seg::Segment, line::Line)
...
+# a single geometry and push then.
+function _pushpiece!(pieces, piece)
+  if !isempty(pieces)
+    prev = last(pieces)
+    p₁, p₂ = vertices(prev)
+    q₁, q₂ = vertices(piece)
+    if p₂ ≈ q₁
+      pieces[end] = Segment(p₁, q₂)
+      return
+    end
```
juliohm: I understand that you are trying to "merge" two consecutive segments/pieces when the end-points coincide, and that this might happen if and only if the polygon has two consecutive co-linear edges. Is that right? Assuming I got it right, is that configuration common enough to deserve special treatment? Why not return the two pieces inside a `Multi` geometry?

If we remove this if statement, I believe we can get rid of `_pushpiece!` and use a simple `push!`?

phchavesmaia: Image

No, that is not the only case in which this can happen. I present a different case in the figure above, where the red segment GE will be "merged" together. I think this kind of intersection can be somewhat common in some workflows.

That said, whether these segments must be joined by `_pushpiece!` is a separate question. At one point, `_pushpiece!` helped with the numerical robustness of the results, but that is no longer the case because we are already deduplicating the `events` points in the loop.

In summary, it can be safely removed if we are willing to report the resulting geometries as `Multi`, as you appropriately pointed out. I think this would be a good change!

juliohm: Interesting! I think we can return a `Multi`-segment without major issues.  Alternatively, the `glue` function could be responsible for checking this co-linearity instead of adding this check in `_pushpiece!`. Intuitively, this check is more related to the "glue" step than it is related to the "push" step, at least that is how I see it.  What do you think?

phchavesmaia: I agree with you. I’m changing it so that we only `push!` the pieces. If this ever becomes an issue, there is already a clear path forward.

src/intersections/segments.jl
```diff
@@ -203,9 +201,106 @@ function intersection(f, seg::Segment, line::Line)
...
+  i = 1
+  depth = 0
+  inpoly = false
+  pieces = typeof(seg)[]
+  for j in 2:length(events)
+    eᵢ = events[i]
+    eⱼ = events[j]
+    if eᵢ[1] ≈ eⱼ[1]
+      # merge coincident events
+      events[i] = (eᵢ[1], eᵢ[2] || eⱼ[2], eᵢ[3] + eⱼ[3])
```
juliohm, line 255: Hi @phchavesmaia! Could you please confirm that `||` is the correct operator? Why not `&&`? We have two events that are considered equal and the algorithm is saying that the "iscross" entry of the resulting event will be `true` whenever one of the "iscross" is `true`. Is it because "iscross" is a stronger condition than "overlap" by design?

src/intersections/segments.jl
```diff
@@ -203,9 +201,106 @@ function intersection(f, seg::Segment, line::Line)
...
+  elseif ncrosses == 1
+    i = findfirst(e -> e[2], events)
+    p = events[i][1]
+    isendpoint = p ≈ a || p ≈ b
+    isvertex = any(v -> p ≈ v, vertices(poly))
+    if isendpoint && !isvertex
+      return @IT Touching p f
+    else
+      return @IT CornerTouching p f
+    end
```
juliohm: Could this logic be refactored to avoid the premature computation of `isendpoint`? Something like

```julia
if any(≈(p), eachvertex(poly))
  return @IT CornerTouching p f
elseif any(≈(p), eachvertex(seg))
  return @IT Touching p f
end
```

I also replaced `vertices(poly)` by `eachvertex(poly)` to avoid allocations for any subtype of `Polygon`. We need to check if that compromises performance.

juliohm: You could also use the short-circuit of `&&` to avoid the premature computation of  ìsvertex` or `isendpoint` if the proposed change above modifies the expected behavior.

src/intersections/segments.jl
phchavesmaia: I still think the correct operator should be `||`, but I may be missing something. Think about the figure I shared in another comment. Point D could be a case where, due to floating-point precision errors in the corner edge segments, you have to decide whether `iscross` should be `true` or `false`.

Conceptually, I think it should be `true`, because the polygon boundary is indeed being crossed. In this example, even if `iscross` is `true`, it would not change the classification because `pieces` is populated, and I suspect a similar situation would occur in most, if not all, cases.

To sum up, I think there is a definitional argument for keeping the current behavior.

src/intersections/segments.jl
phchavesmaia: I think changing the intersection logic would be the best alternative, as you suggested in the first comment. In fact, I think `isendpoint` is redundant. I will implement it right now.

phchavesmaia: Oh, and regarding the use of either `eachvertex` or `vertex`, I went with the former since you have reason to believe it will be more efficient. I think any performance difference will be marginal, but given that we have an argument in its favor, I don't see any reason not to go with it.

juliohm (approved):
One of the most elegant code contributions I reviewed this year! Thank you! 🙂

Issue #535, opened 2022-12-22
Smoothing option in viz function


stla:
Hello,

I exposed my problem [here](https://discourse.julialang.org/t/meshviz-makie-is-smoothing-my-colors/91993). 

In the given answer, the author uses the `Makie.mesh!` function to plot the mesh, with the option `interpolate=false`. Is it possible to have such an option with **MeshViz.jl**? I took a look at the code and it seems you use `Makie.meshscatter!`, which does not have this option.

juliohm:
We provide two coloring options in MeshViz.jl. If you pass a vector of
colors that has the same length of the vector of vertices you get
interpolation. If on the other hand the vector of colors has a length that
is equal to the number of elements you get sharp boundaries. Can you give
it a try?

Em qui., 22 de dez. de 2022 07:28, stla ***@***.***> escreveu:

stla:
Yes, it works. Thanks.

juliohm:
Can you also share the update on Discourse so that others learn about the
solution?

Em qui., 22 de dez. de 2022 07:49, stla ***@***.***> escreveu:

stla:
Actually I still have the same problem: if I provide a color per face, the mesh is not smoothed. So I can't do something like this for example, with sharp colors but smoothed: https://youtu.be/vLkNQh3Sv8w

juliohm:
What exactly is smoothed? I am trying to understand what exactly is the difference between the current output and the desired output. On Zulip we have a thread where I asked if you could share some screenshots of the same geometry with the same angle to compare the different results.

stla:
On Zulip my two screenshots have the same angle.

By "not smoothed" I mean that we can see the faces of the mesh when we plot it, as if the normals were not taken into account.

Here is an example with **rgl**. The sphere at right has no normals, it is not smoothed. 
![Capture du 2022-12-26 19-43-07](https://user-images.githubusercontent.com/4466543/209576047-d924fed3-ab6a-4097-8c14-43dd5110c7f3.png)

juliohm:
Got it. Do you have a PR in mind here that could give this additional flexibility? I don't know how these normals play a role in OpenGL, but Makie probably has an option to use them or not. We just need to forward this option to the low-level Makie function in our recipe.

stla:
[...]

PR #1403, opened 2026-08-02
Generalize normal calculation and add normals option to viz

barche:
Hi,

This is an attempt to address issue #535 in two steps:

1. Generalize `normal` for any `Ngon` using Newell's algorithm (end of https://people.eecs.berkeley.edu/~ug/slide/pipeline/assignments/backfacecull.shtml, see also https://wikis.khronos.org/opengl/Calculating_a_Surface_Normal#Newell's_Method)
2. Add an option to pass normals to `viz`, either per-face or per-vertex

### DIFF
--- a/docs/src/visualization.md
+++ b/docs/src/visualization.md
@@ -35,3 +35,25 @@ grid = CartesianGrid(10, 10, 10)
 
 viz(grid, showsegments = true, segmentcolor = :teal)
 ```
+
+### Normals
+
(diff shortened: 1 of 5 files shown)


juliohm:
That is awesome! 🎉 Could you please split this PR into two PRs? The first one could introduce the new `normal` method with tests, and the second one could consider the `viz` option.

PR #1404, opened 2026-08-02
Normal calculation for any Ngon

barche:
Generalize `normal` for any Ngon using Newell's algorithm (end of https://people.eecs.berkeley.edu/~ug/slide/pipeline/assignments/backfacecull.shtml, see also https://wikis.khronos.org/opengl/Calculating_a_Surface_Normal#Newell's_Method)

This is the first step to address issue https://github.com/JuliaGeometry/Meshes.jl/issues/535, replacing PR #1403

### DIFF
--- a/src/geometries/polytopes/ngon.jl
+++ b/src/geometries/polytopes/ngon.jl
@@ -65,6 +65,23 @@ angles(ngon::Ngon) = angles(boundary(ngon))
 
 innerangles(ngon::Ngon) = innerangles(boundary(ngon))
 
+# Normal calculation using Newell's algorithm
+# https://people.eecs.berkeley.edu/~ug/slide/pipeline/assignments/backfacecull.shtml
+function normal(ngon::Ngon)
+  assertion(embeddim(ngon) == 3, "ngon must be 3-dimensional")
+  v = CircularVector(ngon.vertices)
+  nx = ny = nz = 0.0
+  for i in eachindex(v)
+    x₁, y₁, z₁ = ustrip.(to(v[i + 1]))
+    x₂, y₂, z₂ = ustrip.(to(v[i]))
+    nx += (z₂ + z₁) * (y₂ - y₁)
+    ny += (x₂ + x₁) * (z₂ - z₁)
+    nz += (y₂ + y₁) * (x₂ - x₁)
+  end
+  lt = unit(lentype(ngon))
+  unormalize(Vec(nx*lt, ny*lt, nz*lt))
+end
+
 # ----------
 # TRIANGLES
 # ----------
--- a/src/geometries/polytopes/ngon.jl
+++ b/src/geometries/polytopes/ngon.jl
@@ -65,21 +65,7 @@ angles(ngon::Ngon) = angles(boundary(ngon))
 
 innerangles(ngon::Ngon) = innerangles(boundary(ngon))
 
-# normal of n-gon using Newell's algorithm
-# https://people.eecs.berkeley.edu/~ug/slide/pipeline/assignments/backfacecull.shtml
-function normal(ngon::Ngon)
-  assertion(embeddim(ngon) == 3, "ngon must be embedded in ℝ³")
-  v = CircularVector(ngon.vertices)
-  n = sum(eachindex(v)) do i
-    x₁, y₁, z₁ = ustrip.(to(v[i]))
-    x₂, y₂, z₂ = ustrip.(to(v[i + 1]))
-    nx = (z₁ + z₂) * (y₁ - y₂)
-    ny = (x₁ + x₂) * (z₁ - z₂)
-    nz = (y₁ + y₂) * (x₁ - x₂)
-    Vec(nx, ny, nz)
-  end
-  unormalize(n)
-end
+normal(ngon::Ngon) = newellnormal(vertices(ngon))
 
 # ----------
 # TRIANGLES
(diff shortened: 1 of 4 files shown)


src/geometries/polytopes/ngon.jl
```diff
@@ -65,6 +65,24 @@ angles(ngon::Ngon) = angles(boundary(ngon))
...
 innerangles(ngon::Ngon) = innerangles(boundary(ngon))
 
+# Normal calculation using Newell's algorithm
+# https://people.eecs.berkeley.edu/~ug/slide/pipeline/assignments/backfacecull.shtml
+function normal(ngon::Ngon{N}) where {N}
+  assertion(embeddim(ngon) == 3, "ngon must be 3-dimensional")
+  verts = ngon.vertices
+  nextverts = circshift(ngon.vertices, 1)
+  nx = ny = nz = 0.0 #zero(lentype(ngon))^2
+  for (p₁,p₂) in zip(verts,nextverts)
```
juliohm: We can define `v = CircularVector(ngon.vertices)` to make the implementation more elegant. That allows `v[i], v[i+1]` for any `i`.

src/geometries/polytopes/ngon.jl
```diff
@@ -65,6 +65,24 @@ angles(ngon::Ngon) = angles(boundary(ngon))
 
 innerangles(ngon::Ngon) = innerangles(boundary(ngon))
 
+# Normal calculation using Newell's algorithm
+# https://people.eecs.berkeley.edu/~ug/slide/pipeline/assignments/backfacecull.shtml
+function normal(ngon::Ngon{N}) where {N}
```
juliohm:
```suggestion
function normal(ngon::Ngon)
```

I didn't see `N` used in the implementation.

src/geometries/polytopes/ngon.jl
```diff
@@ -65,6 +65,24 @@ angles(ngon::Ngon) = angles(boundary(ngon))
...
+  assertion(embeddim(ngon) == 3, "ngon must be 3-dimensional")
+  verts = ngon.vertices
+  nextverts = circshift(ngon.vertices, 1)
+  nx = ny = nz = 0.0 #zero(lentype(ngon))^2
+  for (p₁,p₂) in zip(verts,nextverts)
+    x₁, y₁, z₁ = ustrip.(to(p₁))
+    x₂, y₂, z₂ = ustrip.(to(p₂))
+    nx += (z₂+z₁) * (y₂-y₁)
+    ny += (x₂+x₁) * (z₂-z₁)
+    nz += (y₂+y₁) * (x₂-x₁)
```
juliohm: Press `Ctrl+Shift+I` in VSCode to apply the formatting settings of the project.

src/geometries/polytopes/ngon.jl
```diff
@@ -65,6 +65,24 @@ angles(ngon::Ngon) = angles(boundary(ngon))
...
+  end
+  lt = unit(lentype(ngon))
+  return unormalize(Vec(nx*lt, ny*lt, nz*lt))
```
juliohm:
```suggestion
  unormalize(Vec(nx*lt, ny*lt, nz*lt))
```

juliohm (changes requested):
Thank you @barche ! Some minor adjustments and we should be good to go.

barche:
Thanks for the review, should all be fixed!

barche:
There seems to be still a formatting error in the `unormalize` line, but the format in vscode does not seem to correct it.
Edit: I think this is because my vscode uses JuliaFormatter v2, where this is the default:
https://github.com/JuliaEditorSupport/JuliaFormatter.jl/issues/780

I don't see an option to make JuliaFormatter v2 prefer the v1 behavior, but I can of course add the space manually if needed.

juliohm:
Interesting that the version of the formatter is affecting our checks. I will try to fix the issue manually, but we need to keep an eye for it in the future.

Do you have suggestions for more tests for N-gons with N > 4?

JoshuaLampert:
> Interesting that the version of the formatter is affecting our checks. I will try to fix the issue manually, but we need to keep an eye for it in the future.

This is probably because CI uses JuliaFormatter v1, but VSCode uses v2, which can format things differently.

juliohm:
> This is probably because CI uses JuliaFormatter v1, but VSCode uses v2, which can format things differently.

@JoshuaLampert what is your recommendation? Should we port our CI to v2 after some PRs under review are merged?

I plan to review two more PRs over the week. Maybe we could consider this CI update next week after they are incorporated in the project. Changing from v1 to v2 will likely introduce large diffs?

juliohm:
@barche I've moved the implementation to a new `newellnormal` utility to reuse in two places: `Ngon` and `PolyArea`. The latter is a representation that allows polygons with holes, so I simply take the outer vertices to perform the computation. The utility function is useful because the extraction of vertices without allocation differs in these two subtypes of `Polygon`.

JoshuaLampert:
> @JoshuaLampert what is your recommendation? Should we port our CI to v2 after some PRs under review are merged?

Yes, we could at least take a look at the changes and see if the new formatting looks fine. In other repos, I still use v1 of JuliaFormatter.jl because I am not so happy with some formatting rules of v2, but in these other repos I also use other formatting options (e.g., "SciML" style), so it might be fine here.

juliohm:
Allocation behavior is different in Julia LTS, unfortunately. Removing the tests...

juliohm (approved):

juliohm:
Thank you @barche ! We are now ready for the next PR related to `viz` 💯 🙂

barche:
Great! Thanks for all the updates to this!

PR #1405, opened 2026-08-04
Add support for setting normals in visualization

barche:
This is the second part for fixing #535, adding a `normals` option to `viz`. Additionally, I changed the way colors are handled in `vizmesh!`, to treat normals and colors in the same way, since both these attributes can be set either per-vertex or per-element. Both colors and normals are now set as attributes in a `GeometryBasics.Mesh`, which means we no longer need to manually duplicate vertices in `vizmesh!`.


juliohm:
Thank you @barche ! I am wondering if we could simplify this further with a `smooth` boolean option instead. If the two most useful cases for providing normals are the ones where the normals are computed from the geometries themselves, we could hide this low-level detail from end-users and simply set `smooth` to `true` or `false` to switch behavior.

We could also consider _always_ setting normal-per-face when users provide color-per-face because that is more sensible for a discrete approximation. The current behavior where normals are interpolated even when users specify color-per-face could be considered a "bug". That way we would avoid an extra `smooth` option and always do the most reasonable thing for end-users interested in color-per-face.

What are your thoughts? Could we proceed with fixing these normal interpolation internally without additional `viz` options?

barche:
My motivation for adding this is in computer graphics applications, where the normals are sometimes used to control shading. In that context, it is useful to be able to manually set the normals, so I would like to keep the `normals` option for that case.

When `normals` is not set, we could do the following:
- supply per-face normals when per-face colors are set
- otherwise do nothing (so smooth shading) or maybe default to per-face normals if the element count is below a threshold? Other heuristics could be based on the angle between faces, but that is of course more difficult to implement.

juliohm:
> My motivation for adding this is in computer graphics applications, where the normals are sometimes used to control shading.

Could you please elaborate on that? Isn't shading controlled by lights as discussed in the Makie docs? https://docs.makie.org/stable/reference/scene/lighting#Lights

My understanding is that normals are a geometric property of the mesh, and that you can set multiple lights (e.g., `DirectionalLight`, `PointLight`, `SpotLight`, `RectLight`) to control how shades are formed.

barche:
> My understanding is that normals are a geometric property of the mesh, and that you can set multiple lights (e.g., `DirectionalLight`, `PointLight`, `SpotLight`, `RectLight`) to control how shades are formed.

The normals allow control over how the lightsource interacts with the surface in ways that can't really be accomplished by other means. Even though I never exploited this myself, in computer graphics it is expected to be able to control them for certain effects, for example using the normals of a rough mesh on a finer one as is shown here:
https://blenderartists.org/t/customizing-normals-tutorial-series/1369038

(I just watched the beginning)

For me, the practical problem I had is that I wanted to explore a bit how to port parts of K-3D to Julia, and there the normals are stored in the object file, so it makes most sense to load the normals from file and use those, rather than the automatically calculated ones, even if for 99% of cases these will be the same.

juliohm:
> Even though I never exploited this myself,

> it makes most sense to load the normals from file and use those, rather than the automatically calculated ones, even if for 99% of cases these will be the same.

I would like to highlight these two sentences from your reply as they capture the big picture.

The reason why I am reluctant to add a low-level `normals` option to `viz` is that it will be rarely used in practice **and** it only makes sense in a subset of supported domains (e.g., surface meshes). We designed `viz` as a generic function to quickly add geometric objects to the scene with good defaults. If in 99% of the cases people will just provide the normals from the file/mesh, then we better direct our human resources there instead of maintaining code for the 1%. Advanced users who need normal customization could easily extract that `vertices(mesh)` and the `topology(mesh)` to prepare custom visualizations with low-level Makie.jl calls. Makes sense?

> * supply per-face normals when per-face colors are set
> * otherwise do nothing (so smooth shading) or maybe default to per-face normals if the element count is below a threshold? Other heuristics could be based on the angle between faces, but that is of course more difficult to implement.

These suggestions make total sense and are very intuitive. The first one could be considered a bug fix, and the second one where we consider heuristics based on the number of elements could be discussed in a subsequent PR. Would you be willing to adjust the current PR to implement the first suggestion?

Later on, we could reassess the need for `normals` in `viz`. The simple bug fix discussed above will go a long way.

barche:
OK, I agree! Maybe in this PR already I could still ad a `smooth` boolean option, since I think selecting between smooth and flat shading is something many people would expect to be able to do now. Otherwise they would have to manually set a uniform per-face color, which is of course more cumbersome than adding a boolean option. A more advanced auto-selection based on heuristics is not a priority for me.

juliohm:
> OK, I agree!

🎉 

Awesome! Let's start with the simple bug fix, without introducing more options. We could discuss the `smooth` option later on with more brainstorming.

Notice that after a mesh is added to a Makie.jl scene, I believe there are ways to set new normal values. Something like `fig.content[1].mesh.normals = newnormals` could work. Everything in a constructed Makie.jl scene is "hackeable".

barche:
Having now discovered the (very nice BTW) visualization tests, I was surprised that they passed (especially the surf3D-3 test). This showed that the bug was already fixed in the version before this PR, because in the case of per-face colors the vertices were repeated, causing Makie to automatically do the right thing with the normals, at least in the case of planar faces. I still believe this PR has minor advantages compared to the previous implementation:
1. It avoids re-implementing the vertex duplication algorithm here (Makie already does this)
2. It uses the same normal for all triangles that make up a non-planar face, so you don't see the individual triangles of the fan triangulation

juliohm:
> 1. It avoids re-implementing the vertex duplication algorithm here (Makie already does this)

Before I dive into the source code, could you please confirm that Makie.jl also handles meshes with mixed element type? Quadrangle + Triangle + Hexagon + ...?

The last time I looked into it, it didn't support this duplication.

docs/src/visualization.md
```diff
@@ -34,4 +34,27 @@ such as [`Mesh`](@ref) and show facets efficiently:
...
+
+If the number of colors is equal to the number of elements, per-face coloring
+is used. In this case, the normals used in the visualization are also
+calculated per face, resulting in a faceted appearance, even if the same
+color is applied to each face:
+
+```@example viz
+sphere = Sphere((0.,0.,0.), 1.)
+mesh = discretize(sphere, RegularDiscretization(10,10))
+viz(mesh; color=fill(:grey,nelements(mesh)))
```
juliohm: I believe this is already explained in the `viz` docstring. We can omit the `normals` implementation detail in 3D surface meshes.

ext/mesh.jl
```diff
@@ -52,83 +52,88 @@ end
...
-  Makie.map!(plot, [:object, :colorant], [:tmesh, :tcolors]) do mesh, colorant
+  Makie.map!(plot, [:object, :colorant], [:tmesh]) do mesh, colorant
     # relevant settings
     nvert = nvertices(mesh)
     nelem = nelements(mesh)
     verts = map(asmakie, eachvertex(mesh))
     elems = elements(topology(mesh))
 
-    # decide whether or not to reverse connectivity list
-    rev = crs(mesh) <: LatLon && orientation(first(mesh)) == CW ? reverse : identity
```
juliohm, line 63: You are changing too much code. My hope for this PR was simple: if nc == ne, then compute normals and forward to the low-level Makie.jl call.

The new logic will likely introduce issues with LatLon coordinates, and other cases not covered by our visual tests yet.

barche:
> Before I dive into the source code, could you please confirm that Makie.jl also handles meshes with mixed element type? Quadrangle + Triangle + Hexagon + ...?

I think it can with some nudging, but even with the changes in this PR we pass a triangulated mesh to Makie. The only thing I changed is the code to duplicate vertices. I think this way is more future-proof and simpler if we want to add other properties that can be either per-face or per-vertex, since now all attributes (normals and colors) are stored in the GeometryBasics.Mesh.

ext/mesh.jl
barche: I don't think I changed anything in this respect. The only thing I added is a constant to also reverse the normal if the face order is changed. I would be willing to add a test if you can point me in the right direction of an example or similar.

juliohm:
Hi @barche, thank you for clarifying it 🙂 I will pull the PR tomorrow to test the changes locally and will get back to you ASAP. It would be awesome if we could avoid the vertex duplication with the latest Makie.jl version.

ext/mesh.jl
```diff
@@ -125,8 +121,7 @@ function vizmesh!(plot, ::Type, ::Val{2}, edim::Val)
     end
 
     # triangle mesh
-    tmesh = GB.mesh(verts, tris; color=tcolors, normal=tnormals)
-    return (tmesh,)
+    GB.mesh(verts, tris; color=tcolors, normal=tnormals)
```
barche: Note that the function has to return a tuple

ext/mesh.jl
```diff
@@ -52,83 +52,83 @@ end
...
     # relevant settings
     nvert = nvertices(mesh)
     nelem = nelements(mesh)
     verts = map(asmakie, eachvertex(mesh))
     elems = elements(topology(mesh))
 
-    # decide whether or not to reverse connectivity list
-    rev = crs(mesh) <: LatLon && orientation(first(mesh)) == CW ? reverse : identity
+    # decide whether or not to reverse connectivity list and normals
+    rev, sign = crs(mesh) <: LatLon && orientation(first(mesh)) == CW ? (reverse, -1) : (identity, 1)
```
juliohm, line 63: @barche could you please clarify why we need the `sign` besides the `rev` of connectivity? I thought that reversing the connectivity list was enough to correct the normals?

ext/mesh.jl
```diff
@@ -52,83 +52,83 @@ end
...
+
+    facecolors = false
+    if colorant isa AbstractVector
+      colors = Makie.RGBAf.(colorant) # GB.Mesh needs RGBA(f) type
```
juliohm, line 67: I am not sure we need to convert valid `Colorant` to a specific color type (i.e., `RGBAf`). Makie.jl supports all types of Colors.jl, and the `colorant` variable is already preprocessed in our `viz` pipeline, see https://github.com/JuliaGeometry/Meshes.jl/blob/94beefe0bb59de15dcbbdba4ea67c79bbc77a49f/ext/mesh.jl#L6

ext/mesh.jl
```diff
@@ -52,83 +52,83 @@ end
...
         tind += 1
         tris[tind] = GB.TriangleFace(I[1], I[i], I[i + 1])
+        if facecolors
+          if edim == Val(3)
+            tnormals[tind] = sign * asmakie(n)
+          end
+          tcolors[tind] = colors[eind]
+        end
       end
     end
```
juliohm, line 113: Could you please refactor these blocks of code into a single block where you branch once and then return the appropriate vectors of colors and normals filled?

In the current state, this PR splits the initialization from the filling algorithm, which has side effects. In you filling block you are repeatedly checking `if facecolors` in a loop, and that is usually not good for performance.

Try to introduce the minimum amount of change necessary to get the normals in place. The reason I am asking this, is because we optimized these functions in the past, and small changes that seem harmless at first can affect performance, behavior, etc.

juliohm (changes requested):
Attached another round of review.

ext/mesh.jl
juliohm: I see, sorry for the confusion. I had the previous implementation in mind when I suggested the change. In the previous version we were returning multiple objects in a tuple, and now you are suggesting that we return the triangle mesh directly.

juliohm: Fixed it.

ext/mesh.jl
barche: Because I am using `normal(mesh[eind])` the reversed connectivity is not taken into account. I could instead `materialize` the reversed face and compute the normal of that, unless there is an efficiency concern with that approach?

ext/mesh.jl
barche: If the color is passed as a keyword argument to `Makie.mesh`, it gets converted to an opengl-compatible color (`RGBAf` in fact) automatically. With this PR the color is stored in the GeometryBasics.Mesh and passed as-is to OpenGL, which is why we have to convert it here.

ext/mesh.jl
barche: OK, will do this later this week, thanks for the comments!

ext/mesh.jl
juliohm: Got it. The `normal` calculation is done with the original mesh and that makes total sense.

ext/mesh.jl
juliohm: I suggest that we restore the old logic where we simply return the colors as a compute graph node, and use the node in the final plot call. Color conversions can be tricky and we can potentially introduce new bugs by converting colors manually. We created Colorfy.jl to hide part of this complexity.

barche: Well, the conversion is simply a standard conversion performed by Colors.jl, which is actually the same one as used by Makie when colors are passed as an argument to the plot command (see the internal `to_color` function in conversions.jl). It is much simpler than reintroducing the vertex duplication, which is basically a reimplementation of `GeometryBasics.expand_faceviews` and much more complicated than converting one type from Colors.jl to another.

Note also that the way used now to pass per-face colors and normals is as documented by Makie.jl: https://docs.makie.org/dev/reference/plots/mesh#Face-colors-and-normals

juliohm: I see. Please feel free to adjust the PR with the most up-to-date method in Makie.jl and we can continue the review from there. The implementation we currently have was written before Makie.jl supported color-per-face.

barche:
I updated the implementation with a simpler version that I think better matches the coding style of the previous implementation. Ready for the next review round ;)

ext/mesh.jl
```diff
@@ -52,83 +52,63 @@ end
...
-    else # single color
-      # nothing needs to be done
-      tverts = verts
-      telems = tris
-      tcolors = colorant
-    end
+    makefaceview(facevalues, elem4tri) =
+      GB.FaceView(map(eind -> facevalues[eind], elem4tri), GB.GLTriangleFace.(eachindex(elem4tri)))
+    tnormals = facenormals ? makefaceview(sign .* GB.Vec{3,Float32}.(asmakie.(normal.(mesh))), elem4tri) : nothing
+    tcolors = facecolors ? makefaceview(colors, elem4tri) : colors
```
juliohm: Three comments/questions:

1. I believe we should use `TriangleFace` as opposed to the specific `GLTriangleFace`, right? This recipe should work with all Makie backends.
2. Can we get rid of the additional conversion to `Float32`? The call to `asmakie` should be enough, right?
3. You are passing `elem4tri` as a second argument in `makefaceview`, but this variable could be grabbed from the current scope instead. You did it on purpose to avoid some performance issue?

juliohm (commented):
I love PRs that remove tons of lines of code! 💯 This is what I am talking about!

Left some comments/questions. After they are addressed I will clone locally to test the different options.

ext/mesh.jl
barche: 1. The name GLTriangleFace is just to indicate that the fundamental integer and floating point types used are OpenGL-compatible (e.g. `UInt32`, `Float32`). This is the type used at the level of Makie.jl for mesh conversion to all backends, so using it directly here prevents a potential unnecessary extra conversion at the level of Makie.
2. It worked locally with GLMakie, but not converting to `Float32` made the CairoMakie tests fail. The optimal solution would be to make the change in `asmakie` to always make it return Float32, if this is acceptable?
3. Closures make me nervous about performance, but I did a separate benchmark on a representative testcase and it seems the difference is completely negligible, so I can remove the argument if you prefer.

juliohm: 1. Let's use `TriangleFace` and avoid depending too much on Makie internals. It also gives Makie an opportunity to send more computations to the GPU when that is possible. Modern GPUs support double precision and people might one day remove the coercion to Float32 on the backend side. I don't know.
2. It seems like a bug to be reported in CairoMakie.jl? I remember older versions of Makie.jl had issues with double precision, but they got fixed. We don't want to convert coordinates to single precision prematurely.
3. If the performance difference is negligible, I suggest we simplify the code.

barche: I fully agree on 1 and 3. Regarding 2, the type assert that is causing the failure was added in this PR: https://github.com/MakieOrg/Makie.jl/pull/5446

I don't know if it is too strict or not, but it seems to touch on performance-sensitive CairoMakie code. Internally, Makie seems to default to Vec3f for normals everywhere, so I think it is the safest to do the same on our side. After all, it does not impact the Meshes.jl representation of the normals, only the visualization mesh is affected.

I tried removing the assert in CairoMakie and this required one more small change to make CairoMakie swallow the 64-bit normals, but I'm not sure it's worth going down the rabbit hole of making this PR depend on getting that change accepted into Makie?

For reference the Makie diff:

```
diff --git a/CairoMakie/src/mesh.jl b/CairoMakie/src/mesh.jl
index b436f97d8..98a9b697c 100644
--- a/CairoMakie/src/mesh.jl
+++ b/CairoMakie/src/mesh.jl
@@ -280,7 +280,7 @@ function draw_mesh3D(scene, screen, plot::ComputeGraph)
     )
     screen_points = cairo_project_to_screen(plot, output_type = Point3f)::Vector{Point3f}
     meshfaces = plot.faces[]::Vector{GLTriangleFace}
-    meshnormals = plot.normals[]::Union{Nothing, Vector{Vec3f}}
+    meshnormals = plot.normals[]
     _meshuvs = plot.texturecoordinates[]
 
     if (_meshuvs isa AbstractVector{<:Vec3})
@@ -339,7 +339,7 @@ end
 
 to_vec(c::Colorant) = Vec3f(red(c), green(c), blue(c))
 prepare_normals(normalmatrix::Mat3f, normals::Nothing) = nothing
-function prepare_normals(normalmatrix::Mat3f, normals::Vector{Vec3f})
+function prepare_normals(normalmatrix::Mat3f, normals::Vector{<:Vec3})
     return [zero_normalize(normalmatrix * normal) for normal in normals]
 end
```

juliohm: The CairoMakie.jl diff is quite simple. I'd just change `Union{Nothing, Vector{Vec3f}}` to `Union{Nothing, Vector{<:Vec3}}` to avoid possible performance regressions. It deserves a PR in the Makie side.

In the meantime, I will update this PR with the suggested changes to test things locally.

juliohm:
@barche I did run a simple benchmark:

```julia
using Meshes
using BenchmarkTools
import GLMakie

m = discretize(Sphere((0,0,0), 1))
for _ in 1:4
  m = refine(m, QuadRefinement())
end

@btime viz(m)
```

Results:
- 43.954 ms (83870 allocations: 72.47 MiB) on master
- 1.061 s (11948587 allocations: 985.67 MiB) on this PR

The Makie solution with FaceView is much slower, unfortunately. Should we go back to the original implementation, just adding a new `tnormals` variable as I suggested initially?

barche:
Thanks for the test, I will investigate and of course revert if there is no more elegant fix, since such a performance regression is obviously unacceptable.

barche:
Still looking into this. This is clearly a performance issue in GeometryBasics.jl in the `merge_vertex_indices` function. I'd prefer to fix it there and I'll do a little more digging.

juliohm:
> Still looking into this. This is clearly a performance issue in GeometryBasics.jl in the `merge_vertex_indices` function. I'd prefer to fix it there and I'll do a little more digging.

Thank you @barche !

juliohm:
Hi @barche , any progress on the Makie side?

barche:
Yes, I made the GeometryBasics.jl code a lot faster, but I am currently struggling with making a correct comparison with the old code, since `@bench viz(m)` actually only times a part of the pipeline, for example the computations of the normals by Makie (as done in the original way of calling Makie) is not timed that way. Once I get a correct benchmark I'll make a PR to GeometryBasics and we can evaluate here which approach to keep.

barche:
I finally organized the GeometryBasics improvements and submitted the PR, once this goes through we can evaluate the impact here.

juliohm:
Thank you @barche ! I saw the PR got merged there. Time to retry this one? 🙂

barche:
In the previous code version, uniform color resulted in per-face colors, which is slower than per-vertex colors. I switched this back to per-vertex colors in the last commit and rebased everything onto master. Benchmark results using the following code:

```julia
using Meshes
using BenchmarkTools
import GLMakie

m = discretize(Sphere((0,0,0), 1))
for _ in 1:4
  @show nelements(m)
  global m = refine(m, QuadRefinement())
end

function benchviz(m)
  f,__ = viz(m)
  return
end

const colors = fill(GLMakie.RGBf(0,1,1), nelements(m))
function benchviz(m,c)
  f,__ = viz(m; color=c)
  return
end

@benchmark benchviz(m) setup=(GC.gc())
@benchmark benchviz(m, colors) setup=(GLMakie.closeall(); GC.gc())
```

## Per vertex

### Original code

```
BenchmarkTools.Trial: 13 samples with 1 evaluation per sample.
 Range (min … max):  43.117 ms … 47.418 ms  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     44.197 ms              ┊ GC (median):    0.00%
 Time  (mean ± σ):   44.529 ms ±  1.300 ms  ┊ GC (mean ± σ):  0.00% ± 0.00%

  ▁ ▁▁     █ ▁  ▁▁  ▁     ▁         ▁            ▁          ▁  
  █▁██▁▁▁▁▁█▁█▁▁██▁▁█▁▁▁▁▁█▁▁▁▁▁▁▁▁▁█▁▁▁▁▁▁▁▁▁▁▁▁█▁▁▁▁▁▁▁▁▁▁█ ▁
  43.1 ms         Histogram: frequency by time        47.4 ms <

 Memory estimate: 72.45 MiB, allocs estimate: 83553.
```

### New code

```
BenchmarkTools.Trial: 13 samples with 1 evaluation per sample.
 Range (min … max):  47.266 ms … 76.035 ms  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     48.531 ms              ┊ GC (median):    0.00%
 Time  (mean ± σ):   51.422 ms ±  7.711 ms  ┊ GC (mean ± σ):  0.00% ± 0.00%

    █                                                          
  █▁██▁▁▅▁▁▅▁▁▁▁▁▁▅▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▅ ▁
  47.3 ms         Histogram: frequency by time          76 ms <

 Memory estimate: 92.57 MiB, allocs estimate: 83558.
```

## Per face

### Original code

```
BenchmarkTools.Trial: 8 samples with 1 evaluation per sample.
 Range (min … max):  345.183 ms … 374.105 ms  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     354.723 ms               ┊ GC (median):    0.00%
 Time  (mean ± σ):   355.736 ms ±   9.551 ms  ┊ GC (mean ± σ):  0.00% ± 0.00%

  █ █      █     █        █   █       █                       █  
  █▁█▁▁▁▁▁▁█▁▁▁▁▁█▁▁▁▁▁▁▁▁█▁▁▁█▁▁▁▁▁▁▁█▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁█ ▁
  345 ms           Histogram: frequency by time          374 ms <

 Memory estimate: 493.75 MiB, allocs estimate: 83587.
```

### New code

```
BenchmarkTools.Trial: 9 samples with 1 evaluation per sample.
 Range (min … max):  222.031 ms … 950.540 ms  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     311.969 ms               ┊ GC (median):    0.00%
 Time  (mean ± σ):   403.813 ms ± 262.348 ms  ┊ GC (mean ± σ):  0.00% ± 0.00%

  █                                                              
  █▇▁▁▁▁▁▇▁▁▇▁▇▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▇▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▇ ▁
  222 ms           Histogram: frequency by time          951 ms <

 Memory estimate: 527.61 MiB, allocs estimate: 83620.
```

Unfortunately, it is difficult to benchmark the rendering pipeline without noise, as can be seen from the large spread, but I believe the difference is quite minimal and I would be inclined to prefer the new code because of the simplicity. I ran each benchmark multiple times and for each method took the best result with the smallest variance. All were tests done with the latest releases of GeometryBasics.jl (v0.5.13) and StaticArrays.jl (v1.9.22), which contain the performance improvements.

juliohm:
Thank you! I'm far from my computer but saw you mentioned the switch to per
vertex colors by default. Could you please default to per face colors? That
is what we usually do when we have a table of attributes attached to
elements of the mesh.

Em sáb., 19 de set. de 2026, 13:57, Bart Janssens ***@***.***>
escreveu:

barche:
The per-vertex colors are for the default uniform color. In the old version, uniform colors are also applied per-vertex, so I did the same for this new version now.

I also realized I had forgotten about the CairoMakie bug, so I made a PR there too.

PR #1406, opened 2026-08-19
Implement `MoreiraMarch` for concave hulls

souma4:
In #1328 we had discussion and decided to separate the Moreira Santos k-nearest hull algorithm into it's own hull method for simpler maintenance. Code is written much like existing hull methods, but aligned to the original papers pseudocode. I do make some simplifications because we have methods that do what they suggest simpler (such as using `any(intersects())` instead of a `while`-loop. I also simplify their step logic: length of the current hull is more than satisfactory, no need for weird 2-indexing.


juliohm:
Thank you @souma4 ! I will review this PR soon using the pseudo-code in the paper as a reference. If you could fix the formatting issues, that would be great.

juliohm (approved):
Thank you! 🎉❤️ That is very useful addition!

PR #1407, opened 2026-08-20
Add intersection for Box{🌐}

anj1:
Based on the discussion in https://github.com/JuliaGeometry/Meshes.jl/pull/1369#discussion_r3364791648, this is a smaller, less broad PR: Just providing intersections over LatLon boxes.

The code broadly follows the industry-standard logic in https://s2geometry.io/devguide/basic_types.html, but with some improvements (enabled by the very good design of this library): we return true `Multi` boxes in cases of wrapped overlapping boxes.

### DIFF
--- a/src/intersections/boxes.jl
+++ b/src/intersections/boxes.jl
@@ -35,10 +35,9 @@ function intersection(f, box₁::Box, box₂::Box)
 end
 
 function intersection(f, box₁::Box{🌐}, box₂::Box{🌐})
-  # corners in a common LatLon CRS
-  C = manifoldcrs(box₁)
-  cs = (convert(C, coords(p)) for p in (extrema(box₁)..., extrema(box₂)...))
-  m₁, M₁, m₂, M₂ = promote(cs...)
+  # corners in a common CRS
+  m₁, M₁ = coords.(extrema(box₁))
+  m₂, M₂ = coords.(extrema(box₂ |> Proj(crs(box₁))))
 
   # intersect coordinate ranges
   latₛ, latₑ = max(m₁.lat, m₂.lat), min(M₁.lat, M₂.lat)
--- a/src/intersections/boxes.jl
+++ b/src/intersections/boxes.jl
@@ -36,8 +36,9 @@ end
 
 function intersection(f, box₁::Box{🌐}, box₂::Box{🌐})
   # corners in a common CRS
-  m₁, M₁ = coords.(extrema(box₁))
-  m₂, M₂ = coords.(extrema(box₂ |> Proj(crs(box₁))))
+  crs = manifoldcrs(box₁)
+  m₁, M₁ = coords.(extrema(box₁ |> Proj(crs)))
+  m₂, M₂ = coords.(extrema(box₂ |> Proj(crs)))
 
   # intersect coordinate ranges
   latₛ, latₑ = max(m₁.lat, m₂.lat), min(M₁.lat, M₂.lat)
--- a/src/intersections/boxes.jl
+++ b/src/intersections/boxes.jl
@@ -53,15 +53,15 @@ function intersection(f, box₁::Box{🌐}, box₂::Box{🌐})
   isempty(lonranges) && return @IT NotIntersecting nothing f
 
   # classify intersection
-  latpoint = isapproxzero(latₑ - latₛ)
-  lonpoints = all(r -> isapproxzero(last(r) - first(r)), lonranges)
-  corner = latpoint && lonpoints
-  overlap = !latpoint && !lonpoints
+  singlelat = isapproxequal(latₛ, latₑ)
+  singlelon = all(lon -> isapproxequal(lon[1], lon[2]), lonranges)
+  iscorner = singlelat && singlelon
+  overlaps = !singlelat && !singlelon
 
   # construct geometry
   geoms = map(lonranges) do (lonₛ, lonₑ)
     u = withcrs(box₁, (latₛ, lonₛ))
-    if corner
+    if iscorner
       u
     else
       v = withcrs(box₁, (latₑ, lonₑ))
@@ -70,9 +70,9 @@ function intersection(f, box₁::Box{🌐}, box₂::Box{🌐})
   end
   geom = length(geoms) == 1 ? only(geoms) : Multi(geoms)
 
-  if overlap
+  if overlaps
     return @IT Overlapping geom f
-  elseif corner
+  elseif iscorner
     return @IT CornerTouching geom f
   else
     return @IT Touching geom f
--- a/src/intersections/boxes.jl
+++ b/src/intersections/boxes.jl
@@ -49,17 +49,17 @@ function intersection(f, box₁::Box{🌐}, box₂::Box{🌐})
   latₛ > latₑ && return @IT NotIntersecting nothing f
 
   # check longitude coordinate
-  lons = _lonintersects(min₁.lon, max₁.lon, min₂.lon, max₂.lon)
-  isempty(lons) && return @IT NotIntersecting nothing f
+  lonranges = _lonintersects(min₁.lon, max₁.lon, min₂.lon, max₂.lon)
+  isempty(lonranges) && return @IT NotIntersecting nothing f
 
-  # classify
+  # classify intersection
   latpoint = isapproxzero(latₑ - latₛ)
-  lonpoints = all(r -> isapproxzero(last(r) - first(r)), lons)
+  lonpoints = all(r -> isapproxzero(last(r) - first(r)), lonranges)
   corner = latpoint && lonpoints
   overlap = !latpoint && !lonpoints
 
   # construct geometry
-  geoms = map(lons) do (lonₛ, lonₑ)
+  geoms = map(lonranges) do (lonₛ, lonₑ)
     u = withcrs(box₁, (latₛ, lonₛ))
     if corner
       u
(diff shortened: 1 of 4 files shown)


src/intersections/boxes.jl
```diff
@@ -33,3 +33,77 @@ function intersection(f, box₁::Box, box₂::Box)
     return @IT NotIntersecting nothing f
   end
 end
+
+function intersection(f, box₁::Box{🌐}, box₂::Box{🌐})
+  # corners in a common LatLon CRS
+  C = manifoldcrs(box₁)
+  cs = (convert(C, coords(p)) for p in (extrema(box₁)..., extrema(box₂)...))
+  m₁, M₁, m₂, M₂ = promote(cs...)
```
juliohm: You can convert the `CRS` of a geometry using the `Proj` transform. That is more efficient and requires less code. Just do `box2 |> Proj(crs(box1))` to safely convert the CRS of box2 to the CRS of box1 and then extract the `coords` of  the `extrema`.

juliohm (changes requested):
Appreciate the contribution! ❤️ 

If you are basing your implementation on external code, please make sure to check the software license. Is there a webpage or paper explaining the algorithm you implemented?

I will try to review the PR over the weekend.

juliohm:
Also, press `Ctrl+Shift+I` in VSCode to apply the format settings of the project. The format tests are failing.

src/intersections/boxes.jl
anj1: I implemented the math myself after reading the documentation, rather than copying S2 source code. Although the source code is Apache 2.0 licensed as well.

anj1: Also, good point about `|> Proj`! I made the change https://github.com/JuliaGeometry/Meshes.jl/pull/1407/commits/d447160b5bbf67a74a513478c2633d407dfb1cbd

src/intersections/boxes.jl
```diff
@@ -33,3 +33,73 @@ function intersection(f, box₁::Box, box₂::Box)
     return @IT NotIntersecting nothing f
   end
 end
+
+function intersection(f, box₁::Box{🌐}, box₂::Box{🌐})
+  # corners in a common CRS
+  m₁, M₁ = coords.(extrema(box₁))
+  m₂, M₂ = coords.(extrema(box₂ |> Proj(crs(box₁))))
```
juliohm: If you need to enforce the exact same `LatLon` based on the same datum, you need to do something like

```julia
crs = manifoldcrs(box1)
box1 |> Proj(crs)
box2 |> Proj(crs)
```

The current code is just enforcing that both boxes have the same CRS,  which might be something else that is not `LatLon`.

test/predicates.jl
```diff
@@ -99,7 +99,7 @@ end
   hole1 = [merc(0.2, 0.2), merc(0.4, 0.2), merc(0.4, 0.4), merc(0.2, 0.4)]
   hole2 = [merc(0.6, 0.2), merc(0.8, 0.2), merc(0.8, 0.4), merc(0.6, 0.4)]
   poly = PolyArea([outer, hole1, hole2])
-  @test all(p ∈ poly for p in outer)
+  @test all(p in poly for p in outer)
```
juliohm: These changes might be caused by the different formmatter installed locally in VSCode. Please undo. We need to fix the discrepancy of the formmatter version in the repo.

anj1: Interesting; I'm doing it from the REPL:
```julia
using JuliaFormatter
format(".")
```
as I thought this would be the most consistent with how it's configured in the CI. But I made that fix now.

juliohm:
Thank you @anj1! I will review the code more carefully tomorrow.

src/intersections/boxes.jl
```diff
@@ -33,3 +32,79 @@ function intersection(f, box₁::Box, box₂::Box)
...
+  end
+  geom = length(geoms) == 1 ? only(geoms) : Multi(geoms)
+
+  if overlap
+    return @IT Overlapping geom f
+  elseif corner
+    return @IT CornerTouching geom f
+  else
+    return @IT Touching geom f
+  end
```
juliohm, line 75: Please try to base your algorithm in the Euclidean version. I don't think we can have a `Multi`-geometry as the result of an intersection between two boxes. The only possible options are a `Point`, a `Box` or `Nothing`.

juliohm: Nevermind, I am seeing the new option on the sphere now. Two boxes can lead to a `Multi`-`Box` due to the periodicity of the longitude.

anj1: yep! Many other libraries would either return non-exact results or throw an error in this case - in your library we can actually return the true result.

src/intersections/boxes.jl
```diff
@@ -33,3 +32,80 @@ function intersection(f, box₁::Box, box₂::Box)
...
+    return lo == Δ && hi == -Δ ? [(Δ, Δ)] : [(lo, hi)]
+  end
+  ranges
+end
+
+# longitude intervals in the canonical [-180°, 180°] range
+function _lonintervals(lo, hi)
+  Δ = oftype(lo, 180u"°")
+  lo ≤ hi ? [(lo, hi)] : [(-Δ, hi), (lo, Δ)]
+end
```
juliohm: @anj1 our `LatLon` type fixes the longitude range at construction so we don't need to manually adjust it in algorithms. You can assume that all `lon` values extracted from a `LatLon` object will be in the [-180,180] range. I believe we can get rid of this function.

anj1: It's true that the endpoints are already normalized by `LatLon`. But what `_lonintervals` is doing is splitting a wrapped range into linear intervals for intersection. Example: `(170, -170)` has canonical endpoints but represents `[170, 180] ∪ [-180, -170]`. This is also the convention used by `inlonrange`. If it's removed, the code would incorrectly report no intersections for some wrapped boxes.

Here's an example you can run:

```julia
using Meshes, CoordRefSystems

b1 = Box(Point(LatLon(-10, 170)), Point(LatLon(10, -170)))
b2 = Box(Point(LatLon(-5, 175)), Point(LatLon(5, -175)))

east = Point(LatLon(0, 177))
west = Point(LatLon(0, -177))

@show east ∈ b1 && east ∈ b2
@show west ∈ b1 && west ∈ b2
@show type(intersection(b1, b2))
@show b1 ∩ b2
```

Without `_lonintervals`, this reports

```
east ∈ b1 && east ∈ b2 = true
west ∈ b1 && west ∈ b2 = true
type(intersection(b1, b2)) = NotIntersecting
b1 ∩ b2 = nothing
```

whereas the expected intersection is ` Box((-5°, 175°), (5°, -175°))`.

If it would make it clearer, we can rename it to `_splitlonrange` and clarify with a comment. If the issue is the existence of the helper itself, we can inline the logic.

src/intersections/boxes.jl
```diff
@@ -33,3 +32,76 @@ function intersection(f, box₁::Box, box₂::Box)
...
+  _hasπdeg(ranges) = any(lon -> lon[1] == -πdeg || lon[2] == πdeg, ranges)
+  if _hasπdeg(ranges₁) && _hasπdeg(ranges₂) && !_hasπdeg(ranges)
+    push!(ranges, (πdeg, πdeg))
+  end
+
+  # ranges meeting at the antimeridian form a single wrapped interval
+  if length(ranges) ≥ 2 && first(ranges)[1] == -πdeg && last(ranges)[2] == πdeg
+    a, b = last(ranges)[1], first(ranges)[2]
+    return a == πdeg && b == -πdeg ? [(πdeg, πdeg)] : [(a, b)]
+  end
```
juliohm, line 104: Could you please share examples to clarify these two paragraphs? 🙂

src/intersections/boxes.jl
```diff
@@ -33,3 +32,76 @@ function intersection(f, box₁::Box, box₂::Box)
...
+  iscorner = singlelat && singlelon
+  overlaps = !singlelat && !singlelon
+
+  # construct geometry
+  geoms = map(lonranges) do (lonₛ, lonₑ)
+    u = withcrs(box₁, (latₛ, lonₛ))
+    v = withcrs(box₁, (latₑ, lonₑ))
+    iscorner ? u : Box(u, v)
+  end
+  geom = length(geoms) == 1 ? only(geoms) : Multi(geoms)
```
juliohm: Please rebase the PR to the latest master branch to use the new `maybemulti` utility.

juliohm (changes requested):
@anj1 thank you for this nice contribution! ❤️ 

I've made a few changes to improve readability and to facilitate future maintenance.

src/intersections/boxes.jl
anj1: Yep. So first we split wrapped ranges into linear pieces, then we have to handle two seam cases: `[170, 180]` and `[-180, -170]`, which share the `-+180` point despite having no linear overlap, while `[(-180, -175), (175, 180)]` must be recombined as the wrapped range `(175, -175)`.

The degenerate `(180, -180)` case is canonicalized to `(180°, 180°)` so it is correctly classified as a single longitude.

src/intersections/boxes.jl
anj1: done: https://github.com/JuliaGeometry/Meshes.jl/pull/1407/changes/bc668b66d8cb98c4a955b6a71c5794ff4f411aa7

src/intersections/boxes.jl
juliohm: Thank you @anj1 for clarifying it 🙂

juliohm (approved):
🎉

src/intersections/boxes.jl
```diff
@@ -33,3 +32,76 @@ function intersection(f, box₁::Box, box₂::Box)
...
+# 4. do not overlap nor intersect (NotIntersecting -> Nothing)
+function intersection(f, box₁::Box{🌐}, box₂::Box{🌐})
+  # corners in a common CRS
+  crs = manifoldcrs(box₁)
+  min₁, max₁ = coords.(extrema(box₁ |> Proj(crs)))
+  min₂, max₂ = coords.(extrema(box₂ |> Proj(crs)))
+
+  # check latitude coordinate
+  latₛ, latₑ = max(min₁.lat, min₂.lat), min(max₁.lat, max₂.lat)
+  latₛ > latₑ && return @IT NotIntersecting nothing f
```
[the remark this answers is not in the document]
juliohm: We can look into `mayberound` in the future if necessary.

PR #1409, opened 2026-08-20
Add centroid definition for Ellipsoid

### LINKED ISSUE #1391: Find out which tests are emitting warnings
The following warnings are being emitted by our test suite:

```julia
 Warning: maximum number of subdivisions reached `maxsubdiv=32768`, try increasing the keyword argument `maxsubdiv`.
┌ Warning: maximum number of subdivisions reached `maxsubdiv=32768`, try increasing the keyword argument `maxsubdiv`.
```

As far as I know, there is no method to print the exact lines that are emitting the warning, so we need to run tests one by one in the REPL to discover the root of the issue. It is likely emitted by some test that calls `centroid` as we changed the fallback to use h-adaptive integration.

jitendravjh:
Closes #1391.

Both warnings come from the `Scale` test item in test/transforms.jl, at `@test centroid(r) ≈ f(centroid(g))` with `g = Ellipsoid(T.((1, 2, 3)))`. Ellipsoid had no centroid definition, so it went through the generic integral.

That was also inaccurate away from the origin:

```julia
julia> centroid(Ellipsoid((3.0, 2.0, 1.0), Point(1.0, 2.0, 3.0)))
Point(x: 1.0023394395953535 m, y: 2.004678969276033 m, z: 3.0070184557365867 m)
```

An ellipsoid is symmetric about its center, so the centroid is the center.

### DIFF
--- a/src/centroid.jl
+++ b/src/centroid.jl
@@ -19,6 +19,8 @@ centroid(b::Ball) = center(b)
 
 centroid(s::Sphere) = center(s)
 
+centroid(e::Ellipsoid) = center(e)
+
 centroid(d::Disk) = center(d)
 
 centroid(c::Circle) = center(c)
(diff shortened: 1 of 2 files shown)


juliohm (approved):
Very nice @jitendravjh ! Thank you for catching and fixing these warnings! ❤️

PR #1412, opened 2026-08-20
Fix centroid of unbounded geometries

### LINKED ISSUE #1411: `centroid` of `Ray` and `Line` returns `NaN`
```julia
julia> centroid(Ray(Point(1.0, 2.0), Vec(1.0, 1.0)))
┌ Warning: maximum number of subdivisions reached `maxsubdiv=16384`, try increasing the keyword argument `maxsubdiv`.
Point(x: NaN m, y: NaN m)

julia> centroid(Line(Point(0.0, 5.0), Point(1.0, 5.0)))
┌ Warning: maximum number of subdivisions reached `maxsubdiv=16384`, try increasing the keyword argument `maxsubdiv`.
Point(x: NaN m, y: NaN m)

julia> centroid(Line(Point(0.0, 0.0), Point(1.0, 1.0)))
Point(x: -0.0 m, y: -0.0 m)
```

Both have infinite measure, so the generic `centroid` divides by `Inf`. `Plane` has infinite measure too but is handled explicitly and returns its anchor point.

Should `Ray` and `Line` follow `Plane` and return a point on the geometry, or throw instead? Happy to send a PR either way.

jitendravjh:
Closes #1411.

Went with an error rather than returning some point, since Line equality is geometric:

```julia
julia> l1, l2 = Line(Point(0.0, 0.0), Point(1.0, 1.0)), Line(Point(5.0, 5.0), Point(9.0, 9.0));

julia> l1 == l2
true

julia> l1(0), l2(0)
(Point(x: 0.0 m, y: 0.0 m), Point(x: 5.0 m, y: 5.0 m))
```

so any point we pick makes two equal lines differ. Before #1387 both were hitting a MethodError anyway.

Nothing depended on `centroid(::Plane)`. `Stretch` has its own `_origin(p::Plane) = p(0, 0)`, `TransformedGeometry` goes through `isparametrized`, and `supportfun` and `gjkprepare` need `boundarypoints` which errors on a plane. Only the Cone test used it, and `p(0, 0)` works there.

Just removing the method is not enough though, it falls back to the generic integral and returns NaN, so Plane throws too. `Ray`, `Line` and `Plane` are the only geometries with `typemax` measure.

Removed `center(p::Plane)` as well.

### DIFF
--- a/src/centroid.jl
+++ b/src/centroid.jl
@@ -12,7 +12,11 @@ centroid(g::Geometry) = withcrs(g, integral(to, g) / measure(g))
 
 centroid(p::Point) = p
 
-centroid(p::Plane) = center(p)
+centroid(::Ray) = throw(ArgumentError("centroid is not defined for unbounded geometries"))
+
+centroid(::Line) = throw(ArgumentError("centroid is not defined for unbounded geometries"))
+
+centroid(::Plane) = throw(ArgumentError("centroid is not defined for unbounded geometries"))
 
 centroid(b::Box) = center(b)
 
--- a/src/geometries/primitives/plane.jl
+++ b/src/geometries/primitives/plane.jl
@@ -40,8 +40,6 @@ paramdim(::Type{<:Plane}) = 2
 
 normal(p::Plane) = unormalize(ucross(p.u, p.v))
 
-center(p::Plane) = p(0, 0)
-
 ==(p₁::Plane, p₂::Plane) =
   p₁(0, 0) ∈ p₂ && p₁(1, 0) ∈ p₂ && p₁(0, 1) ∈ p₂ && p₂(0, 0) ∈ p₁ && p₂(1, 0) ∈ p₁ && p₂(0, 1) ∈ p₁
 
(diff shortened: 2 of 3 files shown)


src/centroid.jl
```diff
@@ -12,7 +12,7 @@ centroid(g::Geometry) = withcrs(g, integral(to, g) / measure(g))
 
 centroid(p::Point) = p
 
-centroid(p::Plane) = center(p)
+centroid(::Union{Ray,Line,Plane}) = throw(ArgumentError("centroid is not defined for unbounded geometries"))
```
juliohm: Let's keep the methods separate for organizational purposes.

juliohm:
> `center(p::Plane)` is still exported and now unused internally, can drop that too if you prefer, though it would be breaking.

Thank you for the through review! Please feel free to remove the `center` method as well.

jitendravjh:
Both done. The other `center` calls in rand.jl are on a Disk so they are unaffected. Should clear the codecov drop too.

juliohm (approved):
Thank you! 💯

Issue #1410, opened 2026-08-20
`discretize` and `simplexify` for `Ball{𝔼{3}}`, `Cone` and `Frustum`


jitendravjh:
```julia
julia> discretize(Ball(Point(0, 0, 0), 1.0))
ERROR: StackOverflowError:

julia> measure(Cone(Disk(Plane(Point(0, 0, 0), Vec(0, 0, 1)), 1.0), Point(0, 0, 1)))
ERROR: StackOverflowError:

julia> measure(Frustum(Disk(Plane(Point(0, 0, 0), Vec(0, 0, 1)), 2.0), Disk(Plane(Point(0, 0, 1), Vec(0, 0, 1)), 1.0)))
ERROR: StackOverflowError:
```

These two call each other, so anything without a more specific method keeps bouncing between them:

https://github.com/JuliaGeometry/Meshes.jl/blob/94beefe0bb59de15dcbbdba4ea67c79bbc77a49f/src/discretization.jl#L78

https://github.com/JuliaGeometry/Meshes.jl/blob/94beefe0bb59de15dcbbdba4ea67c79bbc77a49f/src/discretization.jl#L224

`Ball{𝔼{3}}`, `Cone` and `Frustum` are the ones ending up there. Ball only needs the `discretize` method. Cone has `isparametrized` already, it needs `discretize` and `isperiodic`. Frustum has the parametrization at [frustum.jl#L44](https://github.com/JuliaGeometry/Meshes.jl/blob/94beefe0bb59de15dcbbdba4ea67c79bbc77a49f/src/geometries/primitives/frustum.jl#L44) but needs `discretize`, `isparametrized` and `isperiodic`.

Can send a PR, let me know how you want Cone and Frustum handled.

jitendravjh:
I tried mirroring `Cylinder` for `Cone` (`isperiodic`, `discretize`, `appendtopo = _appendaxis`). It works and the volume converges, 1.044444 at 2x50x2 and 1.047025 at 8x200x8 against πr²h/3 = 1.047198, but the mesh gets degenerate elements at the apex, 100 out of 300 at 2x50x2 and 1600 out of 14400 at 8x200x8, while `Cylinder` gets none. So the apex needs its own handling next to `_appendcenter` and `_appendpoles`.

[...]

PR #1413, opened 2026-08-21
Implement discretize for Ball{𝔼{3}}

jitendravjh:
Part of #1410.

Added the `discretize` method with `RegularDiscretization(2, 50, 50)`.

`extrapoints(b::Ball, sz)` was only adding the centre, so the poles from the boundary sphere got dropped, and with no `appendtopo` the centre was sampled but never referenced.

`_appendcenteraxis` closes both, so the volume no longer depends on the radial count, 4.17587 at nr=1 and 4.17442 at nr=8, where before it went 4.01 to 4.18.

The ball uses tetrahedra at the centre and the poles, and `_manualconnec(::Tetrahedron)` was missing, so `simplexify` needed that too.

### DIFF
--- a/src/sampling/regular.jl
+++ b/src/sampling/regular.jl
@@ -53,6 +53,14 @@ firstoffset(b::Ball) = (n -> inv(n), firstoffset(boundary(b))...)
 lastoffset(b::Ball) = (n -> zero(n), lastoffset(boundary(b))...)
 extrapoints(b::Ball, sz) = (center(b),)
 
+function extrapoints(b::Ball{𝔼{3}}, sz)
+  T = numtype(lentype(b))
+  n = sz[1]
+  δₛ, δₑ = first(firstoffset(b)), first(lastoffset(b))
+  rs = range(T(δₛ(n)), stop=T(1 - δₑ(n)), length=n)
+  (center(b), (b(r, 0, 0) for r in rs)..., (b(r, 1, 0) for r in rs)...)
+end
+
 _firstoffset(::Sphere, ::Val{3}) = (n -> inv(n + 1), n -> zero(n))
 _lastoffset(::Sphere, ::Val{3}) = (n -> inv(n + 1), n -> inv(n))
 _extrapoints(s::Sphere, ::Val{3}, sz) = (s(0, 0), s(1, 0))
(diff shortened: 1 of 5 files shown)


src/sampling/regular.jl
```diff
@@ -53,6 +53,13 @@ firstoffset(b::Ball) = (n -> inv(n), firstoffset(boundary(b))...)
 lastoffset(b::Ball) = (n -> zero(n), lastoffset(boundary(b))...)
 extrapoints(b::Ball, sz) = (center(b),)
 
+function extrapoints(b::Ball{𝔼{3}}, sz)
+  T = numtype(lentype(b))
+  n = sz[1]
+  rs = range(inv(T(n)), stop=one(T), length=n)
+  (center(b), (b(r, 0, 0) for r in rs)..., (b(r, 1, 0) for r in rs)...)
+end
```
juliohm: I will review the PR more carefully tomorrow, but I think we only need one extra point at the center to discretize the 3D ball. You can think of the discretization as concentric shells forming a grid of hexahedra plus the center point  connected to the first shell with tetrahedra.

Also, try to check the `firstoffset` and `lastoffset` machinery to make sure that we are following it as much as possible. I believe the first dimension (radius) will require a `firstoffset` to insert the extra point  at the center and the other dimensions will have zero offset like the Sphere embedded in 3D.

jitendravjh:
Tried the centre-only version. It is simpler, two element types instead of four, but zero offset on the polar angle puts each pole in the grid once per azimuth, so at 2x50x50 it gives 147 duplicate vertices and 150 degenerate hexahedra along the axis, against 0 and 0 here. Same situation as the cylinder axis, so I kept the axis points and wedges, following `extrapoints(c::Cylinder, sz)` and `_appendaxis`.

On the offsets, `Ball{𝔼{3}}` already inherits (1/n, 1/(n+1), 0) from the boundary sphere and those are unchanged. The radial range inside `extrapoints` was hardcoded, that now derives from them.

juliohm:
> 150 degenerate hexahedra along the axis

Notice that our `GridTopology` was constructed in such a way that you can specify periodic dimensions. We shouldn't have degenerate hexahedra in the final mesh if the parameters are set correctly. Are you already setting the periodicity of the grid dimensions?

jitendravjh:
Yes, `(false, false, true)`, unchanged. This PR has 0 duplicates and 0 degenerate elements, the 147 and 150 were from the centre-only variant.

Periodicity does not reach the poles though, marking the polar angle periodic would glue north to south.

juliohm:
This PR is a masterpiece. Everything looks great 👏🏽 

Could you please confirm that the order of vertices in the elements follows the VTK convention?

image

We adopted that convention in the code to avoid subtle bugs related to face orientation.

jitendravjh:
Confirmed, and this still holds on your latest commits: 0 wrongly oriented faces across all sizes. Deliberately flipped hexahedra and tetrahedra fail the same check.

juliohm (approved):

PR #1417, opened 2026-08-22
Implement discretize for Frustum

jitendravjh:
Part of #1410.

Added `isparametrized`, `isperiodic`, the sampling offsets and the `discretize` method, following `Cylinder`. There is no apex, so `_appendaxis` closes the axis unchanged.

Volume is exactly independent of the radial and axial counts, 7.31110529 at (1,50,1), (2,50,2), (4,50,4) and (8,50,8). Angular refinement converges on πh/3 (R² + Rr + r²), -0.263% at 50 and -0.0164% at 200.


juliohm (approved):
🚀 🎉

PR #1419, opened 2026-08-23
Fix bridge between rings that share a vertex

### LINKED ISSUE #629: `Bridge` returns `NaN`s for `PolyArea` with intersecting `Ring`s

jitendravjh:
Closes #629.

When the outer ring and a hole share a vertex the segment A--B has zero length, so the offset becomes Inf times zero and the mesh comes out with NaN vertices. The bridge is now oriented towards the hole, and since it has zero length a single pair of split points is enough, otherwise the coincident points come back as zero area triangles.

Also tried the Repair(10) route from the thread. It clears the NaN but moves the outer ring by 10atol, so the original vertices no longer appear in the mesh. This way every input point survives.

Checked against measure(poly) over a sweep of hole sizes, corners and scales, 64 cases, no NaN and no area mismatch. One zero area triangle is left at the pinch, that vertex has to appear twice in the bridged ring.

### DIFF
--- a/src/transforms/bridge.jl
+++ b/src/transforms/bridge.jl
@@ -77,8 +77,12 @@ function bridge(rings, rinds, δ)
     A = outer[omax]
     B = inner[imax]
 
+    # the rings may share the vertex, in which case the segment A--B
+    # degenerates and the bridge is oriented towards the hole instead
+    shared = A == B
+
     # direction and normal to segment A--B
-    v = B - A
+    v = shared ? (coordmean(inner) - A) : (B - A)
     u = Vec(-v[2], v[1])
     n = norm(u)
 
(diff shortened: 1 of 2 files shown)


juliohm (approved):
Ninja skills! 🥷 👏🏽

PR #1421, opened 2026-08-24
Fix cartesianrange for geodesic grids

### LINKED ISSUE #1304: Review `cartesianrange` for `Grid` and `Box` in non-Euclidean manifold

jitendravjh:
Closes #1304.

`_geodesicrange` was reading the first grid dimension as longitude, but `LatLon` grids are ordered (lat, lon), so every query returned an empty range. On main this makes `Slice` throw for any geodesic grid.

Also aligned the boundary handling with `_euclideanrange`, which keeps elements that touch the box.

Checked against the Euclidean range on the same grid in (lat, lon) space, 1100 random boxes across regular and rectilinear grids including the decreasing axis case, no mismatches. Added tests, this path had none.

### DIFF
--- a/src/indices.jl
+++ b/src/indices.jl
@@ -151,64 +151,60 @@ function _euclideanrange(grid::OrthoRectilinearGrid, box::Box)
 end
 
 function _geodesicrange(grid::Grid, box::Box)
-  nlon, nlat = vsize(grid)
+  n₁, n₂ = vsize(grid)
+
+  # the latitude may be stored in the first or in the second axis
+  v₁ = convert(LatLon, coords(vertex(grid, (1, 1))))
+  v₂ = convert(LatLon, coords(vertex(grid, (2, 1))))
+  latfirst = abs(v₂.lat - v₁.lat) > abs(v₂.lon - v₁.lon)
+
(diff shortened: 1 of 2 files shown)


juliohm:
Thank you @jitendravjh ! Tthe proposed changes broke the last example in the following tutorial:

https://juliaearth.github.io/GeoStatsDocs/stable/faq/rasters

It is not clear to me yet why this is happening.

juliohm:
I believe it has to do with the fact that GeoIO.jl loads GeoTIFF images preserving the "lon-lat" order of the file. In that case, the corresponding grid has longitude in the first axis. Perhaps we just need to flip the `nlat, nlon = size(grid)` in the new implementation to preserve the axis order?

jitendravjh:
The raster has longitude in the first axis, but a `RegularGrid` from `LatLon` points has latitude there, so flipping would break the other case. It now detects which axis carries the latitude. Tutorial slice gives (3601, 3601), added a test for the longitude-first case.

juliohm (approved):
Another amazing fix! 🫶

Issue #646, opened 2023-12-05
`Ngon`-`Segment` intersection in 2D


tomerarnon:
```julia
julia> r = Rope(Point(1, 0), Point(1, 1), Point(2, 1));

julia> q = Quadrangle(Point(0,0), Point(0, 1), Point(1, 1), Point(1, 0));

julia> intersection(q, r)
ERROR: StackOverflowError:
Stacktrace:
 [1] intersection(f::Function, g₁::Quadrangle{2, Float64}, g₂::Rope{2, Float64, Vector{Point2}})
   @ Meshes ~/.julia/packages/Meshes/w8yhV/src/intersections.jl:100
 [2] intersection(f::Function, g₁::Rope{2, Float64, Vector{Point2}}, g₂::Quadrangle{2, Float64})
   @ Meshes ~/.julia/packages/Meshes/w8yhV/src/intersections.jl:100

julia> b = Box(Point(0,0), Point(1,1));

julia> intersection(b, r)
ERROR: StackOverflowError:
Stacktrace:
 [1] intersection(f::Function, g₁::Box{2, Float64}, g₂::Rope{2, Float64, Vector{Point2}})
   @ Meshes ~/.julia/packages/Meshes/w8yhV/src/intersections.jl:100
 [2] intersection(f::Function, g₁::Rope{2, Float64, Vector{Point2}}, g₂::Box{2, Float64})
   @ Meshes ~/.julia/packages/Meshes/w8yhV/src/intersections.jl:100
```

Related to #644. In my application, I would like to use a `Rope` to represent a "path", and I would like to find intersections between that path and a number of geometries. At the moment, those are segments, ngons (probably only rectangles), and other ropes. 

I guess the stackoverlow is a red herring for any missing method: https://github.com/JuliaGeometry/Meshes.jl/issues/246

tomerarnon:
Intersections of `Box`/`Ngon` with `Segment` is also missing (I wanted to see whether I could process the `Rope` as a series of segments).

juliohm:
In the 2D case we need a new clipping algorithm between `Segment` and `Ngon`. We can then use this algorithm in the method with a `Rope` and a `Ngon`. The `Box` case should fallback to the `Quadrangle` case with `convert(Quadrangle, box)`.

I am busy with other projects at the moment, and won't have the time to add these implementations in the following weeks. Feel free to propose algorithms as PRs for review.

juliohm:
[...]

PR #1422, opened 2026-08-24
Add `Chain`-`Polygon` intersection method

phchavesmaia:
This PR is a continuation of PR #1397 and addresses to Issue #646 by implementing the `Chain`-`Polygon` intersection method. The implementation builds heavily on the previously introduced `Segment`-`Polygon` intersection method, applying it to each segment that composes the chain and combining the resulting intersections.

Importantly, this method does not apply any additional pruning strategy to avoid duplicating effort, as the `Segment`-`Polygon` method already performs pruning by checking for bounding-box intersections.

**Result**: For my simple tests and applications, the method is working just fine, but I have not stress tested it.

### DIFF
--- a/src/intersections/chains.jl
+++ b/src/intersections/chains.jl
@@ -11,11 +11,16 @@ intersection(f, chain₁::Chain, chain₂::Chain) =
 # 3. overlaps at the polygon edges (EdgeTouching)
 # 4. no intersection (NotIntersecting)
 function intersection(f, chain::Chain, poly::Polygon)
-  # assess intersections 
-  pieces = Geometry{manifold(chain), crs(chain)}[]
-  hasintersection = false
+  # retrieve manifold and crs
+  M = manifold(chain)
+  C = crs(chain)
+
+  # collect intersection pieces and
+  # classify final intersection type
   onlytouching = true
   onlyedgetouching = true
+  hasintersection = false
+  pieces = Geometry{M,C}[]
   for seg in segments(chain)
     I = intersection(seg, poly)
     type(I) == NotIntersecting && continue
@@ -29,11 +34,10 @@ function intersection(f, chain::Chain, poly::Polygon)
       push!(pieces, geom)
     end
   end
-
   unique!(pieces)
+
   glue(pieces) = length(pieces) == 1 ? only(pieces) : Multi(pieces)
 
-  # classifying intersections
   if !hasintersection
     return @IT NotIntersecting nothing f
   elseif onlytouching
--- a/src/intersections/chains.jl
+++ b/src/intersections/chains.jl
@@ -10,7 +10,7 @@ intersection(f, chain₁::Chain, chain₂::Chain) =
 # 2. intersect at boundary points (Touching)
 # 3. overlaps at the polygon edges (EdgeTouching)
 # 4. no intersection (NotIntersecting)
-function intersection(f, chain::Chain{𝔼{2}}, poly::Polygon{𝔼{2}})
+function intersection(f, chain::Chain, poly::Polygon)
   # assess intersections 
   pieces = Geometry{𝔼{2},crs(chain)}[]
   hasintersection = false
--- a/src/intersections/chains.jl
+++ b/src/intersections/chains.jl
@@ -36,15 +36,13 @@ function intersection(f, chain::Chain, poly::Polygon)
   end
   unique!(pieces)
 
-  glue(pieces) = length(pieces) == 1 ? only(pieces) : Multi(pieces)
-
   if !hasintersection
     return @IT NotIntersecting nothing f
   elseif onlytouching
-    return @IT Touching glue(pieces) f
+    return @IT Touching maybemulti(pieces) f
   elseif onlyedgetouching
-    return @IT EdgeTouching glue(pieces) f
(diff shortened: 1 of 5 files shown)


src/intersections/chains.jl
```diff
@@ -4,3 +4,43 @@
...
+    geom = get(I)
+    if geom isa Multi
+      append!(pieces, parent(geom))
+    else
+      push!(pieces, geom)
+    end
+  end
+
+  unique!(pieces)
+  glue(pieces) = length(pieces) == 1 ? only(pieces) : GeometrySet(pieces)
```
juliohm: I believe we want to return `Multi` instead of `GeometrySet`. The first is a geometry type and the second is a domain type (i.e., iterable collection of geometries).

src/intersections/chains.jl
```diff
@@ -4,3 +4,43 @@
...
   intersection(f, GeometrySet(segments(chain₁)), GeometrySet(segments(chain₂)))
+
+# intersection between a chain and a polygon
+# 1. intersect at inner points (Intersecting)
+# 2. intersect at boundary points (Touching)
+# 3. overlaps at the polygon edges (EdgeTouching)
+# 4. no intersection (NotIntersecting)
+function intersection(f, chain::Chain{𝔼{2}}, poly::Polygon{𝔼{2}})
+  # assess intersections 
+  pieces = Geometry[]
```
juliohm: I wonder if we could refactor the algorithm to avoid explicit type initialization. Perhaps that is not an issue in this particular case because we feed `pieces` to `glue` later, which gives the compiler another chance to infer a concrete type.

juliohm (changes requested):
Thank you for the extensive set of tests! It is really helpful! 💯 

Attached two minor comments.

src/intersections/chains.jl
phchavesmaia: The problem I was facing was that I could not construct a `Multi` from a `Geometry[]` vector (stackoverflow). To simplify things I went with a `GeometrySet`, but, indeed, a more proper approach is to add the manifold and crs information to the `Geometry` array, hence allowing the `Multi` return.

src/intersections/chains.jl
phchavesmaia: The challenge here is that there can be two geometry types in `pieces` this time around: `Point`s and `Segment`s. Thus, I was having a hard time to think about a concrete type to initialize the vector and went with the explicit type initialization. Maybe `glue` is indeed fixing this up.

src/intersections/chains.jl
```diff
@@ -4,3 +4,43 @@
 
 intersection(f, chain₁::Chain, chain₂::Chain) =
   intersection(f, GeometrySet(segments(chain₁)), GeometrySet(segments(chain₂)))
+
+# intersection between a chain and a polygon
+# 1. intersect at inner points (Intersecting)
+# 2. intersect at boundary points (Touching)
+# 3. overlaps at the polygon edges (EdgeTouching)
+# 4. no intersection (NotIntersecting)
+function intersection(f, chain::Chain{𝔼{2}}, poly::Polygon{𝔼{2}})
```
juliohm:
```suggestion
function intersection(f, chain::Chain, poly::Polygon)
```

I believe there is nothing about this algorithm that is specific to E{2}. I know the underlying method is only implemented for E{2}, but if someone implements more methods in the future, things should work automatically without the restriction.

src/intersections/chains.jl
```diff
@@ -4,3 +4,43 @@
...
+  # assess intersections 
+  pieces = Geometry{𝔼{2},crs(chain)}[]
```
juliohm: Here you can replace E{2} with `manifold(chain)` to make it more generic.

juliohm:
I've refactored some old implementations to always return a valid `Geometry` and have introduced a `maybemulti` utility to use in different methods.

The method for Chain-Chain calls the fallback for Domain-Domain, which treats the arguments as collections of geometries. We can  always `discretize` the arguments of the intersection and send the discretizations to this Domain-Domain fallback. However, sometimes we need to know the specific type of intersection in applications, and that is when method specialization comes in.

In this PR we are adding the Chain-Polygon specialization because it has more type information than the generic fallback. That is something that we can assess more carefully in the future, i.e., when it is important to know the type of intersection, and when the dichotomy `Intersecting` vs. `NonIntersecting` from the Domain-Domain method is enough.

I will reflect more about it over night, but the PR is probably ready to be merged. Please feel free to share additional thoughts. The design of intersections is one of the trickiest ones due to type inference challenges.

juliohm:
The tests are amazing. Really good coverage of difficult configurations 💯 

One thing we could do in a separate PR later is try to improve the final intersection result using heuristics in the `maybemulti` utility. The function could inspect consecutive geometries in the list and decide when it makes sense to merge them as we discussed in the other PR for Segment-Chain.

One heuristic could be as simple as checking if the two geometries are `Segment` that satisfy the `iscollinear` predicate and have a common end point. Another heuristic could be checking if a `Segment` is consecutive to a `Point` that coincides with the corresponding end point. These two heuristics would simplify the result considerably, converting `Multi` geometries into single `Segment`.

juliohm (approved):

juliohm:
The error on Windows is random and unrelated to the proposed changes. Merging it.

Issue #438, opened 2023-05-31
Region trees (quadtrees, octtrees)


ErickChacon:
Just to mention that it would be nice to have implementations for Quadtrees as in CGAL.

![image](https://github.com/JuliaGeometry/Meshes.jl/assets/7862458/33cdd8cf-ff32-4b54-91b2-6590318cfb84)

Image taken from [CGAL](https://doc.cgal.org/latest/Orthtree/index.html#title2)

juliohm:
There are packages implementing these data structures in Julia already, we can probably wrap some of them here with a clean wrapper interface: https://github.com/rdeits/RegionTrees.jl

PR #1424, opened 2026-08-27
Add Bounding Volume Hierarchy (BVH) spatial index

phchavesmaia:
This PR introduces a Bounding Volume Hierarchy (BVH) implementation as a `SpatialIndex`. It is the outcome of previous discussions concluding that a generic spatial indexing interface would improve both the performance and the readability of algorithms that rely on broad-phase pruning. In particular, this provides a foundation for accelerating operations in downstream packages such as `GeoTables.jl`.

A BVH is a static binary tree built over the bounding boxes of a domain's elements. It enables efficient broad-phase spatial queries by recursively partitioning the domain and pruning large portions of the search space during traversal. The implementation is generic and works with any `Domain` whose elements support `boundingbox`.

This PR also establishes the infrastructure for additional spatial index implementations in the future, such as R-trees or KD trees, through the common `SpatialIndex` interface. Thinking about further updates, I would prioritize working on the latter first since it should be easier to implement and it may be more efficient for point-in-polygon searches.

### Implementation Details
The BVH struct stores
1. the element bounding boxes,
2. a permutation vector defining the recursive partition,
3. the hierarchy of internal and leaf nodes.

Internal nodes store the bounding box enclosing their two children, while leaves store contiguous ranges of indices into the permutation vector.  Construction uses recursive median splitting along the axis of maximum spread of the element bounding-box centers. The function takes `leafsize=8` as the default, though the optimal value may depend on the workload. Queries perform a stack-based traversal, pruning subtrees whose bounding boxes do not intersect the query bounding box.

`foreachcandidate` exposes the traversal primitive directly, allowing downstream algorithms to process candidates immediately without materializing an intermediate vector of indices. In specific, it was created so that it can latter be implemented in functions such as `geojoin` and a to-be implemented `overlay`.

Most users are not expected to interact directly with the functions provided by this utility, as it primarily serves as infrastructure for spatial algorithms. When direct use is appropriate, the typical workflow is to construct a spatial index with `BVH()` and query it using `candidates()`. 

Although a more generic function name such as `query()` might seem appropriate, I opted for `candidates(query)` because it more clearly conveys the function's purpose: returning the candidate elements for a given query rather than performing the query itself. Moreover, there should be many functions across different packages that use `query()`, so I tried to avoid such a heavily overloaded naming convention.

### Results
I implemented some tests and the algorithm appears to be working as expected. I already started integrating an `overlay` operation for `GeoTables.jl` that uses the `BVH` spatial index utility and its results are matching those of its equivalent Python implementation. In simple benchmarks, the BVH provided significant improvements in both runtime and memory usage compared to a brute-force approach, reducing both to roughly 10% of the original values.


juliohm:
Thank you @phchavesmaia! I will try to find time next week to share a first review. Excited to see this work starting!

src/spatialindex.jl
```diff
@@ -0,0 +1,31 @@
...
+are not guaranteed to intersect `query` exactly.
+"""
+function candidates end
+
+"""
+    candidates!(inds, query, index)
+
+Empty `inds`, store the broad-phase candidate indices for `query`, and return
+`inds`.
+"""
```
juliohm, line 28: I am noticing similarities with our existing `NeighborSearchMethod` API. The main difference seems to be the query object, which can be any geometry in BVH. Do you think it is worth it to generalize the `NeighborSearchMethod` API to any query object, and then add a new implementation for BVH?

src/spatialindex.jl
```diff
@@ -0,0 +1,31 @@
+# ------------------------------------------------------------------
+# Licensed under the MIT License. See LICENSE in the project root.
+# ------------------------------------------------------------------
+
+"""
+    SpatialIndex
+
+Abstract type for data structures that accelerate spatial queries.
+"""
+abstract type SpatialIndex end
```
juliohm, line 10: We need to be precise with our terminology to make sure that we are not overlooking concepts. The docstring refers to `SpatialIndex` and then mentions "spatial queries". The implementation is a bounding volume "hierarchy". Could we come up with a more consistent name for these methods and choose this name as the abstract type?

src/spatialindex.jl
```diff
@@ -0,0 +1,31 @@
...
+"""
+    candidates(query, index)
+
+Return the indices of elements in `index` whose bounding volumes intersect
+the bounding box of `query`.
+
+The returned indices are broad-phase candidates. The corresponding elements
+are not guaranteed to intersect `query` exactly.
+"""
+function candidates end
```
juliohm, line 21: This is also similar to our `indices` API that takes a `Domain` and a `Geometry` and returns the indices of the domain that intersect with the geometry:

https://github.com/JuliaGeometry/Meshes.jl/blob/master/src/indices.jl

Should we try to rethink the BVH as just a wrapper over a `Domain` and implement `indices` methods for it? I am just brainstorming here to make sure we are converging to a good design.

src/spatialindex/bvh.jl
```diff
@@ -0,0 +1,194 @@
+# ------------------------------------------------------------------
+# Licensed under the MIT License. See LICENSE in the project root.
+# ------------------------------------------------------------------
+
+"""
+    BVH(domain; leafsize=8)
```
juliohm, line 6: We try to be more explicit with names so this could be `BoundingVolumeHierarchy` instead of `BVH`.

src/spatialindex/bvh.jl
```diff
@@ -0,0 +1,194 @@
...
+  last::Int
+end
+
+struct BVH{D,B} <: SpatialIndex
+  domain::D
+  boxes::Vector{B}
+  perm::Vector{Int}
+  nodes::Vector{BVHNode{B}}
+  leafsize::Int
+end
```
juliohm, line 32: A question that is worth asking is if this data structure should be tied to Meshes.jl or it is general enough to consume other objects for which a `boundingbox` definition exists. In the latter case, we could consider registering an independent package a la RegionTrees.jl: https://github.com/JuliaGeometry/Meshes.jl/issues/438

Another related question is if it would be possible to come up with a good generic interface to accommodate BVH and region trees side by side as alternative methods. Coming up with these interfaces is challenging, but it  might be worth to facilitate switching indexes in the future.

src/spatialindex/bvh.jl
```diff
@@ -0,0 +1,194 @@
...
+intersection and are **not** guaranteed to satisfy any exact geometric predicate.
+
+This function is allocation-free apart from its internal traversal stack and is intended as the primitive 
+interface for algorithms that process candidates immediately, avoiding the need to materialize an intermediate
+vector of indices.
+
+See also: [`candidates`](@ref), [`candidates!`](@ref).
+"""
+function foreachcandidate(f, query, bvh::BVH)
+  stack = Int[1]
```
juliohm:
```suggestion
  stack = [1]
```

juliohm (commented):
@phchavesmaia left a short preliminary review about the design space.

PR #1425, opened 2026-08-29
Add geometric distances between points

jitendravjh:
Addresses #98, for pairs of points only. The existing `evaluate` methods are untouched.

```julia
EuclideanDistance()(p₁, p₂)
ManhattanDistance()(p₁, p₂)
GeodesicDistance()(p₁, p₂)
```

Dispatch is on the manifold. On `🌐` the Euclidean distance is the chord, and the geodesic is solved with the series of Karney (2013) on the ellipsoid of the datum. No new dependencies.

Verification:

- 500000 cases from Karney's test set, 167289 near antipodal: max error 7.451 nm
- 47 ellipsoids in CoordRefSystems against PROJ `geod_inverse`: worst 3.7 nm
- on a spherical datum it reproduces the great circle formula to 0.2 nm

1036 ns per pair, no allocations.

Karney, C. F. F. 2013. Algorithms for geodesics. https://doi.org/10.1007/s00190-012-0578-z

### DIFF
--- a/src/distances.jl
+++ b/src/distances.jl
@@ -45,24 +45,6 @@ struct EuclideanDistance <: GeometricDistance end
 (::EuclideanDistance)(p₁::Point{M}, p₂::Point{M}) where {M} = norm(p₂ - p₁)
 
 """
-    ManhattanDistance()
-
-Sum of the absolute differences of the coordinates of two points
-in Euclidean space (`𝔼`), also known as the taxicab distance.
-
-## Examples
-
-```julia
-d = ManhattanDistance()
-
-d(Point(0, 0), Point(1, 1))
-```
-"""
-struct ManhattanDistance <: GeometricDistance end
-
-(::ManhattanDistance)(p₁::Point{𝔼{Dim}}, p₂::Point{𝔼{Dim}}) where {Dim} = sum(abs, p₂ - p₁)
-
-"""
     GeodesicDistance()
 
 Length of the shortest path along the manifold of two points.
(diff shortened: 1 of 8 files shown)


src/distances.jl
```diff
@@ -2,6 +2,126 @@
...
+
+## Examples
+
+```julia
+d = ManhattanDistance()
+
+d(Point(0, 0), Point(1, 1))
+```
+"""
+struct ManhattanDistance <: GeometricDistance end
```
juliohm: I believe we can remove `ManhattanDistance` for now. The Euclidean and geodesic distances are by far the most used ones.

src/distances/geodesic.jl
```diff
@@ -0,0 +1,488 @@
...
+    ε^5 * evalpoly(ε, c₃[5])
+  )
+end
+
+# ----------------
+# INVERSE PROBLEM
+# ----------------
+
+# constants of the ellipsoid that are reused across the iteration
+struct GeodesicEllipsoid{T}
```
juliohm: Could we reuse the ellipsoid of revolution from CoordRefSystems.jl? Every point has a `ellipsoid(datum(crs(point)))`.

If we could rely on the CoordRefSystems.jl interface, that would be perfect. If a utility is missing in the public API, we could consider adding it. Perhaps some of the functions in this PR are already available in CoordRefSystems.jl.

juliohm (changes requested):
Thank you @jitendravjh ! Another amazing contribution! ❤️

jitendravjh:
Removed `ManhattanDistance`. The geodesic now takes the ellipsoid of revolution directly and reads `majoraxis`, `minoraxis`, `flattening` and `eccentricity²` off it.

Karney needs two more quantities, the second eccentricity squared and the third flattening. Both are one liners so I kept them local for now, happy to move them into CoordRefSystems.jl if you would rather have them there. I also went through `utils.jl` to see if anything else here was already available, did not find an overlap.

One side effect worth mentioning: `ellipfromaf⁻¹` throws on a negative flattening, so a prolate ellipsoid cannot be built and those branches were unreachable. I dropped them since there was no way to test them.

Numbers did not change, still 7.451 nm max on Karney's test set and no allocations.

juliohm (approved):
So nice to see this contribution! 🎉 ❤️ 

I am trusting your knowledge on this 💯 

Will just ask Copilot to see if it has more suggestions.

PR #1426, opened 2026-09-02
Add the direct and inverse geodesic problems

jitendravjh:
Follows up #1425 with the two geodesic problems.

```julia
geodesicfwd(p, ϕ, l)    # walk l along azimuth ϕ, return the point
geodesicbwd(p₁, p₂)     # azimuth at p₁ towards p₂
```

`geodesicbwd` returns one azimuth. The length is `GeodesicDistance` and the far end is `geodesicbwd(p₂, p₁)`. `geodesicfwd` keeps the CRS of its input.

The direct problem needed one new series, Karney's reverted C₁′. It is not iterative, 0.34 µs against 1.0 for the distance.

Over all 500000 rows of Karney's test set:

- direct position max 1.27e-8 m, azimuth max 1.75e-8 deg
- round trip back to p₂ max 8.5e-9 m
- distance bit identical to master
- 47 ellipsoids against PROJ `geod_direct`, worst 3.4 nm

The inverse azimuth is ill conditioned near antipodal points, only good to about 0.02 deg there. PROJ has the same error on the same rows, so those tests use a looser tolerance than the distance ones.

I kept both in `src/distances/` with the shared numerics, happy to move them.


docs/src/geodesics.md
```diff
@@ -0,0 +1,48 @@
...
+geodesicfwd
+geodesicbwd
+```
+
+The direct problem walks a given length along a given azimuth:
+
+```@example geodesics
+p = Point(LatLon(-33.8688, 151.2093))
+
+geodesicfwd(p, 45u"°", 1000u"km")
```
juliohm: Could you please add a method that takes unitless arguments and defaults to degree and meter units? i.e., `geodesicfwd(p, 45, 100)` will automatically add the units. That way the documentation will build correctly, without needing an explicit import of Unitful.jl

test/distances.jl
```diff
@@ -160,3 +160,101 @@ end
...
+      # inverse problem: the azimuth at each end
+      @test isapprox(geodesicbwd(p₁, p₂), azi₁ * u"°", atol=τϕ)
+      @test isapprox(geodesicbwd(p₂, p₁), (azi₂ - 180) * u"°", atol=τϕ)
+      # direct problem: the point reached and the distance to it
+      @test isapprox(geodesicfwd(p₁, azi₁, s₁₂), p₂, atol=τ)
+      @test isapprox(GeodesicDistance()(p₁, geodesicfwd(p₁, azi₁, s₁₂)), s₁₂ * u"m", atol=1e-7u"m")
+    end
+
+    # on a datum with a spherical ellipsoid the azimuth is the great circle one
+    S = CoordRefSystems.GRS80S
```
juliohm:
```suggestion
    S = GRS80S
```

juliohm (commented):
Two very minor suggestions 👏🏽 🚀 🥇

docs/src/geodesics.md
juliohm: Regarding this suggestion, we have `aslen` and `asdeg` utilities that can be used to convert any argument to unitful quantities.

jitendravjh:
> Regarding this suggestion, we have `aslen` and `asdeg` utilities that can be used to convert any argument to unitful quantities.

It already takes unitless args, `asdeg` and `aslen` were in there from the start. The docs example was the problem, switched it to plain numbers. GRS80S applied.

docs/src/geodesics.md
jitendravjh: Already does, `asdeg` and `aslen` were there from the start. Example switched to plain numbers.

test/distances.jl
jitendravjh: Done.

juliohm (approved):
❤️ 🚀 💯

Issue #1408, opened 2026-08-20
Require at least three unique points to construct a `Ring`


phchavesmaia:
I was recently working on an I/O workflow in which I read a `.geojson` file and computed the lengths of its geometries. I eventually noticed that some of the results looked suspicious. After comparing them with the results from Python, I confirmed that the Julia results were indeed different and exhibited an upward bias.

While investigating the source of the discrepancy, I noticed that some geometries that Python would characterize as `Segment`-equivalent were represented as `Ring`s in Julia. Consider the following example:

```julia
using Meshes

a = Ring(Point(0, 0), Point(1, 1))

measure(a)
```

which returns:

```
2.8284271247461903 m
```

whereas the equivalent `Segment` has a measure of:

```
1.4142135623730951 m
```

The factor of two comes from interpreting the two points as a closed ring, effectively measuring the segment in both directions.

To the best of my knowledge, a `Ring` should require at least three points. I don't see a benefit in allowing two-point rings, and doing so can lead to unexpected results such as the measurement discrepancy above. That said, I may be wrong and this is a feature rather than a bug. I would love to hear what you think about it.

In my particular case, the geometry was constructed through `GeoIO.load`, but I believe the underlying "issue" starts with `Ring` itself accepting only two unique points.
_____
PS.: The main issue in my specific application is that the source data is somewhat messed up. The source data does imply a `Ring` with 2 unique points and I already found out other strange geometries in there. Regardless, I think the discussion remains valid.

juliohm:
I'm far from my computer, but this might be an issue with the geojson
backend in GeoIO.jl. If the Chain is open it should be loaded as a Rope as
opposed to a Ring.

Em qui., 20 de ago. de 2026, 09:50, Pedro H. Chaves Maia <
***@***.***> escreveu:

phchavesmaia:
In my specific case, the original data contain quite a few inconsistencies and do, indeed, imply `Ring(A,B,A)`; hence, I believe `GeoIO.jl` is working as intended. Among these inconsistencies, for example, the data also imply `Segment(A,B,C,B,D)`, which is unusual, to say the least.

Regardless, I think this issue ultimately comes down to a matter of definition: whether allowing `A-B-A` rings is an intended feature or should be considered a bug.

juliohm:
I believe we allowed `Ring`s with just two points to handle these inconsistencies from industry. Sometimes people store "degenerate holes" inside polygons with two points, for example.

[...]

PR #1427, opened 2026-09-03
Adding `Repair(13)` due to unexpected behavior of `Repair(0)` in `Polytope`s

phchavesmaia:
This PR follows up on Issue #1408.

### Example

Consider the following `Rope` (see the corresponding plot in the issue):
```julia
r = Rope([
    Point(632596.1106999982, 7.466356763999997e6),
    Point(632596.9879999992, 7.466353485000003e6),
    Point(632593.2699999996, 7.466351208799998e6),
    Point(632596.9879999992, 7.466353485000003e6),
    Point(632599.0611,       7.4663482532e6),
    Point(632596.9879999992, 7.466353485000003e6),
    Point(632604.8297999997, 7.466354170700002e6),
])
```
Currently, `Repair(0)` removes duplicated vertices directly:
```julia
julia> r |> Repair(0)
Rope
├─ Point(x: 632596.1106999982 m, y: 7.466356763999997e6 m)
├─ Point(x: 632593.2699999996 m, y: 7.466351208799998e6 m)
├─ Point(x: 632599.0611 m, y: 7.4663482532e6 m)
├─ Point(x: 632596.9879999992 m, y: 7.466353485000003e6 m)
└─ Point(x: 632604.8297999997 m, y: 7.466354170700002e6 m)
```
This changes the original geometry by creating connections that were not present in the input.

This PR changes the duplicate-removal logic so that duplicated edges are removed while the valid parts of the original geometry are preserved.

### Discussion

The above behavior results from the current handling of `Polytope`s. Consequently, `Repair(0)` can create new edges not only for `Chain`s, but also for `Polygon`s in general. If we consider this behavior problematic, then the `Repair` algorithm for these geometries should be reconsidered.

Notice that the new algorithm cannot preserve type stability in general (a `Rope` can become a `Multi`, as in the example above). Therefore, overriding the current behavior of `Repair(0)` leads to failures across several parts of Meshes.jl that rely on its existing behavior. For this reason, I propose introducing `Repair(13)` specifically for the removal of duplicated edges.   

### Result

With the proposed changes:
```julia
julia> r |> Repair(13)
Multi(3×Chain)
├─ Rope((x: 6.32596e5 m, y: 7.46636e6 m), (x: 632597.0 m, y: 7.46635e6 m), (x: 6.32593e5 m, y: 7.46635e6 m))
├─ Segment((x: 632597.0 m, y: 7.46635e6 m), (x: 6.32599e5 m, y: 7.46635e6 m))
└─ Segment((x: 632597.0 m, y: 7.46635e6 m), (x: 6.32605e5 m, y: 7.46635e6 m))
```
Hence, the repaired result preserves the geometry encoded by the raw data while removing duplicated edge information.

### DIFF
--- a/src/transforms/repair.jl
+++ b/src/transforms/repair.jl
@@ -297,7 +297,7 @@ function apply(::Repair{13}, g::PolyArea)
 
   outer = first(repaired)
 
-  # Polygon collapsed to a lower-dimensional geometry
+  # polygon collapsed to a lower-dimensional geometry
   outer isa Ring || return outer, nothing
 
   validholes = filter(r -> r isa Ring, repaired[2:end])
--- a/src/transforms/repair.jl
+++ b/src/transforms/repair.jl
@@ -22,6 +22,7 @@ Perform repairing operation with code `K`.
 - K = 10: outer rings of polygon are expanded
 - K = 11: rings of polygon are coherently oriented
 - K = 12: degenerate rings of polygon are removed
+- K = 13: duplicate edges of chains and polygons are removed (no type stability guarantee)
 
 ## Examples
 
@@ -39,97 +40,6 @@ Repair(K) = Repair{K}()
 # --------------
 
 apply(::Repair{0}, geom::Polytope) = unique(geom), nothing
-
-function apply(::Repair{0}, g::Chain)
-  verts = vertices(g)
-
-  # tuple of already seen edges (a, b) where a < b
-  seen = Set{Tuple{eltype(verts),eltype(verts)}}()
-  # already costructed chain parts
-  parts = Vector{eltype(verts)}[]
-  # currently working chain part
-  current = [first(verts)]
-
-  for i in 1:(length(verts) - 1)
-    a = verts[i]
-    b = verts[i + 1]
-
-    # create a key for the edge that is independent of the order of vertices
-    key = isless(a, b) ? (a, b) : (b, a)
-
-    # if the edge has already been seen, we close the current part and start a new one
-    if key in seen
-      length(current) >= 2 && push!(parts, current)
-      current = [b]
-      # if the edge is new, we add it to the seen set and continue building the current part
-    else
-      push!(seen, key)
-      push!(current, b)
-    end
-  end
-
-  # if the last part has at least two vertices, we add it to the parts
-  length(current) >= 2 && push!(parts, current)
-
-  # construct geometries from the parts
-  geoms = map(parts) do verts
-    if length(verts) == 2
-      Segment(verts...)
-    elseif first(verts) == last(verts)
-      Ring(verts[1:(end - 1)])
-    else
-      Rope(verts)
-    end
-  end
-
-  maybemulti(geoms), nothing
-end
-
-function apply(::Repair{0}, g::PolyArea)
-  repaired = map(rings(g)) do ring
-    first(apply(Repair(0), ring))
-  end
-
-  outer = first(repaired)
-
-  # Polygon collapsed to a lower-dimensional geometry
-  outer isa Ring || return outer, nothing
-
-  validholes = filter(r -> r isa Ring, repaired[2:end])
-
-  PolyArea(outer, validholes), nothing
-end
-
-function apply(::Repair{0}, g::Ngon)
-  repaired = first(apply(Repair(0), Ring(vertices(g))))
-
-  repaired isa Ring || return repaired
-
-  Ngon(vertices(repaired)...)
-end
-
-function _flatten!(out, geom)
-    if geom isa Multi
-        for g in parent(geom)
-            _flatten!(out, g)
-        end
-    else
-        push!(out, geom)
-    end
-end
-
-function apply(::Repair{0}, g::Multi)
-  repaired = map(parent(g)) do geom
-    first(apply(Repair(0), geom))
-  end
-
-  flattened = Vector{eltype(repaired)}()
-  for geom in repaired
-    _flatten!(flattened, geom)
-  end
-  
-  maybemulti(unique(flattened)), nothing
-end
 
 function apply(::Repair{0}, mesh::Mesh)
   # retrieve vertices and connectivities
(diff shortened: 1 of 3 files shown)


juliohm:
@phchavesmaia I understand that the ultimate goal is to convert a rope/ring with duplicated segments into a `Multi`-`Segment` with unique entries. Is that understanding accurate? We could define a type-stable `Repair(13)` operation that collects these unique segments, where segment (A, B) and segment (C, D) are unique iff (A==C && B==D) || (A==D && B==C).

phchavesmaia:
> @phchavesmaia I understand that the ultimate goal is to convert a rope/ring with duplicated segments into a `Multi`-`Segment` with unique entries. Is that understanding accurate? We could define a type-stable `Repair(13)` operation that collects these unique segments, where segment (A, B) and segment (C, D) are unique iff (A==C && B==D) || (A==D && B==C).

Yes, @juliohm , that's pretty much the idea here. In fact, the algorithm for the `Chain` case does exactly it, whereas the other methods are mostly an application of the first one.

The thing about type-stability comes to more edge cases. A `Ring` with two vertices, for example, is reduced to a `Segment`. We could keep it as a `Ring`, but I think that would defeat the purpose.

This `Repair` option, more broadly, remove duplicated segments so that a misspecified geometry can be either expanded or reduced. A`Multi` with overlapping geometries, for example, can become a `Rope` as well.

juliohm:
> The thing about type-stability comes to more edge cases. A `Ring` with two vertices, for example, is reduced to a `Segment`. We could keep it as a `Ring`, but I think that would defeat the purpose.

The `Ring` with two vertices becoming a single `Segment` is the expected result of the operation that we are defining, right? We could approach this problem by first improving the `maybemulti` utility with additional heuristics to glue segments together. The type instability could be handled over there. What do you think? Adding these heuristics to `maybemulti` would also improve some of the recent intersection results.

We need to be careful to add heuristics that are general enough. Different applications might require different "glueing" steps, and the `Repair` and intersection routines should not be application-dependent.

phchavesmaia:
I have been thinking about this, but before we start writing the implementation, I think it is important to agree on the expected behavior of these functions. To me, the following separation of responsibilities makes intuitive sense:

1. `Repair(13)` collects the unique segments of a geometry, i.e., deduplicates its segments, and returns them as a `Multi` (thus preserving type stability);
2. `glue` is a new function that glues a collection of segments according to a parsimonious, application-agnostic heuristic; and
3. `maybemulti` determines whether the resulting collection should be returned as a single geometry or as a `Multi`.

At first, `glue` would only need to operate on segments and reconstruct `Chain`s, since more complex geometries can later be built from these primitives. With this separation, `Repair(13)` remains a relatively simple and type-stable operation, while the topology-dependent reconstruction is handled explicitly by `glue`.

For `glue`, I think we can interpret the collection of segments as an undirected graph. Each vertex has an adjacency degree, allowing us to classify it as a terminal vertex (degree 1), an internal vertex (degree 2), or a branching vertex (degree > 2). We can then construct maximal chains by starting from each unvisited non-internal vertex and following adjacent segments through degree-2 vertices until another terminal or branching vertex is reached. At branching vertices, the chain is terminated rather than trying to infer which outgoing segments should be connected, since such a choice would introduce application-specific assumptions.

After all paths involving non-internal vertices have been visited, any remaining segments must belong to components in which every vertex has degree 2. These components can then be traversed separately and reconstructed as `Ring`s.

One issue is determinism. I don't think the same collection of segments should produce different geometries simply because the input segments were provided in a different order. We can avoid this by assigning each segment a canonical key based on its ordered endpoints and sorting the segments before constructing the adjacency dictionary. We should also process the non-internal starting vertices in a canonical order. This gives us deterministic choices for both the order and orientation of the resulting chains without introducing application-specific gluing rules.

The resulting algorithm therefore has roughly three phases: sorting the segments and building the adjacency dictionary, traversing maximal paths from non-internal vertices, and finally traversing the remaining cyclic components. Although this requires multiple passes over the segments, I don't think this is necessarily problematic: after sorting, the graph construction and traversal themselves should be close to linear, with each segment visited only a small number of times.

This also keeps the intended workflow quite explicit:

`Chain |> Repair(13) |> glue |> maybemulti`

Importantly, `glue` would assume that duplicate segments have already been removed. Otherwise duplicated edges would artificially change the adjacency degree of their endpoints and interfere with the graph-based heuristic. 

It is also important to note that I expect both `glue` and `maybemulti` to be type-unstable in the strict sense. `maybemulti` always returns a concrete geometry at runtime, but its return type depends on the contents of the input collection: it may return either its single element or a `Multi`. Therefore, the compiler cannot generally determine its return type in advance, causing it to fail `@inferred`.

`glue` has a slightly different issue. A single call may produce different kinds of `Chain`s simultaneously so its natural output would be an abstract `Chain{M,C}[]` collection. In this case, the heterogeneity is intrinsic to the operation rather than simply a consequence of choosing between alternative concrete return types.

What do you think about this strategy?

phchavesmaia:
I wrote a version following the principles stated in the previous comment so we can have something more concrete to discuss about. I also tried solving my constant formatting problems by passing the "format" function on the whole project, but it was to no avail.

JoshuaLampert:
> I also tried solving my constant formatting problems by passing the "format" function on the whole project, but it was to no avail.

For the formatting make sure you are using JuliaFormatter.jl v1.0.62 instead of the latest version since this is the one used by Meshes.jl. Especially, do not format using VS Code since it uses a newer JuliaFormatter.jl version, which formats differently.
We should maybe move to JuliaFormatter.jl v2, especially since VS Code uses this nowadays.
Edit: I created #1430 to switch to v2.

phchavesmaia:
> > I also tried solving my constant formatting problems by passing the "format" function on the whole project, but it was to no avail.
> 
> For the formatting make sure you are using JuliaFormatter.jl v1.0.62 instead of the latest version since this is the one used by Meshes.jl. Especially, do not format using VS Code since it uses a newer JuliaFormatter.jl version, which formats differently. We should maybe move to JuliaFormatter.jl v2, especially since VS Code uses this nowadays. Edit: I created #1430 to switch to v2.

Many thanks, @JoshuaLampert !

juliohm:
Hi @phchavesmaia , sorry for the delayed response. I am switching between different projects and finally got some time to read your answer in detail.

I really like the proposed encapsulation with three clear operations `Repair(13)`, `glue` and `maybemulti`. If you could prepare a separate PR for each operation, that would really help 🙂 

Maybe we could start with `maybemulti`? What exactly do we need to change to achieve the ultimate goal?

phchavesmaia:
Great! I think `maybemulti` is in a good place now, and I don't see a need for any further changes. It does what it's supposed to do.

I'll open a PR with the glue proposal.

PR #1428, opened 2026-09-05
Add conversion between azimuths and tangent vectors

jitendravjh:
Follow up to #1426, adds the conversion between azimuths and tangent vectors.

```julia
geodesictangent(p, ϕ)   # unit vector tangent at p along azimuth ϕ
geodesicazimuth(p, v)   # azimuth of v at p
```

This is just the local east/north frame at `p`, so there is no new numerics here. The vector comes back with norm 1 m, the same as `normal` does, and any part of `v` along the normal of the ellipsoid gets ignored.

I checked the tangent against the exact derivative of `geodesicfwd` using autodiff, worst case 2.3e-14 deg, and the round trip is 2.8e-14 deg over 51840 cases. No allocations.

One thing to know, at the poles the frame depends on the stored longitude, so ϕ is relative to that there.

### DIFF
--- a/docs/src/geodesics.md
+++ b/docs/src/geodesics.md
@@ -13,6 +13,8 @@ ellipsoid attached to the datum of the coordinates.
 ```@docs
 geodesicfwd
 geodesicbwd
+geodesictangent
+geodesicazimuth
 ```
 
 The direct problem walks a given length along a given azimuth. Plain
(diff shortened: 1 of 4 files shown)


juliohm (approved):
I lack words to describe how good and clean these contributions are. Thank you! ❤️

PR #1429, opened 2026-09-06
Add Geodesic primitive

juliohm:
So that we can treat geodesics as first-class objects.

For example:

```julia
using Meshes
using CoordRefSystems
import GLMakie

s = Segment(Point(LatLon(0,0)), Point(LatLon(45,90)))
g = Geodesic(Point(LatLon(0,0)), Point(LatLon(45,90)))

viz(s, color="teal")
viz!(g, color="salmon")
```
image

### DIFF
--- a/src/geometries/primitives/geodesic.jl
+++ b/src/geometries/primitives/geodesic.jl
@@ -41,7 +41,7 @@ Base.extrema(g::Geodesic) = g.a, g.b
 Base.isapprox(g₁::Geodesic, g₂::Geodesic; atol=atol(lentype(g₁)), kwargs...) =
   isapprox(g₁.a, g₂.a; atol=atol, kwargs...) && isapprox(g₁.b, g₂.b; atol=atol, kwargs...)
 
-(g::Geodesic{𝔼})(t) = g.a + (t * g.length) * g.dirvec
+(g::Geodesic{<:𝔼})(t) = g.a + t * (g.b - g.a)
 
 (g::Geodesic{🌐})(t) = geodesicfwd(g.a, geodesicazimuth(g.a, g.dirvec), t * g.length)
 
@@ -66,5 +66,5 @@ end
 # HELPER FUNCTIONS
 # -----------------
 
-_geodesicdirection(a::Point{𝔼}, b::Point{𝔼}) = unormalize(b - a)
+_geodesicdirection(a::Point{<:𝔼}, b::Point{<:𝔼}) = unormalize(b - a)
 _geodesicdirection(a::Point{🌐}, b::Point{🌐}) = geodesictangent(a, geodesicbwd(a, b))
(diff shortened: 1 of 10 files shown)


src/geometries/primitives/geodesic.jl
```diff
@@ -0,0 +1,70 @@
...
+Base.maximum(g::Geodesic) = g.b
+
+Base.extrema(g::Geodesic) = g.a, g.b
+
+==(g₁::Geodesic, g₂::Geodesic) = g₁.a == g₂.a && g₁.b == g₂.b
+
+Base.isapprox(g₁::Geodesic, g₂::Geodesic; atol=atol(lentype(g₁)), kwargs...) =
+  isapprox(g₁.a, g₂.a; atol=atol, kwargs...) && isapprox(g₁.b, g₂.b; atol=atol, kwargs...)
+
+(g::Geodesic{𝔼})(t) = g.a + (t * g.length) * g.dirvec
```
jitendravjh: `Geodesic{𝔼}` never matches, 𝔼 is a UnionAll. And after fixing that it still errors, dirvec already carries metres so `(t * g.length) * g.dirvec` comes out m².

```suggestion
(g::Geodesic{<:𝔼})(t) = g.a + t * (g.b - g.a)
```

src/geometries/primitives/geodesic.jl
```diff
@@ -0,0 +1,70 @@
...
+function Base.show(io::IO, ::MIME"text/plain", g::Geodesic)
+  summary(io, g)
+  printfields(io, g, (:a, :b))
+end
+
+# -----------------
+# HELPER FUNCTIONS
+# -----------------
+
+_geodesicdirection(a::Point{𝔼}, b::Point{𝔼}) = unormalize(b - a)
```
jitendravjh: Same thing here, nothing Euclidean constructs without this.

```suggestion
_geodesicdirection(a::Point{<:𝔼}, b::Point{<:𝔼}) = unormalize(b - a)
```

src/geometries/primitives/geodesic.jl
```diff
@@ -0,0 +1,70 @@
...
+function Geodesic(a::Point, b::Point)
+  a′, b′ = promote(a, b)
+  dirvec = _geodesicdirection(a′, b′)
+  length = GeodesicDistance()(a′, b′)
+  Geodesic(a′, b′, dirvec, length)
+end
+
+Geodesic(a::Tuple, b::Tuple) = Geodesic(Point(a), Point(b))
+
+paramdim(::Type{<:Geodesic}) = 1
```
jitendravjh, line 31: measure is missing and you already have the length.

```suggestion
paramdim(::Type{<:Geodesic}) = 1

measure(g::Geodesic) = g.length
```

jitendravjh (commented):
Tried it on the Euclidean side, few suggestions. boundary, reverse and center are missing too, if those were meant to come with the tests.

src/geometries/primitives/geodesic.jl
juliohm: I will add this to measures.jl instead

juliohm:
Close in favor of #1431

PR #1430, opened 2026-09-07
Use v2 of JuliaFormatter.jl

JoshuaLampert:
JuliaFormatter.jl v2 is out already for quite some time and VS Code uses it, too. Therefore it can be annoying to stay with v1. The diff also looks fine to me. What do you think, @juliohm?


juliohm:
Thank you @JoshuaLampert ! The new version is even nicer! It got rid of some unnecessary EOL.

The test failure on Windows is that random failure I think.

JoshuaLampert:
> Thank you @JoshuaLampert ! The new version is even nicer! It got rid of some unnecessary EOL.

Yes, I agree it's nicer.

> The test failure on Windows is that random failure I think.

Yes, it's again due to the bad cache since the storage limit of 10 GB was exceeded. I deleted some of the Windows caches. So hopefully this resolves this after restarting it.

PR #1431, opened 2026-09-07
Fix `Segment` definition on ellipsoid

juliohm:
Now our `Segment{🌐}` is a geodesic, and because of our design, we get geodesic `Rope`s, `Ring`s, `BezierCurve`s, ... for free 🙂  Below is an example with a `Ring` that is made of 3 geodesic segments:

```julia
using Meshes
using CoordRefSystems
import GLMakie

Ring(Point(LatLon(45,0)), Point(LatLon(45,90)), Point(LatLon(90,90))) |> viz
```
image

### DIFF
--- a/src/geodesics.jl
+++ b/src/geodesics.jl
@@ -29,21 +29,28 @@ geodesicfwd(p, 90, 1000000)
 * Karney, C. F. F. 2013. [Algorithms for geodesics](https://doi.org/10.1007/s00190-012-0578-z)
 """
 function geodesicfwd(p::Point{🌐}, ϕ, l)
+  # convert coordinates to LatLon
   c = convert(manifoldcrs(p), coords(p))
-
-  # the ellipsoid comes from the datum of the coordinates
   🌎 = ellipsoid(datum(c))
+  u = unit(majoraxis(🌎))
+
+  # unitful azimuth and length
+  ϕ′ = asdeg(ϕ)
+  l′ = aslen(l)
 
   # the series of Karney need double precision to reach round-off
   T = numtype(lentype(c))
-  S = promote_type(T, Float64)
-  lat, lon = S(ustrip(c.lat)), S(ustrip(c.lon))
-  azi = S(ustrip(u"°", asdeg(ϕ)))
-  len = S(ustrip(unit(majoraxis(🌎)), aslen(l)))
+  U = numtype(typeof(ϕ′))
+  V = numtype(typeof(l′))
+  S = promote_type(T, U, V, Float64)
+  lat = S(ustrip(c.lat))
+  lon = S(ustrip(c.lon))
+  azi = S(ustrip(ϕ′))
+  len = S(ustrip(u, l′))
 
   lat′, lon′, _ = _geodesicdirect(🌎, lat, lon, azi, len)
 
-  withcrs(p, (T(lat′), T(lon′)))
+  withcrs(p, (lat′, lon′))
 end
 
 """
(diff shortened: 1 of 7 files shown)


src/geodesics.jl
```diff
@@ -29,21 +29,28 @@ geodesicfwd(p, 90, 1000000)
...
+  S = promote_type(T, U, V, Float64)
+  lat = S(ustrip(c.lat))
+  lon = S(ustrip(c.lon))
+  azi = S(ustrip(ϕ′))
+  len = S(ustrip(u, l′))
 
   lat′, lon′, _ = _geodesicdirect(🌎, lat, lon, azi, len)
 
-  withcrs(p, (T(lat′), T(lon′)))
+  withcrs(p, (lat′, lon′))
```
juliohm: @jitendravjh I had to remove the `T` coercion at the end in order to make this function differentiable. I understand that we need to operate with `Float64` to get precise results. What would be the impact of removing the `Float64` coercion in the computed geodesic? If users provide coordinates as `Float32` it would be nice to preserve the same type in the output, even if the deviation is large. Users who care about precision will start with `Float64` anyways.

juliohm: Below is an example that works now:

```julia
julia> s = Segment(Point(LatLon(0,0)), Point(LatLon(45,90)))
Segment
├─ Point(lat: 0.0°, lon: 0.0°)
└─ Point(lat: 45.0°, lon: 90.0°)

julia> jacobian(s, (0.0,))
(Vec(0.0 m, 7.090263650863564e6 m, 7.066540628801698e6 m),)
```

jitendravjh: I measured it. Dropping `Float64` costs about 1 m median on `Float32` input, against 0.33 m today, worst case 5.7 m.

You might not have to though. The coercion at the end was what was breaking AD and that is already gone, so `Float64` can stay inside as long as you restore the type on the way out, `R = promote_type(T, U, V)` and then `R(lat′), R(lon′)`. `Float32` still comes out `Float32` and it lands at 0.40 m instead.

jitendravjh:
Couple of things I ran into that are outside the diff.

`geodesictangent` and `geodesicazimuth` still have that same coercion, `geodesics.jl:124` and `:148`, so they blow up on `Dual`s the same way `geodesicfwd` did.

And `jacobian` on an equatorial `Segment` comes back all `NaN`, looks like `hypot(cα₁, sα₁*sβ₁)` with both of them zero.

juliohm:
Thank you @jitendravjh. I've pushed some changes that fix the raised issues. Let's see if tests pass now.

juliohm:
What do you think of the added `promote_type` calls? Do they solve the AD issues you had in mind?

Regarding the Jacobian on the equator, we could propose improvements in a separate PR later.

jitendravjh:
Yes, they do it. `Float32` comes out `Float32` on all four now and the three derivatives that were throwing go through. Accuracy landed at 0.40 m median.

Fine to leave the equator for later.

One thing, Copilot's `center` thread got resolved but the method is still `Segment{<:𝔼}`, so it still throws on 🌐. `center(s::Segment) = s(1 // 2)` gives the same point as `centroid`.

juliohm:
> One thing, Copilot's `center` thread got resolved but the method is still `Segment{<:𝔼}`, so it still throws on 🌐. `center(s::Segment) = s(1 // 2)` gives the same point as `centroid`.

That is intentional because our definition of `center` is for geometries that are axially symmetric like a ball, disk, box, etc. The `centroid` is defined for all manifolds and should be used in the 🌐 case.

jitendravjh:
Ah that makes sense, `centroid` it is. CI is all green now too.

juliohm:
Thank you! 🎉 Another great improvement!

PR #1432, opened 2026-09-09
Add `glue` utility

phchavesmaia:
This PR follows up on #1427 by adding a `glue` utility that takes a vector of segments and glues them into maximal chains according to a canonical segment ordering, ensuring that the result is independent of the input ordering.


src/utils/basic.jl
```diff
@@ -69,3 +69,102 @@ isthreaded(cond=true) = cond && Threads.nthreads() > 1
 Return `only(geoms)` or `Multi(geoms)` depending on the length of `geoms`.
 """
 maybemulti(geoms) = length(geoms) == 1 ? only(geoms) : Multi(geoms)
+
+"""
+    glue(segs)
+Glue unique segments into `Segment`s, `Rope`s and `Ring`s. Segments are sorted.
+"""
+function glue(segs::AbstractVector{<:Segment{M,C}}) where {M,C}
```
juliohm: Before I dive into the review, what is your take on that case where we have `Point`s and `Segment`s as the result of an intersection? Should `glue` also get rid of those points that are vertices of segments in the list?

juliohm:
The format action failed. Did you update the formatter locally? I believe we have the new version in the master branch.

phchavesmaia:
I think the formatting issue has something to do with how I created this branch. The BVH branch, for instance, is now passing the formatting. I made some changes and hopefully it will work now.

src/utils/basic.jl
phchavesmaia: That's a fair point. I added the possibility of `Multi` explicitly containing `Point` objects as well so that they can be removed if they intersect the `Chain`s returned by this `glue` function.

src/utils/basic.jl
```diff
@@ -69,3 +69,119 @@ isthreaded(cond=true) = cond && Threads.nthreads() > 1
...
+
+"""
+    flatten(g)
+
+Return a vector of the geometries contained in `g`. If `g` is a `Multi`, it recursively flattens its parents.
+"""
+function flatten(g)
+  g isa Multi || return [g]
+  mapreduce(flatten, vcat, parent(g))
+end
```
juliohm: Let's use dispatch define the function for `Geometry` and `Multi`. It is more idiomatic than using `isa` in  the function body.

src/utils/basic.jl
```diff
@@ -69,3 +69,119 @@ isthreaded(cond=true) = cond && Threads.nthreads() > 1
 Return `only(geoms)` or `Multi(geoms)` depending on the length of `geoms`.
 """
 maybemulti(geoms) = length(geoms) == 1 ? only(geoms) : Multi(geoms)
+
+"""
+    glue(segs)
+Glue unique segments into `Segment`s, `Rope`s and `Ring`s. Segments are sorted.
```
juliohm, line 75: By reading the implementation I am getting that the return type is always a vector of geometries? It would be nice to document the output format.

src/utils/basic.jl
```diff
@@ -69,3 +69,119 @@ isthreaded(cond=true) = cond && Threads.nthreads() > 1
...
+"""
+function glue(segs::AbstractVector{<:Segment{M,C}}) where {M,C}
```
juliohm: To follow the pattern in other parts of the code, we can remove the `M` and `C` parameters and use `manifold` and `crs` functions in the function body. That way the glue function will work with an heterogeneous list of segments, even if they are represented with different CRS.

src/utils/basic.jl
```diff
@@ -69,3 +69,119 @@ isthreaded(cond=true) = cond && Threads.nthreads() > 1
...
+    push!(chains, Ring(verts[1:(end - 1)]))
+  end
+
+  chains
+end
+
+_segmentkey(seg) = begin
+  a, b = vertices(seg)
+  isless(a, b) ? (a, b) : (b, a)
+end
```
juliohm: This function is only used in one place and its definition is very short. We can make it an anonymous function in the `sort` call.

src/utils/basic.jl
```diff
@@ -69,3 +69,119 @@ isthreaded(cond=true) = cond && Threads.nthreads() > 1
...
+
+    push!(chains, Ring(verts[1:(end - 1)]))
+  end
+
+  chains
+end
+
+_segmentkey(seg) = begin
+  a, b = vertices(seg)
+  isless(a, b) ? (a, b) : (b, a)
```
juliohm:
```suggestion
  a < b ? (a, b) : (b, a)
```

src/utils/basic.jl
```diff
@@ -69,3 +69,119 @@ isthreaded(cond=true) = cond && Threads.nthreads() > 1
...
+  function trace(start, segind)
+    verts = [start]
+    current = start
+    currentind = segind
+
+    while true
+      visited[currentind] = true
+
+      a, b = vertices(segs[currentind])
+      next = current == a ? b : a
```
juliohm: Could you please confirm we need == instead of ≈?

src/utils/basic.jl
```diff
@@ -69,3 +69,119 @@ isthreaded(cond=true) = cond && Threads.nthreads() > 1
...
+      push!(verts, next)
+
+      # stop at terminal or branching vertices
+      length(adj[next]) == 2 || break
+
+      i, j = adj[next]
+      nextind = i == currentind ? j : i
```
juliohm: Probably not a big deal, but if we are planning to use this function with large lists of segments, we could assess if `ifelse` brings more performance. Unlike the ternary operator `cond ? opt1 : opt2`, the function `ifelse(cond, opt1, opt2)` evaluates all arguments before deciding the expression that is returned. It can lead to speedups if the expressions are simple enough.

src/utils/basic.jl
```diff
@@ -69,3 +69,119 @@ isthreaded(cond=true) = cond && Threads.nthreads() > 1
...
+
+      current = next
+      currentind = nextind
+    end
+
+    verts
+  end
+
+  # first trace maximal non-cyclic paths
+  starts = sort([v for (v, inds) in adj if length(inds) != 2])
```
juliohm, line 130: If I am understanding things correctly, each vertex in the original list of segments can belong to an arbitrary number of segments. Is there any assumption in the `adj` list that it should only contain at most 2 indices per vertex? I see the number `2` appearing in a few places, but perhaps we should use a more general condition like `> 1` or `== 1` in these expressions?

juliohm (changes requested):
Very nice code as usual! I could understand it in one go! 💯 

Attached a few comments/suggestions.

src/utils/basic.jl
phchavesmaia: You are correct, a vertex can be in an arbitrary number of segments and, hence, have different adjacency degrees. You are seeing a lot of `2`s because it denotes an "internal vertex", as we defined it here:

> Each vertex has an adjacency degree, allowing us to classify it as a terminal vertex (degree 1), an internal vertex (degree 2), or a branching vertex (degree > 2).

In the first loop, for example, we avoid starting from an internal vertex so that we are either starting from terminal or branching vertexes. This avoids constructing a `Rope` starting from a middle point, for example. 

In the second loop, since we already covered all `Segment`s and `Ropes`, we start from the non-used internal vertices instead. Necessarily, these must be part of a `Ring`, so we build them accordingly.

I think there are other ways to build this logic, but this one seemed the most immediate to me.

src/utils/basic.jl
phchavesmaia: I think we can use ≈ without any problems. I will change it.

src/utils/basic.jl
phchavesmaia: Sure thing! I will change it across the code. I made some benchmarks and it does make the process a little (0.84%) faster.

src/utils/basic.jl
phchavesmaia: Sure! I think this was a function that was originally being used in Repair(13), hence I though about reusing it there. However, I don't think it is necessary anymore since glue will impose a "canonical ordering" of the segments.

src/utils/basic.jl
phchavesmaia: Sure!

src/utils/basic.jl
phchavesmaia: That's correct. I added the output type information to the docstring and a reference to `maybemulti` for users who want to aggregate the result into a single geometry.

src/utils/basic.jl
phchavesmaia: Sure!

phchavesmaia (commented):
It follows the version after the first round of review

src/utils/basic.jl
```diff
@@ -66,6 +66,126 @@ isthreaded(cond=true) = cond && Threads.nthreads() > 1
...
     maybemulti(geoms)
 
-Return `only(geoms)` or `Multi(geoms)` depending on the length of `geoms`.
+Return `only(geoms)` or `Multi(geoms)` depending on the length of the `geoms` vector.
 """
-maybemulti(geoms) = length(geoms) == 1 ? only(geoms) : Multi(geoms)
+maybemulti(geoms::AbstractVector{<:Geometry}) = length(geoms) == 1 ? only(geoms) : Multi(geoms)
+
+"""
+    glue(segs)
```
juliohm, line 74:
```suggestion
    glue(segs)

```

src/utils/basic.jl
```diff
@@ -66,6 +66,126 @@ isthreaded(cond=true) = cond && Threads.nthreads() > 1
...
+  segs = sort(segs; by=seg -> begin
+    a, b = vertices(seg)
+    ifelse(a < b, (a, b), (b, a))
+  end)
+
+  # determine the manifold and crs of the segments
+  M = manifold(eltype(segs))
+  C = crs(eltype(segs))
+
+  # build adjacency dictionary
```
juliohm:
```suggestion
  # build adjacency dictionary
  S = eltype(segs)
  M = manifold(S)
  C = crs(S)
```

juliohm: What about

```julia
s = first(segs)
P = typeof(s(0))
```

for the point type? I believe we can initialize the dictionary with `P` and Julia will convert the points of other segments upon assignment. That way we avoid type inference issues with `eltype(segs)` when `segs` is quite heterogeneous.

src/utils/basic.jl
```diff
@@ -66,6 +66,126 @@ isthreaded(cond=true) = cond && Threads.nthreads() > 1
...
+  adj = Dict{Point{M,C},Vector{Int}}()
+  for (i, seg) in enumerate(segs)
+    a, b = vertices(seg)
+    push!(get!(adj, a, Int[]), i)
+    push!(get!(adj, b, Int[]), i)
+  end
+
+  visited = falses(length(segs))
+  chains = Chain{M,C}[]
```
juliohm: I see that you tend to write algorithms with empty initialization and abstract types. I understand that some algorithms really rely on this code pattern. Is this the case here? Can't we use the `map(iter) do x` syntax to avoid explicit type initialization? If we can let compiler figure out the exact type, that is always the preferred approach. If the algorithm is more inefficient without explicit initialization, then we have a good reason to keep it that way.

src/utils/basic.jl
```diff
@@ -66,6 +66,126 @@ isthreaded(cond=true) = cond && Threads.nthreads() > 1
...
+  # first trace maximal non-cyclic paths
+  starts = sort([v for (v, inds) in adj if length(inds) != 2])
+
+  for start in starts
+    for segind in adj[start]
+      visited[segind] && continue
+
+      verts = trace(start, segind)
+
+      geom = length(verts) == 2 ? Segment(verts...) : Rope(verts)
```
juliohm:
```suggestion
      geom = length(verts) == 2 ? Segment(verts) : Rope(verts)
```

src/utils/basic.jl
```diff
@@ -66,6 +66,126 @@ isthreaded(cond=true) = cond && Threads.nthreads() > 1
...
+"""
+    flatten(g)
+
+Return a vector of the geometries contained in `g`. If `g` is a `Multi`, it recursively flattens its parents.
+The output is a vector containing all the geometries within `g`, with any nested `Multi` geometries recursively flattened.
+
+See [`maybemulti`](@ref) for turning this vector into a single geometry. 
+"""
+flatten(g::Multi) = mapreduce(flatten, vcat, parent(g))
+
```
juliohm:
```suggestion
```

test/utils.jl
```diff
@@ -178,3 +178,183 @@ end
   end
   @test !Meshes.isthreaded(false)
 end
+
+@testitem "glue" setup = [Setup] begin
+  # single segment is preserved
+  a = cart(0, 0)
+  b = cart(1, 0)
+  seg = Segment(a, b)
+
```
juliohm, line 189:
```suggestion
```

test/utils.jl
```diff
@@ -178,3 +178,183 @@ end
...
+  glued = Meshes.glue([seg])
+  @test glued == [seg]
+
+  # connected segments are glued into a rope
+  a = cart(0, 0)
+  b = cart(1, 0)
+  c = cart(2, 0)
+  segs = [Segment(a, b), Segment(b, c)]
+
```
```suggestion
```

test/utils.jl
```diff
@@ -178,3 +178,183 @@ end
...
+  glued = Meshes.glue(segs)
+  @test glued == Chain[Rope(a, b, c)]
+
+  # input ordering does not affect the result
+  glued = Meshes.glue(reverse(segs))
+  @test glued == Chain[Rope(a, b, c)]
+
+  # segment orientation does not affect the result
```
juliohm, line 205: So we usually reserve paragraphs for the same "testset". Each paragraph has a set of inputs and a set of tests. If a set of inputs is used across multiple paragraphs, that is more difficult to track and adjust. If you could make sure that these paragraphs have their own inputs clearly defined (even if copied/pasted), that would be great.

src/utils/basic.jl
juliohm: Just noticed that you also use `M` and `C` to initialize an abstract collection of `Chain{M,C}`.

juliohm (changes requested):
Thank you @phchavesmaia ! Attached another round.

src/utils/basic.jl
phchavesmaia: `verts` is a vector. If we just pass it to `Segment` directly it will fail, hence we must use `verts...` and I pushed back against this commit.

src/utils/basic.jl
phchavesmaia: > What about
> 
> ```julia
> s = first(segs)
> P = typeof(s(0))
> ```
> 
> for the point type? I believe we can initialize the dictionary with `P` and Julia will convert the points of other segments upon assignment. That way we avoid type inference issues with `eltype(segs)` when `segs` is quite heterogeneous.

Oh yes! this completely passed by me. I incorporate your suggestion.

src/utils/basic.jl
phchavesmaia: This tendency is likely a Pythonic mannerism of mine, combined with an OOP mindset. In fact, that's one of the reasons I'm enjoying working on these PRs (and your reviews!) so much: they're helping me, little by little, learn to approach Julia on its own terms.

I will implement your suggestion and compare the results!

src/utils/basic.jl
juliohm: Oh you are right. Segment has a constructor with vector, but it needs to be `SVector`.

src/utils/basic.jl
phchavesmaia: I am struggling to find an effective way to use the `map(iter) do x` approach here.

The main difficulty is that each `start` may generate zero, one, or several geometries because there is another loop over `adj[start]`. A plain `map(starts) do start` therefore does not match the structure of the algorithm very naturally. I can think of two options:

1. Create a temporary vector inside each call to collect the geometries, which only moves the type initialization to another level; or
2. Use a nested `map`, effectively producing a nested collection.

Following the second approach, I need something like `mapreduce(vcat, starts) do start` to flatten the result, and I also need to filter out the `nothing`s produced when `visited[segind] == true`. In my benchmarks, this introduces a noticeable performance penalty.

An alternative is to use comprehensions:

```julia
chains = [
  begin
    verts = trace(start, segind)
    length(verts) == 2 ? Segment(verts...) : Rope(verts)
  end
  for start in starts
  for segind in adj[start]
  if !visited[segind]
]

rings = [
  begin
    a, b = vertices(segs[segind])
    start = ifelse(a < b, a, b)

    verts = trace(start, segind)
    Ring(verts[1:(end - 1)])
  end
  for segind in eachindex(segs)
  if !visited[segind]
]

[chains..., rings...]
```

This avoids the explicit `Chain{M,C}[]` initialization, but at some cost. In a branching structure with 10,000 branch centers (the worst case I tested) the comprehension version is about 5% slower and uses roughly 15% more memory than the current explicit-loop implementation. The number of allocations is almost identical.

The difference seems to grow with the complexity of the resulting geometry and the number of segments. For the other cases I tested (long paths producing `Rope`s, collections of `Ring`s, and disconnected segments) the results are much closer. The branching case is the only one where I would consider the performance loss significant.

Is there another `map`-based formulation that I am overlooking? Otherwise, do you think avoiding the explicit output-type initialization is worth this slight loss in performance, or would the explicit `for`/`push!` version be preferable here?

juliohm: Thank you for diving into these alternatives! I need to read the code again to make sure I am not missing something, but couldn't the algorithm delay `Segment` vs `Rope` vs `Ring` type information to the very end so that these vectors are always vectors of vectors of points? That could save us from this extra logic to find the type to initialize lists?

At the end I am assuming that it should be easy to convert a vector of vector of points into a vector of final geometries based on `length`, `first` and `last` of each chain?

> In a branching structure with 10,000 branch centers (the worst case I tested) the comprehension version is about 5% slower and uses roughly 15% more memory than the current explicit-loop implementation. The number of allocations is almost identical.

That seems like a very extreme case that we won't encounter in practice, or do you see applications where this will happen? Also, I don't understand how 15% more memory is needed and at the same time the number of allocations is almost identical. Could you please clarify?

phchavesmaia: Yes, that is definitely a possibility! I just implemented a version of it and it is working fine. We lose little to no efficiency while avoiding the need to use the abstract type 🚀

Regarding the benchmark between the current version and the comprehension-based one, I completely agree that this would be an extreme case, serving mostly to assess the differences between the coding strategies. For most applications, the difference between the two would probably be imperceptible. However, I believe the strategy you just proposed takes the best of both worlds, allowing us to work with concrete types whenever possible at zero cost.

On the allocations and memory usage discussion, as far as I understand, the number of allocations does not directly determine the amount of memory allocated. A single allocation may correspond to a very small object or to a large vector, for example. Therefore, two implementations can have nearly identical allocation counts while allocating substantially different amounts of memory overall.

test/utils.jl
phchavesmaia: I tried to make it more conformant, though avoiding to rewrite every single point every time. Please, check if it is as expected.

src/utils/basic.jl
```diff
@@ -66,6 +66,125 @@ isthreaded(cond=true) = cond && Threads.nthreads() > 1
...
+  # convert traced vertex paths into appropriate geometries (Ring, Segment, Rope)
+  map(paths) do verts
+    if first(verts) ≈ last(verts)
+      Ring(verts[1:(end - 1)])
+    elseif length(verts) == 2
+      Segment(verts...)
+    else
+      Rope(verts)
+    end
+  end
```
juliohm: That is very clean! 💯

src/utils/basic.jl
```diff
@@ -66,6 +66,125 @@ isthreaded(cond=true) = cond && Threads.nthreads() > 1
...
+
+  # collect all segments from the chains
+  segs = collect(Iterators.flatten(segments(g) for g in chains))
+  glued = glue(segs)
+
+  # if there are no points, we can return the glued geometry immediately
+  isempty(points) && return glued
+
+  # remove points already represented by the glued 1D geometry
+  points = filter(p -> !any(chain -> p ∈ chain, glued), points)
```
juliohm: Perhaps we should only check if the point is a vertex of a chain? In the intersection algorithm the points in the list coincide with vertices of chains. Checking `any(≈(p), eachvertex(chain))` is more efficient and is implemented for all manifolds (𝔼{Dim} and 🌐). On the other hand, checking if a point is anywhere within the segment is a harder problem, which is currently only implemented for straight segments in 𝔼{Dim}. A geodesic (curved) segment makes a non-trivial set of points.

phchavesmaia: Conceptually, I think that checking for the intersection is more aligned with what the `glue` function intends to accomplish. That said, for its current purpuses, I agree with you that checking for the vertices is probably enough.

src/utils/basic.jl
```diff
@@ -66,6 +66,124 @@ isthreaded(cond=true) = cond && Threads.nthreads() > 1
...
+
+Glue unique segments into `Segment`s, `Rope`s and `Ring`s. Segments are sorted.
+"""
+function glue(segs::AbstractVector{<:Segment})
+  length(segs) == 1 && return [only(segs)]
+
+  # sort segments independently of input order
+  segs = sort(segs; by=seg -> begin
+    a, b = vertices(seg)
+    a < b ? (a, b) : (b, a)
```
phchavesmaia, line 85: Why did you push back on using the ` ifelse `   function? If I recall correctly, it indeed led to increased performance.

juliohm: Oh sorry, you benchmarked it? Was it a noticeable speedup?

phchavesmaia: It wasn't anything dramatic. I think the speedup amounted to something like 1%. If you think that using the ternary operator makes it more readable, I think it is a worthwhile change.

juliohm: I switched back to `?` to match other functions, but we could certainly use `ifelse` if it increases the performance. I should have benchmarked it myself before pushing the commits, sorry.

phchavesmaia: Oh, don't worry about it! A 1% difference in runtime in such a fast function is, for most intents and purposes, an insignificant change. Even for consistency's sake it is worth to keep the ternary operator IMO.

juliohm: Fully agree! Let's use `ifelse` with intention in other places where it really makes a difference 💯

juliohm (approved):
🎉 💯 🙂

PR #1434, opened 2026-09-14
Making `Segment`/`Chain`-`Polygon` intersections more topologically robust

phchavesmaia:
This PR follows on PRs #1397, #1422, and #1432 by making the intersections between `Segment`s and `Chain`s with `Polygon`s more topologically robust. It follows a list of the updates:
1. `Segment`-`Polygon` intersections now checks for collinear and adjacent segment intersections, dissolving them into a single segment (i.e. where an output was `Multi([Segment(p1,p2), Segment(p2,p3)])` it is now `Segment(p1,p3)`);
2.  `Chain`-`Polygon` intersection now applies the `glue` function to simplify the final intersection topology (see #1432); and
3. The tests were changed to match these changes.


src/intersections/chains.jl
```diff
@@ -39,10 +39,10 @@ function intersection(f, chain::Chain, poly::Polygon)
...
   elseif onlytouching
-    return @IT Touching maybemulti(pieces) f
+    return @IT Touching maybemulti(glue(Multi(pieces))) f
   elseif onlyedgetouching
-    return @IT EdgeTouching maybemulti(pieces) f
+    return @IT EdgeTouching maybemulti(glue(Multi(pieces))) f
   else
-    return @IT Intersecting maybemulti(pieces) f
+    return @IT Intersecting maybemulti(glue(Multi(pieces))) f
   end
```
juliohm, line 50: Please introduce an intermediate variable, e.g., `geom = maybemulti(glue(Multi(pieces)))`, and reuse it in the if-else branches.

src/intersections/segments.jl
```diff
@@ -247,6 +247,7 @@ function intersection(f, seg::Segment{𝔼{2}}, poly::Polygon{𝔼{2}})
   depth = 0
   inpoly = false
   pieces = typeof(seg)[]
+  haspieces = false
```
juliohm: Was this boolean variable introduced just to check if the `pieces` vector has items? Why not just use `length(pieces) > 0`?

src/intersections/segments.jl
```diff
@@ -257,11 +258,29 @@ function intersection(f, seg::Segment{𝔼{2}}, poly::Polygon{𝔼{2}})
...
+        keep = true
+      else
+        keep = false
+      end
+      # add the current segment piece to the list if it should be kept
+      if keep
+        if haspieces
+          a, b = vertices(last(pieces))
+          c, d = vertices(piece)
+          if b ≈ c
```
juliohm, line 276: Isn't this logic handled by `glue` already? Why do we need to modify the intersection code? Is this referring to the collinearity check?

juliohm (changes requested):
Thank you for working on these improvements!

Attached a first round of review.

src/intersections/segments.jl
phchavesmaia: Precisely. I am taking advantage that all the pieces must be collinear to extend adjacent segments. If I simply applied `glue` here I would get `Rope([a,b,d])` instead of `Segment(a,d)`.

src/intersections/segments.jl
phchavesmaia: Exactly. I thought about checking `isempty(pieces)`, but thought that the boolean would be easier to assess. Maybe that's a ultra minor concern, but I think it would be cheaper to reuse a boolean than evaluate a condition every single time a piece must be pushed.

juliohm: I believe that `isempty` is more readable, and saves the reader from figuring out yet another boolean variable.

src/intersections/segments.jl
juliohm: Shouldn't we use `iscollinear` somewhere to check for collinearity? Is it implicit in this branch of the algorithm?

phchavesmaia: Yes, it is implicit. Any intersections between a segment and a polygon must be a part of the original segment, hence being collinear.

test/intersections.jl
```diff
@@ -704,38 +708,43 @@ end
...
   p = Quadrangle(cart(0, 0), cart(4, 0), cart(4, 4), cart(0, 4))
   @test intersection(r, p) |> type == Intersecting
-  @test r ∩ p ≈ p ∩ r ≈ Multi([Segment(cart(1, 1), cart(2, 2)), Segment(cart(2, 2), cart(3, 1))])
+  @test r ∩ p ≈ p ∩ r ≈ r
 
   # chain entirely inside polygon
   r = Ring([cart(1, 1), cart(3, 1), cart(3, 3), cart(1, 3)])
   p = Quadrangle(cart(0, 0), cart(4, 0), cart(4, 4), cart(0, 4))
   @test intersection(r, p) |> type == Intersecting
+  @test r ∩ p ≈ p ∩ r ≈ Ring([cart(1, 1), cart(1, 3), cart(3, 3), cart(3, 1)])
```
juliohm, line 735: Would it be easy to adjust the `glue` function to make sure that rings are counter-clockwise? Clockwise rings represent holes.

juliohm:
Thank you @phchavesmaia for the quick updates! 🚀 

I left a small comment regarding the orientation of rings returned by `glue`. Also, should we consider using `glue` in the `Segment`-`Polygon` intersection method? It also calls `maybemulti` with `pieces`. Could these pieces be further simplified?

PR #1437, opened 2026-09-16
Adding `approxunique` utility

souma4:
When trying to implement a concavehull algorithm I found a case where if two points are exceptionally close to each other (within atol), they are treated as separate points, but then when we perform the later intersection checks for candidate segments, they get called "intersecting" because they fall within the tolerance. This can lead to invalid hulls up to k == n points and then an error for well specified pointsets. More generally, approxunique can be useful for simplifying any set of points and ensuring correctness of algorithms. 

Currently this uses a `BallSearch` to have O(nlogn) scaling rather than O(n^2). That can be amended if necessary.


src/utils/cmp.jl
```diff
@@ -24,3 +24,27 @@ Round `λ` to `x` if it is within the tolerance `tol`.
...
+
+"""
+    approxunique(points)
+
+Return a collection of points that are unique up to the default tolerance for the type of the points.
+"""
+function approxunique(points)
+  p = collect(points)
+
+  length(p) ≤ 1 && return p
```
juliohm: Is  this needed to avoid errors downstream? Why not build a `BallSearch` with a single point?

src/utils/cmp.jl
```diff
@@ -24,3 +24,27 @@ Round `λ` to `x` if it is within the tolerance `tol`.
...
   isapprox(λ, x, atol=atol) ? x : λ
 end
+
+"""
+    approxunique(points)
+
+Return a collection of points that are unique up to the default tolerance for the type of the points.
+"""
+function approxunique(points)
+  p = collect(points)
```
juliohm: Why do we need to create a local copy? Can't we assume that the function argument is already a `AbstractVector{<:Point}`?

src/utils/cmp.jl
```diff
@@ -24,3 +24,27 @@ Round `λ` to `x` if it is within the tolerance `tol`.
...
+  length(p) ≤ 1 && return p
+
+  # spatial index to avoid comparing all pairs of points
+  searcher = BallSearch(p, MetricBall(atol(lentype(first(p)))))
+
+  keep = trues(length(p))
+  for i in eachindex(p)
+    keep[i] || continue
+    for j in search(p[i], searcher)
+      j > i && isapprox(p[i], p[j]) && (keep[j] = false)
```
juliohm:
```suggestion
      j > i && p[i] ≈ p[j] && (keep[j] = false)
```

test/utils.jl
```diff
@@ -37,6 +37,20 @@ end
...
+  points = [cart(0, 0), cart(1, 1), cart(1, 1), cart(2, 2)]
+  pts = Meshes.approxunique(points)
+  @test length(pts) == 3
+  @test all(p in pts for p in [cart(0, 0), cart(1, 1), cart(2, 2)])
+
+  points = [cart(0, 0), cart(1, 1 - eps(T)), cart(1, 1), cart(2, 2)]
+  box = Box(cart(0, 0), cart(2, 2))
+  pts = Meshes.approxunique(points)
+  @test length(pts) == 3
+  @test all(any(isapprox(i, p) for i in pts) for p in [cart(0, 0), cart(1, 1), cart(2, 2)])
```
```suggestion
  @test all(any(p ≈ p′ for p′ in pts) for p in [cart(0, 0), cart(1, 1), cart(2, 2)])
```

juliohm (changes requested):
Thank you @souma4 ! ❤️  A few minor suggestions attached.

src/utils/cmp.jl
souma4: I guess we don't. In my head I was thinking in terms of "generator or vector" but we could just lean into vector

src/utils/cmp.jl
souma4: In hindsight, I am not sure why I added that. It doesn't cause issues.

juliohm (approved):

PR #1438, opened 2026-09-17
Add `EdgeRefinement` method

### LINKED ISSUE #1118: Add refinement method based on edge predicates
A new `EdgeRefinement(pred)` to speed up the `refine`ment of general `Polygon`al `Mesh`es.

In this method, edges of the mesh are refined if the `pred`icate holds `true`. By setting `pred(e) = length(e) > maxlen()`, we can refine the mesh adaptively for accurate visualization, without exploding the number of elements.

jitendravjh:
Closes #1118.

Splits the edges where the predicate holds and subdivides only the triangles touching them into 2, 3 or 4. Midpoints are shared between adjacent triangles, so no hanging vertices.

Bounding every edge of a 2000 vertex polygon on 🌐 to `maxlen()` gives 36458 elements instead of the 95904 from the `QuadRefinement` loop.

New vertices come from `centroid` of the split edge, so they sit on the geodesic.


juliohm:
❤️ ❤️ ❤️

juliohm:
@jitendravjh I wonder if it would be too challenging to generalize the algorithm to work on general polygonal meshes, not just triangle meshes. The example you added in the docs is a good example. I would expect that the right half of the grid to be untouched, i.e., all elements remained quadrangles.

If the implementation becomes too complex to handle general `Ngon` elements, we can leave this as future improvement.

juliohm:
Thank you @jitendravjh ! I've simplified the code in terms of `Tuple`s again. Please let me know if you see any issue with this refactoring step. I believe it shouldn't compromise performance as we still need to allocate the intermediate vector before calling the final `map` to create `newconnec` with inferred types.

jitendravjh:
> Thank you @jitendravjh ! I've simplified the code in terms of `Tuple`s again. Please let me know if you see any issue with this refactoring step. I believe it shouldn't compromise performance as we still need to allocate the intermediate vector before calling the final `map` to create `newconnec` with inferred types.

Looks good. Types still come out concrete, `Connectivity{Quadrangle,4} `when nothing splits and `Connectivity{Triangle,3}` when everything does, abstract only when the mesh really is mixed. Tests pass in both float types.

PR #1439, opened 2026-09-17
Fix `TriRefinement` when elements are preserved

jitendravjh:
`refine(mesh, TriRefinement(pred))` throws when a preserved element is not a triangle, for example `refine(cartgrid(4, 4), TriRefinement(e -> false))`. Collects tuples like `EdgeRefinement` does.


juliohm:
I believe that in this case we always return a triangle mesh by design. Users might find it surprising that a mesh has other element types after triangle refinement. Same for `QuadRefinement` that always returns meshes of quadrangles.

Issue #1367, opened 2026-06-02
Optimize visualization of non-Euclidean geometries and domains


gdkrmr:
The following never finishes for me, the julia process just keeps spinning at 100%:
```julia_repl
julia> import GeoArtifacts

julia> using GeoTables

julia> using Meshes
       #using Plots

julia> using CairoMakie

julia> BRA = GeoArtifacts.GADM.get("BRA", depth = 0)
               1×3 GeoTable over 1 GeometrySet
┌─────────────┬─────────────┬───────────────────────────────┐
│    GID_0    │   COUNTRY   │           geometry            │
│ Categorical │ Categorical │         MultiPolygon          │
│  [NoUnits]  │  [NoUnits]  │ 🖈 GeodeticLatLon{WGS84Latest} │
├─────────────┼─────────────┼───────────────────────────────┤
│     BRA     │   Brazil    │     Multi(1007×PolyArea)      │
└─────────────┴─────────────┴───────────────────────────────┘

julia> BRA_domain = domain(BRA)
1 GeometrySet
└─ Multi(1007×PolyArea)

julia> viz(BRA_domain)
```

```julia_repl
(jl_gIjdF8) pkg> st
Status `/tmp/jl_gIjdF8/Project.toml`
  [13f3f980] CairoMakie v0.15.11
  [2a3f1dca] GeoArtifacts v1.3.4
  [e502b557] GeoTables v1.28.5
  [eacbb407] Meshes v0.57.12

julia> VERSION
v"1.11.9"
```

juliohm:
Hi @gdkrmr , thank you for reporting the issue. This is happening because the geometries in the dataset have `LatLon` coordinates and are interpreted as spherical geometries. Our pipeline to visualize spherical geometries needs some optimizations. We automatically refine polygons until all edges are smaller than a predefined (adjustable) [threshold](https://juliageometry.github.io/MeshesDocs/stable/tolerances/#Maximum-length-for-default-refinement-of-meshes-on), but the refinement method that is currently applied by default is greedy `QuadRefinement`:

https://github.com/JuliaGeometry/Meshes.jl/blob/7aad27adaecbb58e28193b26371c8d546205debc/src/refinement.jl#L19

We need to replace this default by a new adaptive  refinement method based on edge predicates, which I didn't have time to implement yet: https://github.com/JuliaGeometry/Meshes.jl/issues/1118

If you have time and skills to contribute, that would be super helpful.

[...]

PR #1440, opened 2026-09-17
Refine only the long edges in `MaxLengthRefinement`

jitendravjh:
For #1367. `MaxLengthRefinement` refined the whole mesh with `QuadRefinement` until no element was too long, so short edges were halved along with the long ones. It now splits only the edges that exceed the length.

A polygon on 🌐 with 2000 vertices ends up with 36458 triangles after `simplexify` instead of 191808.

Grids keep the old path, so `Box` and `Quadrangle` keep their `GridTopology`. The unstructured cases in `discretization.jl` now come out as triangle meshes.


jitendravjh:
Closing in favor of #1441.

PR #1441, opened 2026-09-17
Refactor MaxLengthRefinement

juliohm:
Related to #1367

The `MaxLengthRefinement` of general `Mesh`es is performed with the new `EdgeRefinement`.


jitendravjh:
`refinemaxlen` doesn't bound the edges anymore, it splits each long edge once and stops. A big triangle on 🌐 ends up with 4601 km edges while `maxlen()` is 500 km. Repeating until nothing is longer fixes it.

juliohm:
> `refinemaxlen` doesn't bound the edges anymore, it splits each long edge once and stops. A big triangle on 🌐 ends up with 4601 km edges while `maxlen()` is 500 km. Repeating until nothing is longer fixes it.

That is true. I will try to reintroduce the `while` loop. If you have suggestions, please feel free to share.

jitendravjh:
`segments` throws on 1D meshes with `SimpleTopology`, so comparing element counts is easier:

```julia
function refine(mesh::Mesh, method::MaxLengthRefinement)
  islong(e) = length(e) > method.length
  while true
    rmesh = refine(mesh, EdgeRefinement(islong))
    nelements(rmesh) == nelements(mesh) && return mesh
    mesh = rmesh
  end
end
```

Tried it on your branch, everything lands within `maxlen()` and the 2000 vertex polygon stays at 36458 elements.

juliohm:
@jitendravjh do you have any suggestion to retain the `GridTopology` of the tests involving `Box` and `Quadrangle` with `LatLon` coordinates?

The `discretize(box)` generates an efficient 1x1 RegularGrid with `GridTopology` and the `refine(mesh, MaxLengthRefinement(maxlen))` step throws this information away. We have a specialization for `OrthoRegularGrid` already, and could consider introducing a new specialization for other meshes with `GridTopology`.

juliohm:
Figured out a method for general `Grid` based on `RegularRefinement`. The heuristic to check if elements of the `Grid` are small enough is not perfect, but should work in practice. It only inspects the first, middle and last element of the grid as opposed to checking all elements.

